PR #1, opened 2020-09-07
Initial server setup

CarlosNZ:
The following info is copied from the README as it gives a basic overview of what's been set up here:

---- 
Install dependencies:  
`yarn install`

Initialise database:  
`yarn database_init`  
(Needs `psql` command-line tool to be installed, and a super-user named `postgres` -- [here](https://gist.github.com/ibraheem4/ce5ccd3e4d7a65589ce84f2a3b7c23a3) for instructions.  
Creates a database named `irims` and populates it with minimal data.)

To run in dev mode:  
`yarn dev`

To launch Postgraphile server (in a new Terminal process):  
`yarn pg`

To launch in dev mode _with_ Postgraphile server:  
`yarn dev-pg`

To build (compiles all .ts files to .js in `build` folder):  
`yarn build`

To build and run compiled files:  
`yarn start`

In dev mode, uses `nodemon` to monitor changes, and `ts-node` to compile typescript files on the fly.

Note: Plugins (in `src/plugins`) are **standalone packages** and must be compiled individually. However, they should come pre-compiled and are copied to the `build` folder when building the server app. To automatically re-compile all plugins, run `yarn build_plugins`.

This repo has the following functionality implemented in basic form:

- Fastify server with endpoints for:
  - File upload (and register in database)
  - File download (by database ID)
- Triggers and Actions, including:
  - Register plugins on startup
  - Run actions in response to Database triggers
  - Create scheduled actions (using `node-schedule` which I'm not sure is the best way to go)
- `evaluateExpression` module

## Triggers and Actions

In order to see Triggers and Actions working, you'll need to run the `createSchema.sql` then `insertData.sql` scripts in the database folder.

The application with ID = 3 (Company registration: Company C) has one Action associated with it, the Console logger. All it does is write a message to the console, and you can see the details on the `template_action` table.

To cause a Trigger to initiate the Action, change the `trigger` field on the application:  
`UPDATE application SET trigger = 'onApplicationSubmit' WHERE id = 3`

You should see the message printed to the console.

Also, check the `action_plugin` table to see a couple of extra dummy plugins that automatically get registered on startup.

## evaluateExpression

Documented in the `Query-Syntax.md` in the documentation folder.

Code is in the `modules` folder.

To test, run: `yarn test` (uses installed Jest)

## Documentation

Developer documentation available in documentation folder -- intended for publication to this repo's wiki. See the documentation's own README for more information.


nmadruga (commented):
Overall comment `LICENSE` file should be reverted.

src/components/triggersAndActions.ts
```diff
@@ -0,0 +1,176 @@
...
+  actionSchedule: any[]
+) {
+  // const actionSchedule: any[] = [];
+
+  console.log('Loading Scheduled jobs');
+  // Load from Database
+  client
+    .query(
+      `SELECT id, action_code, parameters, time_executed FROM action_queue WHERE status = 'Scheduled' ORDER BY time_executed`
+    )
```
nmadruga, line 75: Util Pg function:
* retrieveScheduledActions

src/components/triggersAndActions.ts
```diff
@@ -0,0 +1,176 @@
...
+              code: action.action_code,
+              parameters: action.parameters,
+            },
+            actionLibrary
+          );
+        }
+      });
+    });
+  // return actionSchedule;
+};
```
nmadruga, line 109: I think I'll continue on Monday from here 😆

nmadruga (changes requested):
Wow! It's pretty good how much you have built so far! 
After start looking at I can see how good this is. I'm happy about how this is going, my main suggestions are trying to keep the code cleaner adding more modularisation and using Types/interfaces instead of raw scalar types to improve the code readiness.

CarlosNZ:
> Overall comment `LICENSE` file should be reverted.

Okay, have just copied your one from front-end

README.md
```diff
@@ -1,2 +1,62 @@
-# application-manager-server
```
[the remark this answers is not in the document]
CarlosNZ: 👍

src/components/postgresConnect.ts
```diff
@@ -0,0 +1,50 @@
...
+
+// Postgres Database listeners for Triggers/Actions
+export const pgClient = new Client({
+  user: 'postgres',
+  host: 'localhost',
+  database: 'tmf_app_manager',
+  password: '',
+  port: 5432,
+});
+
```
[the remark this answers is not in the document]
CarlosNZ: Yup, done. In `./src/config.json` (along with a couple of other configurable values)

src/modules/evaluateExpression/evaluateExpression.ts
```diff
@@ -0,0 +1,128 @@
...
+
+      case 'objectProperties':
+        try {
+          return params[childrenResolved[0].object][childrenResolved[0].property];
+        } catch {
+          return "Can't resolve object";
+        }
+
+      case 'pgSQL':
+        if (!params.connection) return 'No database connection provided';
```
nmadruga, line 89: I was going to suggest to access the Pg util functions here as well. No need to pass the client here. 
I might just check how does the test work to define a mock database to use, if that can be changed if not passed as an argument to the function I guess it's alright to leave it for now

src/modules/evaluateExpression/evaluateExpression.ts
```diff
@@ -0,0 +1,128 @@
...
+        return processPgSQL(childrenResolved, query.type, params.connection);
+
+      case 'graphQL':
+        if (!params.connection) return 'No database connection provided';
```
nmadruga, line 93: Same thing, could be accessing the GraphQL client directly inside the function/module.

src/modules/evaluateExpression/evaluateExpression.ts
```diff
@@ -0,0 +1,128 @@
...
+        return processGraphQL(childrenResolved, params.connection);
+
+      // etc. for as many other operators as we want/need.
+    }
+  }
+  return 'No matching operators';
+}
+
+async function processPgSQL(queryArray: any[], queryType: string, connection: IConnection) {
```
nmadruga, line 102: Move this function to Pg util function:
* processQuery

src/modules/evaluateExpression/evaluateExpressionTestData.ts
```diff
@@ -0,0 +1,708 @@
...
+          children: [
+            { value: 'SELECT COUNT(*) FROM user_organisation WHERE user_id = $1' },
+            {
+              operator: 'objectProperties',
+              children: [{ value: { object: 'user', property: 'id' } }],
+            },
+            {
+              operator: 'objectProperties',
+              children: [{ value: { object: 'organisation', property: 'id' } }],
+            },
```
nmadruga, line 609: I don't think this object is in use on this query

src/components/triggersAndActions.ts
```diff
@@ -0,0 +1,176 @@
...
+  condition: { [key: string]: any };
+  parameter_queries: { [key: string]: any };
+}
+
+const pluginFolder = path.join(getAppRootDir(), 'plugins'); //Change this to config file
+
+// Load actions from Database at server startup
+export const loadActions = async function (
+  client: { [key: string]: Function },
+  actionLibrary: { [key: string]: Function }
```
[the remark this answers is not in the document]
CarlosNZ: Okay, I've defined ActionLibrary as an Interface. Left the client one for now, because it should be defined by the package it came with `node-postgres`, but doesn't seem to be defined. There's probably a community-defined type for it somewhere, but not sure how to import and work with this yet. Leaving it as is as a reminder to re-look at it (I'll make an issue too).

src/components/triggersAndActions.ts
```diff
@@ -0,0 +1,176 @@
...
+  // const actionLibrary: { [key: string]: Function } = {};
+
+  client
+    .query('SELECT code, path, function_name FROM action_plugin')
+    .then((res: DatabaseResult) => {
+      res.rows.forEach((row) => {
+        const action = require(path.join(pluginFolder, row.path));
+        actionLibrary[row.code] = action[row.function_name];
+      });
+    });
```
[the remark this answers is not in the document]
CarlosNZ: Yeah, will do this as part of issue for [database abstraction layer](https://github.com/orgs/openmsupply/projects/6#card-45395961).

src/modules/evaluateExpression/evaluateExpressionTestData.ts
```diff
@@ -0,0 +1,708 @@
...
+    {
+      operator: '=',
+      children: [
+        {
+          operator: 'objectProperties',
+          children: [{ value: { object: 'application', property: 'stage' } }],
+        },
+        {
+          value: 1,
+        },
```
nmadruga, line 633: Hmmm I think this one can be more complex. It seems to me that the `application.stage` property might be an object instead, with name and order (1). Something like that. In that case, this query might need a little tweak.

src/modules/evaluateExpression/evaluateExpressionTestData.ts
```diff
@@ -0,0 +1,708 @@
...
+                  "object": "application",
+                  "property": "stage"
+                }
+              }
+            ]
+          },
+          {
+            "value": 1
+          }
+        ]
```
nmadruga, line 677: The same problem described in the previous `complex2` query

src/plugins/action_console_dummy_plugin_1/src/dummy1.js
```diff
@@ -0,0 +1,15 @@
...
+      status: 'Success',
+      error_log: '',
+    };
+  } catch (error) {
+    return {
+      status: 'Fail',
+      error_log: 'There was a problem',
+    };
+  }
+};
```
nmadruga, line 15: Delete this one?

src/components/triggersAndActions.ts
```diff
@@ -0,0 +1,176 @@
...
+  console.log('Actions loaded');
+
+  // return actionLibrary;
+};
+
+// Load scheduled jobs from Database at server startup
+export const loadScheduledActions = async function (
+  client: { [key: string]: Function },
+  actionLibrary: { [key: string]: Function },
+  actionSchedule: any[]
```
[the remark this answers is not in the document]
CarlosNZ: First two as mentioned above. 

For the `actionSchedule`, I'm not sure how to define the type of object contained in the array, as it's an object from a 3rd-party library. And also, I'm not totally sure if this library is what we're going to be going with, so will re-visit this once we've got further with [how to handle scheduled actions](https://github.com/orgs/openmsupply/projects/6#card-45400590).

src/components/triggersAndActions.ts
CarlosNZ: Other issue, as above.

src/plugins/action_console_dummy_plugin_1/src/plugin.json
```diff
@@ -0,0 +1,19 @@
...
+    "author": {
+      "name": "The mSupply Foundation",
+      "url": "http://msupply.foundation"
+    },
+    "links": [],
+    "screenshots": [],
+    "version": "1.0.0",
+    "updated": "2020-08-27"
+  }
+}
```
nmadruga, line 19: Very cool! Note: if we don't have a documentation for this one yet. It will be great to add to the **public docs** reminder.

src/components/triggersAndActions.ts
CarlosNZ: Okay, sure thing.

Maybe we merge this PR so we've got something start with, and you can do a more thorough review of Triggers and Actions later?

src/plugins/action_console_dummy_plugin_2/src/dummy2.js
```diff
@@ -0,0 +1,15 @@
...
+      status: 'Success',
+      error_log: '',
+    };
+  } catch (error) {
+    return {
+      status: 'Fail',
+      error_log: 'There was a problem',
+    };
+  }
+};
```
nmadruga, line 15: Delete this one?

src/plugins/action_console_log/src/consoleLog.js
```diff
@@ -0,0 +1,16 @@
...
+      status: 'Success',
+      error_log: '',
+    };
+  } catch (error) {
+    return {
+      status: 'Fail',
+      error_log: 'There was a problem',
+    };
+  }
+};
```
nmadruga, line 16: Delete this one?

src/plugins/action_console_log/src/dummy2.js
```diff
@@ -0,0 +1,15 @@
...
+      status: 'Success',
+      error_log: '',
+    };
+  } catch (error) {
+    return {
+      status: 'Fail',
+      error_log: 'There was a problem',
+    };
+  }
+};
```
nmadruga, line 15: Another one to delete?

src/plugins/action_console_log/src/dummy2.ts
```diff
@@ -0,0 +1,14 @@
+module.exports['dummy2'] = function (parameters: any) {
```
nmadruga, line 1: I don't see what is the different between this one and the `src/plugins/action_console_dummy_plugin_2/src/dummy2.ts` file. Is this to show a difference on how plugins can export similar function names?

src/server.ts
```diff
@@ -0,0 +1,65 @@
...
+// Bare-bones Fastify server
+
+const startServer = async () => {
+  await loadActionPlugins(); // Connects to Database and listens for Triggers
+
+  createFilesFolder();
+
+  const server = fastify();
+
+  server.register(require('fastify-static'), {
```
nmadruga, line 21: Can `require('fastify-static')` be imported as one of the packages on the top please?

src/server.ts
```diff
@@ -0,0 +1,65 @@
...
+    root: path.join(getAppRootDir(), filesFolderName),
+  });
+
+  server.register(require('fastify-multipart'));
```
nmadruga, line 25: Also import `require('fastify-multipart')` on the top please. : )

src/server.ts
```diff
@@ -0,0 +1,65 @@
...
+
+  // File download endpoint (get by Database ID)
+  server.get('/file', async function (request: any, reply: any) {
```
nmadruga, line 28: Types for request and reply

src/server.ts
```diff
@@ -0,0 +1,65 @@
...
+    const filename = await getFilename(request.query.id);
+    // TO-DO Check for permission to access file
+    return reply.sendFile(filename);
+  });
+
+  // File upload endpoint
+  server.post('/upload', async function (request: any, reply) {
```
nmadruga, line 35: Same types fro request and reply here

src/server.ts
```diff
@@ -0,0 +1,65 @@
...
+    // TO-DO: Check if logged in
+    const data = await request.files();
+    await saveFiles(data, request.query);
```
nmadruga, line 38: Great job, keeping these File utils in a separated module! I like that : D

src/components/fileHandler.ts
```diff
@@ -0,0 +1,76 @@
...
+export function createFilesFolder() {
+  try {
+    fs.mkdirSync(path.join(getAppRootDir(), filesFolderName));
+  } catch {
+    // Folder already exists
+  }
+}
+
+export async function getFilename(id: string) {
+  const result = await pgClient.query('SELECT path, original_filename FROM file WHERE id = $1', [
```
nmadruga, line 20: Pg util function detected : )

src/components/fileHandler.ts
```diff
@@ -0,0 +1,76 @@
...
+    id,
+  ]);
+  const folder = result.rows[0].path; // Not currently used
+  const filenameOrig = result.rows[0].original_filename;
```
nmadruga, line 24: That can be already returned in the structured way from the Pg util function

src/components/fileHandler.ts
```diff
@@ -0,0 +1,76 @@
...
+  const ext = path.extname(String(filenameOrig));
+  const basename = path.basename(filenameOrig, ext);
+  return `${basename}_${id}${ext}`;
+}
+
+const pump = util.promisify(pipeline);
+export async function saveFiles(data: any, queryParams: any) {
```
nmadruga, line 31: Types for `data` and `queryParams` please

src/components/fileHandler.ts
```diff
@@ -0,0 +1,76 @@
...
+  for await (const file of data) {
+    await pump(
+      file.file,
+      fs.createWriteStream(path.join(getAppRootDir(), filesFolderName, file.filename))
+    );
+    const parameters: any = {
```
nmadruga, line 37: I wonder if there is a nicer way to initialise all of a type undefined instead of having to create like this...

src/components/fileHandler.ts
```diff
@@ -0,0 +1,76 @@
...
+    });
+    const fileID = await registerFileInDB(file, parameters);
+  }
+}
+
+async function registerFileInDB(file: any, parameters: any) {
+  // Insert record into Db and get back ID
+  const query = {
+    text:
+      'INSERT INTO file (user_id, original_filename, path, mimetype, application_id, application_response_id) VALUES ($1, $2, $3, $4, $5, $6) RETURNING id;',
```
nmadruga, line 56: Another pg util function to be called:
* registerFileInDB

src/components/triggersAndActions.ts
nmadruga: Ah, I think it might be good to have a run on the way it is to change to use the Pg util functions in a separated place. Also some missing interface/types make it more difficult to understand. I think it's better to address points of improvement first, otherwise they can be left behind while we rush other things in.

src/modules/evaluateExpression/evaluateExpression.test.ts
```diff
@@ -0,0 +1,345 @@
...
+const { Client } = require('pg');
+
+//CONFIG -- DATABASE SETUP:
+const pgConnect = new Client({
+  user: 'postgres',
+  host: 'localhost',
+  database: 'irims_playground',
+  password: '',
+  port: 5432,
+});
```
nmadruga, line 15: Could use the same config file

nmadruga (changes requested):
I see now that you have added another issue to deal with the Database abstraction layer. The code is a bit hard to understand the way it is now without that. So I'll be happy to see that one get added this week to have another look.

Prettier notes:
- Should we not use `;` for ending lines? It isn't required in Javascript, I suggest adding this rule.

Overall notes:
- The jest in each plugin, I'm not sure exactly how that works. Is it a test that get added to the suite of all tests done when we run the tests for the backend files?

CarlosNZ:
> Prettier notes:
>     * Should we not use `;` for ending lines? It isn't required in Javascript, I suggest adding this rule.
That's already happening for me. I think it's default behaviour. I think you only need to add rules that over-ride the defaults, is that not correct?

CarlosNZ:
>  The jest in each plugin, I'm not sure exactly how that works. Is it a test that get added to the suite of all tests done when we run the tests for the backend files?

The idea is that you should be able to work on each plugin as it's own package, so they each need their own configs etc. So if you run `test` from within each plugin folder, it'll run the tests associated with that plugin. But if you run `test` from the main repo folder, it'll run all the tests including the plugins (I think -- although this can be configured in the `jest.config.js` file.

There's currently no tests set up for the plugins, but I've added it to their package configs anyway as a kind of "template" which will make setting up tests easier in new plugins.

src/components/triggersAndActions.ts
```diff
@@ -0,0 +1,176 @@
...
+  console.log('Loading Scheduled jobs');
+  // Load from Database
+  client
+    .query(
+      `SELECT id, action_code, parameters, time_executed FROM action_queue WHERE status = 'Scheduled' ORDER BY time_executed`
+    )
+    .then((res: DatabaseResult) => {
+      res.rows.forEach((action) => {
+        // console.log(action);
+        const date = new Date(action.time_executed);
```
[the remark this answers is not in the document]
CarlosNZ: Yes, you're right it is -- the only reason I didn't change it because I didn't want to change it everywhere (database, etc.) just yet. But yeah, makes sense to make it clearer -- I'll change it, but it's probably going to cause a merge conflict when merging my other branch with more substantial database changes #7 .

Originally the field was going to be just a log of when the Action had run, but then when thinking about scheduled actions, I realised it makes sense to use that value as the scheduled time as well, and just search for which ones are in the future at startup.

src/modules/evaluateExpression/evaluateExpression.ts
```diff
@@ -0,0 +1,128 @@
+interface IParameters {
+  [key: string]: any;
```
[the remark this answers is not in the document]
CarlosNZ: I had trouble when I tried to restrict this value's type. It can be any of `string`, `number` or `boolean`. And even when I defined it as being allowed to be any one of these, I was getting errors further down with the functions I tried to call on them if the method didn't work on ALL of these types (i.e. functions that only operate on strings gave an error saying it can't run on `boolean`, etc.)

What's the best practice for this kind of thing? Do a `if typeOf` check before calling each method and throw a custom error if not a match? Will typescript allow this?

src/modules/evaluateExpression/evaluateExpression.ts
```diff
@@ -0,0 +1,128 @@
...
+  connection?: IConnection;
+}
+
+interface IQueryNode {
+  value?: string | number | boolean | object;
+  type?: string;
+  operator?: string;
```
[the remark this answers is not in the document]
CarlosNZ: But those values are just strings. Do you mean create a new type and provide a list of values that they must be?

src/modules/evaluateExpression/evaluateExpression.ts
CarlosNZ: Not sure exactly what you mean. I need to check the client object exists here or typescript gives an error on the next line saying that the connection parameter can't be `undefined`. But I have to have undefined as an option for that field as all the other operators don't require a `connection` field. So I'm just doing a type-check before calling a function. (I guess I'm also answering my own previous question though lol)

src/modules/evaluateExpression/evaluateExpression.ts
CarlosNZ: See above response.

src/modules/evaluateExpression/evaluateExpression.ts
CarlosNZ: This `evaluateExpression` module is going to be moved out into its own package eventually, so best not rely on unconnected modules, right? (Also, the PG function here is quite unique as the actual `query` is provided by the caller -- and it could be anything).

src/modules/evaluateExpression/evaluateExpressionTestData.ts
CarlosNZ: The test that uses this object is currently commented out, since it uses the as-yet-unimplemented GraphQL operator. But it will use this object -- I just haven't written the test query properly yet.

src/modules/evaluateExpression/evaluateExpressionTestData.ts
CarlosNZ: Yes, true, but I've specified in my spec for this module that the `objectProperties` operator is still under consideration and we'll have a better idea of its requirements once we know how local state is being represented.

src/modules/evaluateExpression/evaluateExpressionTestData.ts
CarlosNZ: Same answer 😜

src/plugins/action_console_dummy_plugin_1/src/dummy1.js
CarlosNZ: I think it's a good idea to keep these two dummy plugins until we have a couple more real plugins. Because they allow us to check that all the functionality of the `registerPlugins` component is working as expected. (i.e. loading their info to database, skipping if doesn't have required info, etc.)

src/plugins/action_console_dummy_plugin_1/src/plugin.json
CarlosNZ: You mean the plugin spec?

src/plugins/action_console_dummy_plugin_2/src/dummy2.js
CarlosNZ: See above.

src/plugins/action_console_dummy_plugin_1/src/dummy1.js
CarlosNZ: Oh, did you mean delete the `.js` file?
Yeah, could do, but the idea was these should be compiled independently of the main project, and so when you download the repo you get the plugins "pre-built". (Otherwise you'd have to remember to run `yarn build_plugins` before `yarn build`, or you won't get compiled plugins in the build folder.)

src/plugins/action_console_dummy_plugin_2/src/dummy2.js
CarlosNZ: See previous.

src/plugins/action_console_log/src/consoleLog.js
CarlosNZ: Same.

src/plugins/action_console_log/src/dummy2.js
CarlosNZ: Same.

src/plugins/action_console_log/src/dummy2.ts
CarlosNZ: Yup, these dummy files shouldn't be in this folder -- will delete. 😊

src/server.ts
CarlosNZ: Yes!

src/server.ts
CarlosNZ: I'm not sure how to Type these. The actual request and reply object are enormous, complicated things, so wouldn't want to do it manually. Fastify is supposed to have a @types file and I've installed it. And when I specifically install @types packages for these plugins I get "fastify-multipart provides its own type definitions, so you do not need this installed." 

Do you know what I'm missing here?

src/server.ts
CarlosNZ: See prev. 🙁

src/components/fileHandler.ts
CarlosNZ: Have defined an interface for `queryParams`, but how would I go about defining the type for `data`? -- it's a collection of files passed from an http POST request. 🤷‍♂️

src/components/fileHandler.ts
CarlosNZ: Changed to `const parameters = { ...queryParams };`

src/modules/evaluateExpression/evaluateExpression.test.ts
CarlosNZ: Yes!

src/components/registerPlugins.ts
```diff
@@ -0,0 +1,105 @@
...
+    .filter((item) => fs.statSync(path.join(pluginFolder, item)).isDirectory())
+    .map((item) => {
+      return { path: path.join(item, 'src') };
+    })
+    .filter((item) => fs.existsSync(path.join(pluginFolder, item.path, 'plugin.json')))
+    .map((item) => {
+      const pluginObject = {
+        ...item,
+        ...JSON.parse(fs.readFileSync(path.join(pluginFolder, item.path, 'plugin.json'), 'utf8')),
+      };
```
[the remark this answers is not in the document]
CarlosNZ: It won't because the previous `.filter` removes any items from the array that don't have the .`json` file at the expected path. ;)

(For testing, I originally had additional dummy plugin folders that were missing `plugin.json` files, and they were skipped gracefully).

src/components/triggersAndActions.ts
```diff
@@ -0,0 +1,176 @@
...
+}
+
+export interface Action {
+  code: string;
+  path: string;
+  name: string;
+  trigger: string;
+  condition: { [key: string]: any };
+  parameter_queries: { [key: string]: any };
+}
```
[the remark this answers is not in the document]
CarlosNZ: Later? (I'll probably pull the two Database ones out when I make the postgres abstraction layer anyway.)

src/components/registerPlugins.ts
```diff
@@ -0,0 +1,105 @@
...
+// The unique identifier for a plugin is its "code".
+
+import * as fs from 'fs';
+import path from 'path';
+import { DatabaseResult } from './triggersAndActions';
+import getAppRootDir from './getAppRoot';
+
+const pluginFolder = path.join(getAppRootDir(), 'plugins'); //Change this to config file
+
+export default async function registerPlugins(client: { [key: string]: Function }) {
```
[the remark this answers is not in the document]
CarlosNZ: Okay, I now realise I need to import `@types/pg` (and for any other 3rd-party packages) to have predefined types.

src/modules/evaluateExpression/evaluateExpression.ts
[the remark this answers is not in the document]
CarlosNZ: And also, if I try and restrict the type here, I also get an error on the `connection` field, as the `Client` type is not one of the allowed types for the `[key: string]` field(s). I can see the logic of why it's doing this, but not sure the best way to deal with it.

src/components/registerPlugins.ts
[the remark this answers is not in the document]
nmadruga: Great! That's nice touch. It would be good to add to the documentation (if you haven't already) what what is expected to have inside each plugin folder.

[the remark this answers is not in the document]
CarlosNZ: Okay have made [an issue](#21) for it. :)

src/modules/evaluateExpression/evaluateExpression.ts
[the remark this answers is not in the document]
nmadruga: Yes, sorry, exactly. I'm also learning about Typescript syntax now. What I was thinking is a way to defined an enum of possible operators and types to be used. And then type as that. So even though you pass the operator and type as string it should be part of this subset of accepted types. I'll find the syntax for that now.

src/components/fileHandler.ts
nmadruga: Yea. Fair enough. We can check later if there is any way to have a type for `data`.

src/modules/evaluateExpression/evaluateExpression.ts
nmadruga: Oh. I see your point. I was thinking if this would be part of what is shared between the Back-end core and the plugins. Not sure how to do this one either, so I think we should leave it for now and improve later on a PR dedicated of creating the basic structure to be accessible by the `evaluator` and `querySyntax` module.

src/plugins/action_console_dummy_plugin_1/src/plugin.json
nmadruga: Yep

src/plugins/action_console_dummy_plugin_1/src/dummy1.js
nmadruga: I see. Thanks!

src/plugins/action_console_dummy_plugin_1/src/plugin.json
CarlosNZ: [Issue](#21 ) made :)

src/modules/evaluateExpression/evaluateExpression.ts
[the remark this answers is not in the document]
CarlosNZ: Got it -- can use String Literal types:

> You can use these features together to get enum-like behavior with strings.

https://www.typescriptlang.org/docs/handbook/literal-types.html#string-literal-types

CarlosNZ:
FYI -- just updated `.prettierrc` to match front-end one (with semi-colons off) and reformatted all code.

src/modules/evaluateExpression/evaluateExpression.ts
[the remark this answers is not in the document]
nmadruga: I think there are two available options:
1) You define an enum an then use it as the type for this field inside your `interface` definition:
```
enum QUERY_SYNTAX_TYPE {
  STRING = 'string',
  ARRAY = 'array',
  NUMBER = 'number'
}

enum QUERY_SYNTAX_OPERATOR {
  PLUS = '+',
  EQUALS = '=',
  NOT_EQUAL = '!=',
  AND = 'AND',
  OR = 'OR',
  CONCAT = 'CONCAT',
  REGEX = 'REGEX',
  LOCAL_DATA = 'objectProperties',
  PG_SQL = 'pgSQL',
  GRAPH_QL = 'graphQL'
}

interface IQueryNode {
  value?: string | number | boolean | object;
  type?: QUERY_SYNTAX_TYPE;
  operator?: QUERY_SYNTAX_OPERATOR;
  children?: Array<IQueryNode>;
}
```

2) You can define the subset to be checked for strings inside the `interface` definition:
```
interface IQueryNode {
  value?: string | number | boolean | object;
  type?: 'array' | 'string' | 'number' ;
  operator?: '+' | '=' | '!=' | 'AND' | 'OR' | 'CONCAT' | 'REGEX' | 'objectProperties' | 'pgSQL' | 'graphQL' ;
  children?: Array<IQueryNode>;
}
```

nmadruga (approved):
That's great! I think that besides a few types that we still need to investigate how to change from using `any` and the Pg utils functions which will be address in another PR it seems good to me.

PR #7, opened 2020-09-10
#3 Database inheritance and other changes

### LINKED ISSUE #3: Schema design problems -- inheritance and linked list
There's a couple of related problems with the current schema design that I've encountered.

### 1. Inherited tables don't play well with postGraphile/GraphQL

The tables `template_question` and `template_information` inherit common fields from `template_element`. The problem is the resultant GraphQL schema doesn't expose the sub-tables correctly. When drilling down through `allTemplates` or `templateById` through `Section` etc, only `templateElementsBySectionId` is available, and within that only the parent table (Elements) fields are available. 

Similarly when using mutations to drill down into Elements from Template or Section level, the sub-tables are not exposed. This is a big problem for using nested mutations to create Templates, etc.

There's probably a way to write some SQL functions to force postGraphile to expose the sub-tables properly (it's beyond me right now), but is it worth it? I'd be inclined to just change the schema to have the one `template_element` table and do away with the inheritance. We'd just add one additional field to distinguish between "Question" or "Information" types and there'd be a couple of NULL values for the "Information" types. This seems the simplest solution -- any other suggestions?

### 2. Template element "Next element" shouldn't use auto-generated primary key as its link.

The problem with this is that the link becomes meaningless outside the current state of the database (i.e if exported). There needs to be a way to define the **next** question in terms of something that persists outside the database. 

I'd suggest just using the `code`, and having it as a "soft" link rather than a proper foreign key (the same way I've done for linking `action_plugin`s to `template`s). So instead of `next_element_id` we'd have `next_element_code`. Codes must be unique for a given template, but could be duplicated between templates, so when identifying the next question we'd have to match on Template ID AND code. But that seems fine to me.

The other thing is that GraphQL nested mutations won't work well with the current schema, if you wanted to create a bunch of questions in one query -- there's no meaningful way to refer to a database ID of an element that doesn't yet exist, so you'd have to write a tedious loop to generate many individual queries to create a proper linked list of questions. Again, linking by "code" seems like a reasonable solution. 

Any objections to making this change, or any better suggestions?

CarlosNZ:
Fixes #3.

1. Removed inheritance from `template_element` and the child tables `template_question` and `template_information`. Now all in single `template_element` table, with one extra field called `category`, which can be either `Question` or `Information`.
2. `next_element_id`  is now `next_element_code` as per the issue discussion.
3. Also have updated the `yarn database_init` script. It now uses GraphQL queries to add the mock data, which can be found in `./database/insertData.js`
4. New version of Schema diagram (draw.io) in DESIGN Google Drive folder.

Run `yarn database_init` to recreate the the above changes.


yarn.lock
```diff
@@ -3566,6 +3566,11 @@ nice-try@^1.0.4:
   resolved "https://registry.yarnpkg.com/nice-try/-/nice-try-1.0.5.tgz#a3378a7696ce7d223e88fc9b764bd7ef1089e366"
   integrity sha512-1nh45deeb5olNY7eX82BkPO7SSxR5SSYJiPTrTdFUVYwAl8CKMA5N9PjTYkHiRjisVcxcQ1HXdLhx2qxxJzLNQ==
 
+node-fetch@^2.6.1:
+  version "2.6.1"
+  resolved "https://registry.yarnpkg.com/node-fetch/-/node-fetch-2.6.1.tgz#045bd323631f76ed2e2b55573394416b639a0052"
+  integrity sha512-V4aYg89jEoVRxRb2fJdAg8FHvI7cEyYdVAh94HH0UIK8oJxUfkjlDQN9RbMx+bEjP7+ggMiFRprSti032Oipxw==
+
```
nmadruga, line 3573: should we commit this file? I always remote from my commits. So I'm not sure.

database/insertData.js
```diff
@@ -0,0 +1,592 @@
...
+        ) {
+          user {
+            email
+            password
+            username
+          }
+        }
+      }`,
+  `
+            mutation {
```
nmadruga, line 247:
```suggestion
  `mutation {
```

database/insertData.js
```diff
@@ -0,0 +1,592 @@
...
+        ) {
+          user {
+            email
+            password
+            username
+          }
+        }
+      }`,
+  `      
+      mutation {
```
nmadruga, line 261:
```suggestion
  `mutation {
```

database/insertData.js
```diff
@@ -0,0 +1,592 @@
...
```
nmadruga, line 275:
```suggestion
  `mutation {
```

nmadruga (approved):

yarn.lock
CarlosNZ: I think so. They did on mobile. I think it's important to make sure all users stay up do date with the correct versions of all their node packages, isn't it?

database/insertData.js
CarlosNZ: Done

Issue #8, opened 2020-09-11
Create separated sql files for each table and script to import in order 


nmadruga:
Only for creating tables.

Goal: Having a clearer way to check for schema updates.

CarlosNZ:
What do you mean by "a clearer way to check for schema updates"?

CarlosNZ:
Question -- what do you guys think @andreievg @nmadruga 

There are two ways to do this:

1. Name all `.sql` scripts with a numeric prefix (e.g. `01_user_table`), then make a script to loop over all files in folder (the prefix will ensure they execute in the correct order)

2. No numeric prefix on files, but keep correct order in an array of filenames in build script.

Advantage of **1** is that any name changes only have to be done to one file, and the build script is simpler. Downside is if there are any files inserted in the middle of the sequence, all later numbers will have to be incremented (which is not really a big deal auto-renaming).

I've started with Option 1, but can easily switch to Option 2 later if you guys think that's preferable.

PR #20, opened 2020-09-15
#8 Database setup script -- seperate tables

CarlosNZ:
1. Moved each table into its own `.sql` script. Put them all in `./database/buildSchema` with a numeric prefix specifying the order of execution
2. Wrote shell script `./database/initialise_database.sh` to:
    * Create database (requires `psql`)
    * Build schema using above .sql files (requires `psql`)
    * Insert data using existing `insertData.js` file (requires postGraphile to be running on port 5000)
3. Changed npm script `yarn database_init` to use new `initialise_database` script.


nmadruga (approved):
I think there are a couple of changes to be done ((i.e: user_role), probably while we are working on the documentation of each Schema area to be added in the documentation.

For the sake of this PR I think it's great. All files separated, I ran the script. All seems good.

PR #25, opened 2020-09-18
#15 schema documentation application group

### LINKED ISSUE #15: Schema Documentation - Application

nmadruga:
Fixes #15 .
Schema change on <b>application_stage_history</b> table field `stage` => `stage_id` and remove static options for the stage, since this will be entered by the admin user.  
Documentation for schema regarding <b>application</b>.
The image for the schema group was printed using DBeaver for [Custom Diagrams](https://dbeaver.com/docs/wiki/Custom-Diagrams/)


database/buildSchema/18_application_stage_history.sql
```diff
@@ -1,11 +1,9 @@
 -- application stage history
 
-CREATE TYPE public.application_stage AS ENUM ('Screening', 'Assessment', 'Final Decision'); 
-
 CREATE TABLE public.application_stage_history (
     id serial primary key,
     application_id integer references public.application(id),
-    stage public.application_stage,
+    stage_id integer references public.template_stage(id),
     time_created timestamp with time zone,
```
CarlosNZ, line 7: Shouldn't it just be `timestamp`, not `timestamp with time zone`?
(I realise this isn't actually a change in this PR, but just noticed it -- should be just `timestamp` throughout, right?)

documentation/Database-Schema-Application.md
```diff
@@ -0,0 +1,56 @@
+# Database Schema Application
+
+![Database Schema](images/database-schema-application.png)
+
+## Tables description: Application
```
CarlosNZ, line 5: Maybe "Database area: Application" (or whatever you ended up using for #24 )

documentation/Database-Schema-Application.md
```diff
@@ -0,0 +1,56 @@
...
+
+### Application instance
+The application has all the visualisation and logic defined in the <b>application template</b> [(described in Database Schema Template)](Database-Schema-Template.md). When the Applicant clicks to apply for a application type (aka template) a new instance of one application is created for them. The Applicant user can make changes until submitting the application to be reviewed.
```
CarlosNZ, line 8: Typo: **an** application (not "a")

documentation/Database-Schema-Application.md
```diff
@@ -0,0 +1,56 @@
...
+
+* <b>application</b>
```
CarlosNZ, line 10: Prefer **\*\*markdown\*\*** formatting rather than HTML tags (same as #24)

CarlosNZ, line 10: Also -- see [my comment](https://github.com/openmsupply/application-manager-server/pull/24#discussion_r491740184) on #24 -- applies to all in list.

documentation/Database-Schema-Application.md
```diff
@@ -0,0 +1,56 @@
...
+Representation of the application instance. All nested elements are accessible via joined tables and can be created or queried in the same call using the GraphQL engine.
+The `template_id` links to the template used to display all sections and elements (questions) to the user.
+The `user_id` links to the Applicant user that created this application.
+The `serial` is a number used to display the application number to users. 
+The `name` will be deduced by the template name + user + company names.
```
CarlosNZ, line 15: Ah, okay!
We can do this in the front-end, but do you think it'd be better to write a postGres function to fill this field automatically?

documentation/Database-Schema-Application.md
```diff
@@ -0,0 +1,56 @@
...
+The `is_active` is set to `'true'` for all applications that are still in the review process. 
+The `is_active` is set to `'false'` for applications that have finished the review process, have expire or were withdrawn by the Applicant.
```
CarlosNZ, line 17: Typo: "have **expired**"

documentation/Database-Schema-Application.md
```diff
@@ -0,0 +1,56 @@
...
+
+* <b>application</b>
+Representation of the application instance. All nested elements are accessible via joined tables and can be created or queried in the same call using the GraphQL engine.
+The `template_id` links to the template used to display all sections and elements (questions) to the user.
+The `user_id` links to the Applicant user that created this application.
+The `serial` is a number used to display the application number to users. 
+The `name` will be deduced by the template name + user + company names.
+The `is_active` is set to `'true'` for all applications that are still in the review process. 
+The `is_active` is set to `'false'` for applications that have finished the review process, have expire or were withdrawn by the Applicant. 
+The `trigger` is updated everytime the application has changes done by users or a scheduler. The trigger would be one of the options: `'onApplicationCreate'`, `'onApplicationSubmit'`, `'onApplicationSave'`, `'onApplicationWithdrawn'`, `'onReviewStart'`, `'onReviewEditComment'`, `'onReviewSave'`, `'onReviewAssign'`, `'onApprovalSubmit'`, `'onScheduleTime'`.
```
CarlosNZ, line 18: Agghh -- not more places to maintain the list of Triggers 🤣

See my comment on #24 -- maybe best to just say "See here for detailed list of available triggers" with a link to the Triggers & Actions page.

documentation/Database-Schema-Application.md
```diff
@@ -0,0 +1,56 @@
...
+The `outcome` is either `'Pending'`, `'Approved'`, `'Rejected'`. The application is 'Pending' during the review process, each stage will store also the outcome, so it will only be updated to here once the review process is finished resulting in either 'Approved' or 'Rejected'.
+<b>To be considered:</b>
+   * The `unique_identifier` is a new field. What is it used for? Should is replace the `serial`?
```
CarlosNZ, line 21: Yes, indeed, we certainly don't need a `serial no`, a `unique_id` _and_ an `id`!!

The idea was to allow the country to give each application a "label"/"code" according to their local style/convention. Maybe "id_string"? I dunno, pick one I guess?

documentation/Database-Schema-Application.md
```diff
@@ -0,0 +1,56 @@
...
+   * An application with Draft should be flagged as `is_active`?
```
CarlosNZ, line 22: Yes, I think so. Because the main filter we'll be applying at the front end (by default) will limit the display to Active applications, so you'd want the Drafts in that, I'd say.

documentation/Database-Schema-Application.md
```diff
@@ -0,0 +1,56 @@
...
+
+* <b>application_section</b>
+Sections of the application that contain some elements. This is exactly the same as the `template_section`.
+The `application_id` links to the application.
+The `template_section_id` links to the template.
+<b>To be considered: Why so we need an `application_section` and not `application_element`. We should use the same logic for both.</b>
```
CarlosNZ, line 28: Yes, good point. `application_response` (which is the equivalent of `template_element`) should be related to `application_section` rather than `application` to keep consistency between Template and Application.

You want to go ahead and change it, or will you make a new issue?

documentation/Database-Schema-Application.md
```diff
@@ -0,0 +1,56 @@
...
+The `application_id` links to the application.
+The `stage_id` links to the template stage, which will have the `name` and number of this stage.
+The `time_created` defined when this stage was created.
+The `is_current` is set to `'true'` while this is the current stage of this application. Otherwise is set to `'false'`.
+<b>To be considered: When the application review is not in progress anymore do we keep the last `is_active` stage as `'true'`?</b>
+
+* <b>application_status_history</b>
+For each stage of the application it can have associated statuses. The application status history will keep records for when the application changed to each particular status of a stage. 
+The `application_stage_history_id` links to the application stage (and the application itself).
+The `status` is either `'Draft'`, `'Withdrawn'`, `'Submitted'`, `'Changes Required'`, `'Re-submitted'`, `'Completed'` and self-explanatory,
```
CarlosNZ, line 41: I think you can lose the "and self-explanatory" from the end. If it's self-explanatory (which it is), why explain that it's self-explanatory? Just let it explain itself -- it's self-explanatory after all! 🤣

documentation/Database-Schema-Application.md
```diff
@@ -0,0 +1,56 @@
...
+The `time_created` defined when this status of a stage was created.
+The `is_current` is set to `'true'` while this is the current status of this application stage. Otherwise is set to `'false'`.
+
+### Question response
+For every new answer that the Applicant give to each resquested elements (questions) in the application a new <b>application response</b> is created. The Applicant user can make changes to responses until submitting or re-submitting the application to be reviewed. The
```
CarlosNZ, line 46: Typo: "requested"

CarlosNZ: Also, an orphan "The" at the end of that line. 🤷‍♂️

documentation/Database-Schema-Application.md
```diff
@@ -0,0 +1,56 @@
...
+The `status` is either `'Draft'`, `'Withdrawn'`, `'Submitted'`, `'Changes Required'`, `'Re-submitted'`, `'Completed'` and self-explanatory,
+The `time_created` defined when this status of a stage was created.
+The `is_current` is set to `'true'` while this is the current status of this application stage. Otherwise is set to `'false'`.
+
+### Question response
+For every new answer that the Applicant give to each resquested elements (questions) in the application a new <b>application response</b> is created. The Applicant user can make changes to responses until submitting or re-submitting the application to be reviewed. The 
+
+* <b>application_reponse</b>
+The `template_question_id` links to the template element, which defines what is the question and how to render in the application.
+The `application_id` links to the application.
```
CarlosNZ, line 50: And as discussed above, this with become `application_section_id` if we change it.

documentation/Database-Schema-Application.md
```diff
@@ -0,0 +1,56 @@
...
+The `value` is what the Applicant user enter as a response.
+The `time_created` defined when this response was created.
+More detailed description of response coming soon: `value`.
+<b>To be considered:</b>
+  * Should we change `template_question_id` to `template_element_id` to keep consistent with the table name?
```
CarlosNZ, line 55: Yes, I think so.

documentation/Database-Schema-Application.md
```diff
@@ -0,0 +1,56 @@
...
+  * How can we flag a question to be editable by the user when Changes are required if it's only defined per template in `template_element`? Considering what was also flagged in the `application_section` we might need another `application_question` table to deal with that.
```
CarlosNZ, line 56: Wouldn't this be defined by `review_response.review_decision`? If that value is "Observations" (and the associated Review has been submitted by reviewer), then this question become editable by Applicant.

CarlosNZ (changes requested):
Cool, just a few typos. Good points about the suggested changes -- I think they all make sense to me. You want to change them in this PR, or make a new issue?

database/buildSchema/18_application_stage_history.sql
nmadruga: Yea. No problem. I'll fix this.

documentation/Database-Schema-Application.md
nmadruga: Well spotted! I keep getting this wrong. : O

documentation/Database-Schema-Application.md
nmadruga: Yes, for sure! I'll start a document where we keep the possible useful functions to be added to postGres.

documentation/Database-Schema-Application.md
nmadruga: But wouldn't we use the `serial` field for that. I guess we could not use the serial and just have one `id_string` or `number` or `unique_identifier` and use an action to set this value. By default it can be a very simple thing and the country using it can add more specific stuff to the `identifierGenerator`.

documentation/Database-Schema-Application.md
nmadruga: Cool. I'll add this to the explanation of `is_active` field. Thanks

documentation/Database-Schema-Application.md
nmadruga: By what was discussed in today's meeting we will keep as it is.

documentation/Database-Schema-Application.md
nmadruga: hahahhaha. Yes. I agree. Don't need to explain the inexplainable.

documentation/Database-Schema-Application.md
nmadruga: Yes. As discussed in today's meeting. We would be using a SQL view or function to deduce this from the review decisions. 👍

documentation/Database-Schema-Application.md
```diff
@@ -0,0 +1,84 @@
...
+
+The `outcome` is either `'Pending'`, `'Approved'`, `'Rejected'`. The application is 'Pending' during the review process, each stage will store also the outcome, so it will only be updated to here once the review process is finished resulting in either 'Approved' or 'Rejected'.
+
+**To be considered:**
+
+- The `unique_identifier` is a new field. What is it used for? Should is replace the `serial`?
+
+### application_section
+
+Sections of the application. It will point to the `template_section` that stores all the elements (questions) to be displayed. We required to have this records to have those associated to reviews, which are done per section of each application.
```
CarlosNZ, line 37: > We required to have this records to have those associated to reviews

Can't quite make sense of this. Looks like you might have left too many words in. ;)

documentation/Database-Schema-Application.md
```diff
@@ -0,0 +1,84 @@
...
+
+### application_stage_history
+
+There is one or more stages per application template. Each one is defined by `template_stage`. The application stage history will keep records for when the application changed to each particular stage (previously defined for this application template).
+
+The `application_id` links to the application.
+
+The `stage_id` links to the template stage, which will have the `name` and number of this stage.
+
+The `time_created` defined when this stage was created.
```
CarlosNZ, line 51:
```suggestion
The `time_created` records when this stage was created.
```

CarlosNZ (changes requested):
Tiny little fixes. Cheers :)

database/insertData.js
```diff
@@ -407,17 +407,17 @@ const queries = [
...
                 timeCreated: "NOW()"
                 value: "{text: 'Smith'}"
-                templateElementToTemplateQuestionId: { connectById: { id: 3 } }
+                templateElementToTemplateElementId: { connectById: { id: 3 } }
               }
               {
                 timeCreated: "NOW()"
                 value: "{option: '1'}"
-                templateElementToTemplateQuestionId: { connectById: { id: 6 } }
+                templateElementToTemplateElementId: { connectById: { id: 6 } }
```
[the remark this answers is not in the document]
nmadruga: Oops! And I ran the thing. Couldn't see the problem. I'll fix now. Thanks!

documentation/Database-Schema-Application.md
nmadruga: Hahaha The classic too many words for a sentence. You know that Tolstoi really cared about that. All his sentences are well thought to have the minimum number of words required. : )

documentation/Functions.md
```diff
@@ -0,0 +1,5 @@
+# List of useful functions to be added to Database
```
CarlosNZ, line 1: Good idea :)

documentation/Database-Schema-Application.md
```diff
@@ -34,7 +34,7 @@ The `outcome` is either `'Pending'`, `'Approved'`, `'Rejected'`. The application
 
 ### application_section
 
-Sections of the application. It will point to the `template_section` that stores all the elements (questions) to be displayed. We required to have this records to have those associated to reviews, which are done per section of each application.
+Sections of the application. It will point to the `template_section` that stores all the elements (questions) to be displayed. We required to have this records to have those associated to reviews.
```
CarlosNZ, line 37:
```suggestion
Sections of the application. It will point to the `template_section` that stores all the elements (questions) to be displayed. This table is required as it provides a way to associate reviews (which are per-section) to specific application sections.
```

CarlosNZ (approved):
Sweeeeet :)

PR #37, opened 2020-09-22
#21 Action plugins spec

CarlosNZ:
First version of **action plugin** spec.

A few things to think about/discuss in **Needs consideration**

### DIFF
--- a/documentation/Action-plugin-specification.md
+++ b/documentation/Action-plugin-specification.md
@@ -11,7 +11,7 @@ Action plugins reside in `/src/plugins`. Each plugin is contained within its own
 Within a plugin's folder are the following files:
 
 - `jest.config.js` -- Contains automated tests for the plugin that run whenever `test` is run in the project.
-- `package.json` -- A standard `npm` package.json file which treats this plugin's folder as a distinct package.
+- `package.json` -- A standard `npm` package.json file which treats this plugin's folder as a distinct package. (more details in **Development** section below)
 - `tsconfig.json` -- Typescript configuration file. Shouldn't need to be changed.
 - `/src` -- folder containing plugin source code and metadata
 - `/src/plugin.json` -- contains metadata for the plugin. The system reads this file (at startup) to load all the relevant information about the plugin into the database, so make sure it is accurate and up-to-date. Further details below.
--- a/documentation/Action-plugin-specification.md
+++ b/documentation/Action-plugin-specification.md
@@ -4,6 +4,8 @@ Please see [Triggers & Actions](./Triggers-and-Actions.md) for an explanation of
 
 This document details the specification for the **Action plugins** themselves.
 
+See [here](./List-of-Action-plugins.md) for a list of currently implemented Action plugins.
+
 ## Location & Folder contents
 
 Action plugins reside in `/src/plugins`. Each plugin is contained within its own subfolder. The name of the folder does not matter in terms of behaviour, but should be named something consistent with existing plugins.
@@ -44,7 +46,7 @@ module.exports['consoleLog'] = function (parameters: any) {
 
 In this case `consoleLog` is the name of the function called by the Action module. Relevant code sits within the `try` block. The `catch (error)` block should be left as is (unless you wish to break it down into more specific errors) -- this ensures the `action_queue` table keeps a record of the success or failure of each action.
 
-All parameters are passed in as keys/values in the `parameters` object. The plugin specifies the names of the parameter fields it is expecting (in `plugin.json`, below). The application template that is associated with the  Action plugin should store [expressions/queries](./Query-Syntax.md) to generate the values for the parameter fields. These values are evaluated when the action is triggered and stored in the action_queue. The evaluated parameters are passed to the Action when its function is called.
+All parameters are passed in as keys/values in the `parameters` object. The plugin specifies the names of the parameter fields it is expecting (in `plugin.json`, below). The application template that is associated with the Action plugin should store [expressions/queries](./Query-Syntax.md) to generate the values for the parameter fields. These values are evaluated when the action is triggered and stored in the action_queue. The evaluated parameters are passed to the Action when its function is called.
 
 In the simple example above, the only parameter expected is `message`, which the function prints to the Console.
 
@@ -118,21 +120,18 @@ The `action_plugin` table is the primary record of what actions are available to
 
 Plugins are intended to be stand-alone units, in that a user should be able to simply copy a plugin folder to another system and have it work straight away. (Eventually, we envisage that plugins could be imported directly via the front-end UI.) To that end, each plugin is developed as its own package, with its own `package.json` and `node_modules` folder, as well as Typescript and testing configurations.
 
-While developing a plugin, you can run the main app with:
-
-`yarn dev` 
+While developing a plugin, you can run the main app with:  
+`yarn dev`
 
 Any changes you make to the plugin `.ts` file will be reflected immediately in the dev environment.
 
-After development work is complete, the plugin should be built independently by running _in the plugin's root folder_, which will compile the typescript code into `.js` file(s) with:
+After development work is complete, the plugin should be built independently by running the following command _in the plugin's root folder_, which will compile the typescript code into `.js` file(s):  
+ `yarn build`
 
- `yarn build` 
- 
-Alternatively to build all plugins from the project root folder with:
-
+Alternatively, to build _all_ plugins from the project root folder:  
 `yarn build_plugins`
 
-When the main project is built (`yarn build`), the plugins are not re-compiled, but simply copied directly to the build folder, so its expected they will each have their own compiled `.js` file already in place.
+When the main project is built (`yarn build`), the plugins are not re-compiled, but simply copied directly to the build folder, so it is expected they will each have their own compiled `.js` file already in place.
 
 ## Needs consideration
 
@@ -141,7 +140,3 @@ When the main project is built (`yarn build`), the plugins are not re-compiled, 
 - Should there be (optional) **default values** for the `required_parameters`? I'm thinking, for things like "Send notification", the user's email address will be required, but it's always going to be the same query (`{operator: "objectProperties", children: [{value: {object: "user", property: "email"}}]}`) so it seems like it should there by default so the Admin doesn't have to specify an obvious value in the Template Builder.
 
 - Currently, parameters are evaluated immeditely before the Action is saved to the `action_queue`. For immediate actions, this is fine, but for Scheduled Actions, can we imagine a use case where it would be better for them to be evaluated at the time they run (which could be much later, even years)?
-
-## List of Action plugins created to-date:
-
-- **Console Log**: Just prints a message to the console. For demo purposes only.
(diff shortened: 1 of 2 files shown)


documentation/Action-plugin-specification.md
```diff
@@ -0,0 +1,133 @@
...
+- `package.json` -- A standard `npm` package.json file which treats this plugin's folder as a distinct package. (more details in **Development** section [below](#development))
+- `tsconfig.json` -- Typescript configuration file. Shouldn't need to be changed.
+- `/src` -- folder containing plugin source code and metadata
+- `/src/plugin.json` -- contains metadata for the plugin. The system reads this file (at startup) to load all the relevant information about the plugin into the database, so make sure it is accurate and up-to-date. Further details [below](#plugin).
+- `/src/<plugin-name>.ts` -- Typescript source code
+- `/src/<plugin-name>.js` -- Javascript code compiled from Typescript. This is what is actually loaded and run in the built server app.
+
+## Source code file
+
+The main source code for the plugin is simply an exported function that the server app dynamically loads via `require` at run-time. Plugin code should follow this basic structure:
```
nmadruga, line 23:
```suggestion
The main source code for the plugin is simply an exported function that the server app dynamically loads via `require` at run-time. Calling the function exported by each action plugin, sequentially is the job of the **Action module**. 

Plugin code should follow this basic structure:
```

documentation/Action-plugin-specification.md
```diff
@@ -0,0 +1,133 @@
...
+      status: 'Fail',
+      error_log: 'There was a problem',
+    }
+  }
+}
+```
+
+In this case `consoleLog` is the name of the function called by the Action module. Relevant code sits within the `try` block. The `catch (error)` block should be left as is (unless you wish to break it down into more specific errors) -- this ensures the `action_queue` table keeps a record of the success or failure of each action.
+
+All parameters are passed in as keys/values in the `parameters` object. The plugin specifies the names of the parameter fields it is expecting (in `plugin.json`, below), and the template associated with this Action stores [expressions/queries](./Query-Syntax.md) to generate the values. These values are evaluated when the action is triggered and stored in the action_queue. The evaluated parameters are passed to the Action when its function is called.
```
nmadruga, line 45:
```suggestion
All parameters are passed in as keys/values in the `parameters` object. The plugin specifies the names of the parameter fields it is expecting (in `plugin.json`, below). The application template that is associated with the  Action plugin should store [expressions/queries](./Query-Syntax.md) to generate the values for the parameter fields. These values are evaluated when the action is triggered and stored in the action_queue. The evaluated parameters are passed to the Action when its function is called.
```

nmadruga: I trie to add more words to it. But don't know how to make it clearer. Maybe it's not clear to me yet:

The plugin define which is the object for parameters and fields expected. The application template has queries to get this values from the database or local data. Then this fields/values get evaluated by the evaluator?

Maybe a diagram would be handy here.

documentation/Action-plugin-specification.md
```diff
@@ -0,0 +1,133 @@
...
+
+```
+{
+  "type": "action_plugin",
+  "code": "cLog",
+  "name": "Console Logger",
+  "file": "consoleLog",
+  "function_name": "consoleLog",
+  "description": "All it does is print a message to the console. That's it.",
+  "required_parameters": ["message"],
```
nmadruga, line 65: Should we define also what is the type expected for each field?

documentation/Action-plugin-specification.md
```diff
@@ -0,0 +1,133 @@
...
+- `function_name`
+- `required_parameters`
+
+<a name="development"></a>
+
+## Development and build process
+
+Plugins are intended to be stand-alone units, in that a user should be able to simply copy a plugin folder to another system and have it work straight away. (Eventually, we envisage that plugins could be imported directly via the front-end UI.) To that end, each plugin is developed as its own package, with its own `package.json` and `node_modules` folder, as well as Typescript and testing configurations.
+
+When developing a plugin, you can run the main app with `yarn dev` as normal, and changes you make to the plugin `.ts` file will be reflected immediately in the dev environment.
```
nmadruga, line 119:
```suggestion
While developing a plugin, you can run the main app with:

`yarn dev` 

Any changes you make to the plugin `.ts` file will be reflected immediately in the dev environment.
```

documentation/Action-plugin-specification.md
```diff
@@ -0,0 +1,133 @@
...
+
+However, when development work is complete, the plugin should be built independently by running `yarn build` _in the plugin's root folder_, which will compile the typescript code into `.js` file(s). (There is also a shortcut `yarn build_plugins` to build all plugins from the project root.) When the main project is built (`yarn build`), the plugins are not re-compiled, but simply copied directly to the build folder, so its expected they will each have their own compiled `.js` file already in place.
```
nmadruga, line 121:
```suggestion
After development work is complete, the plugin should be built independently by running _in the plugin's root folder_, which will compile the typescript code into `.js` file(s) with:

 `yarn build` 
 
Alternatively to build all plugins from the project root folder with:

`yarn build_plugins`

When the main project is built (`yarn build`), the plugins are not re-compiled, but simply copied directly to the build folder, so its expected they will each have their own compiled `.js` file already in place.
```

documentation/Action-plugin-specification.md
```diff
@@ -0,0 +1,133 @@
...
+
+## Needs consideration
+
+- Should the `required_parameters` specify an expected type so that the Template Builder UI knows what kind of options to present?
```
nmadruga, line 125: Yes. I think so, we should give errors from what is selected to be queried for each field checking if the final result is the same as the expected.

documentation/Action-plugin-specification.md
```diff
@@ -0,0 +1,133 @@
...
+
+- Should there be (optional) **default values** for the `required_parameters`? I'm thinking, for things like "Send notification", the user's email address will be required, but it's always going to be the same query (`{operator: "objectProperties", children: [{value: {object: "user", property: "email"}}]}`) so it seems like it should there by default so the Admin doesn't have to specify an obvious value in the Template Builder.
```
nmadruga, line 127: Yes. Indeed. I'm just trying to figure how we can check if the default values are correct. When the person developing the plugin we would give some sort of error as: `"this objectProperty is not defined in this context"` But I agree that those very basic objects should be defined wiht a default value. And in case that is changed it would just be passed in the `parameters` by something defined in the application template queries, right?

documentation/Action-plugin-specification.md
```diff
@@ -0,0 +1,133 @@
...
+
+- Currently, parameters are evaluated immeditely before the Action is saved to the `action_queue`. For immediate actions, this is fine, but for Scheduled Actions, can we imagine a use case where it would be better for them to be evaluated at the time they run (which could be much later, even years)?
```
nmadruga, line 129: I think it makes all the sense to be evaluated at the time they will run. Otherwise the parameters might have changed. And **I read somewhere** about us storing the queried values so when the action runs it doesn't need to get it from the database. So that enforces even more, the evaluation needs to be right before the action runs.

Any reason not to?

documentation/Action-plugin-specification.md
```diff
@@ -0,0 +1,133 @@
...
+
+## List of Action plugins created to-date:
+
+- **Console Log**: Just prints a message to the console. For demo purposes only.
```
nmadruga, line 133: I think this would be better on a separated file. Just so it's easier to find and easier to refer.

nmadruga (changes requested):
Very good to have this. Thanks for working on it.

I have added a few suggestions. Not a big deal, but I think it would make the meaningful references easier to see. I think the workflow on how the parameters end up getting to the plugins would need a sort of diagram. It's kind of hard to get from only the explanation. I can help with that, but just tought about flagging this here.

documentation/Action-plugin-specification.md
CarlosNZ: Yes, I mentioned that in the **Needs consideration** section.

documentation/Action-plugin-specification.md
CarlosNZ: I think the main purpose of the "type" here would be so the UI can know what kind of options to make available. So, for example, if the required parameter was `email_address`, the expected type might be "email address", so then the UI in the Template Builder might present these options:
- user's email address (saves query to lookup `user.email`)
- other email (user fills in a free-text string, which is saved as a literal value.

And then there would be mappings stored somewhere that link "types" to "presentation options", or something like that. 

We should probably have a discussion around this stuff soon to get some clarification before getting in too deep.

documentation/Action-plugin-specification.md
CarlosNZ: I think plugin developers (i.e. us! for now) should be expected to handle this correctly. But yes, an invalid query should return something like what you suggested.

And yeah, the "default" value would be the "placeholder" value shown in the Template Builder UI when selecting this Action.

documentation/Action-plugin-specification.md
CarlosNZ: Yeah, I guess we just need to try and think of use cases and see which is more likely. 

In terms of code, they're currently evaluated by the Trigger Queue processor and the evaluated results saved into the action queue.

To do it the other way, it makes the scheduled jobs code somewhat more complicated, but since we're not yet sure the best way to handle scheduled actions (I have an prototype, but I'm not sure it's the best way), we should probably have a discussion about this before making new changes.

documentation/Action-plugin-specification.md
CarlosNZ: Sure -- have made a new page for them.

documentation/Action-plugin-specification.md
nmadruga: Yes. It would be better to add as part of next meeting. : )

documentation/Action-plugin-specification.md
nmadruga: Ok. I agree to leave this one for now as a **To be considered** here

nmadruga (approved):

documentation/Action-plugin-specification.md
```diff
@@ -0,0 +1,142 @@
...
+      status: 'Fail',
+      error_log: 'There was a problem',
+    }
+  }
+}
+```
+
+In this case `consoleLog` is the name of the function called by the Action module. Relevant code sits within the `try` block. The `catch (error)` block should be left as is (unless you wish to break it down into more specific errors) -- this ensures the `action_queue` table keeps a record of the success or failure of each action.
+
+All parameters are passed in as keys/values in the `parameters` object. The plugin specifies the names of the parameter fields it is expecting (in `plugin.json`, below). The application template that is associated with the Action plugin should store [expressions/queries](./Query-Syntax.md) to generate the values for the parameter fields. These values are evaluated when the action is triggered and stored in the action_queue. The evaluated parameters are passed to the Action when its function is called.
```
nmadruga, line 49: Maybe the only thing to add here would be a diagram to show exactly what is described here. Is just hard to understand the way it is, and it might be helpful to further conversation on this topic.

CarlosNZ: Okay, I'll have a think about how to best present it.

PR #38, opened 2020-09-23
#35 Template schema updates

### LINKED ISSUE #35: Schema -- update Template/Version table structure as per discussion
The problem with the current schema is that there is confusion around the `template` and `template_version` tables. It wasn't clear which one's record gets duplicated when a new "version" is created, and the naming was quite ambiguous. After discussion we found a solution that simplifies the schema and makes the naming of tables and their relations clear.

Agreed `template` sections schema:
Templates and versions-agreed

When a template is edited/modified a new "copy" of it is stored in the `template` table with a new `timestamp` (and any other changed fields). The only field that cannot be changed is `code`, which is how different versions of the same template are connected. Most of the time the application is only going to be interested in template records which have `status = Available` -- this is the currently "active" version of the template.

### To Do
- Edit the `buildSchema` scripts to represent the above structure
- Update `insertData` so the templates and applications conform to modified schema
- Update documentation accordingly.

CarlosNZ:
Fixes #35, removing the `template_version` table.

Note: even though I have removed `06_template_version.sql`, I have not yet renumbered the rest of the files, as I'm concerned that doing so will make a mess when trying to reconcile merge conflicts with @nmadruga 's changes in other PRs (eg #25 ). Let's wait until all database-changing PRs are merged to master and then deal with renumbering?

Agreed `template` schema:
Templates and versions-agreed


database/buildSchema/07_template.sql
```diff
@@ -4,8 +4,8 @@ CREATE TYPE public.template_status AS ENUM ('Draft', 'Available', 'Disabled');
 
 CREATE TABLE public.template (
     id serial primary key,
-    version_id integer references public.template_version,
     name varchar,
-    code varchar NOT NULL,
-    status public.template_status
+    code varchar,
```
CarlosNZ, line 8:
```suggestion
code varchar NOT NULL,
```

nmadruga (approved):

nmadruga:
@CarlosNZ Just need to resolve the conflicts before we merge this one with the `#16-schema-documentaion-template-group` branch.

CarlosNZ:
> @CarlosNZ Just need to resolve the conflicts before we merge this one with the `#16-schema-documentaion-template-group` branch.

Okay, I'll wait till you merge #16 into `master` and then resolve the conflicts.

nmadruga:
@CarlosNZ it's merged now.

PR #41, opened 2020-09-23
#14 schema documentation user group

### LINKED ISSUE #14: Schema Documentation - User/Organisation
User/Organisation

nmadruga:
Fixes #14 

Basic documentation for users. Still missing a lot of fields for the user and organisation table, but the idea is to be incremental, specially for this 2 since we are currently working with the code to create users and companies and more fields would be more difficult to work with. Anyway a few things **to be be considered**. One that is not added to the documentation - which I can add is should we have a link between user/organisation with files uploaded with profs of identity, address and that kind of thing?

## Database changes:

* Removes `user.role` => Since this can be more than one (e.g. Same user can be Reviewer and Applicant depending on template). What the user can do is basically only defined by its permissions.
* Rename `user_organisation.job` to `user_organisation.user_role` to be used as something that can be displayed to the user when they are logged in that company. What is the user role can be defined either in the application the user applies to company or deduced by the permissions the user have in the company (so when a permission is revoke we can update the user_role). Please, few free to suggest if this information needs to be clearer in the documentation on this PR.


documentation/Database-Schema-User.md
```diff
@@ -0,0 +1,42 @@
+# Database Schema User
+
+![Database Schema](images/database-schema-user.png)
+
+# Database Area description: User
+
+## User instance
+
+An user is either the Applicant that wants to apply to a application (Registration) in the system or the other entities (i.e. Reviewer) that will be the evaluator of the workflow of the application. What will define exact the type of user are permissions (permission_join) which is associated with the user and the organisation of the user. So the same user may have different actions available depening on the company selected on the login.
```
CarlosNZ, line 9:
```suggestion
A user is either the Applicant that wants to apply for an application (Registration) in the system or the other entities (i.e. Reviewer) that will evaluate the application. What will define exact the type of user are permissions (`permission_join`) which is associated with the user and the organisation of the user. So the same user may have different operations available depending on the company selected on login.
```

documentation/Database-Schema-User.md
```diff
@@ -0,0 +1,42 @@
...
+- A few more fields will be required in the user table, to be added as needed.
+- The `license_number` could be stored in a different table that have the `organisation_id`, the license number, expiry and if it's valid. This way when a company license is renewed we can create a new entry and link to the company, keeping track of old licenses.
+
+### user organisation
+
+The `user_id` links to the user table.
+
+The `organisation_id` links to the organistion.
+
+The `user_role` is just a literal for now that states what is this user role in the organisation. It is created based on what permissions the user has to act as the organisation. For example a user that creates the company in the system can be the **company owner** and another user that joins the compnay by default can be the **employee**. Other example is the user that is part of FDA and when associated with the main organisation in the system the default job is **revisor**.
```
CarlosNZ, line 42:
```suggestion
The `user_role` is just a literal for now that states this user's role in the organisation. It is created based on what permissions the user has to act as the organisation. For example a user that creates the company in the system can be the **Company Owner** and another user that joins the company by default can be an **Employee** or **Member**. Another example is the user that is part of FDA and when associated with the main organisation in the system the default job is **Reviewer**.
```

CarlosNZ (changes requested):
Sweet, just minor typo things really, plus one suggested database field type change.

documentation/Database-Schema-User.md
```diff
@@ -0,0 +1,42 @@
...
+
+**To be considered:**
+
+- A few more fields will be required in the user table, to be added as needed.
+
+### organisation
+
+The `name` is how the company/organisation is defined in the system.
+
+The `license_number` is a number generated by the system and stored in the organisation. It shoudl also have a expiry.
```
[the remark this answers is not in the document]
nmadruga: Yes. Indeed. I have changed on `04_organisation.sql` file.

I think we will probably need more than this to be stored - for example the `application_id` where all the details for a company approved is stored. But I think this can be added when we get to the Epic to add the "Create company" application.

CarlosNZ (approved):
👍

Issue #40, opened 2020-09-23
Add plugin in order to allow nested conditions


CarlosNZ:
This plugin required:
https://github.com/graphile-contrib/postgraphile-plugin-connection-filter

PG docs: https://www.graphile.org/postgraphile/filtering/

Also: migrate command line options into `.postgraphilerc.js`

Issue #2, opened 2020-09-08
Serve postGraphile endpoint through Fastify


CarlosNZ:
Even though we can run postGraphile as a separate server in its own process (the current setup), it will probably be better in the long run to serve it through Fastify so we only have the one server to interact with.

I suggest using [middie](https://github.com/fastify/middie) as middleware for registering postGraphile as a Fastify plugin.

Other advantages to incorporating it into Fastify server:
- The terminal command for launching postgraphile is getting increasingly verbose an will continue to do so as we add more functionality and configurations. It would be easier to read and manage if each plugin was part of the main codebase in a structured manner.
- Some plugins (such as [`postgraphile-plugin-nested-mutations`](https://github.com/mlipscombe/postgraphile-plugin-nested-mutations) look like they need to have postGraphile running as a library in order to use the full range of options. For example, the `nestedMutationsSimpleFieldNames` (which would be useful) can only be enabled in this way (as far as I can tell).

PR #42, opened 2020-09-24
#40 Advanced Filtering plugin

CarlosNZ:
Addresses #40 

Simple changes:
- moved most of PostGraphile's command line options into a seperate `.postgraphilers.js` file, as described [here](https://www.graphile.org/postgraphile/usage-cli/#rc-file-options). Main reason for doing this is that the connection-filter plugin (below) required the `connectionFilterRelations: true` flag set in order to allow filtering through relations, and there didn't seem to be a way to add flags to plugins via the command line.
- added [connection-filter plugin](https://github.com/graphile-contrib/postgraphile-plugin-connection-filter) as suggested [here](https://www.graphile.org/postgraphile/filtering/#filter-plugin). This enables the query filter to dive deeper into the relation tree, so we can do queries like "Return all applications with template name = 'Drug Registration'". Without this, you could only filter by TemplateID, and we're definitely going to be wanting queries like this (in the Application List for example).  
The plugin adds an additional `filter` argument to the schema's connections, which open up a _huge_ range of additional query options.

Note: this plugin does raise some [security considerations](https://github.com/graphile-contrib/postgraphile-plugin-connection-filter#performance-and-security) so we should make note to apply some of the suggested remediations (such as whitelisted queries) when nearing production release.


nmadruga (approved):
Great. And the config looks so much better with the .postgraphile config file.

Do you think it's worth it adding the issue #2 to be done now so we have this all tidy up?

CarlosNZ:
> Do you think it's worth it adding the issue #2 to be done now so we have this all tidy up?

Maybe, although @andreievg doesn't seem keen to rush that for now. I think if it can be done in less than a couple of hours it'd be worth it, but we can do without it for now.

PR #48, opened 2020-09-29
#30 database abstraction layer

### LINKED ISSUE #30: Database Abstraction Layer
Before starting on this issue, it might pay to check out this library, as it may save us the trouble of creating our own abstraction layer:  https://vincit.github.io/objection.js/

nmadruga:
Fixes #30 

[EDITED]
## Changes
Remove the Database code to be encapsulated on singleton component `postgresConnect.js`. It uses a pool connection for each query. Just uses the client now for adding the notification for triggers. 
Have also added many types for each payload and returned object for queries. Not sure how to go with that, and to make it in sync with GraphQL types. There are duplications so consider if we can use the generated/graphql.ts with types here.
Any suggestion for variable/types names will be very welcome, I haven't think too much about it during development. 

## Tests:
- [ ] Import plugins
- [ ] Run trigger see existing simple action checking result - it means the plugins are getting uploaded and triggered correctly
- [ ] Change action to use pgSQL operation and run trigger to check result - it means that the `evaluateExpression` still works 
- [ ] Upload a file - check that the file has been correctly uploaded (not sure how to do this).


src/modules/evaluateExpression/evaluateExpression.ts
```diff
@@ -1,45 +1,9 @@
...
-  | '+'
-  | 'REGEX'
-  | 'objectProperties'
-  | 'pgSQL'
-  | 'graphQL'
-
-const defaultParameters: IParameters = {
-  connection: {},
-}
+import { IQueryNode, IParameters } from '../../types'
```
CarlosNZ, line 1: Can we NOT put the types for this module into the project `types` file, since we want to extract this as its own package at some point? Maybe make a separate `types` file in this modules folder?

src/modules/evaluateExpression/evaluateExpression.ts
```diff
@@ -1,45 +1,9 @@
...
+import PostgresDB from '../../components/postgresConnect'
```
CarlosNZ, line 2: I think having the Connection object passed in as a parameter is better for this module. Because if its being used without PostGres, there's no need to import or install any packages. For example, when the front-end uses this module, we don't want to have to have a post-gres package installed as we'll never use it. So the connection should be handled outside this module and just passed in if/when required.

src/modules/evaluateExpression/evaluateExpression.ts
nmadruga: Sure. Indeed. I was considering putting every type inside this file, but you are right this will be divided in different projects/packages soon. I'll move to a separated file inside the same module : )

src/modules/evaluateExpression/evaluateExpression.ts
nmadruga: Ah, that's a good point. Sorry I wasn't seen that. : ) Will revert this right away.

nmadruga (commented):
I think this can be reviewed again now. I'll also updated the PR description : )

One thing that is not well tested yet is the pgSQL `operation` I tried but I think I'm not structuring it correctly, What I was trying was to edit the existing action (using pgAdmin) to this:
`UPDATE template_action SET "condition" = '{"operator":"=","children":[{"value":3},{"operator":"pgSQL","children":[{"text":"SELECT count(*) FROM application"},{"values":[]}]}]}' WHERE id = 1`
And then running the trigger:
`UPDATE application SET "trigger" = 'onApplicationSubmit' WHERE template_id = 2`

See in the console that the `childrenResolved` are getting undefined. I'm not sure how to structure that, maybe we need more details on the documentation.

documentation/Tests.md
```diff
@@ -0,0 +1,11 @@
+# List of useful tests to be added
+
+- [ ] Upload file:
+  - [ ] Succesful added
+  - [ ] Connection error
+  - [ ] Permission error
+  - [ ] Overwrite error
```
CarlosNZ, line 7: Good idea to add these tests.
As it's currently implemented, it shouldn't overwrite anything, since the filename is appended with the Database ID on upload. Not sure that this is the best way, but it seemed like a reasonable, simple way of ensuring filename uniqueness.

CarlosNZ:
Hey, I'm getting this error when I start the server. It outputs the following 3 times (one for each plugin its trying to scan):
```
[0] error: bind message supplies 9 parameters, but prepared statement "" requires 6
[0]     at Parser.parseErrorMessage (/Users/carl/Github/openmsupply-application-manager/application-manager-server/node_modules/pg-protocol/src/parser.ts:357:11)
[0]     at Parser.handlePacket (/Users/carl/Github/openmsupply-application-manager/application-manager-server/node_modules/pg-protocol/src/parser.ts:186:21)
[0]     at Parser.parse (/Users/carl/Github/openmsupply-application-manager/application-manager-server/node_modules/pg-protocol/src/parser.ts:101:30)
[0]     at Socket.<anonymous> (/Users/carl/Github/openmsupply-application-manager/application-manager-server/node_modules/pg-protocol/src/index.ts:7:48)
[0]     at Socket.emit (events.js:315:20)
[0]     at Socket.EventEmitter.emit (domain.js:485:12)
[0]     at addChunk (_stream_readable.js:302:12)
[0]     at readableAddChunk (_stream_readable.js:278:9)
[0]     at Socket.Readable.push (_stream_readable.js:217:10)
[0]     at TCP.onStreamRead (internal/stream_base_commons.js:186:23)
```

src/components/fileHandler.ts
```diff
@@ -3,8 +3,8 @@ import fs from 'fs'
 import util from 'util'
 import { pipeline } from 'stream'
 import getAppRootDir from './getAppRoot'
-import { pgClient } from './postgresConnect'
 import * as config from '../config.json'
+import PosgresDB from '../components/postgresConnect'
```
CarlosNZ, line 7: Spelling on Import module -- should be `PostgresDB` ?

src/modules/evaluateExpression/evaluateExpression.ts
```diff
@@ -93,13 +56,17 @@ export default async function evaluateExpression(
...
+        if (Object.entries(params).length === 0)
+          return 'No parameters received for objectProperties node'
         try {
           return params[childrenResolved[0].object][childrenResolved[0].property]
         } catch {
           return "Can't resolve object"
         }
 
       case 'pgSQL':
+        console.log('pgSQL', childrenResolved)
```
CarlosNZ, line 68: Remove `console.log()` ;)

CarlosNZ (changes requested):
Hey, didn't do a proper review here -- will put it through its paces properly once we've figured out what that startup error is all about. 

Just a couple of tiny things here.

CarlosNZ:
Also -- when I run set the Trigger, which should launch the Console logger Action, it gives this error:
```
[0] Error: Query values must be an array
[0]     at Query.submit (/Users/carl/Github/openmsupply-application-manager/application-manager-server/node_modules/pg/lib/query.js:153:14)
[0]     at Client._pulseQueryQueue (/Users/carl/Github/openmsupply-application-manager/application-manager-server/node_modules/pg/lib/client.js:472:45)
[0]     at Client.query (/Users/carl/Github/openmsupply-application-manager/application-manager-server/node_modules/pg/lib/client.js:567:10)
[0]     at PostgresDB.<anonymous> (/Users/carl/Github/openmsupply-application-manager/application-manager-server/src/components/postgresConnect.ts:66:27)
[0]     at step (/Users/carl/Github/openmsupply-application-manager/application-manager-server/src/components/postgresConnect.ts:52:23)
[0]     at Object.next (/Users/carl/Github/openmsupply-application-manager/application-manager-server/src/components/postgresConnect.ts:33:53)
[0]     at fulfilled (/Users/carl/Github/openmsupply-application-manager/application-manager-server/src/components/postgresConnect.ts:24:58)
[0]     at processTicksAndRejections (internal/process/task_queues.js:97:5)
```

nmadruga:
@CarlosNZ I have fixed those errors. Please have another look in this PR when you have time today : )

For some queries that I'm just sending the array of values coming from an object -> What do you think of crating a function to this with the object passed? `connection.query('UPDATE user SET ? WHERE ?', [{ Name: name }, { UserId: userId }])`. This would be to fix the possible case that the values are in a different order then what is expected in the query.

database/buildSchema/22_action_plugin.sql
```diff
@@ -1,12 +1,14 @@
 -- action plugins
 
 CREATE TABLE public.action_plugin (
---    id serial primary key,
     code varchar primary key,
     name varchar,
+    type varchar,
     description varchar,
     path varchar,
+    file varchar,
```
CarlosNZ, line 9: Why do we need the `file` field? The `path` field already contains the filename in it (it created using `folder + path` when loaded), so we're just storing the same thing twice.

It should either be `folder_path` and `file` which we combine at run-time, or just `path` which contains the full path. I'd opt for the latter, since the full path is the actual relevant info and only ever used in its joined state.

documentation/Database-Schema-Template.md
```diff
@@ -88,14 +88,22 @@ The actions logic are defined inside action plugins, which execute a function to
 
 The `code` is unique per action plugin.
 
+The `type` is storing the plugin type. For now always `'action_plugin'`.
+
 The `name` and `description` are for visualisation to help Admin users selecting what action to use for an application template.
 
 The `path` is the local path in the server where the imported plugin is stored.
 
+The `file` is the name of the file where the function are defined.
```
CarlosNZ, line 97: See file comment above -- `path` already includes this.

src/components/fileHandler.ts
CarlosNZ, line 7: Spelling -- `PostgresDB` ?

src/components/postgresConnect.ts
```diff
@@ -1,45 +1,210 @@
...
+      // Make sure to release the client before any error handling,
+      // just in case the error handling itself throws an error.
+      client.release()
+    }
+  }
+
+  public addActionQueue = async (action: ActionQueuePayload) => {
+    try {
+      const result = await this.query(
+        `INSERT into action_queue (${Object.keys(
```
CarlosNZ, line 80: I guess the reason you're building this query string yourself rather than passing substitution parameters is because the substitutions are arrays rather than strings?

It's highly recommended to not do this, as the built-in parameter parsing is much more robust against SQL injection. I think you should be able to pass arrays as parameters -- node-pg will convert them to Postgres arrays. (See the "parameterized queries" section here: https://node-postgres.com/features/queries)

You might need to use this module if you can't do it with the current tools: https://www.npmjs.com/package/pg-format

src/components/postgresConnect.ts
```diff
@@ -1,45 +1,210 @@
...
-    case 'trigger_notifications':
-      processTrigger(pgClient, JSON.parse(msg.payload))
-      break
-    case 'action_notifications':
-      executeAction(pgClient, JSON.parse(msg.payload), actionLibrary)
-      break
+  public addFile = async (payload: FilePayload): Promise<Pick<File, 'id'>> => {
+    try {
+      const result = await this.query(
+        `INSERT INTO file (${Object.keys(payload)}) VALUES () RETURNING id`,
```
CarlosNZ, line 118: Parameterized queries if possible.

src/components/postgresConnect.ts
```diff
@@ -1,45 +1,210 @@
...
+      return result.rows[0] as File
+    } catch (err) {
+      return err.stack
+    }
+  }
+
+  public addActionPlugin = async (plugin: ActionPluginPayload) => {
+    try {
+      await this.query(
+        `INSERT INTO action_plugin (${Object.keys(plugin)}) VALUES (${this.getValuesPlaceholders(
```
CarlosNZ, line 142: Parameterized queries

src/components/postgresConnect.ts
```diff
@@ -1,45 +1,210 @@
...
+    try {
+      await this.query('DELETE FROM action_plugin WHERE code = $1', Object.values(payload))
+    } catch (err) {
+      console.log(err.stack)
+    }
+  }
+
+  public getActionPlugins = async (): Promise<ActionPlugin[]> => {
+    try {
+      const result = await this.query('SELECT * FROM action_plugin', [])
```
CarlosNZ, line 162: Is the empty array necessary? (I thought it you just pass a single query parameter if there are no parameter substitutions).

src/components/postgresConnect.ts
```diff
@@ -1,45 +1,210 @@
...
+    tableName: string,
+    payload: ActionInTemplateGetPayload
+  ): Promise<ActionInTemplate[]> => {
+    try {
+      const result = await this.query(
+        `SELECT action_plugin.code, action_plugin.path, action_plugin.name, trigger, condition, parameter_queries 
+          FROM template 
+         JOIN template_action ON template.id = template_action.template_id 
+         JOIN action_plugin ON template_action.action_code = action_plugin.code 
+         WHERE template_id = (SELECT template_id FROM ${tableName} WHERE id = $1) AND trigger = $2`,
```
CarlosNZ, line 179: This `${tableName}` substitution is the one place where the built-in parameter parser won't work. I had this problem too, but I now see that this is expected behaviour:

> PostgreSQL does not support parameters for identifiers. If you need to have dynamic database, schema, table, or column names (e.g. in DDL statements) use pg-format package for handling escaping these values to ensure you do not have SQL injection!

It's probably not worth installing `pg-format` for just this one thing (and it's a read query, so not particularly risky), but if you can't make the earlier substitutions work with proper parameterized queries, I'd suggest that it probably is worth it: https://www.npmjs.com/package/pg-format

src/components/triggersAndActions.ts
```diff
@@ -1,137 +1,130 @@
 import path from 'path'
 import getAppRootDir from './getAppRoot'
 import * as config from '../config.json'
-import { ActionLibrary, Action, DatabaseResult, TriggerPayload, ActionPayload } from '../types'
+import { ActionLibrary, ActionInTemplate, TriggerPayload, ActionPayload } from '../types'
 import evaluateExpression from '../modules/evaluateExpression/evaluateExpression'
-import { Client } from 'pg'
+import PosgresDB from '../components/postgresConnect'
```
CarlosNZ, line 6: Spelling `PostgresDB` ! (and all references, obv) ;)

src/components/triggersAndActions.ts
```diff
@@ -1,137 +1,130 @@
...
+      }
+    } catch (err) {
+      console.log(err.stack)
+      return
+    }
+  }
+
+  actionSchedule.length > 0
+    ? console.log(`${actionSchedule.length} scheduled jobs loaded.`)
+    : console.log('There was no jobs to be loaded.')
```
CarlosNZ, line 79:
```suggestion
: console.log('There were no jobs to be loaded.')
```

CarlosNZ (changes requested):
Great, thanks for doing this and troubleshooting it all.

Just a few nits, typos, and I'd recommend making sure you use parameterized queries as much as possible.

I'm just posting this review now so you can see it, but I'm going to run a few more substantial tests soon too. Will report back if I discover any issues.

src/modules/evaluateExpression/evaluateExpression.ts
```diff
@@ -1,46 +1,11 @@
...
-
-const defaultParameters: IParameters = {
-  connection: {},
-}
+import { IConnection, IQueryNode, IParameters } from './types'
 
 export default async function evaluateExpression(
   inputQuery: IQueryNode | string,
-  params = defaultParameters
+  params: IParameters
```
CarlosNZ, line 5: Ah, this seems to have broken the `evaluateExpression` test suite. Gets the error ` Expected 2 arguments, but got 1.` for all tests. I think the `default_parameters` was supplying an empty connection object, which is now missing.

I suggest don't worry about this for now. I'm going to re-write the input parameter structure anyway, when I revisit the `objectProperties` operator, so I can fix it with that.

CarlosNZ:
## Tests:
- [x] Import plugins
- [x] Run trigger see existing simple action checking result - it means the plugins are getting uploaded and triggered correctly
- [ ] Change action to use pgSQL operation and run trigger to check result - it means that the `evaluateExpression` still works [evaluateExpression test suite is broken, so will re-test after fix (see previous comment inline with code) 
- [x] Upload a file - check that the file has been correctly uploaded (not sure how to do this). -- _Use Postman -- in the Body, add the key `file` type "file", with value the uploaded file. (Can also add other parameters according to the API)_
- [ ] Download a file -- **FAIL** doesn't seem to work. Will investigate and report back.

src/components/fileHandler.ts
nmadruga: Yes. Thanks!

database/buildSchema/22_action_plugin.sql
```diff
@@ -1,12 +1,14 @@
 -- action plugins
 
 CREATE TABLE public.action_plugin (
---    id serial primary key,
     code varchar primary key,
     name varchar,
+    type varchar,
```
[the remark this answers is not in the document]
nmadruga: This is coming from the `plugin.json` config file. Since the INSERT now accepts all fields that are part of the plugin this was one of them. I also though it was odd to have that there. I think we have 2 options:
1. Remove this type form the plugin.json file.
2. Remove the type form the plugin before sending it to the DB.

database/buildSchema/22_action_plugin.sql
nmadruga: I agree. Again, coming from the `plugin.json`

src/components/postgresConnect.ts
```diff
@@ -1,45 +1,231 @@
...
-      break
+  public addFile = async (payload: FilePayload): Promise<number> => {
+    try {
+      const result = await this.query(
+        `INSERT INTO file (${Object.keys(payload)}) VALUES (${this.getValuesPlaceholders(
+          payload
+        )}) RETURNING id`,
+        Object.values(payload)
+      )
+      return result.rows[0]
```
CarlosNZ, line 127: I think this is what's causing the problem with file upload. (See my comment on main thread for test result)

The `registerFileInDB` method (filehandler.ts) is expecting a `number` to be returned, but its actually receiving an object with an `id` field. My original code was returning `result.rows[0].id`, so it'll need to return that value instead.

src/components/postgresConnect.ts
nmadruga: The keys and objects passed have to be in the format expected by `ActionQueuePayload`. I tried to see this package `pg-format` before but it wasn't clear to me how to use in the same way that we would like to: with keys and values that are variable, not always the full structure. I need some more input on this one to continue.

src/components/postgresConnect.ts
nmadruga: Ah, yea. I can fix that now. : )

src/components/postgresConnect.ts
nmadruga: Hmm... Yes, I spent about 1 hour trying to understand this one the other day. But I still haven't seen how the `pg-format` could help with passing a dynamic set of { key, values } to the query. I'll see if I find something about it on the weekend.

src/modules/evaluateExpression/evaluateExpression.ts
nmadruga: Ok. Sorry about that. thanks

src/components/postgresConnect.ts
nmadruga: Ah! Yes. That will be definitely a problem. Fixing now!

database/buildSchema/22_action_plugin.sql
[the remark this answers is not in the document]
CarlosNZ: I don't think we want to put all fields from the `plugin.json` file into the database. Apart from the compulsory ones, a plugin author should be able to add any fields they want, for other purposes (such as some other software that might want to catalogue them, I don't know, but they should be pretty open-ended), and we don't want to cause a database error whenever a field is added that isn't in the database already. The database fields were only intended to capture the essential information that the system needs to run the plugin, whereas the `plugin.json` is open to being a much more comprehensive metadata store (and there definitely isn't a one-to-one mapping between meta-data and database fields (e.g. the `path` field is a combination of the discovered path (not in `plugin.json`) and the filename.))

src/components/postgresConnect.ts
CarlosNZ: Okay, I think this works adequately:
```
const query = `INSERT into action_queue (${Object.keys(action
      )}, time_queued) VALUES (${this.getValuesPlaceholders(action)}, CURRENT_TIMESTAMP)`
const result = await this.query(query, Object.values(action))
```

src/modules/evaluateExpression/evaluateExpression.ts
```diff
@@ -1,46 +1,11 @@
...
 ): Promise<string | number | boolean | any[]> {
+  console.log('inputQuery', inputQuery)
```
CarlosNZ, line 7: `console.log` !

CarlosNZ:
Hmm, I'm also getting this error when I run the Action trigger from PostGres:
```
(node:37415) UnhandledPromiseRejectionWarning: SyntaxError: Unexpected token T in JSON at position 0
[0]     at JSON.parse (<anonymous>)
[0]     at Object.<anonymous> (/Users/carl/Github/openmsupply-application-manager/application-manager-server/src/modules/evaluateExpression/evaluateExpression.ts:10:55)
[0]     at step (/Users/carl/Github/openmsupply-application-manager/application-manager-server/src/modules/evaluateExpression/evaluateExpression.ts:33:23)
[0]     at Object.next (/Users/carl/Github/openmsupply-application-manager/application-manager-server/src/modules/evaluateExpression/evaluateExpression.ts:14:53)
[0]     at /Users/carl/Github/openmsupply-application-manager/application-manager-server/src/modules/evaluateExpression/evaluateExpression.ts:8:71
[0]     at new Promise (<anonymous>)
[0]     at __awaiter (/Users/carl/Github/openmsupply-application-manager/application-manager-server/src/modules/evaluateExpression/evaluateExpression.ts:4:12)
[0]     at Object.evaluateExpression [as default] (/Users/carl/Github/openmsupply-application-manager/application-manager-server/src/modules/evaluateExpression/evaluateExpression.ts:40:12)
[0]     at Object.<anonymous> (/Users/carl/Github/openmsupply-application-manager/application-manager-server/src/components/triggersAndActions.ts:99:65)
[0]     at step (/Users/carl/Github/openmsupply-application-manager/application-manager-server/src/components/triggersAndActions.ts:52:23)
[0] (Use `node --trace-warnings ...` to show where the warning was created)
[0] (node:37415) UnhandledPromiseRejectionWarning: Unhandled promise rejection. This error originated either by throwing inside of an async function without a catch block, or by rejecting a promise which was not handled with .catch(). To terminate the node process on unhandled promise rejection, use the CLI flag `--unhandled-rejections=strict` (see https://nodejs.org/api/cli.html#cli_unhandled_rejections_mode). (rejection id: 2)
[0] (node:37415) [DEP0018] DeprecationWarning: Unhandled promise rejections are deprecated. In the future, promise rejections that are not handled will terminate the Node.js process with a non-zero exit code.
```

I'm just looking into it now. It seems like for the `message` part, the `evaluateExpression` is getting a string as input, and not an object. And it's not a JSON string, it's just the raw message.

Seems like the plugin is working -- it prints the message to the console -- but we probably need to take another look at the Action parsing to see why it's sending a string rather than the object.

src/components/postgresConnect.ts
```diff
@@ -1,45 +1,231 @@
...
+        Object.values(action)
+      )
+      return true
+    } catch (err) {
+      console.log(err.stack)
+      return false
+    }
+  }
+
+  public executeActionQueued = async (payload: ActionQueueExecutePayload): Promise<boolean> => {
```
CarlosNZ, line 91: Perhaps this function name is somewhat unclear. How about `executedActionStatusUpdate` ?

database/buildSchema/22_action_plugin.sql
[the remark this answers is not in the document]
nmadruga: Yes, I agree. I'll remove this fields from the DB now and documentation.

I think the ideal would be that only the fields expected to be included in the database will be passed to the method. I'll check how this can be done using Typescript and casting. I'm sure there is a simple way to do this.

src/components/postgresConnect.ts
nmadruga: Nice. Thanks for looking into this with more detail. Is always better to be sure the core is making safe stuff.

nmadruga: Ok. I have updated in every INSERT query that was using the dynamic keys and placeholders to build the query. :)

src/components/postgresConnect.ts
nmadruga: Cool. I'll update it now.

src/modules/evaluateExpression/evaluateExpression.ts
nmadruga: Oh, I'm actually returning this since the tests aren't working. I'll add some commits.

nmadruga:
@CarlosNZ I think most of the problems are now fixed. Please have another look

src/components/triggersAndActions.ts
```diff
@@ -1,137 +1,129 @@
 import path from 'path'
 import getAppRootDir from './getAppRoot'
 import * as config from '../config.json'
-import { ActionLibrary, Action, DatabaseResult, TriggerPayload, ActionPayload } from '../types'
+import { ActionLibrary, ActionInTemplate, TriggerPayload, ActionPayload } from '../types'
 import evaluateExpression from '../modules/evaluateExpression/evaluateExpression'
-import { Client } from 'pg'
+import PosgresDB from '../components/postgresConnect'
```
CarlosNZ, line 6: Still 1 typo here (Pos**t**gres) ;)

src/modules/evaluateExpression/evaluateExpression.ts
```diff
@@ -120,7 +96,7 @@ async function processPgSQL(queryArray: any[], queryType: string, connection: IC
     rowMode: queryType ? 'array' : '',
   }
   try {
-    const res = await connection.query(query)
+    const res = await connection.query(query.text, query.values)
```
CarlosNZ, line 99: Change this back to just `connection.query(query)` -- it needs the whole object including `rowMode`

src/components/postgresConnect.ts
```diff
@@ -1,45 +1,223 @@
...
+    Object.keys(object).map((key, index) => `$${index + 1}`)
+
+  public static get Instance() {
+    return this._instance || (this._instance = new this())
+  }
+
+  public query = async (text: string, payload: any[] = []): Promise<QueryResult> => {
+    const client = await this.pool.connect()
+    try {
+      return await client.query(text, payload)
```
CarlosNZ, line 67: So yeah, add back `rowMode` as third parameter here -- **pgSQL** needs it :)

CarlosNZ (changes requested):
Just one `Posgres` typo left ;)

Plus what we just talked about :)

src/modules/evaluateExpression/evaluateExpression.ts
nmadruga: Done. Great find! Wow, and sorry for missing that one there. Big mistake.

src/components/postgresConnect.ts
nmadruga: Cool. Done!

CarlosNZ (approved):
Sweet -- tests all pass 👍

CarlosNZ:
> I'll let you do the Merge honours.

I tell a lie -- I merged this into my PR already, so best if we merge this, then that. 🤷‍♂️

PR #59, opened 2020-10-04
#51 objectProperties operator update

### LINKED ISSUE #51: evaluateExpression -- update objectProperties operator to agreed object handling process
Do after discussion as to the best way to interact with local state properties.

**UPDATE** (after discussion with @nmadruga):

Recommended implementation (for now -- we can add more complex object parsing later if required).

- All objects to be queried are passed in inside the single `parameters` object, in an array of objects in the field `objects`. (Array because an expression may require values from multiple objects (on different nodes of the tree).)
- In the (single) child node of the `objectProperties` operator node, an object is required with the following properties:
   - `objectIndex` (optional) -- the index of the `objects` array that contains the object this node is interested in. If this property is not defined, it defaults to `0`.
   - `property` -- the name of the field whose value is to be extracted. Can be a nested property, written in dot notation (e.g. `questions.q2`)

#### Example input expression:  
```
evaluateExpression(  
    { operator: 'objectProperties',  
      children: [  
         { value: { property: 'questions.q1' } }  
              ]  
     },  
     { objects: [application] } )
```  
where `application` is the local state object: 
```
application = {
  id: 1,
  name: 'Drug Registration',
  status: 'Draft',
  stage: 1,
  questions: { q1: 'What is the answer?', q2: 'Enter your name' },
}
```

So, now, overall, the `parameters` object has (up to) three available fields (defined in the Interface `IParameters`):
- `objects` (array of input objects, as above)
- `pgConnection` (PostGres Client object)
- `graphQLConnection` (GraphQL connection parameters)

CarlosNZ:
Fixes #51.

Update `objectProperties` operator according to discussion. The only thing I've made different from what we discussed is that the objects are to be passed in in an array, since it's possible that more than one object might be needed for a complex query. I've also updated the documentation to specify how to use it.

**EDIT**
Added/updated tests in `evaluateExpressionTestData.ts` to use the new input format.

Updated documentation to explain simplified parameters object and new `objectProperties` operator.


documentation/Query-Syntax.md
```diff
@@ -119,7 +119,36 @@ Compares an input string with a regular expression string and returns whether it
 
 ## objectProperties
 
-Used to extract values from current state, for example, from `user`, `organisation` or `form` (i.e. previously entered responses) data. Any object to be evaluated must be passed in to the `evaluateExpression` function in its parameters object, with the field name identical to the object name (see below).
+For extracting values from local state objects (e.g. `user`, `organisation` or `form`) as part of queries. This might be needed if a question's visibility condition relies on the answer to a previous question, for example. All objects to be queried are to be passed in in an array in the field `objects`. See **Usage** below for detailed overview of arguments.
```
nmadruga, line 122:
```suggestion
For extracting values from local state objects (e.g. `user`, `organisation` or `form`) as part of queries. E.g to determine a question's visibility condition based on an answer to a previous question. 

The evaluateExpression expects each instance of the `objectProperties` to be passed along with all objects that are required in the array field `objects`. See **Usage** below for detailed overview of arguments.
```

documentation/Query-Syntax.md
```diff
@@ -119,7 +119,36 @@ Compares an input string with a regular expression string and returns whether it
...
+
+- Input: A single child node whose `value` is an object with the following properties:
+  - `objectIndex` (optional) -- the index of the `objects` array that contains the object this node is interested in. If this property is not defined, it defaults to `0`, so you'd usually only include it if you were supplying more than one object to the whole expression.
+  - `property` -- the name of the field whose value is to be extracted. Can be a nested property, written in dot notation (e.g. `questions.q2`) (but cannot yet )
+- Output: the value specified in `property`, so could be any type.
```
nmadruga, line 127:
```suggestion
 - Input (i.e. children `array`):
_Note: Each node can only be dealing with one object property!_ 
The array `objects` is passed for more than one evaluationExpression object and each node will always associated to a single property only.
    - 1st child node returns an object with the following properties:
      - `objectIndex` (optional) -- index of one object in the `objects` array of the object this node is interested in. Defaults to `0` -- only include it if you were supplying more than one object to the whole expression.
      - `property` -- the name of the field whose value is to be extracted. Can be a nested property, written in dot notation (e.g. `questions.q2`) **(but cannot yet )**
- Output:
  - Value specified in `property` of any type.
```

src/modules/evaluateExpression/evaluateExpressionTestData.ts
```diff
@@ -691,11 +703,11 @@ testData.complex1 = {
...
             {
               operator: 'objectProperties',
-              children: [{ value: { object: 'user', property: 'id' } }],
+              children: [{ value: { objectIndex: 0, property: 'id' } }],
             },
             {
               operator: 'objectProperties',
-              children: [{ value: { object: 'organisation', property: 'id' } }],
+              children: [{ value: { objectIndex: 1, property: 'id' } }],
             },
```
nmadruga, line 711: It is a good idea to uncomment the text for this `complex1` QuerySyntax to test this data changed

src/modules/evaluateExpression/types.ts
```diff
@@ -20,7 +20,8 @@ export interface IGraphQLConnection {
 
 export interface IParameters {
   [key: string]: any
```
nmadruga, line 22: Wouldn't this have the key `objects` now?

src/modules/evaluateExpression/evaluateExpression.ts
```diff
@@ -79,17 +70,21 @@ export default async function evaluateExpression(
         if (Object.entries(params).length === 0)
           return 'No parameters received for objectProperties node'
         try {
-          return params[childrenResolved[0].object][childrenResolved[0].property]
+          const objectIndex = childrenResolved[0].objectIndex ? childrenResolved[0].objectIndex : 0
+          const inputObject = params && params.objects ? params.objects[objectIndex] : {}
+          const property = childrenResolved[0].property
```
nmadruga, line 75: I haven't understood this line. Why you are just checking the `first` child?

src/modules/evaluateExpression/evaluateExpression.ts
```diff
@@ -79,17 +70,21 @@ export default async function evaluateExpression(
...
         }
 
       case 'pgSQL':
-        if (!params.pgConnection) return 'No database connection provided'
+        if (!params || !params.pgConnection) return 'No database connection provided'
         return processPgSQL(childrenResolved, query.type, params.pgConnection)
 
       case 'graphQL':
-        if (!params.graphQLConnection) return 'No database connection provided'
+        if (!params || !params.graphQLConnection) return 'No database connection provided'
```
nmadruga, line 86:
```suggestion
        if (!params.graphQLConnection) return 'No database connection provided'
```

nmadruga (changes requested):
Ok. Just a few small changes and now I'll start using this one to test using some mocked local states for another PR on the Front-end.

documentation/Query-Syntax.md
CarlosNZ: Okay, have reworded it and tried to include your added information. The (but cannot yet) was an incomplete sentence 😊 -- have fixed to "(but cannot yet traverse into arrays.)"

src/modules/evaluateExpression/evaluateExpressionTestData.ts
CarlosNZ: Good call. The result of `complex1` is currently `false`, since there is no organisations in the database, but we should add some and make this test return `true` sometime soon.

src/modules/evaluateExpression/types.ts
CarlosNZ: Yes, good catch. It should be:
```
export interface IParameters {
  objects?: object[]
  pgConnection?: IConnection
  graphQLConnection?: IGraphQLConnection
}
```

src/modules/evaluateExpression/evaluateExpression.ts
CarlosNZ: The `objectProperties` operator only contains one child node, with:
```{ objectIndex: <number>, property: <nameOfProperty> }```

I've tried to make it clear in the code by assigning the relevant items to the variable names `objectIndex`, `inputObject` and `property`.

CarlosNZ: But yes, all the array indexing and nested object properties containing names of other properties does do one's head in. 🙄

src/modules/evaluateExpression/evaluateExpression.ts
CarlosNZ: Hmm, I put it like this because I was getting a type error, since `params` could have additional properties. But it seems okay now that we've restricted the `IParameters` type more effectively. Nice, thanks :)

src/modules/evaluateExpression/evaluateExpression.ts
nmadruga: Ah. Ok, this is inside the `case 'objectParameters'`.

nmadruga (approved):

Issue #61, opened 2020-10-06
(Ongoing) documentation of question type parameters


CarlosNZ:
**NOTE: Please keep this spec updated in the documentation.** Wiki page: https://github.com/openmsupply/application-manager-web-app/wiki/Element-Type-Specs

Let's use this post to start an authoritative description of `template_element` fields, and the input and response parameters of each question/element type, so we can be confident we're naming and defining things correctly when working on different areas.

I'd suggest that as each plugin is implemented, we add its spec to a wiki page and make further updates there. 

For now, let's only add the question types that we're planning on implementing initially?

----

## Template Element fields
- **id, section_id** : database references
- **code**: `string` -- unique (per template) identifier
- **index**: `integer` -- element ordering sequence (starting at `0`)
- **title**: `string` -- describes the element, shown in template builder (probably)
- **category**: `enum` -- either "Information" or "Question"
- **visibility_condition**: `JSON` -- dynamic query determining whether element appears to the user
   - default: `{"value": true}` (i.e. it always appears) _--We should add this as a PostGres default so it doesn't need to be provided in templates (remove this comment when done)_
- **element_type_plugin_code**: `string` -- code of the question/element plugin that presents this element to the UI
- **is_required**: `JSON` -- dynamic query determining whether question is required to be answered _(Currently it is `boolean` -- we should change this to dynamic field)_
   - default: `{"value": true}`  _--TO-DO: Add default to database schema_
- **is_editable**: `JSON` -- dynamic query determining whether can be edited (Would only be false in rare circumstances)
   - default: `{"value": true}`  _--TO-DO: Add default to database schema_
- **parameters**: `JSON` -- the parameters specific to each question/element type. See individual plugins below for parameter breakdown
- **default_value**: `JSON` -- _TO-DO: Remove this field and use in `parameters` if required_ 
- **validation**: `JSON` -- a dynamic expression for checking if the user's response is a valid input.
   - default: `{"value": true}`  _--TO-DO: Add default to database schema_
- **validation_message**: `string` -- the message that shows in the UI when validation fails. _TO-DO: Add field to schema_

## Question/Element types

**Note**: all parameter fields can also have a dynamic query object instead of a primitive. The `evaluateExpression` function will return literal strings (or numbers, booleans) as is. The types described for the parameters below are the type that is expected to be _returned_ from a query expression.

### Short Text Input _(not yet implemented -- highest priority)_
- **type/code**: `shortText`
- **category**: `Question`

[...]

Issue #60, opened 2020-10-06
yarn build plugins fix


CarlosNZ:
There is a problem with building the action plugins that I only just discovered. Even when you run `yarn build` from the plugin subfolder, it creates `.js` builds of *all* `.ts` files in the whole project. It should only compile `.ts` files from its own subfolder. The subfolder `tsconfig` files must have an incorrect notion of what its own project root should be.

CarlosNZ:
Ah -- solved it. Each plugin needs a `node_modules` folder in it, or else the compiler goes looking up the tree for packages.

So you just have to make sure you've run `yarn install` once at least in each folder to create it, and then it'll work as it's supposed to. 

Might just add this to the `yarn build_plugins` script.

CarlosNZ:
Nope -- that didn't actually solve it. Now it's behaving inconsistently. Hmmm...

CarlosNZ:
Okay, got it, but don't know how to fix just yet.

It's compiling all files in the dependency tree, so the dummy plugins and consoleLog don't cause higher files to compile, as they don't import anything, whereas the new "createUser" plugin is what's causing higher-level files to be compiled to .js.

CarlosNZ:
Okay, this won't work as planned for the above reasons. Typescript won't compile isolated files without also compiling any dependent files, regardless of the include/exclude settings, or else the resultant .js file wouldn't actually work.

I guess we'll go back to just a single build, including all plugins. It just means that plugin updates will require a full build everytime. Sound okay? Or any other suggestions?

CarlosNZ:
Another update -- if we pass in any objects (such as `PostgresDB`) to Action plugins rather than import them, then they won't be compiled up the tree.

For now it's just `PostgresDB`, but we need to add any other objects required by later plugins to this call in triggersAndActions:

`const actionResult = await actionLibrary[payload.code](payload.parameters, PostgresDB)`

Will push a PR now to make this change.

PR #63, opened 2020-10-07
#55 Create User Action plugin (and associated changes)

### LINKED ISSUE #55: New Action to create users
Add unit tests for new method created (Database Abstraction Layer)

CarlosNZ:
Fixes #55 

The plugin itself is only a few lines of code, but there was quite a bit of additional tweaking and configuring to get it all working nicely, and I picked up a few other bugs/deficiencies along the way.

To see the **CreateUser** plugin in action:
- reset the database with `yarn database_init`. This reloads the templates according to [this](#61) question spec. I've also modified the fields in the `user` table (plus a few other little tweaks -- check the diff)
- it also loads a new "User Registration" application into the database, with outcome "Pending". This is how it would be if the applicant had saved their draft application, but not yet Submitted it.
- run this command to simulate Submitting the application:  
`UPDATE application SET trigger = 'onApplicationSubmit' WHERE id = 1`
- you should see "Adding new user: c_drown" printed to console, and the new user record created in the `user` table.

There's also a basic test suite for the action -- `createUser.test.ts` -- which just confirms that the Action returns the database response expected.

To run just this test (rather than all test suites), use `yarn test src/plugins/action_create_user/src/createUser.test.ts`

However, the trickiest part of this issue was creating the `template_action` parameter queries -- since only the application ID is passed in via the trigger, the parameter queries had to do SQL lookups to get the `user` data from the application responses. I've updated the `triggersAndActions` component so that the `payload` object (that's the data returned by PostGres when the trigger runs) and the pgConnection are _always_ passed into the `evaluateExpression` calls in `processTrigger`, so that parameter queries can get info from the payload object if needed (like in this case).

### Other things changed/added as part of this issue:

- fixed a bug in `updateActionPlugin` method of database abstraction class -- the parameter substitutions were out of order.
- added `createUser` method in database abstraction class
- the `registerPlugins` module now compares _all_ fields of scanned vs database plugins for changes, and updates the database accordingly.
- started a `utilityFunctions` file -- I figured this would be a good place to collect any non-specific, useful stand-alone functions that might be used elsewhere (and I moved the `getAppRootDir()` function into this file)
- created a `deepEquality` function (in `utilityFunctions`) for comparing objects or arrays by value. I know Lodash has a deep equality function, but I didn't think we wanted to include a whole other library just for a minor feature. (If we end up needing lodash for other things, we can remove this function, although many of lodash's features are native in JS now, aren't they?)
- test suite for above `deepEquality`. (run `yarn test` to run all test suites)
- remove `default_value` field from `template_element` table (as suggested [here](https://github.com/openmsupply/application-manager-server/issues/61#issue-716043733) -- defaults should be question type parameters, as their type will vary.
- removed plugins' compiled `.js` files -- the separated plugin build that I'd planned [will not work in its current state](#60) so will need to build all plugins as part of the main build (to do in #60)

#### Couple of things to consider/note:

- When running the `createUser` test, this warning appears:  
`A worker process has failed to exit gracefully and has been force exited. This is likely caused by tests leaking due to improper teardown. Try running with --runInBand --detectOpenHandles to find leaks.`. 
Have not been able to figure out what's going on here, so any suggestions would be helpful.
- Note that any user registration application won't have anything its `user_id` field, since its been filled in by a non-registered user. The user exists after submitting the application. Do you think it should go back and put the newly-created user_id in the application? This would be a "special case" and would probably need a dedicated Action just to do it. I think it's probably fine not to do it, just thought it's worth pointing out.


src/types.ts
```diff
@@ -68,6 +68,15 @@ export interface ActionPluginPayload {
...
 
+export interface User {
+  first_name?: string
+  last_name?: string
+  username: string
+  date_of_birth?: Date
+  password_hash: string
+  email: string
+}
+
```
nmadruga, line 79: I think it might be better to use the types from the generated `schema.graphql` that was included in the previous version of this postgraphile server. I think using this would be better than redefining the types here.

src/plugins/action_create_user/src/createUser.test.ts
```diff
@@ -0,0 +1,39 @@
...
+})
+
+test('Test: Invalid user -- should fail', () => {
+  return Action.createUser(invalidUser).then((result: any) => {
+    expect(result).toEqual({
+      status: 'Fail',
+      error_log: 'There was a problem creating new user.',
+    })
+  })
+})
```
nmadruga, line 39: all tests passed, but there is one error output that I'll investigate the reason why:
Screen Shot 2020-10-13 at 2 14 40 PM

nmadruga (changes requested):
Great PR. Has some nice improvements in it and from what I understood the `visibility_criteria` for user registration is also included here. Good one to get approved before I move on to show the questions on the User Registration form.

Made a few comments, see if you agree. Otherwise there is just a minor problem with tests that I'll see if I find where is the problem and will post here.

src/plugins/action_create_user/src/createUser.test.ts
CarlosNZ: Just quickly -- this is intentional. The test sends invalid data to the action, so it passes by throwing an error ;)

We can probably make it more elegant by changing what the error returns -- I think the whole error object is probably better than `error.stack`, since it allows the calling method to handle it as it sees fit, and I can also adjust this test to accommodate. Will tweak.

database/insertData.js
```diff
@@ -224,335 +411,374 @@ const queries = [
...
-                  isCurrent: true
-                  applicationStatusHistoriesUsingId: {
-                    create: {
-                      status: SUBMITTED
-                      timeCreated: "NOW()"
-                      isCurrent: true
-                    }
+                }
+              }
+              {
```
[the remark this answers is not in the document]
CarlosNZ: You mean having no comma between elements? Yeah, it's weird, huh? Just one of those GraphQL quirks. I think you can actually put commas in, but when you click "Prettify" in the Query builder UI it strips them out.

src/components/postgresConnect.ts
```diff
@@ -199,7 +204,7 @@ class PostgresDB {
 
   public updateActionPlugin = async (plugin: ActionPlugin): Promise<boolean> => {
     const text =
-      'UPDATE action_plugin SET name = $1, description = $2, path = $3, function_name = $4, required_parameters = $5 WHERE code = $6'
+      'UPDATE action_plugin SET name = $2, description = $3, path = $4, function_name = $5, required_parameters = $6 WHERE code = $1'
```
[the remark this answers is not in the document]
CarlosNZ: Yeah, it's a potential problem. But I think, at least in this case, the only time its used is in the `registerPlugins` component, so we can keep ensuring they match. Maybe make a new issue to generalise these queries to dynamically match the keys, no matter the order? (It's a bit tricky to do it in such a way that guarantees proper parameterized query strings, so should really be its own issue, I think)

src/components/utilityFunctions.ts
```diff
@@ -0,0 +1,35 @@
...
+          if (!deepEquality(obj1[i], obj2[i])) return false
+        } else if (!(obj1[i] === obj2[i])) return false
+      }
+      return true
+    case false:
+      if (!(obj1 instanceof Object && obj2 instanceof Object)) return false
+      if (isArray1 || isArray2) return false // one array slipped through
+      // Both objects
+      if (!deepEquality(Object.keys(obj1), Object.keys(obj2))) return false
+      return deepEquality(Object.values(obj1), Object.values(obj2))
```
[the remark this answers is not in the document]
CarlosNZ: It's not quite the same thing -- if the keys don't match then return `false`, but if the keys _do_ match, then we still have to check the values all match. (so we can't just return the result of the key comparison, we still have one more comparison to do)

To be sure, I switched this code to your suggestion and two tests failed. :( (and they were the ones with matching keys but differing values.)

src/plugins/action_create_user/src/createUser.test.ts
```diff
@@ -0,0 +1,39 @@
...
+test('Test: add User to database', () => {
+  return Action.createUser(testUser).then((result: any) => {
+    expect(result).toEqual({
+      status: 'Success',
+      error_log: '',
+    })
+  })
+})
+
+test('Test: Invalid user -- should fail', () => {
```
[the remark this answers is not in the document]
CarlosNZ: Nah, the `date_of_birth` and `username` fields are misnamed. Have put details in the test message now..

src/types.ts
CarlosNZ: Yeah, good point -- have you imported any of the `schema.graphql` types elsewhere, just so I can see how it's done?

Just one problem though maybe -- will the generated types file know which fields should be optional? (e.g. `date_of_birth?: Date`)

src/plugins/action_create_user/src/createUser.test.ts
CarlosNZ: Okay, I've updated `postgresConnect` to `throw` errors for each method. That way the calling method can handle the error as it sees fit.

In this case, the `try/catch` block in `createUser` handles the error and updates the database accordingly.

Do you think it's too much to change all of these? If so, I can change them back and just leave the `createUser` one.

Personally I think throwing the error up the call chain is the best approach as it gives the calling method control to handle the error as it sees fit (which might vary in different contexts), (e.g write log to database, console output, UI message, etc.) Plus it makes writing error condition unit tests easier.

src/components/postgresConnect.ts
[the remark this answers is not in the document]
nmadruga: Yes. I agree, have added a new note in the Backlog issues

src/components/utilityFunctions.ts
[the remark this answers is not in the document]
nmadruga: Oh! That's right. Sorry I missed the second one was comparing values.

src/types.ts
nmadruga: I haven't used here in the back-end, just front-end. Which is done via Graphql `code-gen` package. But it should be similar here. When the postgraphile runs it creates this file and it contains all these types in there.
Ah, I don't know how to do this optional with the existing types... There is probably a way to define a type a make one or more fields optional.

src/plugins/action_create_user/src/createUser.test.ts
nmadruga: No, I totally agree with that. The solution before was kind of temporary, and now I see why it is a bad one. We can continue like that and work on a errorHandler that should be called within the calls to the DB layer.

nmadruga (approved):

PR #64, opened 2020-10-07
#56 Endpoint to check for unique values

### LINKED ISSUE #56: Fastify new endpoint for checking uniqueness of user email/username
Without permission

CarlosNZ:
Fixes #56

Endpoint is provided at `/check-unique`.
Query parameters are:
- `type`: available options: `username`, `email`, `organisation`
- `value`: the value being checked for uniqueness

Example request URL: `/check-unique/type=email&value=carl@sussol.net`

Check is case-insensitive, so `carlsmith` will return Not Unique if `CarlSmith` is in the database.

Returns a boolean.

I created a unit test suite. Run: `yarn test src/server.test.ts`

And please try out your own queries with Postman.

Also: added one organisation to database in `insertData.js`

### DIFF
--- a/src/server.ts
+++ b/src/server.ts
@@ -49,8 +49,7 @@ const startServer = async () => {
 
   // Unique name/email/organisation check
   server.get('/check-unique', async (request: any, reply) => {
-    const type = request.query.type
-    const value = request.query.value
+    const { type, value } = request.query
     if (value === '' || value === undefined) return false
     let table, field
     switch (type) {
--- a/src/server.ts
+++ b/src/server.ts
@@ -10,6 +10,7 @@ import {
   filesFolderName,
 } from './components/fileHandler'
 import { getAppRootDir } from './components/utilityFunctions'
+import PostgresDB from './components/postgresConnect'
 
 // Bare-bones Fastify server
 
@@ -44,6 +45,31 @@ const startServer = async () => {
   server.get('/', async (request, reply) => {
     console.log('Request made')
     return 'This is the response\n'
+  })
+
+  // Unique name/email/company check
+  server.get('/check-unique', async (request: any, reply) => {
+    const type = request.query.type
+    const value = request.query.value
+    let table, field
+    switch (type) {
+      case 'username':
+        table = 'user'
+        field = 'username'
+        break
+      case 'email':
+        table = 'user'
+        field = 'email'
+        break
+      case 'organisation':
+        table = 'organisation'
+        field = 'name'
+        break
+      default:
+        return null
+    }
+    const isUnique = await PostgresDB.isUnique(table, field, value)
+    return isUnique
   })
 
   server.listen(8080, (err, address) => {
--- a/src/server.ts
+++ b/src/server.ts
@@ -50,7 +50,13 @@ const startServer = async () => {
   // Unique name/email/organisation check
   server.get('/check-unique', async (request: any, reply) => {
     const { type, value } = request.query
-    if (value === '' || value === undefined) return false
+    if (value === '' || value === undefined) {
+      reply.send({
+        unique: false,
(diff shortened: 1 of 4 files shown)


src/server.ts
```diff
@@ -46,6 +47,32 @@ const startServer = async () => {
     return 'This is the response\n'
   })
 
+  // Unique name/email/organisation check
+  server.get('/check-unique', async (request: any, reply) => {
+    const type = request.query.type
+    const value = request.query.value
```
nmadruga, line 53:
```suggestion
    const { type, value } = request.query
```

src/server.ts
```diff
@@ -46,6 +47,32 @@ const startServer = async () => {
...
+        break
+      case 'email':
+        table = 'user'
+        field = 'email'
+        break
+      case 'organisation':
+        table = 'organisation'
+        field = 'name'
+        break
+      default:
```
nmadruga, line 69: I think it would be good to have some alert if some type passed is not treated, so basically send back the result with result + message. It can still return false, but the message would say: `not a recognised type`.

nmadruga (changes requested):
That's good. Will be useful, and I liked the reuse of the Abstraction layer.
The tests in my option would rather be done in a separated issue because we need a mock database to use, other wise when we start changing the existing data these tests will fail.

CarlosNZ:
> The tests in my option would rather be done in a separated issue because we need a mock database to use, other wise when we start changing the existing data these tests will fail.

Yes, they should be using a mock database, and will need to migrate the tests over in a new issue. But in the meantime, these tests are still useful, no? (And they're actually more likely to have persistent results than the ones currently in the expression-evaluator, which will also need to be migrated).

We should do all test migrations in a single issue when the mock database is configured, right?

src/server.ts
CarlosNZ: How do you suggest structuring this? Currently, the endpoint always returns a boolean, so are you suggesting returning another type (like an object) as the result for those situations?

nmadruga: Yes. I think it would be nicer to have the result as an object and the queryResult is returned in the body. It's also preferable to return the status in the api call as what can be checked first and then the result in the body. For example for success = `{ status: 200, resultBody = { unique: "false" }`. That's basically my opinion from what I have seen before, not 100% certain. what do you think @andreievg?

nmadruga:
> We should do all test migrations in a single issue when the mock database is configured, right?

Yes. I get your point and I think that's better than my ideal of having the test task done at some point that would just add all these tests in it. Migrating will be easier to maintain and check that we cover everything that was covered before.

src/server.ts
CarlosNZ: I don't think we need the status as that's included in the HTTP response headers anyway.

I agree that returning a pure boolean is probably not best practice in a general sense. Only problem is that in this case writing a dynamic expression to parse the return object will make for a more complex query. Or maybe it won't be too bad, will have to try and see. How about a response like this:
```
{ 
  unique: true/false
  message: "Not a recognised type" (or <empty string> if type is okay)
}
```

I'll try that and then re-write the email validation query and see how much more complex it is.

CarlosNZ: Okay, I've changed it to the above and it didn't actually make the queries much more complex (just one extra `{value: "unique"}` node in the API lookup :) I've updated the API tests in #68 to reflect this change too.

Let me know if that works for you.

nmadruga: > I don't think we need the status as that's included in the HTTP response headers anyway.

That's what I was trying to say. It looks like an object though. It was supposed to be a wrapper of what is returned by the API.

> :) I've updated the API tests in #68 to reflect this change too.

Thanks! Will have a look.

nmadruga (approved):

Issue #53, opened 2020-10-04
Actions for basic user registration


nmadruga:
- Identify Actions
- Make graphQL

CarlosNZ:
Actions that I can think of:
- createUser -- almost done in #55.
- changeOutcome (to COMPLETED) -- will do next.

Issue #39, opened 2020-09-23
Problem with PostGraphile --watch flag not updating schema on PostGres schema changes


CarlosNZ:
The `--watch` flag is supposed to make PostGraphile listen for changes to the postgres database schema and update its schema accordingly. However, this doesn't seem to be happening. I'm having to quit the PostGraphile server and re-launch it to get it to recognise the schema changes.

We are currently launching PostGraphile from the `post_graphile.sh` script, which executes the following command:

`npx postgraphile -c "postgres://postgres@localhost/tmf_app_manager" -C postgres --watch --enhance-graphiql --dynamic-json --cors --export-schema-graphql "database/schema.graphql" --append-plugins @graphile-contrib/pg-simplify-inflector,postgraphile-plugin-nested-mutations`

nmadruga:
I was having a little research regarding this one today. It seems to me that keeping the postgraphile in a separated server is not worth, since we can't use some of the features that running it in the same fastify server as a middleware. In this link: https://www.graphile.org/postgraphile/usage-schema/ it explains how to start one. 
With the Postgraphile as a middleware that would be another endpoint on the server  would also allow us to use `watchPostGraphileSchema` which could be running the `insert_data.js` after receives that the schema update is finished.

PR #65, opened 2020-10-08
Action plugin -- change outcome

CarlosNZ:
Extends work done in #55. One more action for #53.

Updated the User Registration template to also trigger this action -- it sets the outcome to "Approved" as soon as its submitted (and also sets `is_active` to `false`).

To see it in action:
- reset database to load new template (`yarn database_init`), and start the server.
- run `UPDATE application SET trigger = 'onApplicationSubmit' WHERE id = 1`
- you should see activity written to console, and the `outcome` and `is_active` fields have been changed
(the other actions also run)


src/components/postgresConnect.ts
```diff
@@ -238,6 +239,24 @@ class PostgresDB {
       return false
     }
   }
+
+  public setApplicationOutcome = async (
+    appId: number,
+    outcome: ApplicationOutcome
+  ): Promise<boolean> => {
+    // Assumes that `is_active` should become False when outcome is set to "Approved" or "Rejected"
```
nmadruga, line 247: Do you think this could be a function in SQL? So when we change the outcome to `Approve` or `Rejected` the application is automatically changed to have `is_active` as false. I was thinking that leaving this to be an action might be not the best option. In other words: something to be remembered all the time that should be default in the system ideally is a database trigger.

database/insertData.js
```diff
@@ -288,6 +288,18 @@ const queries = [
...
                 }
               }
+              {
+                actionCode: "changeOutcome"
+                condition: { value: true }
+                trigger: ON_APPLICATION_SUBMIT
+                parameterQueries: {
+                  application_id: {
+                    operator: "objectProperties"
+                    children: [{ value: { property: "record_id" } }]
```
nmadruga, line 298: I saw this object in the other PR but wasn't sure how it is receiving the application_id. Is this something done by the evaluator? I couldn't find where this object is being passed.

nmadruga:
Something unexpected too: I have updated the data, but after running the trigger: `UPDATE application SET trigger = 'onApplicationSubmit' WHERE id = 1` I haven't seen any console.log print something and although the `outcome` is now APPROVED the `is_active` is still true for the application. I wonder if something is missing in this PR or the data update didn't work.

src/components/postgresConnect.ts
CarlosNZ: Good point, I'll add the postgres trigger now and remove it from this file.

CarlosNZ: Done :)

database/insertData.js
CarlosNZ: I know, it's confusing isn't it. The object that gets passed in is the Payload object from the Trigger queue, in Triggers&Actions:
```
for (const action of actions) {
    for (const key in action.parameter_queries) {
      action.parameter_queries[key] = await evaluateExpression(action.parameter_queries[key], {
        objects: [payload],
        pgConnection: PostgresDB,
      })
    }
```
This expression tells the evaluator to look at the `record_id` field of that object, which in this case is the `application_id`, since it comes from a trigger on the `application` table.

CarlosNZ:
> Something unexpected too: I have updated the data, but after running the trigger: `UPDATE application SET trigger = 'onApplicationSubmit' WHERE id = 1` I haven't seen any console.log print something and although the `outcome` is now APPROVED the `is_active` is still true for the application. I wonder if something is missing in this PR or the data update didn't work.

That's odd. It's working for me. Try it again now that I've moved the `is_active` update to a Postgres trigger. (and reset the database)

From a clean start, when you execute the SQL statement, you should get this in the console:
```
The Console Log action is running...
Action has been executed on User Registration template
 
Adding new user: c_drown

Updating application: Approved
```

The application in the `application` table should be changed to "Approved" with `is_active` = `false`, and the user `c_drown` added to the user table.

src/components/triggersAndActions.ts
```diff
@@ -122,7 +122,7 @@ export async function executeAction(
   actionLibrary: ActionLibrary
 ): Promise<boolean> {
   // TO-DO: If Scheduled, create a Job instead
-  const actionResult = actionLibrary[payload.code](payload.parameters)
+  const actionResult = await actionLibrary[payload.code](payload.parameters)
```
CarlosNZ, line 125: Added this as result was not being written to `action_queue`. Needs to wait for result to resolve.

database/insertData.js
nmadruga: I see. But how does it know that `the record_id field of that object, which in this case is the application_id`?

src/types.ts
```diff
@@ -112,3 +112,5 @@ export interface TriggerQueueUpdatePayload {
   id: number
   status: TriggerStatus
 }
+
+export type ApplicationOutcome = 'Pending' | 'Approved' | 'Rejected'
```
nmadruga, line 116: Another case for using the exported schema with types?

database/insertData.js
CarlosNZ: The `TriggerPayload` (object sent to Trigger notification listener) has the following fields:
```
{
  id: number (id of this trigger in the trigger_queue table)
  trigger: TriggerStatus (e.g OnApplicationSubmit)
  table: string (which table was triggered -- in this case "application"
  record_id: number (the id of the record that was triggered from the above table)
}
```
So it's basically sending the information about the database record (`table`, `record_id`) that had its `trigger` field updated and what the trigger was (`trigger`).

This payload object is sent as the object passed into to any `evaluateExpression` calls needed to evaluate the parameters of the Action when it runs. So the evaluator, in this case, returns the `record_id` of the first (and only) object in the `objects` field of parameters, which is the `id` of the application in the `application` table that was triggered. It knows the `record_id` will refer to the `application` table, because it's acting on the `OnApplicationSubmit` trigger.

Basically, everything that the Actions need must be deduced from this payload and are evaluated using the `parameter_queries` of each template action before being sent to the action plugin function.

Sorry, I know that's confusing. 🤯 There's probably a better way to make this all clearer. Does the diagram [here](https://github.com/openmsupply/application-manager-server/wiki/Triggers-and-Actions) help at all? Happy to talk this through with you in person when you're in and hopefully it'll make it clearer for the both of us, and maybe we can see potential improvements.

src/types.ts
CarlosNZ: I think I'll have to get you to show me how to do this. The auto-generated `schema.graphql` file is not in a format that can be imported into Typescript file. How did you generate the `graphql.tsx` file in the front-end? 🤷‍♂️

CarlosNZ: Cool -- I've updated to auto-generate and use types from Schema. Cheers. 👍

CarlosNZ:
> Something unexpected too: I have updated the data, but after running the trigger: `UPDATE application SET trigger = 'onApplicationSubmit' WHERE id = 1` I haven't seen any console.log print something and although the `outcome` is now APPROVED the `is_active` is still true for the application. I wonder if something is missing in this PR or the data update didn't work.

I think I know what's happening here, cos it just happened to me too. When you reset the database with `yarn database_init` there are no plugins registered in the database. So you have to quit and restart the server so the plugins get re-registered, then it should work.

src/types.ts
CarlosNZ: Also updated the `database_init` script to auto-generate the types file.

nmadruga: Nice!

nmadruga: I was thinking it would be good to have this generator automated somewhere in the front-end as well. Not sure maybe the pre-commit hook

database/insertData.js
nmadruga: Yes it helped. Thanks. Is just so much to remember! But I think I understood the idea now. Good explanation

src/types.ts
```diff
@@ -68,14 +68,14 @@ export interface ActionPluginPayload {
...
-  email: string
-}
+// export interface User {
+//   first_name?: string
+//   last_name?: string
+//   username: string
+//   date_of_birth?: Date
+//   password_hash: string
+//   email: string
+// }
```
nmadruga, line 78: Better delete than leave comments. If it is required the code can be checked in a previous version.

nmadruga (changes requested):
It's nice to have the `code-gen` here. After the presentation about monorepos I suffer to see every duplication haha. But it's better to have it here than re-create the types.

Also liked what you have done to regenerate when we run the database script.

Just a few commented codes to remove that's all.

database/createDatabase.sql
```diff
@@ -1,5 +1,5 @@
-DROP DATABASE IF EXISTS tmf_app_manager;
+-- DROP DATABASE IF EXISTS tmf_app_manager;
```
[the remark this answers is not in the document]
CarlosNZ: Yeah -- it wasn't usually dropping anyway -- it just paused for ages while it tried, but then didn't because it still had active connections. So I just took this out to speed up the process. 

I think I only had the `DROP DATABASE` command because I thought the next `CREATE DATABASE` command would give an error if it already existed, but it doesn't, so probably don't need it. Just DROPping public schema seems to be enough for now. Might be a good idea to leave it commented though, as a reminder to turn it back on if we start doing more drastic things during the init?

nmadruga:
> I think I know what's happening here, cos it just happened to me too. When you reset the database with yarn database_init there are no plugins registered in the database. So you have to quit and restart the server so the plugins get re-registered, then it should work.

Yes... This is related to what we are trying to fix on #39. So, not a problem related to this issue.

database/createDatabase.sql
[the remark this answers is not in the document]
nmadruga: Ok. Good to know that the CREATE DATABASE will recreate one. I wasn't aware either.
Let's leave this comment for now.

nmadruga (changes requested):
Almost!

I tried running the action again and the result is looking like that: 
````
[0] Server listening at http://127.0.0.1:8080
[0] 
[0] The Console Log action is running...
[0] Action has been executed on User Registration template
[0] 
[0] Adding new user: [object Object]
[0] 
[0] Updating application: Approved
````
Missing the JSON Stringify?

nmadruga (approved):

Issue #68, opened 2020-10-11
Make a generic API (GET) operator for evaluateExpression


CarlosNZ:
Use the same kind of node structure as `objectProperties` and `graphQL`:
- first node returns the endpoint URL
- 2nd node returns an array of query field names
- next `n` nodes return the values for those query fields
- last node returns return node from response (optional -- will return raw response if not specified)

PR #70, opened 2020-10-12
#68 API operator for expression evaluator

CarlosNZ:
Addresses #68. 

Note: a couple of the changed files here are from branch `#56-endpoint-unique-name-check`
 (that I needed to merge into here to make one of the endpoints work).

- Added **API** operator, which follows basically the same format as the **GraphQL** operator (See included documentation for full spec)

- Also added more error handling throughout. Any errors should result in an `Error` being thrown to the calling method for handling with an `error.message` informing what kind of failure occured. Note: it will throw an error and stop when encountering the first error, so there could still be errors in other nodes which are not evaluated.

- Added additional tests for the new operator, including a more complex Email validation check like what we will might use in the front-end.

### DIFF
--- a/database/insertData.js
+++ b/database/insertData.js
@@ -462,6 +462,22 @@ const queries = [
       }
     }
   }`,
+  //   Add one organisation
+  `mutation {
+    createOrganisation(
+      input: {
+        organisation: {
+          address: "123 Nowhere St\\nAuckland"
+          licenceNumber: "XYZ1234"
+          name: "Drugs-R-Us"
+        }
+      }
+    ) {
+      organisation {
+        name
+      }
+    }
+  }`,
   //   User registration application 1
   `mutation {
     createApplication(
--- a/documentation/Query-Syntax.md
+++ b/documentation/Query-Syntax.md
@@ -126,7 +126,7 @@ The `evaluateExpression` function expects each expression to be passed along wit
 - Input: A single child node whose `value` is an object with the following properties:
   - `objectIndex` (optional) -- the index of the `objects` array that contains the object this node is interested in. If this property is not defined, it defaults to `0`, so you'd usually only include it if you were supplying more than one object to the whole expression.  
     _Note: Each node can only return **one** object property_
-  - `property` -- the name of the field whose value is to be extracted. Can be a nested property, written in dot notation (e.g. `questions.q2`) (but cannot yet traverse into arrays.)
+  - `property` -- the name of the field whose value is to be extracted. Can be a nested property, written in dot notation (e.g. `questions.q2`) (but cannot yet get specific elements from arrays.)
 - Output: the value specified in `property` of any type.
 
 **Example**:
@@ -193,7 +193,7 @@ Performs queries on connected GraphQL interface.
   - 1st child node returns a **string** representing the GraphQL query
   - 2nd child node returns an **array** of field names for the query's associated variables object. If no variables are required for the query, pass an empty array (i.e. `{ value: [] }`).
   - 3rd...N-1 child nodes return the values of the fields for the variables object -- one node for each field in the previous node's array.
-  - The Nth (last) child node returns a **string** stating the node in the returned GraphQL object that is required. E.g. `applications.name` Because GraphQL returns results as nested objects, to get an output in a "simple type", a node in the return object tree is needed. (See examples below and in `TestData`)
+  - The Nth (last) child node returns a **string** stating the node in the returned GraphQL object that is required. E.g. `applications.name` Because GraphQL returns results as nested objects, to get an output in a "simple type", a node in the return object tree is needed. (See examples below and in `TestData`). This last node is optional -- if not provided, the whole result object will be returned unmodified.
 - Output: the returned GraphQL node can be either `string`, `number`, `boolean`, `array`, or `object`. If the output is an object, it will be returned as follows:
 
   - If there is only one field, only the value of the field will be returned.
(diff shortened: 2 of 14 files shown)


database/insertData.js
```diff
@@ -462,6 +462,22 @@ const queries = [
       }
     }
   }`,
+  //   Add one organisation
```
CarlosNZ, line 465: Note: not done as part of this PR -- part of #55

src/components/postgresConnect.ts
```diff
@@ -238,6 +238,17 @@ class PostgresDB {
       return false
     }
   }
+
+  public isUnique = async (table: string, field: string, value: string): Promise<boolean> => {
```
CarlosNZ, line 242: Not part of this PR. Done in #64

src/server.test.ts
```diff
@@ -0,0 +1,92 @@
+// Test suite for Fastify server endpoints
```
CarlosNZ, line 1: Not part of this PR -- done in #64

src/server.ts
```diff
@@ -46,6 +47,32 @@ const startServer = async () => {
     return 'This is the response\n'
   })
 
+  // Unique name/email/organisation check
```
CarlosNZ, line 50: Not part of this PR -- done in #64

src/modules/expression-evaluator/src/types.ts
```diff
@@ -22,6 +22,7 @@ export interface IParameters {
   objects?: object[]
   pgConnection?: IConnection
   graphQLConnection?: IGraphQLConnection
+  APIfetch?: any // Don't know type of fetch object
```
nmadruga, line 25: I'll have a look if I can find out what type could be used here. Would probably be a combination of types, because like you said the front-end and back-end use different APIfetch obejcts.

nmadruga (changes requested):

documentation/Query-Syntax.md
```diff
@@ -218,6 +218,38 @@ Performs queries on connected GraphQL interface.
...
 
+## API
+
+Performs GET requests to public API endpoints.
+
+- Input: _(note: basically the same as GraphQL)_
+  - 1st child node returns a **string** containing the url of the API endpoint
+  - 2nd child note returns an **array** of field names for the url query parameters. If no query parameters are required, pass an empty array (i.e. `{ value: [] }`).
+  - 3rd...N-1 child notes return the values for the query parameters in the 2nd node -- one node for each field name
+  - Nth (last) child node returns a **string** stating the field in the return JSON object that is required. This is optional -- if not supplied the complete response data is returned (which is what you'd want if expecting something other than JSON object). Note: an array is returned, the array _elements_ will be reduced to just this field (assuming they are objects)
```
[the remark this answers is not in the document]
CarlosNZ: Okay, have moved it to Output and re-written it to be (hopefully) clearer.

src/modules/expression-evaluator/src/evaluateExpression.ts
```diff
@@ -76,16 +79,45 @@ export default async function evaluateExpression(
...
+        }
+
+      case 'API':
+        let urlWithQuery, returnNode
+        try {
+          const url = childrenResolved[0]
+          const queryFields = childrenResolved[1]
+          const queryValues = childrenResolved.slice(2, queryFields.length + 2)
+          returnNode =
+            childrenResolved[queryFields.length + 2] && childrenResolved[queryFields.length + 2]
```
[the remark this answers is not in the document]
CarlosNZ: Hmm, I think I thought that it would thrown an error if the index referred to a position that doesn't exist (it does in other languages), so I was checking it existed before assigning it. Never mind, I've simplified it to this now, and the tests still all pass.

`returnNode = childrenResolved[queryFields.length + 2]`

src/modules/expression-evaluator/src/evaluateExpression.ts
```diff
@@ -113,25 +145,31 @@ async function processPgSQL(queryArray: any[], queryType: string, connection: IC
...
 async function processGraphQL(queryArray: any[], connection: IGraphQLConnection) {
-  const query = queryArray[0]
-  const variableNames = queryArray[1]
-  const variableNodes = queryArray.slice(2, queryArray.length - 1)
-  const returnNode = queryArray[queryArray.length - 1]
+  try {
+    const query = queryArray[0]
+    const variableNames = queryArray[1]
+    const variableNodes = queryArray.slice(2, variableNames.length + 2)
+    const returnNode = queryArray[variableNames.length + 2] && queryArray[variableNames.length + 2]
```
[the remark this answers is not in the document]
CarlosNZ: Okay, have pulled them into function called `assignChildNodesToQuery`. Feel free to suggest a better name.

src/modules/expression-evaluator/src/evaluateExpression.ts
```diff
@@ -113,25 +145,31 @@ async function processPgSQL(queryArray: any[], queryType: string, connection: IC
...
+    const selectedNode = returnNode ? extractNode(data, returnNode) : data
 
-  return Array.isArray(selectedNode)
-    ? selectedNode.map((item) => simplifyObject(item))
-    : simplifyObject(selectedNode)
+    return Array.isArray(selectedNode)
+      ? selectedNode.map((item) => simplifyObject(item))
+      : returnNode
+      ? simplifyObject(selectedNode)
+      : selectedNode
```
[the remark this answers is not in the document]
CarlosNZ: Yeah, I thought there was a slight difference in the type of data that might be returned from each, but I just hadn't come across in my testing data (know of any API endpoints that return a single object with an array in one of the fields?)

I've made them use the same function now. :)

src/modules/expression-evaluator/src/evaluateExpression.ts
```diff
@@ -145,11 +183,13 @@ const zipArraysToObject = (variableNames: string[], variableValues: any[]) => {
...
 const extractNode = (
-  data: BasicObject,
+  data: BasicObject | BasicObject[],
   node: string
 ): BasicObject | string | number | boolean | BasicObject[] => {
   const returnNodeArray = node.split('.')
-  return extractNodeWithArray(data, returnNodeArray)
+  if (Array.isArray(data)) {
+    return data.map((item) => extractNodeWithArray(item, returnNodeArray))
+  } else return extractNodeWithArray(data, returnNodeArray)
```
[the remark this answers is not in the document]
CarlosNZ: Okay, I've tried to add more explanatory variable names, and I've gone through and changed all references to "node/returnNode" to "property/returnProperty" (node really only makes sense in a GraphQL context). Will update the docs, too.

Have also added additional comments to clarify some of the more confusing stuff. Hope it makes it easier to follow.

src/modules/expression-evaluator/src/types.ts
CarlosNZ: Looks like `APIfetch?: Function` is acceptable. 👍

src/modules/expression-evaluator/src/evaluateExpression.ts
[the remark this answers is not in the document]
CarlosNZ: Actually, I realised this function could be re-written without the inner function with a simple type-check at the top. 🤦‍♂️ Should make it clearer still.

src/modules/expression-evaluator/src/evaluateExpression.ts
```diff
@@ -74,18 +77,44 @@ export default async function evaluateExpression(
...
-          return extractNode(inputObject, property)
+          return extractProperty(inputObject, property)
+        } catch {
+          throw new Error("Can't resolve object")
+        }
+
+      case 'API':
+        let url, urlWithQuery, queryFields, queryValues: string[], returnProperty
+        try {
+          ;[url, queryFields, queryValues, returnProperty] = assignChildNodesToQuery(
```
nmadruga, line 88: `; ` Maybe a mistake?

nmadruga (approved):
Looks way better. Thanks!

src/modules/expression-evaluator/src/evaluateExpression.ts
CarlosNZ: Nah, prettier puts these in sometimes. I don't fully understand why (not in all cases anyway) but there's some discussion [over here](https://github.com/prettier/prettier/issues/1907) if you're interested. 

(It's a consequence of us not having semi-colons at the ends of lines. Sometimes they are required, and it puts them in at the start of the next line. Not quite sure why it would be required in this case though 🤷‍♂️)

PR #71, opened 2020-10-13
#60 Remove Imports from plugins

### LINKED ISSUE #60: yarn build plugins fix

CarlosNZ:
Fixes #60 by removing imports of other project components from plugin code and passing them in as additional parameters when each Action is called.

For now, it's only `PostgresDB`, but there may be more down the line.

Also, amended plugin's tsconfig.json files to exclude test files.


nmadruga (changes requested):
I would like to see more details about why this code is changed to not use the ES6. Thinking that we would need something similar to what is now defined in `src/plugins/action_change_outcome/src/changeOutcome.js` for every plugin it's not ideal. What was the need behind this change?

CarlosNZ:
> I would like to see more details about why this code is changed to not use the ES6. Thinking that we would need something similar to what is now defined in `src/plugins/action_change_outcome/src/changeOutcome.js` for every plugin it's not ideal. What was the need behind this change?

Hang on, I think I've actually pushed the wrong branch to this PR -- got mixed up between two different fixes for the same issue. Sorry 😳

Will sort it out and post back here when it's fixed.

nmadruga (approved):
Well done.
I saw that injection of the database was best thing to do for tests so that's a very good change included in this PR.

PR #75, opened 2020-10-15
Expression evaluator UI query builder

CarlosNZ:
This is a quick-and-dirty dev tool I whipped up last weekend to make building complex expressions/queries easier. It's just a simple `.js` app using `create-react-app`, but it may as well be included here -- I think it will make trying to write complex queries for templates easier.
Screen Shot 2020-10-15 at 1 03 23 PM

I've stashed it deep down within the `/modules/expression-evaluator` project, but I've added a launching script from the main repo root:  
`yarn gui`

It won't work though until you've run `yarn install` from within both `src/modules/expression-evaluator` and `/modules/expression-evaluator/expression-evaluator-gui` so each standalone packages downloads their dependencies.

Couple of notes:
- It uses the `evaluateExpression` package from the dev folder (not the published package) as this will enable live testing when actually making new changes to the evaluator
- create-react-app won't let me import local modules outside the src folder. My workaround is to just re-copy the relevant files into this app's src folder on launch. Works fine, it just means that any changes made to the original `evaluateExpression.ts` won't show up with a hot reload -- you'll have to stop and start the app.
- the `node-postgres` package won't run directly from the browser, so there's a little Express server that also runs with this app that simply relays postgres queries from this app to postgres
- API calls to local server (e.g. `http://localhost:8080/check-unique)` don't work. I don't fully understand why, but I got round it by specifying a proxy to `http://localhost:8080` in `package.json` -- you just have to put relative links as the URL. i.e. `/check-unique`
- You'll also need to add a .env file to the expression-evaluate-gui folder with the following content:  
`SKIP_PREFLIGHT_CHECK=true` 
This prevents errors being thrown due to different projects having different versions of various dependencies.


nmadruga:
> create-react-app won't let me import local modules outside the src folder. My workaround is to just re-copy the relevant files into this app's src folder on launch. Works fine, it just means that any changes made to the original evaluateExpression.ts won't show up with a hot reload -- you'll have to stop and start the app.

This needs to be in some documented page for using this new cool tool!

nmadruga:
> API calls to local server (e.g. http://localhost:8080/check-unique) don't work. I don't fully understand why, but I got round it by specifying a proxy to http://localhost:8080 in package.json -- you just have to put relative links as the URL. i.e. /check-unique

I haven't fully understood this, but I guess also need to be documented somewhere easier to find.

nmadruga:
After running `yarn install` in both `src/modules/expression-evaluator` and `/modules/expression-evaluator/expression-evaluator-gui`, I went back to the root project folder and tried running `yarn gui`, but got this problem:
Screen Shot 2020-10-22 at 10 03 23 AM

src/modules/expression-evaluator/package.json
```diff
@@ -30,7 +30,8 @@
   },
   "scripts": {
     "build": "rimraf ./lib && tsc",
-    "test": "jest"
+    "test": "jest",
```
nmadruga, line 33: Does this really need a test environment?

nmadruga: Oh! Thaat's the copy of the expression-evaluator. Got it!

nmadruga: I'm still not convinced that having a copy instead of using the package is better.

nmadruga (commented):
Just a few comments. I wasn't able to try it yet. But I think it would be better to have the expression-evaluator included as a package. I don't understand why it would get the changes loaded after it was implemented on the real package. And I really think it's not worth having the code + setup duplicated.

CarlosNZ:
> Screen Shot 2020-10-22 at 10 03 23 AM

I know, it's annoying, don't really know a "proper" way to fix this  (i.e. allow different packages to use different versions of dependencies), but I used this workaround suggested by the error message.

Add a `.env` file in the `expression-evaluate-gui` folder with the following text:

`SKIP_PREFLIGHT_CHECK=true`

(.env files don't automatically sync with Github, otherwise you'd already have this.)

I'll document all these details once you approve it -- just didn't want to get ahead of myself if you had some better ideas.

src/modules/expression-evaluator/package.json
CarlosNZ: > I'm still not convinced that having a copy instead of using the package is better.

Because if we're doing development on the module itself and want to test queries, we the GUI to display using the in-development version so we can see the effect, not just what would be shown with the current published package.

CarlosNZ:
Okay, I've added all the notes and information to the documentation -- have added a new section at the bottom of the [Query syntax page](https://github.com/openmsupply/application-manager-server/wiki/Query-Syntax) called **Development**

andreievg:
I am happy for this to just be merged, it doesn't touch any of the server functionality.

Issue #77, opened 2020-10-15
Figure out best way to handle sequential Actions.


CarlosNZ:
My proposed way to handle it is to add a new `Trigger` column to the `action_queue` table, and secondary Actions (ones that rely on completion of earlier action) have this specified as their trigger. Call it `onActionCompleted`, say.

Just need to figure out exactly all the outputs of 1st action can become parameters of the 2nd using our existing framework, and will mock up a proof of concept.

CarlosNZ:
Incorporated into #86

CarlosNZ:
Closing, as superceded by #86.

PR #89, opened 2020-10-28
#86 Improvements to handle sequential and async Actions properly

### LINKED ISSUE #86: Update Triggers/Actions
1. Handle sequential Actions:
   - sequence no. or next/prev info in template
   - Trigger processor evaluates and queues all actions at once
   - No listener on Action table -- trigger processor executes all in sequence once processed.

2. Trigger fields set to "Processing" initially, then back to NULL once last Action is completed (done by Trigger processor)

The reasoning for (2) is to avoid a potential inconsistency when the front-end queries the the application after Submission (or Review, etc.). If the Actions are not yet complete, the Status will still show "Draft", even though it should be "Submitted". The plan now is for the front-end to also query the Trigger status and, if not NULL, then keep trying periodically until it is. Then the front-end can be sure the Actions are completed.

CarlosNZ:
Fixes #86 (Branch name is #77, due to starting with a previous issue)

Had to do some additional improvements before implementing this properly, namely: 
- adding an additional Database abstraction layer (`DBConnect`) which specifies which queries are sent to `PostgresDB` layer (most of them, at this point) or a `graphQLConnect` layer. This way, more complex queries (which will be required later) can be implemented with GraphQL while hiding implementation from main code.
- I realised the subquery to get the template ID from the incoming Trigger notification was only correct for the `application` table and wouldn't work for `review`, etc. Have pulled this subquery out of the `getActionsByTemplateId` method and created a new DB method called `getTemplateId`.
- changed the `action_queue` table to record the template_id, and made seperate fields for `parameter_queries` and `parameters_evaluated`, which contain the raw parameter expressions and evaluated expressions, respectively.
- the parameter expressions (`parameter_queries`) are evaluated during execution (which will make a difference for Scheduled actions and Sequential actions that need the result of an earlier action in their parameter_queries)
- renamed a few postgresConnect methods to more accurately reflect their function

**And now for the main issue:**

Actions can now be defined as **Sequential** or **Async**. The default is Async, which works pretty much like before -- they are added to the `action_queue` and picked up by the Action listener and executed immediately, so they can run in parallel.

Sequential Actions are implemented by putting an integer in the (new) `sequence` field in the `template_action` table, specifying the order in which simultaneously triggered actions should run.

When a trigger is triggered, the `processTrigger` method (in triggersAndActions.ts) does the following:

1. Sets the original trigger to an intermediate state: "Processing" _(new behaviour)_
2. Finds all Actions for the specified template and trigger _(same as previous)_
3. Remove all actions that don't match the `condition` expression. _(same as previous)_
4. Actions are separated into Sequential and Async arrays (if `sequence === null` then Async) _(new behaviour)_
5. Async actions are evaluated and sent to action_queue with a status of "Queued" _(new behaviour)_
6. Sequential actions are evaluated, sorted into the correct sequence, then written to action_queue with a status of "Processing"
7. The Action Listener service is notified of action_queue insertions, but it only acts on ones with a status of "Queued" (i.e. the Async actions), which are evaluated and executed immediately, regardless of the state of any other actions. Their evaluated parameter and Success/Error result is written back to the ActionQueue (along with completion timestamp)
8. Sequential actions are reloaded from the action_queue and evaluated/executed in the correct sequence, with the result status written back to action queue as above. Each Action must complete before the next starts.
9. Finally, the original trigger is set back to `null`.

The purpose of the trigger remaining in a "Processing" state is so the front end can tell when changes (e.g. status change from Draft to Submitted) are done -- i.e. the front-end should always query the trigger field and ensure that is is NULL whenever it needs to make another query as a result of a trigger (e.g. on Submission)

### To test:

The initial `User Registration` template is configured with 4 Actions -- 3 which are sequential, and one Async. Once the database has been re-initialised, run the following SQL command (simulating a user registration application submission) to trigger the Actions:

` UPDATE application SET trigger = 'onApplicationSubmit' WHERE id = 1`

### Still To-Do:

- [x]  I'll update the documentation for this PR before we merge it [DONE -- although diagram won't be updated until #90]
- [x]  For an Action to query the results of the previous action, it will have to re-query the database. This could result in some unreasonably complicated queries, since the only information that is passed through is the original trigger payload (id, table, record_id, trigger). I think it'd be a good idea to have an "output" object for each plugin that is written back to the action_queue and passed along to each subsequent action to be available for the evaluator. For example, the `createUser` action should output (at least) the username and user table id, making querying information about this much simpler for subsequent actions. I'll do a basic implementation to show how it'll work. [Initial implementation DONE in #90]

**Note**: Scheduled Actions are probably somewhat broken as a result of these changes. Will update them separately at some point once we have a discussion about how they should work.


src/components/graphQLConnect.ts
```diff
@@ -0,0 +1,59 @@
...
+          table
+          recordId
+          triggerType
+        }
+      }
+    }`
+    const variables = { record_id }
+    const result = await this.gqlQuery(query, variables)
+    const table = result.actionQueue.triggerQueueByTriggerEvent.table
+    const origRecord_id = result.actionQueue.triggerQueueByTriggerEvent.record_id
```
andreievg, line 45: wouldn't this be recordId ?

src/components/graphQLConnect.ts
```diff
@@ -0,0 +1,59 @@
+import fetch from 'node-fetch'
+import * as config from '../config.json'
+
+const endpoint = config.graphQLendpoint
+
+class GraphQLdb {
```
andreievg, line 6: I can see some use cases where GraphQL could be simpler when retrieving deep child relationships. 

not sure if this particular use case warrants a GraphQL statement, but I guess good start. Although I think would be nicer to review two PR's one for sequential actions and one for introducing GraphQL. 
No changes required just a minor comment

src/components/postgresConnect.ts
```diff
@@ -107,11 +109,11 @@ class PostgresDB {
     }
   }
 
-  public getActionsQueued = async (
+  public getActionsScheduled = async (
     payload: ActionQueueGetPayload = { status: 'Scheduled' }
   ): Promise<ActionQueue[]> => {
     const text =
-      'SELECT id, action_code, parameters, execution_time FROM action_queue WHERE status = $1 ORDER BY execution_time'
+      'SELECT id, action_code, trigger_payload, parameter_queries, time_completed FROM action_queue WHERE status = $1 ORDER BY time_completed'
```
andreievg, line 116: Hmm, this one is looking for actions that are scheduled for the future ? If so, kind of weird to have 'time_completed' in that case

src/components/postgresConnect.ts
```diff
@@ -123,6 +125,46 @@ class PostgresDB {
...
+    const text =
+      "SELECT id, action_code, trigger_payload, parameter_queries FROM action_queue WHERE template_id = $1 AND status = 'Processing' ORDER BY sequence"
+    try {
+      const result = await this.query({
+        text,
+        values: [templateId],
+      })
+      return result.rows as ActionQueue[]
+    } catch (err) {
+      throw err
```
andreievg, line 138: Was just looking at this file, and was wondering about catch(err) and throw err, is it there for future sakes ? when we want to handle errors in a more specific way ?

src/components/databaseConnect.ts
```diff
@@ -0,0 +1,73 @@
...
+class DBConnect {
+  private static _instance: DBConnect
+
+  constructor() {}
+
+  public static get Instance() {
+    return this._instance || (this._instance = new this())
+  }
+
+  public query = PostgresDB.query
```
andreievg, line 13: should we explicitly say this is pgQuery ?

src/components/triggersAndActions.ts
```diff
@@ -78,55 +89,103 @@ export const loadScheduledActions = async function (
...
+
+  // Execute Actions one by one
+  for (const action of actionsToExecute) {
+    const actionPayload = {
+      id: action.id,
+      code: action.action_code,
+      trigger_payload: action.trigger_payload,
+      parameter_queries: action.parameter_queries,
+    }
+    await executeAction(actionPayload, actionLibrary)
```
andreievg, line 146: does this really happen synchronously ? I thought for loops like this would need to be for await

src/components/triggersAndActions.ts
```diff
@@ -78,55 +89,103 @@ export const loadScheduledActions = async function (
...
+    }
   }
 
-  // Evaluate parameters for each Action
-  for (const action of actions) {
-    for (const key in action.parameter_queries) {
-      action.parameter_queries[key] = await evaluateExpression(action.parameter_queries[key], {
-        objects: [payload],
-        pgConnection: PostgresDB,
+  for (const actionList of [actionsAsync, actionsSequential]) {
```
andreievg, line 116: you can destructure both arrays and avoid nested for

src/components/triggersAndActions.ts
```diff
@@ -78,55 +89,103 @@ export const loadScheduledActions = async function (
...
 
   for (const action of result) {
     const condition = await evaluateExpression(action.condition, {
       objects: [payload],
-      pgConnection: PostgresDB,
+      pgConnection: DBConnect,
     })
-    if (condition) actions.push(action)
+    if (condition) {
+      if (action.sequence) actionsSequential.push(action as ActionSequential)
```
andreievg, line 111: can avoid extra code with different arrays by just checking for sequence for status, in loop below

andreievg (changes requested):
some minor changes and some questions, overall look pretty good

database/buildSchema/23_action_queue.sql
```diff
@@ -1,28 +1,35 @@
...
-    execution_time timestamp,
+    time_completed timestamp,
     error_log varchar
 );
 
 -- Function to Notify Action service of ActionQueue insert
 CREATE OR REPLACE FUNCTION public.notify_action_queue()
 RETURNS trigger as $action_event$
 BEGIN
+IF NEW.status = 'Queued' THEN
```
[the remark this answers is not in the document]
CarlosNZ: Yeah, true, although it's a WHEN clause in the trigger:
```
CREATE TRIGGER action_queue AFTER INSERT ON public.action_queue
FOR EACH ROW
WHEN (NEW.status <> 'Processing')
EXECUTE FUNCTION public.notify_action_queue();
```

Done in 3831574

database/buildSchema/24_template_action.sql
```diff
@@ -4,8 +4,8 @@ CREATE TABLE public.template_action (
     id serial primary key,
     template_id integer references public.template(id),
     action_code varchar,
-    previous_action_id integer references public.template_action(id),
```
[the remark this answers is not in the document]
CarlosNZ: `insertData.sql` is obsolete -- we've been using `insertData.js` for a while.

I've deleted the `.sql` file to avoid future confusion. (8b68e07)

src/components/databaseConnect.ts
```diff
@@ -0,0 +1,73 @@
+import PostgresDB from './postgresConnect'
```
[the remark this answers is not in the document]
CarlosNZ: Nicole recommended this structure as it makes it easy to switch a certain abstract query (e.g. `getTemplateId(table, record_id`) from Postgres to GraphQL without making any changes to the calling method.

We can quickly discuss this at our next meeting if you like, I don't think it'll be much work to change it either way. (New issue though)

src/components/graphQLConnect.ts
CarlosNZ: You mean it should be named `origRecordId`? Sure, done in 8b68e07.

src/components/graphQLConnect.ts
CarlosNZ: Yeah, I only started making the GraphQL abstraction when I was writing the `getTemplateId` method and realised that the query to get the original template id from some of the deeper tables (e.g. `review_response`) would be much easier to write with GraphQL.

However, I haven't actually implemented those queries yet, as we don't have Triggers on the tables nor any mock data to work with, so left them as To-Dos for now. (Only `application` and `review` table triggers have `getTemplateId` implemented so far, in SQL.)

src/components/postgresConnect.ts
CarlosNZ: Yeah, fair point. I originally had this field called `execution_time`, but I didn't like the inconsistent naming with `time_queued`. I guess we could have an additional column called `time_scheduled` (and just leave it `null` for non-scheduled actions), but let's discuss it before I re-address Scheduled actions (they're somewhat broken as a result of these changes -- have added a clarifying note to the original post here).

src/components/postgresConnect.ts
CarlosNZ: Yes, we can add more specific error messaging at some point, but my understanding is that this would simply pass any error thrown by the `node-postgres` module back up to the calling method, which can decide how to handle it. Is that correct?

src/components/databaseConnect.ts
CarlosNZ: Not sure what you mean here. 🤷‍♂️

src/components/triggersAndActions.ts
CarlosNZ: Yeah, that's why I'm using a `for...of` loop rather than `forEach`. ForEach doesn't work well within an async function, but `for...of`/`for...in` does (as well as traditional `for (let i; i<n; i++)` loops.

src/components/triggersAndActions.ts
CarlosNZ: Good call -- updated in 80103cd.

src/components/triggersAndActions.ts
CarlosNZ: Yes, done. :)

CarlosNZ:
Thanks for the comments, @andreievg. Have addressed them with either fixes or additional comments. 

Added an additional paragraph to the documentation, too. Still need to update the diagram, but I'll probably just do it once properly with #90.

src/components/graphQLConnect.ts
andreievg: i mean triggerQueueByTriggerEvent.`recordId`

src/components/postgresConnect.ts
andreievg: i like that idea, 'scheduled_time or time_scheduled`

src/components/postgresConnect.ts
andreievg: Yeah, I guess what I was trying to say if the error is caught and then thrown straight away, it's the same as not having the try/catch.

src/components/databaseConnect.ts
andreievg: If it's a single class encompassing both Postgres and GraphQL, and if it exposes query method, we probably want to differentiate if it's query method for postgres or GraphQL

src/components/triggersAndActions.ts
andreievg: great

src/components/triggersAndActions.ts
```diff
@@ -78,55 +89,100 @@ export const loadScheduledActions = async function (
...
+      trigger_event: trigger_id,
+      template_id: templateId,
+      sequence: action.sequence,
       action_code: action.code,
-      parameters: action.parameter_queries,
-      status: 'Queued',
+      trigger_payload: payload,
+      parameter_queries: action.parameter_queries,
+      parameters_evaluated: {},
+      status: action.sequence ? 'Processing' : 'Queued',
```
andreievg, line 127: would assume sequence would start at 1 ?

src/components/triggersAndActions.ts
```diff
@@ -78,55 +89,100 @@ export const loadScheduledActions = async function (
...
+  const evaluatorParams = {
+    objects: [payload.trigger_payload],
+    pgConnection: DBConnect, // Add graphQLConnection, Fetch (API) here when required
+  }
+  // Evaluate parameters
+  const parametersEvaluated = await evaluateParameters(payload.parameter_queries, evaluatorParams)
+
   // TO-DO: If Scheduled, create a Job instead
-  const actionResult = await actionLibrary[payload.code](payload.parameters, PostgresDB)
+  const actionResult = await actionLibrary[payload.code](parametersEvaluated, DBConnect)
```
andreievg, line 180: is it possible that actions will use evaluator expression ? If so they would need GraphQL ad Postgres instance injected. I guess we can store them in dbConnectInstance class

src/components/graphQLConnect.ts
CarlosNZ: Okay, have deleted this whole method, since it's not needed anymore.

src/components/postgresConnect.ts
CarlosNZ: Okay, I have added a `time_scheduled` column. However, I have NOT changed the scheduled actions to use this as they're currently broken and beyond this issue and current requirements. (So that column is just a placeholder for now.)

src/components/postgresConnect.ts
CarlosNZ: Okay, point taken. Since this is done like this in a lot of places, let's re-factor them all later on once we have a better strategy for error handling?

src/components/databaseConnect.ts
CarlosNZ: Okay, I see what you mean now. Only problem is that the DBConnect object that is passed to the evaluator requires a `.query` method on it to match the type requirements of the `node-postgres` connection object used by the evaluator.

Don't see a straightforward way around this right now.

src/components/triggersAndActions.ts
CarlosNZ: It actually doesn't matter. It just orders them, it doesn't care about the actual numbers. Sequence 2,4,10,12 would be the same as  0,1,2,3.

src/components/triggersAndActions.ts
CarlosNZ: Actions shouldn't need the evaluator. All parameters are evaluated before being passed in, and there's no need for it within the plugin as its only needed where expressions need to be user-facing.

src/components/triggersAndActions.ts
andreievg: 0 ? 'processing' :'queued' === 'queued'

src/components/databaseConnect.ts
andreievg: Why would evaluator need databaseConnect ? It should just get a singleton of Postgres and Graphql

To be honest I think we need to be quite clear about databaseConnect, and what it's there for. Would have preferred it to be in a seperate issue, so we can discuss what it does and why create it. 

If it's an abstraction, so we can combine both PG and GQL queries, then it should really be named something like dbQueryHelper. In which case we put actual logic of the queries in that class. And in that case Postgres and Graphql singletons are there to abstract connection and basic query logic.

CarlosNZ: Sure, yeah sorry, I should have discussed it before shoving it into this issue. It's just that I was wanting to write a GraphQL query instead of a complicated SQL query to get the templateIds of some of the more distant tables, and asked Nicole if it'd be okay to make some of the methods in PostgresConnect use GraphQL and she wasn't keen to mix-and-match and suggested this approach. In the end I didn't actually write the GraphQL queries yet (haven't implemented triggers on those tables), but I'm happy to re-do this if we want.

The evaluator doesn't need DBConnect per se, it just needs a valid PostGres connection (with a `.query` method, and a GraphQL object (with endpoint, fetch, etc.) if it's using those operators. DBConnect fits the bill for both, and the calling method (e.g. TriggersAndActions) will have only the DBConnect object available to pass in.

Shall I make a new issue to address this, or shall we have a chat about it first?

src/components/triggersAndActions.ts
CarlosNZ: Ah, I see -- you're saying that if `action.sequence ===0` then it'll evaluate to False, whereas only null/undefined should. Good point, will fix.

CarlosNZ: How about:
`status: typeof action.sequence === 'number' ? 'Processing' : 'Queued'`

andreievg (approved):
As per discussion in one of the comment, it would be great to make an issue to discuss DB abstraction. Especially when both access layers (node pg and graphql) can be used interchangeably from one class.

Issue #90, opened 2020-10-28
Output parameters for Action plugins


CarlosNZ:
See "To-Do" section of #89

PR #91, opened 2020-10-28
#90 Output parameters for Action plugins

CarlosNZ:
Implements Action output handling, as mentioned in previous PR #89.

I think having easy access to output properties of Actions in sequential actions will make writing parameter queries much simpler.

I also wrote a variant of the `createUser` plugin called `createUserFromApplication`. The new version handles the heavy lifting of retrieving the application responses to create the new user, which makes the parameter queries for the template_action in the UserRegistration template _much_ simpler. (Check the diff for `insertData.js`)

Note that this plugin in specifically tailored to the UserRegistration template and will need to be amended if the template changes. I think this is something we will continue to face -- the more specific the plugin, the more it's tied to specific template implementations, but with the upside of much simpler parameter handling, so useful to have a comparison here.

To see the output parameters being passed through, reset the database and run the same query as in #89:
`UPDATE application SET trigger = 'onApplicationSubmit' WHERE id = 1`

This time there is an additional `consoleLog` action at the end of the sequence which prints a message based on the output from the `createUserFromApplication` and `changeOutcome` Actions.

Updates to schema and plugin spec:
- `action_queue` has additional `output` field.
- `action_plugin` has additional `output_properties` field (this will just let future TemplateBuilder know what is available in subsequent plugins)
- `plugin.json` has additional `output_properties` field (for loading info into database)
- Plugins return object (previously containing only `success` and `error_log`) also has an `output` property, containing and object with the prescribed output values
- deleted the two "dummy" plugins, since we have enough real ones now. 

~~**Note:** Have merged #75 (UI for expression evaluator) into here because I needed to use it, and I assumed it would be merged to master before this PR, so the diff for this PR will be unnecessarily large until that is merged, sorry.~~ --Merged now.


nmadruga:
> deleted the two "dummy" plugins, since we have enough real ones now.

That's good! I'll start my review now after having read the description of PRs #89 and #91. Good explanations btw!

src/modules/expression-evaluator/src/evaluateExpression.test.ts
```diff
@@ -501,13 +501,10 @@ test('Input is an array', () => {
...
-    async () =>
-      await evaluateExpression(
-        '{"operator":"=", "children":[{"value":6},{"operator":"+", "children":[{"value":6},{"value":6}]}]}}'
-      )
-  ).rejects.toThrow('Invalid JSON String')
+test('Input is an string', () => {
+  return evaluateExpression('Friday drinks?').then((result: any) => {
+    expect(result).toEqual('Friday drinks?')
+  })
 })
```
nmadruga, line 508: I was wondering if there is any tests that can be added to cover the changes on action with output.

src/modules/expression-evaluator/expression-evaluate-gui/Z_To-Do.todo
```diff
@@ -0,0 +1,18 @@
+Important:
```
nmadruga, line 1: Oh no! I think all the `expression-evaluator-gui` changes end-up been added to this PR. Can you remove those please?

src/components/registerPlugins.ts
```diff
@@ -87,7 +88,8 @@ export default async function registerPlugins() {
         plugin.description !== dbPlugin.description ||
         plugin.path !== dbPlugin.path ||
         plugin.function_name !== dbPlugin.function_name ||
-        !deepEquality(plugin.required_parameters, dbPlugin.required_parameters))
+        !deepEquality(plugin.required_parameters, dbPlugin.required_parameters) ||
+        !deepEquality(plugin.output_properties, dbPlugin.output_properties, true))
```
nmadruga, line 92: Can this checks be moved to a utility function? I don't understand what this is checking. So I thought it would be easier to see if this was just a call a simple utility function (in the same file) so the code here looks clearar and the name of this function already says what this is doing.

nmadruga (commented):
Will finish this one tomorrow. And test it. I think I get the concept. Important one.

Just added a few comments now.

src/modules/expression-evaluator/expression-evaluate-gui/Z_To-Do.todo
CarlosNZ: It's merged with `master` and `#77-sequential-actions` now, so they no longer show up in the diff.

src/components/registerPlugins.ts
CarlosNZ: It's checking to see if the plugin details its just scanned from the folders have changed compared with what's in the database (then updates the database if so).

Moved comparison to own function at the bottom.

src/modules/expression-evaluator/expression-evaluate-gui/Z_To-Do.todo
nmadruga: Nice! I think Andrei had a look too. That's good.

src/components/registerPlugins.ts
nmadruga: Thanks!

src/plugins/action_create_user_from_application/src/createUserFromApplication.ts
```diff
@@ -0,0 +1,64 @@
+import { BasicObject } from '@openmsupply/expression-evaluator/lib/types'
+
+const responseMap = {
+  // Change these when question codes change
+  username: 'Q3',
+  first_name: 'Q1',
+  last_name: 'Q2',
+  email: 'Q4',
+  password_hash: 'Q5',
+}
```
nmadruga, line 10: I'm trying to understand how would this work when we allow different templates created.
Would we have one action per template to would store those fields hardcoded?
There would be a way to send which fields { title, codes } to use? 
Thinking on when we allow people to create a template or even when we have to maintain more than one template, how do we keep those up-to-date?

These are only some concerns, I don't think we need to make any changes to this PR - we can add new issues if we see any improvements that can be done and - maybe add some instructions to the wiki about how to create and maintain the actions (specially regarding the output parameters).

nmadruga (approved):
The code seems to be doing what it should. So approving this now! LGTM.

I just have concerns on how do we keep these output parameters up-to-date in actions (when any changes on fields may happen), but that can be in another issue.

I'll wait for your thought to create another issue for that - if required.

src/plugins/action_create_user_from_application/src/createUserFromApplication.ts
CarlosNZ: This Action is very much meant to be specifically tied to the current UserRegistration template. As I said in the original post, it just makes writing the lookup queries much simpler, but the generic `createUser` Action works fine too -- it just requires a lot more fiddling in the template config to extract the values required for input.

Since the User Registration template is going to be a "core" template, with probably not a lot of variation, I thought it made sense to have a dedicated handler for it.

CarlosNZ:
> The code seems to be doing what it should. So approving this now! LGTM.

👍 Will hold off on merging until Monday. @andreievg  wanted to have a bit more of a think about best way of handling passing through values between actions (if at all). Can clarify at the next Sprint meeting on Mon. Personally, I think it's useful to have the option, even if it doesn't get used that much.

> I just have concerns on how do we keep these output parameters up-to-date in actions (when any changes on fields may happen), but that can be in another issue.
> 
> I'll wait for your thought to create another issue for that - if required.

Yeah, although I'm not sure that's any more of an issue than keeping input parameters up-to-date in templates, is it? Let's discuss next week. :)

PR #94, opened 2020-11-02
#85 Actions for Change Application Stage and Status

CarlosNZ:
Created Actions "incrementStage" and "changeStatus", and inserted them into existing templates. Note that "incrementStage" automatically creates an associated "Draft" status record if none exists, so when creating a new application, only one call of "incrementAction" is required.

Also addressed the following as part of this PR:
- created Postgres functions to ensure that there can only ever be one `is_current = true` per stage and status: whenever `is_current` is set to `true`, the function automatically sets `is_current = false` for all other status or stage records for that application. (The plugin(s) also assumes that any new status or stage record will be the new `is_current` record, so that's applied automatically for every update)
- a few additional PostgresDB methods for getting and setting application statuses and stages.
- started a shared `types` file for plugins.

Also, I really don't understand the need for the adding the "Completed" status_history record when incrementing the Stage. It's never `is_current`, and it doesn't provide any additional information (and it's not a real record of any history). It'd be good if reviewer could suggest a situation/query where it could be at all helpful, because it just seems to add unnecessary confusion to me and serves no purpose. (Ideally I'd get rid of lines 50-58 of `incrementStage.ts`) 

**To test:**
In order to test from the front-end, you'll need my branch that adds the "Create" trigger. PR [here](https://github.com/openmsupply/application-manager-web-app/pull/79)
- reset the database and reload the server.
- Proceed to start a new user registration application from the web browser front-end.
- You should see a new application with Status "Draft" and Stage "Automatic" in the database (see the `application_stage_status` VIEW to see in combined view)
- Complete and Submit the application.
- The application should now be set to "Status: Completed" (and Outcome: Approved). Note: The user information is NOT added to the database correctly, as the current input plugin is not sending the response as correctly structured JSON, but the `createUserFromApplication` Action is working fine when responses are manually input.

To check other possible stage/status updates, use manual SQL queries against the other templates/applications:
- the Company Registration template has two stages, which you can test by updating the Trigger in the partial Company Registration application.
   - "onApplicationCreate" increments the stage (and creates new "Draft" status). You can just run this trigger multiple times to see it increment through all the stages -- not realistic, but demonstrates the plugin working.
   - "onApplicationSubmit" changes the status to "Submitted".
   - "onReviewSave" trigger (on Review table) also increments the stage.


src/plugins/action_change_status/src/changeStatus.ts
```diff
@@ -0,0 +1,86 @@
...
+        console.log(
+          `WARNING: Application ${applicationId} already has status: ${newStatus}. No changes were made.`
+        )
+        returnObject.status = 'Success'
+        returnObject.error_log = 'Status not changed'
+        returnObject.output = { status: newStatus, statusId: currentStatus.id }
+        break
+      default:
+        // Create a new application_status_history record
+        const result = await DBConnect.addNewStatusHistory(
```
andreievg, line 60: would prefer an early return of the returnObject and then add new status operation:

i.e
```
if(newStatus === currentStatus?.status)
   ...
  return returnObject
if(!currentStageHistoryId)
  ...
  return returnObject

... do the addition.
```

It's a stylistic comment, so not insisting it's changed to match above, I've suggested this way because of repetition of addition logic.

src/plugins/action_change_status/src/changeStatus.ts
```diff
@@ -0,0 +1,86 @@
...
+          console.log("Couldn't create new application_status_history")
+        }
+    }
+
+    // Create output object and return
+    if (returnObject.status !== 'Fail')
+      returnObject.output = { ...returnObject.output, applicationId }
+    return returnObject
+  } catch (err) {
+    console.log('Unable to change Status')
```
andreievg, line 80: would prefer message that are logged to be stored in the returnObject, and the whole object console.log in the caller of the action (if status == FAIL that is)

src/plugins/action_increment_stage/src/incrementStage.ts
```diff
@@ -0,0 +1,93 @@
...
+module.exports['incrementStage'] = async function (
+  parameters: IParameters,
+  DBConnect: any
+): Promise<ActionPluginOutput> {
+  const { applicationId } = parameters
+  const returnObject: ActionPluginOutput = { status: null, error_log: '' }
+  console.log(`Incrementing the Stage for Application ${applicationId}...`)
+
+  try {
+    const templateId: number = await DBConnect.getTemplateId('application', applicationId)
```
andreievg, line 22: I think the logic here can be simplified a bit, following this:

`getCurrentStatusHistory` (add number of application_stage_history to query)

if status is not undefined, create completed (should have everything for that from previous query), and set next status = draft, otherwise next status = status.

`getNextStageID` new SQL
```
select min(stageID) from template_stage
 where (number >  {number from `getCurrentStatusHistory` or 0} and template_id = {templateId} } 
```
if stageID undefined, then its last stage, return (can add some msg to returnObject)

add stage
add status

I personally think early returns of returnObject (on failure etc..) is much easier to follow

andreievg (changes requested):
Overall looks good, I think incrementStage can be much simpler, as per review. let me know what you think

nmadruga:
Testing this now:

1. Test Create the application will set **stage** + **status** to `Draft`:
- reset the database and reload the server.
- Proceed to start a new user registration application from the web browser front-end.
- [x] You should see a new application with Status "Draft" and Stage "Automatic" in the database (see the application_stage_status VIEW to see in combined view). **-PASS-**
2. Test Submit the application will keep the same **stage** and set **status** to `Completed`: 
- Complete and Submit the application.
- [x] The application should now be set to "Status: Completed" (and Outcome: Approved). **-PASS-**

nmadruga:
> Note: The user information is NOT added to the database correctly, as the current input plugin is not sending the response as correctly structured JSON, but the createUserFromApplication Action is working fine when responses are manually input.

Do we need to create another issue to fix that?

src/components/postgresConnect.ts
```diff
@@ -316,6 +316,82 @@ class PostgresDB {
...
+    try {
+      const result = await this.query({ text, values: [stageHistoryId] })
+      return result.rows[0]
+    } catch (err) {
+      console.log(err.message)
+      throw err
+    }
+  }
+
+  public addNewStatusHistory = async (stageHistoryId: number, status = 'Draft') => {
```
nmadruga, line 383:
```suggestion
  public addNewStatusHistory = async (stageHistoryId: number, status: ApplicationStatus = 'Draft') => {
```

And remember to import the type `ApplicationStatus` from the `graphql.ts`file. This would prevent using this method passing some status not included in the enum.

src/plugins/action_increment_stage/src/incrementStage.ts
```diff
@@ -0,0 +1,93 @@
...
+
+type IParameters = {
+  applicationId: number
+}
+
+type Stage = {
+  id: number
+  number: number
+  title: string
+}
```
nmadruga, line 11: Consider moving this to the `types` file

src/plugins/action_change_status/src/changeStatus.ts
```diff
@@ -0,0 +1,86 @@
+import { ActionPluginOutput } from '../../types'
+
+type Status =
+  | 'Draft'
+  | 'Withdrawn'
+  | 'Submitted'
+  | 'Changes Required'
+  | 'Re-submitted'
+  | 'Completed'
```
nmadruga, line 9: Consider moving this to the `types` file

nmadruga (changes requested):
Just a few comments. I think this is looking good overall, I like how it's organised and the `databaseConnect` although it's not a common design pattern - we though about using it to simplify this DAL with 2 different database access layers. I like how easy to read this is.

I think understanding why we create the new stage passing a `stage_id` which is a foreign key from `template_stage` is a bit confusing. I'm not sure if we can rename this to be clearer, but it's confusing on my mind. I'll revisit the schema documentation for that.

CarlosNZ:
> > Note: The user information is NOT added to the database correctly, as the current input plugin is not sending the response as correctly structured JSON, but the createUserFromApplication Action is working fine when responses are manually input.
> 
> Do we need to create another issue to fix that?

I'll probably try and include it when I do the issue to mutate Response after user interaction.

database/buildSchema/19_application_status_history.sql
```diff
@@ -7,5 +7,24 @@ CREATE TABLE public.application_status_history (
...
+    is_current bool DEFAULT true
+);
+
+
+-- FUNCTION to set `is_current` to false on all other status_histories of current application
+CREATE OR REPLACE FUNCTION public.status_is_current_update()
+RETURNS trigger as $application_status_history_event$
+BEGIN
+	UPDATE public.application_status_history SET is_current = false
+	WHERE application_stage_history_id = NEW.application_stage_history_id AND id<>NEW.id;
```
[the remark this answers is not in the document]
CarlosNZ: Okay, great -- I've used your auto-computed application_id column [below](https://github.com/openmsupply/application-manager-server/pull/94#discussion_r515854808) and modifed this function to use the new `application_id` column rather than `application_stage_history_id`.

So no gnarly query, and works regardless of connected stage. 👍

src/components/postgresConnect.ts
```diff
@@ -316,6 +316,82 @@ class PostgresDB {
...
+      throw err
+    }
+  }
+
+  public getCurrentStatusHistory = async (applicationId: number) => {
+    const text = `SELECT application_status_history.id,
+      application_stage_history_id, status FROM
+      application_status_history JOIN application_stage_history ON
+      application_stage_history_id = application_stage_history.id WHERE
+      application_id = $1 AND application_status_history.is_current = true;`
```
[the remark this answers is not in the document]
CarlosNZ: > This is linked to what I commented on `getCurrentStageHistory` method. Why not using the view `application_stage_status` instead?

Because you can't WRITE to a VIEW table, so the query to update the status_history `is_current` is still pretty nasty even when referring to the `application_stage_status` View.  I've used @andreievg 's auto-generated `application_id` column, and that works a treat :) 

However, I will use the VIEW table for the `get...` query below, so thanks for that suggestion.

src/plugins/action_change_status/src/changeStatus.ts
CarlosNZ: Righto, I've done something like that now and removed the addition repetition.

src/plugins/action_change_status/src/changeStatus.ts
CarlosNZ: Yeah makes sense. Have removed these and added a console log if `status==='Fail'` in the TriggersAndActions caller.

database/insertData.js
```diff
@@ -365,17 +391,58 @@ const queries = [
...
-              parameterQueries: {
-                message: {
-                  value: "The Action has been executed. Automated Actions FTW!!!"
+            create: [
+              {
+                actionCode: "cLog"
+                condition: { value: true }
+                trigger: ON_APPLICATION_SUBMIT
+                parameterQueries: {
+                  message: { value: "Company Registration submission" }
```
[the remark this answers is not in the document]
CarlosNZ: Yeah, sure.

database/insertData.js
```diff
@@ -365,17 +391,58 @@ const queries = [
...
+                    children: [
+                      { value: "SELECT application_id FROM review WHERE id = $1" }
+                      {
+                        operator: "objectProperties"
+                        children: [{ value: { property: "record_id" } }]
+                      }
+                    ]
+                  }
+                }
+              }
```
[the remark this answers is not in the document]
CarlosNZ: Just another example, really, for testing the actions work fine with different triggers on different tables.

The query is different, because from the Review table the `record_id` is not the `application_id`, so it needs to do a bit more logic to deduce the application id. I'm actually thinking it might be a good idea to add some additional functionality to the triggers so that `application_id` and `template_id` (and maybe some other values?) always get passed to the Trigger Processor in the notification payload, to avoid having to do these additional lookups for common values.

Have made a card in the backlog: https://github.com/orgs/openmsupply/projects/6#card-48677106

src/components/postgresConnect.ts
```diff
@@ -316,6 +316,82 @@ class PostgresDB {
...
+      return result.rows
+    } catch (err) {
+      console.log(err.message)
+      throw err
+    }
+  }
+
+  public getCurrentStageHistory = async (applicationId: number) => {
+    const text =
+      'SELECT id, stage_id FROM application_stage_history WHERE application_id = $1 and is_current = true'
```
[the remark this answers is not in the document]
CarlosNZ: Good point.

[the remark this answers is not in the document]
CarlosNZ: I think the reason I didn't was because the current `application_stage_status` view doesn't include the id's and stuff, which we need for creating the subsequent status_history record.

I've made another view now called `application_stage_status` that has all the relevant fields from the `stage`, `application_stage_history` and `application_status_history` tables, which will be more useful for querying.

src/components/postgresConnect.ts
```diff
@@ -316,6 +316,82 @@ class PostgresDB {
...
+    const text =
+      'SELECT id, status FROM application_status_history WHERE application_stage_history_id = $1 AND is_current = true'
+    try {
+      const result = await this.query({ text, values: [stageHistoryId] })
+      return result.rows[0]
+    } catch (err) {
+      console.log(err.message)
+      throw err
+    }
+  }
```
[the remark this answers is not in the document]
CarlosNZ: This searches by stageHistoryId, the other one searches by applicationId

src/components/postgresConnect.ts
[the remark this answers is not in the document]
CarlosNZ: > And this one could also return the current status altogether.

Only if they're correctly linked (i.e every current status history is linked to a current stage history), but if they're not, getting them both together gets very complicated.

src/components/postgresConnect.ts
CarlosNZ: I just tried this, and the problem is that the GraphQL enum is expecting "DRAFT" (in caps), whereas Postgres is expecting "Draft". It seems like in general it's quite difficult to use the auto-generated GraphQL types in other contexts which haven't undergone the inflections.

src/plugins/action_increment_stage/src/incrementStage.ts
CarlosNZ: Hmm, I thought this one would probably be unique to this plugin. How about moving it once we use it in another plugin?

src/plugins/action_change_status/src/changeStatus.ts
CarlosNZ: Done. :)

src/plugins/action_increment_stage/src/incrementStage.ts
CarlosNZ: Thanks for this suggestion. I've refactored it quite a bit and it's a lot clearer. The new VIEW which includes all stage_status data combined (including IDs, which the old one lacked) makes it simpler as well.

Will push shortly :)

CarlosNZ:
Okay, have re-factored this quite a bit now to take into account both your suggestions @nmadruga  and @andreievg .

Basically:
- Changed the VIEW `application_stage_status` to a more comprehensive view called `application_stage_status_all`, which includes all relevant columns from `template_stage`, `application_stage_history` and `application_status_history`, renamed appropriately. (Note that the original View is not particularly useful now, as we have computed columns showing current stage_num, stage and status in the application table in GraphQL)
- Used the above View in queries where appropriate
- Added computed/generated column `application_id` to `application_status_table` (which makes querying status by application_id _waay_ easier
- Simplified plugin logic substantially, including more use of early returns in error conditions

andreievg (approved):
Looks great! Thanks for changes

database/insertData.js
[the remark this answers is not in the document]
CarlosNZ: Have actually done this #96, with quite a bit of extra documentation, which should make it easier to follow.

src/components/postgresConnect.ts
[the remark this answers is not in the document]
CarlosNZ: I've actually removed one of them now, since it's not needed.

src/components/postgresConnect.ts
CarlosNZ: > > > Note: The user information is NOT added to the database correctly, as the current input plugin is not sending the response as correctly structured JSON, but the createUserFromApplication Action is working fine when responses are manually input.
> > 
> > 
> > Do we need to create another issue to fix that?
> 
> I'll probably try and include it when I do the issue to mutate Response after user interaction.

It's done in #34 (Mostly, it only sets the `text` field, we need to see how to handle the other two once we make plugins that use them)

PR #95, opened 2020-11-02
Template element schema adjustments

CarlosNZ:
Implemented the things marked "TO-DO" in "Template element fields" in #61 (and updated Templates accordingly).

Will be useful for implementing Response local state in front-end (working on it now).

Have updated front-end docs with the specs, and added a link in these docs to that (not active yet, but will be once front-end issue is published).

### DIFF
--- a/database/buildSchema/14_template_element.sql
+++ b/database/buildSchema/14_template_element.sql
@@ -9,10 +9,11 @@ CREATE TABLE public.template_element (
     index integer,
     title varchar,
     category public.template_element_category,
-    visibility_condition jsonb,
+    visibility_condition jsonb DEFAULT '{"value":true}'::jsonb,
     element_type_plugin_code varchar,
-    is_required boolean,
-    is_editable jsonb,
+    is_required jsonb DEFAULT '{"value":true}'::jsonb,
+    is_editable jsonb DEFAULT '{"value":true}'::jsonb,
     parameters jsonb,
-    validation jsonb
+    validation jsonb DEFAULT '{"value":true}'::jsonb,
+    validation_message varchar
 );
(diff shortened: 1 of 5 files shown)


database/insertData.js
```diff
@@ -39,33 +38,23 @@ const queries = [
                       index: 1
                       title: "First Name"
                       elementTypePluginCode: "shortText"
-                      visibilityCondition: { value: true }
                       category: QUESTION
-                      isRequired: true
```
andreievg, line 44: isn't it jsonb now ?

database/buildSchema/14_template_element.sql
```diff
@@ -9,10 +9,11 @@ CREATE TABLE public.template_element (
...
+    visibility_condition jsonb DEFAULT '{"value":true}'::jsonb,
     element_type_plugin_code varchar,
-    is_required boolean,
-    is_editable jsonb,
+    is_required jsonb DEFAULT '{"value":true}'::jsonb,
+    is_editable jsonb DEFAULT '{"value":true}'::jsonb,
     parameters jsonb,
-    validation jsonb
+    validation jsonb DEFAULT '{"value":true}'::jsonb,
+    validation_message varchar
```
andreievg, line 18: Hmm, I have a feeling we might decommission these two fields. but doesn't hurt having them here now

andreievg (changes requested):
just one little comment.

i rebased to 85, so would need to base base to master once 85 is merged

database/insertData.js
CarlosNZ: Yes, it is, but single literals are actually valid JSON, and the evaluator accepts and returns them correctly, so this would be fine.

But I actually took all these out of the script, since I made `{"value": true}` the default, so not sure why you're seeing this..

nmadruga (approved):

andreievg (approved):

Issue #84, opened 2020-10-27
Back-end creation of JWT with user permissions + login endpoint


nmadruga:

andreievg:
Base on the epic and summary in https://github.com/openmsupply/application-manager-web-app/issues/65

Two end points:

GET /userPermission (Authorisation with token)

Response: 

```JSON
{
  " templatePermissions" : {
     "applicationType": [ 
          {
               "permissionType": "type",
               "stage": "number",
               "anything else"
          }
     ]  
   },
   "username": "username",
   "JWT": "JWT token with just username for now"
}
```
i.e.
```JSON
{
  " templatePermissions" : {
        "updateUserDetail": [ 
            {
                "permissionType": "apply"
           }
         ] ,

       "drugRegistration": [ 
          {
               "permissionType": "review",
               "canConsolidate": "true",
               "stage": 1,
               "level": 1
          },
          {
               "permissionType": "review",
               "canConsolidate": "false",
               "stage": 2,
               "level": 1
          }
       ]   
   },
   "username": "username",
   "JWT": "JWT token with just username for now"
}
```


POST /login BODY : { "username": ... , "password":... { already hashed} }

Response:

`on success`: same as above, but add a key, "success": true

`o failure`: { "success": false }

PR #101, opened 2020-11-12
#84 create login and user permission end points

andreievg:
Fixes: https://github.com/openmsupply/application-manager-server/issues/84

### Dependencies 

+ jsonwebtoken and types (not sure that i've used types though)

### Data

Added two new users ('nonRegistered', 'userWithMultiplePermissions').

Added three policies: oneTimeApply, basicApply, basicReview (no rules etc.., just types)

Added three permissions names: applyUserRegistration, applyCompanyRego, reviewCompanyRego (linked to policies in same order as mentioned)

Added three template permission: permission names joined to templates (as per names of permission name), no restrictions just join

Added join between users and permission names:
nonRegisterd -> applyUserRegistration
All others apart from multi permission user -> applyCompanyRego
userWithMultiplePermissions -> all permission names

### End Point

`GET` userPermission

* Expects a JWT in authorisation header (Authorisation: Bearer `token`), it can be missing. If token is missing or not valid, then return nonRegistered user permissions
* Response below

`POST` login

* Expects json body with username and passwordHash
* Fail response: {  success: false }

userPermission and successful login return the same result (except login would add success: true), as follows:

```JSON
{
   "username": "{username}",
   "JWT": "{signed jwt, for now just object with username}",
   "templatePermissions": { "templateName": [ "permission types, form policy" ] }
}
```
Can just base64 decode JWT to see what's inside

### Postman collection

[IRIMS permissions.postman_collection.json.zip](https://github.com/openmsupply/application-manager-server/files/5528070/IRIMS.permissions.postman_collection.json.zip)

### Extra

Secret is currently in config.json

Can run `database_init`, and check with postman collection


src/components/permissions.ts
```diff
@@ -0,0 +1,76 @@
...
+
+  templatePermissionRows.forEach(({ permissionType, templateName }: any) => {
+    if (!templatePermissions[templateName]) templatePermissions[templateName] = []
+    templatePermissions[templateName].push(permissionType)
+  })
+
+  return templatePermissions
+}
+
+const getJWT = async (username: String, templatePermissionRows: Array<PermissionRow>) => {
```
CarlosNZ, line 62: `templatePermissionRows` is not used in this function -- is that intended for later (To-Do)?

CarlosNZ (commented):
Cool -- will get you to talk us through some aspects of this on Monday :)

src/components/permissions.ts
andreievg: yep, it was a copy paste and node used currently but most likely will be needed.

great, Monday chat it is.

src/components/permissions.ts
```diff
@@ -0,0 +1,76 @@
...
+  return await signPromise({ username }, config.jwtSecret) // TODO gaphile/pg row lvl, and token
+}
+
+const getUserPermissions = async (username: string) => {
+  const templatePermissionRows = await databaseConnect.getUserTemplatePermissions(username)
+
+  return {
+    username,
+    templatePermissions: getTemplatePermissions(templatePermissionRows),
```
nmadruga, line 71: I think this method could be renamed to clear out the meaning - it's not getting the template permission, is transforming into a object to be returned. Maybe `getTemplatePermissions` => `buildTemplatePermissionsObject`

nmadruga: And maybe move the method to be after this one, it would make the workflow clearer

nmadruga (commented):
Nice work. I made just 2 comment with improvements.

src/components/permissions.ts
andreievg: yeah, that's way better thanks: 
08f4e194c7722923162233ab66fb94437684ab27

src/components/permissions.ts
```diff
@@ -0,0 +1,76 @@
+import databaseConnect from './databaseConnect'
+import config from '../config.json'
+import { verify, sign } from 'jsonwebtoken'
+import { promisify } from 'util'
+
+type PermissionType = 'Apply' | 'Review' | 'Assign'
```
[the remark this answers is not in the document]
andreievg: With some help from ssnzdevs: 506f97ad7ba58bfb6e3945e223a71ae0fad4b1e5

nmadruga (approved):
Great. Appreciated the changes, and tested now too. Everything is working well.

Just one overall comment: the issue description says the endpoint is called `userPermission` but it's actually `userPermissions`. Not sure which one you would prefer to use, but if we are sticking with the current, just need to remember to use the correct one in the docs.

andreievg:
> Just one overall comment: the issue description says the endpoint is called userPermission but it's actually userPermissions. Not sure which one you would prefer to use, but if we are sticking with the current, just need to remember to use the correct one in the docs.

Yep, I'll make sure to use correct end point when doing docs

PR #102, opened 2020-11-12
Bug fix -- template definition tweak

CarlosNZ:
Needed to go with front-end bug-fixes. (https://github.com/openmsupply/application-manager-web-app/pull/112)

Just changes the fields sent to `createUser` Action to `responses.Q1.text` rather than `responses.Q1`, so it handles correctly formatted Response object.

Reset database (`yarn database_init`) and restart server before testing.

### DIFF
--- a/database/insertData.js
+++ b/database/insertData.js
@@ -207,23 +207,23 @@ const queries = [
                 parameterQueries: {
                   first_name: {
                     operator: "objectProperties"
-                    children: [{ value: { property: "responses.Q1" } }]
+                    children: [{ value: { property: "responses.Q1.text" } }]
                   }
                   last_name: {
                     operator: "objectProperties"
-                    children: [{ value: { property: "responses.Q2" } }]
+                    children: [{ value: { property: "responses.Q2.text" } }]
                   }
                   username: {
                     operator: "objectProperties"
-                    children: [{ value: { property: "responses.Q3" } }]
+                    children: [{ value: { property: "responses.Q3.text" } }]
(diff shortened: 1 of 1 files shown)


database/insertData.js
```diff
@@ -207,23 +207,23 @@ const queries = [
                 parameterQueries: {
                   first_name: {
                     operator: "objectProperties"
-                    children: [{ value: { property: "responses.Q1" } }]
+                    children: [{ value: { property: "responses.Q1.text" } }]
```
andreievg, line 210: What is the difference between this and:

https://github.com/openmsupply/application-manager-server/blob/1e151905e1851c39c0a1d076a0a3ef5676f98717/database/insertData.js#L112

Would be great to get both of them aligned.

Maybe plugins onChange callback should insist on taking 'textValue', 'value' parameters. Thus whenever we reference question response in evaluator we know to reference textValue ?

CarlosNZ: The difference is mainly in the object that is being passed in. Previously (in the other code you linked to), we didn't yet have an established shape for the reponses object in the front-end, so I was just making an assumption.

Now we have a clearer idea of how it will be exposed. And yes, we could always have just the "textValue" value, but there may be times when we need to know a little more information from one of the other fields in the response object (e.g. "reference" -- to link a choice with a specific database entity), so I thought it was safer to just include the whole field, with the added expense of having to add `.text` to the query.

andreievg (approved):

PR #120, opened 2020-11-18
#81F - Add new field is_linear in the Template

nmadruga:
PR to be merge along [Front-end#85](https://github.com/openmsupply/application-manager-web-app/issues/81) issue - in a PR soon!

Added new `is_linear` field on the Template entity.

### Insert data
Use the `isLinear` to set one application as linear and the other as non-linear for testing.
Also have added a few elements with `isRequired` equals false for testing purposes.

### DIFF
--- a/database/buildSchema/07_template.sql
+++ b/database/buildSchema/07_template.sql
@@ -6,6 +6,7 @@ CREATE TABLE public.template (
     id serial primary key,
     name varchar,
 	code varchar NOT NULL,
+    is_linear boolean,
     status public.template_status,
     version_timestamp timestamp   
 );
--- a/database/buildSchema/07_template.sql
+++ b/database/buildSchema/07_template.sql
@@ -6,7 +6,7 @@ CREATE TABLE public.template (
     id serial primary key,
     name varchar,
 	code varchar NOT NULL,
-    is_linear boolean,
+is_linear boolean DEFAULT true,
     status public.template_status,
     version_timestamp timestamp   
 );
--- a/database/insertData.js
+++ b/database/insertData.js
@@ -12,6 +12,7 @@ const queries = [
         template: {
           code: "UserRego1"
           name: "User Registration"
+          isLinear: true 
           status: AVAILABLE
           versionTimestamp: "NOW()"
           templateSectionsUsingId: {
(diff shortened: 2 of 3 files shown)


database/buildSchema/07_template.sql
```diff
@@ -6,6 +6,7 @@ CREATE TABLE public.template (
     id serial primary key,
     name varchar,
 	code varchar NOT NULL,
+    is_linear boolean,
```
CarlosNZ, line 9:
```suggestion
is_linear boolean DEFAULT true,
```

database/insertData.js
```diff
@@ -97,6 +98,7 @@ const queries = [
                         label: "What category of organisation do you wish to join"
                         options: ["Manufacturer", "Distributor", "Importer"]
                       }
+                      isRequired: false
```
CarlosNZ, line 101: Did you mean to make this one `isRequired: false`? Wouldn't it be Q7 (Select Manufacturer) rather than this one, since Q7 is dependent on this (and you've made the other two conditional questions `isRequired: false`

database/insertData.js
```diff
@@ -12,6 +12,7 @@ const queries = [
         template: {
           code: "UserRego1"
           name: "User Registration"
+          isLinear: true
```
CarlosNZ, line 15: If you set the default as suggested above, then you can omit this line.

CarlosNZ (changes requested):
Suggest using default value = true.

database/insertData.js
nmadruga: I think that's what I wanted. Then I'll be checking the following (for the progress bar):
- Q6: `{ isRequired: false + isVisible: True }`  => So this can be empty and the progress can continue.
- Q7: `{ isRequired: true + isVisible: false }` => If the Q6 is empty this one isn't visible - so the process can continue.
- Q7: `{ isRequired: true + isVisible: true }` => If the Q6 has an answer now this question is expected to be filled.

database/insertData.js
nmadruga: Done! Thanks

database/insertData.js
CarlosNZ: Okay, just not sure why Q7 is required and Q8/9 are not, when all three of them are dependent on the selection in Q6.

nmadruga: Just for testing purposes. I wanted to test one question that `isRequired` but can be visible or not. Instead of adding the same to all the 3 I have made only one example to check.

CarlosNZ: Ah, okay. Sweet.

CarlosNZ (approved):

Issue #103, opened 2020-11-13
Create action to grant permissions to user


nmadruga:
Create basic action to give gran **any** permission name to **any** user. This will be used later on for linking the users with templates that they can apply for. Also needs to consider other use cases, for example linking reviewers to templates to review, etc.

PR #123, opened 2020-11-20
#103 grant permission action

andreievg:
Fixes: #103 

Created action to grant permissions to a user.

It accepts 'username' and 'permissionNames' <- array, creates permission join.

In this branch userRego is configured to grant applyCompanyRego right to the user.

To test, check user_permission_join table, create new user via registration (btw I've noticed on current master second page is not quite working, but can skip it and go straight to summary). After submission of registration a new record will be added to user_permission_join table.

PR itself is quite small but the build js and yarn lock show a huge diff


src/plugins/action_grant_permissions/src/grantPermissions.ts
```diff
@@ -0,0 +1,27 @@
...
+    )
+
+    for (const permissionName of permissionNames) {
+      if (currentUserPermisionsNames.includes(permissionName)) continue
+      await DBConnect.joinPermissionNameToUser(username, permissionName)
+      currentUserPermisionsNames.push(permissionName)
+    }
+    return {
+      status: 'Success',
+      error_log: '',
```
CarlosNZ, line 18:
```suggestion
      error_log: '',
      output: {
        permissionNames,
     },
```
I think a minimal output object is useful and consistent with the other plugins. It's also useful for debugging purposes.

src/components/postgresConnect.ts
```diff
@@ -499,6 +499,38 @@ class PostgresDB {
...
+      throw err
+    }
+  }
+
+  public joinPermissionNameToUser = async (username: string, permissionName: string) => {
+    const text = `
+    insert into permission_join (user_id, permission_name_id) 
+    values (
+        (select id from "user" where username = $1), 
+        (select id from permission_name where name = $2))
```
CarlosNZ, line 525:
``` suggestion
        (select id from permission_name where name = $2))
        RETURNING id
```
Useful if subsequent plugins wish to interact with this record (see output additions below).

src/components/postgresConnect.ts
```diff
@@ -499,6 +499,38 @@ class PostgresDB {
...
+    `
+    try {
+      const result = await this.query({ text, values: [username, permissionName] })
```
CarlosNZ, line 528:
```suggestion
    const result = await this.query({ text, values: [username, permissionName] })
    return { permissionJoinId: result.rows[0].id, success: true }
```

src/plugins/action_grant_permissions/src/grantPermissions.test.ts
```diff
@@ -0,0 +1,19 @@
...
+const testParams = {
+  username: 'valerio',
+  permissionNames: ['reviewCompanyRego'],
+}
+
+test('Test: Add permission to Valerio', () => {
+  return Action.grantPermissions(testParams, PostgresDB).then((result: any) => {
+    expect(result).toEqual({
+      status: 'Success',
+      error_log: '',
```
CarlosNZ, line 16: If you add the `output` field (see below), then you'll need to adjust this accordingly.

database/insertData.js
```diff
@@ -316,9 +316,23 @@ const queries = [
...
               }
               {
-                actionCode: "changeStatus"
+                actionCode: "grantPermissions"
                 trigger: ON_APPLICATION_SUBMIT
                 sequence: 3
+                parameterQueries: {
+                  username: {
+                    operator: "objectProperties"
+                    children: [{ value: { property: "responses.Q3.text" } }]
```
CarlosNZ, line 325: Personally, I prefer to use the output of the previous Action to provide the username, but this works fine in this case, so not to worry.

CarlosNZ (changes requested):
Also, need to increment the "Sequence" number of the last Action from 5 to 6:

https://github.com/openmsupply/application-manager-server/blob/b833c78a80e1f60668e03c8c002e8e2af3e9546e/database/insertData.js#L359

src/plugins/action_grant_permissions/src/grantPermissions.ts
```diff
@@ -0,0 +1,27 @@
+module.exports['grantPermissions'] = async function (
+  { username, permissionNames }: any,
+  DBConnect: any
+) {
+  try {
+    console.log('\nGranting permission/s:')
```
CarlosNZ, line 6:
```suggestion
   console.log(`\nGranting permission/s: ${permissionNames}`)
   console.log(`\nto user: ${username}`)

**Comment by CarlosNZ** [comment]
Also, you probably want to update the docs for the "List-of-Action-plugins" page:
https://github.com/openmsupply/application-manager-server/wiki/List-of-Action-plugins

File: database/insertData.js (hunk 5 above)
Comment (andreievg) [reply database/insertData.js:325]: I don't think actions should be coupled and reliant on each other, even in configurations

File: src/components/postgresConnect.ts (hunk 2 above)
Comment (andreievg) [reply src/components/postgresConnect.ts:525]: While we are in development phase, I don't really see the need to over optimise plugins. What would be a use case for action needing permission join id ?

File: src/plugins/action_grant_permissions/package.json [hunk 7]
```diff
@@ -0,0 +1,21 @@
+{
+  "name": "console_log_action",
+  "version": "0.1",
+  "description": "All it does is print a message to the console.",
+  "main": "grantPermissions.ts",
+  "author": "The mSupplyFoundation",
+  "license": "MIT",
+  "scripts": {
+    "dev": "echo 'Please run yarn dev from main app folder'",
```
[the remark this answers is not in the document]
andreievg: 7c406f0e6add387ad9a840fd563cd05942a1cf56

src/plugins/action_grant_permissions/src/grantPermissions.ts
andreievg: 7c406f0e6add387ad9a840fd563cd05942a1cf56

andreievg:
@CarlosNZ thanks for review

Let me know how strongly you feel that plugin should return output. I think it should be done by use case bases while we in dev mode.

> Also, you probably want to update the docs for the "List-of-Action-plugins" page:
https://github.com/openmsupply/application-manager-server/wiki/List-of-Action-plugins

Sweet, I've taken a note to update this

src/components/postgresConnect.ts
CarlosNZ: In case subsequent Actions needed to do something with the new permission_join record. But sure, it's not needed for now.

CarlosNZ:
> Let me know how strongly you feel that plugin should return output. I think it should be done by use case bases while we in dev mode.

I agree that in this case any output values are probably not going to be very useful. But as a general practice, I think providing minimal useful outputs from Actions is a good idea. Even if they don't really get used much, I don't see a downside.

CarlosNZ (approved):

andreievg:
> I agree that in this case any output values are probably not going to be very useful. But as a general practice, I think providing minimal useful outputs from Actions is a good idea. Even if they don't really get used much, I don't see a downside.

Ok, will do for next action

PR #125, opened 2020-11-23
#126F Plugin generalisation & Dropdown plugin

CarlosNZ:
Back-end counterpart to front-end issue #126/PR132: https://github.com/openmsupply/application-manager-web-app/pull/132 

NOTE: Don't merge this until that issue is merged to `master`.

All this does is update the UserRego template to correctly use the new DropdownSelector plugin.

### DIFF
--- a/database/insertData.js
+++ b/database/insertData.js
@@ -261,7 +261,11 @@ const queries = [
                       parameters: {
                         label: "Select Importer"
                         placeholder: "Select"
-                        options: ["Importer A", "Importer B", "Importer C"]
+                        options: [
+                          "Importer A", 
+                          "Importer B", 
+                          "Importer C"
+                        ]
                         validation: { value: true }
                       }
                       isRequired: false
--- a/database/insertData.js
+++ b/database/insertData.js
@@ -172,11 +172,14 @@ const queries = [
                       code: "Q6"
                       index: 7
                       title: "Organisation Category"
-                      elementTypePluginCode: "radioChoice"
+                      elementTypePluginCode: "dropdownChoice"
+                      # Change this to "radioChoice" once we've made the plugin
                       category: QUESTION
                       parameters: {
-                        label: "What category of organisation do you wish to join"
+                        label: "Organisation Type"
+                        description: "Select which type of organisation you belong to."
                         options: ["Manufacturer", "Distributor", "Importer"]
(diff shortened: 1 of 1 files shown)


database/insertData.js
```diff
@@ -172,11 +172,14 @@ const queries = [
...
-                      elementTypePluginCode: "radioChoice"
+                      elementTypePluginCode: "dropdownChoice"
+                      # Change this to "radioChoice" once we've made the plugin
                       category: QUESTION
                       parameters: {
-                        label: "What category of organisation do you wish to join"
+                        label: "Organisation Type"
+                        description: "Select which type of organisation you belong to."
                         options: ["Manufacturer", "Distributor", "Importer"]
+                        validation: { value: true }
```
nmadruga, line 182: Why do you need this? I thought this would be the default behaviour

database/insertData.js
```diff
@@ -252,11 +259,9 @@ const queries = [
                       category: QUESTION
                       parameters: {
                         label: "Select Importer"
-                        options: [
-                          "Importer A",
-                          "Importer B",
-                          "Importer C"
-                        ]
+                        placeholder: "Select"
+                        options: ["Importer A", "Importer B", "Importer C"]
```
nmadruga, line 263:
```suggestion
                        options: [
                          "Importer A", 
                          "Importer B", 
                          "Importer C"
                        ]
```

I think that's easier to read

nmadruga (commented):
Hi, I'm going to test the front-end issue now.

Thanks for fixing this dropdown, I'm very happy to see this one is done now! Just made 2 comments, please have a look

database/insertData.js
CarlosNZ: Need to be careful here as the Query editor doesn't want commas when listing vertically, so it's non-standard JSON.

Also -- I don't think we want to get in the habit of over-riding the auto-formatting. I always copy this whole query into the Web GUI to check it's correct, and it auto-formats it -- and we don't want to have to remember to do a bunch of manual changes every time we paste it back.

database/insertData.js
CarlosNZ: Yes, it wasn't -- but it is now.

nmadruga (approved):
Good to go : D

PR #127, opened 2020-11-23
#81F Add many elements and new pages to Create company application template

nmadruga:


database/insertData.js
```diff
@@ -442,6 +453,57 @@ const queries = [
...
+                      title: "Intro Section 1 - Page 2/2"
+                      elementTypePluginCode: "textInfo"
+                      category: INFORMATION
+                      parameters: { title: "Company nationality" }
+                    },
+                    {
+                      code: "S1Q3"
+                      index: 5
+                      title: "Organisation national or international"
+                      elementTypePluginCode: "radioChoice"
```
CarlosNZ, line 477: For now can we use `dropdownChoice` so we can actually see the plugin on the page?

database/insertData.js
```diff
@@ -442,6 +453,57 @@ const queries = [
...
                         ]
                       }
+                    },
+                    {
+                      code: "S1PB1"
+                      index: 3
+                      title: "Page Break"
+                      elementTypePluginCode: "pageBreak"
+                      category: INFORMATION
+                      parameters: { previousValidityCheck: true }
```
CarlosNZ, line 463: Just FYI -- I think we've removed the need for this parameter since we've gone with `linear`/`non-linear` setting for whole application.

database/insertData.js
```diff
@@ -442,6 +453,57 @@ const queries = [
...
+                          "National",
+                          "International"
+                        ]
+                      }
+                    },
+                    {
+                      code: "S1Q4"
+                      index: 6
+                      title: "Import permit upload"
+                      elementTypePluginCode: "textInfo"
```
CarlosNZ, line 491:
```suggesition
elementTypePluginCode: "textInfo"
category: QUESTION
```
Needs category -- otherwise doesn't show on page.

database/insertData.js
```diff
@@ -442,6 +453,57 @@ const queries = [
...
+                      # Remember to pass Responses object into visibilityCondition
```
CarlosNZ, line 492: Can delete this line now.

database/insertData.js
```diff
@@ -453,25 +515,132 @@ const queries = [
...
+                    {
+                      code: "S2Q1"
+                      index: 1
                       title: "Organisation Street"
                       elementTypePluginCode: "shortText"
                       category: QUESTION
-                      parameters: { label: "Start typing the company street" }
+                      parameters: { label: "Enter the company street" }
+                      isRequired: false
+                      # Remember to pass Responses object into visibilityCondition
```
CarlosNZ, line 544: Delete

database/insertData.js
```diff
@@ -453,25 +515,132 @@ const queries = [
...
+                        ]
+                      }
+                    },
+                    {
+                      code: "S2PB1"
+                      index: 3
+                      title: "Page Break"
+                      elementTypePluginCode: "pageBreak"
+                      category: INFORMATION
+                      parameters: { previousValidityCheck: true }
```
CarlosNZ, line 582: Don't need this parameter (as above)

database/insertData.js
```diff
@@ -453,25 +515,132 @@ const queries = [
...
+                      # Remember to pass Responses object into visibilityCondition
```
CarlosNZ, line 583: Delete comment

CarlosNZ (changes requested):
Cool, thanks for this.

Can't really test it fully in the front-end until the Drop-Down plugin is merged, so maybe you want to have a look at that next?

database/insertData.js
nmadruga: Sure. I'm changing to this now.

Should be able to look have another look on [PR issue#126](https://github.com/openmsupply/application-manager-web-app/pull/132) today or tomorrow so we can see all running together. That will be nice indeed.

database/insertData.js
nmadruga: GReat. I'll remove from all.

database/insertData.js
nmadruga: Oh! You solved a mystery to me! 🌚

CarlosNZ (approved):
Sweet. The queries just to check response is not empty are a bit unwieldy. I think going forward we can probably replace with a REGEX comparison, so it doesn't need two sepearate comparison queries (`null` AND `""`), but not essential for now.

PR #128, opened 2020-11-25
119 Better error handling for failed Actions

### LINKED ISSUE #119: Bug fix -- Action throwing error when submitting registration form
To replicate:
- Start new registration
- Leave all responses empty
- Submit

Back-end gives an "UnhandledPromiseRejection" error, and never resets the trigger back to Null, so front-end keeps loading indefinitely next time you try and go to that application.

This does NOT happen when form responses have been entered for the fields required to create user, so must be something to do with having null/undefined values sent to Action.

CarlosNZ:
Fixes #119.

The simplest way to force an error situation is to use the current front-end `master` (Commit: f6bdeb9) and submit the UserRego application without filling in any content. (This won't be possible after validation checking is merged to master)

Previously, after an Action failure, an unhandled error was thrown, meaning the `processTrigger` method was aborted and so the Trigger on the application table was being stuck perpetually in "Processing"

The changes here achieve the following:
- Better error handling in back-end: Error messages are written to the action_queue `error_log` and the `processTrigger` method does not abort
- On failure of a sequential action, any subsequent Actions are cancelled, with a "Cancellation" message written to their record in the action_queue.
- The Trigger field on the triggering table is set to (a new value) "Error". This way the front-end will behave as before (perpetual loading due to "Not NULL" trigger), but with the potential to handle error state differently when we want to implement this.

To test, submit UserRego form with no user details provided. Note the errors printed to console, and check the `error_log`s in the action_queue table. Also, note the Trigger is set to 'Error' state.

### DIFF
--- a/database/buildSchema/16_application.sql
+++ b/database/buildSchema/16_application.sql
@@ -2,7 +2,7 @@
 
 CREATE TYPE public.application_outcome AS ENUM ('Pending', 'Approved', 'Rejected');
 
-CREATE TYPE public.trigger as ENUM ('onApplicationCreate', 'onApplicationSubmit', 'onApplicationSave', 'onApplicationWithdrawn', 'onReviewStart', 'onReviewEditComment', 'onReviewSave', 'onReviewAssign', 'onApprovalSubmit', 'onScheduleTime', 'Processing');
+CREATE TYPE public.trigger as ENUM ('onApplicationCreate', 'onApplicationSubmit', 'onApplicationSave', 'onApplicationWithdrawn', 'onReviewStart', 'onReviewEditComment', 'onReviewSave', 'onReviewAssign', 'onApprovalSubmit', 'onScheduleTime', 'Processing', 'Error');
 
 CREATE TABLE public.application (
     id serial primary key,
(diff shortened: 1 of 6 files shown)


CarlosNZ:
TO-DO: Check behaviour with Async Actions. (Will delete this comment once I've confirmed it's working properly)

andreievg (approved):
Looks good to me, I haven't tested it thought.

Some suggestions, not really required, but nice to have (and just trying to add some sort of feedback):

* Do we log parameters for every action before it's fired ? (most plugins would most likely log it within, but if one doesn't and causes an error, might be easier to replicate), maybe saving parameters_evaluated in DB before running plugin might be enough.
* actionFailed could just be a string of action that was failed (empty for all good), although object is probably better in case more things need to be added

CarlosNZ:
>     * Do we log parameters for every action before it's fired ? (most plugins would most likely log it within, but if one doesn't and causes an error, might be easier to replicate), maybe saving parameters_evaluated in DB before running plugin might be enough.

Yes, although if the failure is due to an error in the evaluator (looking up object properties that don't exist, say), then the evaluated parameters will never get logged. The unevaluated queries will be logged first, though in the `parameter_queries` field.

>     *actionFailed could just be a string of action that was failed (empty for all good), although object is probably better in case more things need to be added

Yeah, true. I'll update it.

CarlosNZ:
> TO-DO: Check behaviour with Async Actions. (Will delete this comment once I've confirmed it's working properly)

One thing to note -- Async Actions will write the error to the action_queue log just the same as sequential ones will, but they are not able to set the Trigger on the application to 'Error', as there's no way to force the Async code to wait until all sequential code is done before setting trigger, or even informing the sequential code. I think this is okay for now, just something to be aware of. 

Upshot is -- only use async actions for non-critical events.

PR #129, opened 2020-11-27
#115F New user registration template

CarlosNZ:
This is conjunction with the login functionality for the front-end. [Issue 115](https://github.com/openmsupply/application-manager-web-app/issues/115) /  [PR](https://github.com/openmsupply/application-manager-web-app/pull/144)

I've created a **new** user registration form that should be come the "proper" user registration form, as opposed to the current `UserRego1` which is for testing more complex functionality. 

The new one is just one page, appropriate validation, and only essential questions. It has code `UserRegistration`, so we can access it at `/applications/new?type=UserRegistration`. This will be the URL that the Login/Register process re-directs to.

We can keep using and amending UserRego1 to test functionality etc.

### DIFF
--- a/database/insertData.js
+++ b/database/insertData.js
@@ -5,13 +5,13 @@ const config = require('../src/config.json')
 const graphQLendpoint = config.graphQLendpoint
 
 const queries = [
-  // Template A -- User Registration
+  // Template A -- Test - General Registration
   `mutation {
     createTemplate(
       input: {
         template: {
-          code: "UserRego1"
-          name: "User Registration"
+          code: "TestRego"
+          name: "Test -- General Registration"
           status: AVAILABLE
           versionTimestamp: "NOW()"
           templateSectionsUsingId: {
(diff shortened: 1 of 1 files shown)


nmadruga (approved):
Cool. That looks good. I was going to comment on the need to use the `maskedInput` for the password in the summary, but that's part of the front-end work.

Another small thing: Should we remove the `old` user registration from here, so we don't keep two very similar applications?

CarlosNZ:
> Cool. That looks good. I was going to comment on the need to use the `maskedInput` for the password in the summary, but that's part of the front-end work.

Yeah, we'll get the password display working properly when we get all the hashing functionality working, etc. It may need to be its own "password" plugin type, but I'm *hoping* we can incorporate all the password functionality into shortText. 🤞
> 
> Another small thing: Should we remove the `old` user registration from here, so we don't keep two very similar applications?

I think it's useful for testing/demoing new functionality and complex evaluator queries, which we'll need to be able to do as we write new plugins. Maybe I'll rename it and give it new code so it's clearly for "testing" purposed.

PR #130, opened 2020-11-30
#130F Use ResponsesFullByCode

CarlosNZ:
Companion to front-end issue 130. (https://github.com/openmsupply/application-manager-web-app/pull/152)

We're only using the one "ResponsesByCode" object in the front-end now, which contains the whole response object (not just "text" field, so some queries needed to be re-written to explicitly refer to `.text` property of questions.

### DIFF
--- a/database/insertData.js
+++ b/database/insertData.js
@@ -57,7 +57,7 @@ const queries = [
                               children: [
                                 {
                                   operator: "objectProperties"
-                                  children: [{ value: { property: "Q1" } }]
+                                  children: [{ value: { property: "Q1.text" } }]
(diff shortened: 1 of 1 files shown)


nmadruga (approved):
LGTM

nmadruga:
Do we need to update the documentation to reflect this change?

CarlosNZ:
> Do we need to update the documentation to reflect this change?

I don't think we've documented this at all currently. Was waiting till we got something a bit more stable. I'll try and do a docs update at the end of the week now that the Application process is settling down.

PR #133, opened 2020-12-01
Renumber database setup files

CarlosNZ:
Just taking the opportunity to do this while there's no back-end issues in progress.

Renumbered database build files to get rid of the `_b` files and increment the numbers correctly from 01 to 34.

(And look -- Github is smart enough to recognise that these files have been "renamed without changes" -- and I didn't even use `git mv`!)


nmadruga (approved):
Looks good.

PR #142, opened 2020-12-07
#135 new review test template & application

CarlosNZ:
Implements #135.

_Requires [front-end PR #195](https://github.com/openmsupply/application-manager-web-app/pull/195) to make timestamp fields match, and new template to appear in sidebar._

Added new "ReviewTest" template, with submitted application form into database. Application appears in Application List -- named "Test Review -- Vitamin C"

Also:
- Changed "time_created" in "application_response" to "timestamp", and set a default value of current time, which should also close [front-end issue #172](https://github.com/openmsupply/application-manager-web-app/issues/172).
- Added additional `touch` command to `initialise_database.sh` to force the server to restart after a database init.

### DIFF
--- a/database/insertData.js
+++ b/database/insertData.js
@@ -1491,7 +1491,7 @@ const queries = [
                       elementTypePluginCode: "shortText"
                       category: QUESTION
                       isRequired: false
-                      parameters: { label: "Please list any side effect" }
+                      parameters: { label: "Please list any side effects" }
                     }
                   ]
                 }
@@ -1606,6 +1606,21 @@ const queries = [
     createUser(
       input: {
         user: { email: "valerio@nra.org", passwordHash: "7110EDA4D09E062AA5E4A390B0A572AC0D2C0220", username: "valerio" }
+      }
+    ) {
+      user {
+        email
+        passwordHash
+        username
+      }
+    }
+  }`,
+  `mutation {
+    createUser(
+      input: {
+        user: { email: "js@nowhere.com", passwordHash: "123456", username: "js",
+        firstName: "John", lastName: "Smith"
+       }
       }
     ) {
       user {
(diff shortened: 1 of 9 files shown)


database/insertData.js
```diff
@@ -1299,6 +1299,267 @@ const queries = [
...
+                        placeholder: "Enter a country"
+                      }
+                    }
+                  ]
+                }
+              }
+              {
+                code: "S2"
+                title: "Product Info"
+                index: 7
```
nmadruga, line 1414:
```suggestion
                index: 1
```

database/insertData.js
```diff
@@ -1299,6 +1299,267 @@ const queries = [
...
+                      title: "Product Intro"
+                      elementTypePluginCode: "textInfo"
+                      category: INFORMATION
+                      parameters: {
+                        text: "In this section, we require your **PRODUCT information**"
+                      }
+                    }
+                    {
+                      code: "Q20"
+                      index: 8
```
nmadruga, line 1429:
```suggestion
                      index: 1
```

database/insertData.js
```diff
@@ -1299,6 +1299,267 @@ const queries = [
...
+                      title: "Product Name"
+                      elementTypePluginCode: "shortText"
+                      category: QUESTION
+                      parameters: { label: "Name of Product" }
+                    }
+                    {
+                      code: "Q21"
+                      index: 9
```
nmadruga, line 1437:
```suggestion
                      index: 2
```

database/insertData.js
```diff
@@ -1299,6 +1299,267 @@ const queries = [
...
+                      category: QUESTION
+                      parameters: {
+                        label: "What type of product are you registering?"
+                        placeholder: "Select"
+                        options: ["Medicine", "Natural Product"]
+                      }
+                    }
+                    {
+                      code: "PB2"
+                      index: 10
```
nmadruga, line 1449:
```suggestion
                      index: 3
```

database/insertData.js
```diff
@@ -1299,6 +1299,267 @@ const queries = [
...
+                      title: "Page Break"
+                      elementTypePluginCode: "pageBreak"
+                      category: INFORMATION
+                    }
+                    {
+                      code: "Q22"
+                      index: 11
```
nmadruga, line 1456:
```suggestion
                      index: 4
```

database/insertData.js
```diff
@@ -1299,6 +1299,267 @@ const queries = [
...
+                      elementTypePluginCode: "shortText"
+                      category: QUESTION
+                      parameters: {
+                        label: "Size of single dose"
+                        placeholder: "Metric units please"
+                      }
+                    }
+                    {
+                      code: "Q23"
+                      index: 12
```
nmadruga, line 1467:
```suggestion
                      index: 5
```

database/insertData.js
```diff
@@ -1299,6 +1299,267 @@ const queries = [
...
+                            }
+                            { value: "^[0-9]+$" }
+                          ]
+                        }
+                        validationMessage: "Must be a number"
+                      }
+                    }
+                    {
+                      code: "Q24"
+                      index: 13
```
nmadruga, line 1489:
```suggestion
                      index: 6
```

database/insertData.js
```diff
@@ -1690,6 +1950,75 @@ const queries = [
...
+                value: { text: "Natural Product", optionIndex: 1 }
+              }
+              { isValid: null, value: null, templateElementId: 49 }
+              { isValid: true, templateElementId: 50, value: { text: "50mg" } }
+              { isValid: true, templateElementId: 51, value: { text: "100" } }
+              {
+                isValid: true
+                templateElementId: 52
+                value: { text: "Turning orange" }
+              }
```
nmadruga, line 2009: I was thinking maybe not using the templateElementId... If that can be easily replaced by the combination of templateId + elementCode it could be clearer. And maybe less dependent on the order of elements, thinking if we add other elements (even in another template) these `Id` would be changed.

database/insertData.js
```diff
@@ -1690,6 +1950,75 @@ const queries = [
...
+            ]
+          }
+        }
+      }
+    ) {
+      query {
+        applicationBySerial(serial: "12345") {
+          id
+        }
```
nmadruga, line 2018: Probably not needed

nmadruga (changes requested):
I have discovered the problem for the 2nd bug found [here](https://github.com/openmsupply/application-manager-web-app/issues/196). I'm pointing out which indexes have been wrongly set. Basically, we use indexes from 0-N for elements within a section and each section should have indexes from 0-N. We probably need a better way to validate if the sections are correctly setup.

CarlosNZ:
> We probably need a better way to validate if the sections are correctly setup.

Ideally it should respect the order and not care about the specific numbers. Like, the Sections could be numbered "1, 10, 999" (and same for questions) and as long as it can be sorted, it works.

database/insertData.js
CarlosNZ: Okay, have done all these in 1 commit. 91008560

database/insertData.js
CarlosNZ: I agree, but I'm not sure how to do that. Even the other applications in this file use `templateElementToTemplateElementId: { connectById: { id: 12 } }`. It's not exactly the same, but it still refers to a specific templateElementID.

Any ideas? It'd be great if we could reference the template elements by code, I'm just sure how to.

nmadruga: I'll have a look now see if I can find anything to replace that.

nmadruga (approved):

database/insertData.js
CarlosNZ: I don't think it's possible without (at the very least) additional plugins. Because you'd need to provide a connection to a unique identifier. Which would need to be the combination of Section_id and code. And postgraphile doesn't provide "create" connections to unique fields anyway, only in queries. There may be a plugin to open up this functionality, but will have to look for that in another job.

Issue #146, opened 2020-12-08
Add action to change status of review 


andreievg:
Similar to application:

GraphQL mutation would set onReviewCreate in the trigger. An action will dispatch to add history, similar to `action_change_status`, default to 'Draft'

Considering to either add another parameter to existing `action_change_status` (application or review), or just making another action. I think the former would be easier, for now. 

Looks like action will be trigger and processed without any changes, need to double check though and make required changes (part of this issue, if needed)

PR #147, opened 2020-12-08
#146 add action to change review status

andreievg:
Fixes: #146 

Needs to be tested, waiting for insertData with assignments.

**Changes**

getCurrentStatusHistory -> getApplicationCurrentStatusHistory
addNewStatusHistory -> addNewApplicationStatusHistory

getReviewCurrentStatusHistory (new method)
addNewReviewStatusHistory (new method)

Updated changeStatus.ts, extracted application status change logic to `changeApplicationStatus` and created `changeReviewStatus`. Small check in module method to determine which one to run


CarlosNZ:
> Needs to be tested, waiting for insertData with assignments.

Done [here](https://github.com/openmsupply/application-manager-server/pull/148) if you want to merge it into this branch?

src/components/postgresConnect.ts
```diff
@@ -476,6 +476,32 @@ class PostgresDB {
     }
   }
 
+  public getReviewCurrentStatusHistory = async (reviewId: number) => {
+    const text = `SELECT id, status, id FROM
```
nmadruga, line 480:
```suggestion
    const text = `SELECT id, status FROM
```

src/components/postgresConnect.ts
```diff
@@ -476,6 +476,32 @@ class PostgresDB {
...
+    try {
+      const result = await this.query({ text, values: [reviewId] })
+      return result.rows[0]
+    } catch (err) {
+      console.log(err.message)
+      throw err
+    }
+  }
+
+  public addNewReviewStatusHistory = async (reviewId: number, status = 'Draft') => {
```
nmadruga, line 492: It would be better to use the graphql type here instead. so we keep it up-to-date with the one in the database - in case we change it later.

src/plugins/action_change_status/src/changeStatus.ts
```diff
@@ -2,19 +2,38 @@ import { ActionPluginOutput, Status } from '../../types'
...
 module.exports['changeStatus'] = async function (
   parameters: IParameters,
   DBConnect: any
 ): Promise<ActionPluginOutput> {
-  const { applicationId, newStatus } = parameters
+  const { applicationId, reviewId, newStatus } = parameters
+
+  if (applicationId) {
+    return await changeApplicationStatus(applicationId, newStatus, DBConnect)
+  } else if (reviewId) {
```
CarlosNZ, line 17: Haven't tested yet either, just making a couple of quick comments:
`if (applicationId)` -- if you're wanting to change the review status, you'll still get an applicationId passed in (I think), so not sure this is a good way to determine whether we're updating review status or application status.

2 other ways I can think of:
- check the Trigger
- add a new parameter and explicitly declare it in the Action definition of the template.

src/components/postgresConnect.ts
```diff
@@ -450,7 +450,7 @@ class PostgresDB {
     }
   }
 
-  public getCurrentStatusHistory = async (applicationId: number) => {
+  public getApplicationCurrentStatusHistory = async (applicationId: number) => {
```
CarlosNZ, line 453: Assume you did a global search for anywhere else this might be called?

nmadruga (changes requested):
Nice! The reuse of the existing action was a clever solution.

Just small changes here:

database/insertData.js
```diff
@@ -1542,6 +1542,16 @@ const queries = [
...
+              {
+                actionCode: "changeStatus"
+                trigger: ON_REVIEW_CREATE
+                parameterQueries: {
+                  applicationId: {
+                    operator: "objectProperties"
+                    children: [{ value: { property: "reviewId" } }]
+                  }
+                  newStatus: { value: "Draft" }
+              }
```
CarlosNZ, line 1554:
```suggestion
            {
              actionCode: "changeStatus"
              trigger: ON_REVIEW_CREATE
              parameterQueries: {
                applicationId: {
                  operator: "objectProperties"
                  children: [{ value: { property: "reviewId" } }]
                }
                newStatus: { value: "Draft" }
              }
            }
```
Missing one closing `}`

CarlosNZ (changes requested):
Error in insertData. Suggested fix.

database/insertData.js
CarlosNZ: This is the object collected by the Trigger processor that is available to evaluate parameters from:
```
ApplicationData {
   trigger_id: 1,
   trigger: 'onReviewCreate',
   table: 'review',
   record_id: 1,
   applicationId: 4,
   templateId: 4,
   stageId: 5,
   stageNumber: 1,
   stage: 'Screening',
   stageHistoryId: 3,
   stageHistoryTimeCreated: null,
   statusHistoryId: 3,
   status: 'Submitted',
   statusHistoryTimeCreated: 2020-12-08T23:42:34.307Z,
   userId: 5,
   firstName: 'John',
   lastName: 'Smith',
   username: 'js',
   dateOfBirth: null,
   email: 'js@nowhere.com',
   responses: {
     PB1: null,
     PB2: { text: 'Turning orange' },
     Q1: null,
     Q2: { text: '39' },
     Q20: { text: '50mg' },
     Q21: { text: '100' },
     Q3: { text: 'New Zealand' },
     Q4: { text: 'Vitamin C' },
     Q5: { text: 'Natural Product', optionIndex: 1 },
     Text1: { text: 'js@nowhere.com' },
     Text2: null
   }
 }
```

src/plugins/action_change_status/src/changeStatus.ts
CarlosNZ: Okay, have tested now. I see how it works, but the Action definition in the template query had some errors, have made suggestions below, and it's working now.

src/components/postgresConnect.ts
CarlosNZ: You mean the `status` type? There's a problem with that, which I also faced when you suggested the same thing on the Application Status update: GraphQL converts enums to ALLCAPS, so 'Draft' is not valid, it's expecting 'DRAFT'. But postGres expects 'Draft'. Not sure there's any way round this, unless we can somehow export Postgres types directly?

CarlosNZ: We'd need something like this, I think: https://github.com/adelsz/pgtyped

Might be worth the effort?

src/plugins/action_change_status/src/changeStatus.ts
andreievg: Thanks for fixing

src/components/postgresConnect.ts
andreievg: Yeah i did a global replace

src/components/postgresConnect.ts
andreievg: Hmm, a bit skeptical atm, of how much we want to typescript. I'll try to check out that package as suggested (if time allows atm).

database/insertData.js
andreievg: I complete stuffed that up, thanks for the suggested fix

src/components/postgresConnect.ts
CarlosNZ: Yeah, definitely not a priority. Could be useful when we change some enums, then Typescript would quickly point out inconsistencies.

CarlosNZ (approved):

src/components/postgresConnect.ts
nmadruga: Oh, that's not nice. I thought it would work the same way as we have on the front-end code.
Well, let's leave it for now then...

nmadruga (approved):

PR #148, opened 2020-12-08
#136 Assign application to reviewer

CarlosNZ:
Addresses #136. Updated `insertData.js` to create a review assignment of the new "Vitamin C" test review application.

Created two new users as well ("Reviewer" and "Assigner") -- unfortunately I accidentally pushed them to master before I realised what branch I was on (Ooops 😖), so you won't see them in this diff, but its line 1641–1668.

Once you've run the database setup script, you can see the details of the review assignment and the questions assigned (all of them) using the following GraphQL query:
```
query ShowReviewAssignment {
  reviewAssignment(id: 1) {
    application {
      name
      user {
        firstName
        lastName
      }
    }
    assigner {
      firstName
      lastName
    }
    reviewer {
      firstName
      lastName
    }
    reviewQuestionAssignments {
      nodes {
        templateElement {
          title
        }
      }
      totalCount
    }
  }
}
```

Question: I've currently only assigned questions that are actual "questions" rather than textInfo or pageBreak. Do you think this is the way to go, or will assigning a Section just assign everything regardless of its type and we filter in the front-end?

### DIFF
--- a/database/insertData.js
+++ b/database/insertData.js
@@ -2054,6 +2054,40 @@ const queries = [
       }
     }
   }`,
+  // Assign above application to reviewer (all questions)
+  `mutation {
+    createReviewAssignment(
+      input: {
+        reviewAssignment: {
+          applicationId: 4
+          assignerId: 7
+          reviewerId: 6
+          stageId: 5
+          reviewQuestionAssignmentsUsingId: {
+            create: [
+              { templateElementId: 43 }
+              { templateElementId: 44 }
+              { templateElementId: 45 }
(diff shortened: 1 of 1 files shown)


andreievg (approved):
Looks good

nmadruga:
> Question: I've currently only assigned questions that are actual "questions" rather than textInfo or pageBreak. Do you think this is the way to go, or will assigning a Section just assign everything regardless of its type and we filter in the front-end?

My think is to ONLY assign questions, since those are going to have responses after all.

PR #149, opened 2020-12-09
bug fix review to review_assignment reference

andreievg:
note review appears in review_assignment
Screen Shot 2020-12-09 at 5 22 01 PM
Screen Shot 2020-12-09 at 5 21 19 PM


CarlosNZ (approved):
👍

PR #153, opened 2020-12-14
#151 Template fixes

### LINKED ISSUE #151: Bug: Mock "review" template -- mapping between responses and template IDs is messed up.
When you view the Summary page for the Review Testing "Vitamin C" application, the responses don't match up to the questions. Must the template_ids mapped incorrectly in template definition.

CarlosNZ:
Fixes #151. 

Template element ids were incorrect. Must have occurred because of other changes. That's the trouble with hard-coding these links, but we don't really have a better way currently. 

Also:
- added `isActive: true` to this template.
- updated format and default time to application_stage_history `time_created` field (Missed last week).
- Rebuilt plugins (.js) since we noticed that (at least) one of them didn't match the lastest `.ts` source.

### DIFF
--- a/database/buildSchema/18_application_stage_history.sql
+++ b/database/buildSchema/18_application_stage_history.sql
@@ -4,7 +4,7 @@ CREATE TABLE public.application_stage_history (
     id serial primary key,
     application_id integer references public.application(id),
     stage_id integer references public.template_stage(id),
-    time_created timestamp,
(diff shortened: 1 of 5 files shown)


nmadruga (approved):
Thanks! That was annoying me so much.

CarlosNZ:
> Thanks! That was annoying me so much.

Ha ha me too!

PR #156, opened 2020-12-16
#177F Adding 2nd reviewer and update password to be the same one

nmadruga:
Based on PR #152, please merge that one before this one.


CarlosNZ (approved):
All good, but would you mind doing one thing before merging?

I've just updated the jest tests to account for changes in the database (#157) -- would you mind merging that branch into here, then updating the tests to work with your changes? I think you'll only need to increment the id for the `createUser` test. Let me know if it's any more complicated than that.

PR #160, opened 2020-12-17
permission docs

andreievg:
PR is mainly diagrams, as per below.

Also slight variation in JWT content as per second diagram

https://github.com/openmsupply/application-manager-server/blob/permission-docs/documentation/images/jwt-flow.png

https://github.com/openmsupply/application-manager-server/blob/permission-docs/documentation/images/jwt-and-permissions-detailed.png


andreievg:
I've just made a slight change to diagram and fixed the doc.

The change in diagram was, from: `p_${permission_policy_id}_${permision_name_id}` to `p_${permission_policy_id}_${template_permission_id}`

documentation/Database-Schema-Permission.md
```diff
@@ -28,7 +28,10 @@ The `name` and `description` of the policy helps the Admin users identify the po
 
 The `type` is the action type, that should be one of the options: `'Apply'`, `'Review'`, `'Assign'`, ...(few more to be added).
 
-The `rules` will define the list of when the user is allowed to be acting on the linked application. For example, the policy to Apply for an application is that the associated User ID is the same as your and the current status set as Draft or Changes required.
+The `rules` will define the list of when the user is allowed to be acting on the linked application. For example, the policy to Apply for an application is that the associated User ID is the same as your and the current status set as Draft or Changes required. Rules will link directly to row level permissions in postgres, to make admin task easier, we are using JSON SQL format for this field (`json-sql-builder2`). When permission policies are edited, the JSON SQL will be converted to SQL and row permissions will be edited and/or added to respective tables. We use mechanism provided by ~[Graphile](https://graphile.org/postgraphile/security/) to expose JWT parameters to be used in SQL queries inside row level permissions. As per below diagrams:
```
CarlosNZ, line 31:
```suggestion
The `rules` will define the list of when the user is allowed to be acting on the linked application. For example, the policy to Apply for an application is that the associated User ID is the same as the logged-in user and the current status set as "Draft" or "Changes required". Rules will link directly to row level permissions in postgres; to make admin task easier, we are using JSON SQL format for this field (`json-sql-builder2`). When permission policies are edited, the JSON SQL will be converted to SQL and row permissions will be edited and/or added to respective tables. We use mechanism provided by **[Graphile](https://graphile.org/postgraphile/security/)** to expose JWT parameters to be used in SQL queries inside row level permissions. As per below diagrams:
```

CarlosNZ: > current status set as Draft or Changes required

Not sure that I understand "Apply" needs to be these statuses. Surely when you apply, there's no status at all, since there's no application yet?

CarlosNZ (commented):
Perhaps you could clarify a couple of things at our next meeting?

documentation/Database-Schema-Permission.md
andreievg: I think this was meant to be for 'UPDATE' rule, I'll amend this doc further

Issue #159, opened 2020-12-17
Evaluator: Refactor existing templates and front-end code to comply with named objects


CarlosNZ:
To be done after the evaluator changes in #158 have been merged and published to evaluator package.

Needs front and back-end refactoring.

CarlosNZ:
Note: basically done, just need to double check everything is working correctly. Will push PR shortly.

Issue #158, opened 2020-12-17
Evaluator: new operator -- stringSubstitution


CarlosNZ:
Child nodes:
0: String with %1, %2 parameters embedded
1-n: values to be substituted into %1, %2

Will be useful for creating more complex message strings for start and end of template, etc. Could be done with "CONCAT", but the queries would get ridiculous.

--- 
Just putting this here now so I have it for later:

Regex for matching parameters: `(?<![\\])%([\d]+)`
- ~~Allows `$` or `%` prefix~~
- ~~Ignores literal % (if you prefix them with `\` in string)~~
- Captures the number part in Group for further processing

andreievg:
In the docs from this pr: https://github.com/openmsupply/application-manager-server/pull/161, there is suggestion of dollar sign, maybe just percentage sign ?

CarlosNZ:
> In the docs from this pr: #161, there is suggestion of dollar sign, maybe just percentage sign ?

Sure, is that more "standard"? Just noticed that SQL uses $1. Was just thinking that because we'll probably write a regex to find the parameters, it would be pretty simple to allow both options.

Question: do you think ordering should follow strict numerical order, or just the order in the string? So if the string has `"First name: %2, Last name: %1"`, do we replace them in inverse order that the values are supplied, or just in the order it finds the parameters and ignore the actual numbers?

I only ask because matching the numerical order seems more "correct", but quite a bit more complex code-wise, so just wondering if you think it's worth it?

andreievg:
I think numeric value after the '%' should match child index. Btw in implementation I would just go through each child and do global replace of %{childnumber} (regex should make sure no other digits after %{childnumber})

The only reason for not wanting $ is cause $1 might be a price whereas percentage always follows a number

Issue #154, opened 2020-12-16
Evaluator: update objectProperties operator to handle named objects rather than indexed


CarlosNZ:
Low priority, but Andrei recommended this would be better if possible. Will take a look at it at some point.

Issue #145, opened 2020-12-08
Bug Fix: evaluator module -- when type is set to "array" CONCAT operator doesn't concatenate literal strings into an array.


CarlosNZ:
Discovered that the CONCAT operator is not returning an Array, even when explicitly declared in `type`, if the "children" were all strings  (it'll correctly concatenate multiple arrays).

Hacky workaround is to supply an empty array as the first child, but should fix this properly.

e.g.
```
{
  operator: "CONCAT",
  type: "array",
  children: [
    [],
    {
      operator: "objectProperties",
      children: [
        {
          value: {
            property: "Q1.text"
          }
        }
      ]
    },
    {
      operator: "objectProperties",
      children: [
        {
          value: {
            property: "Q2.text"
          }
        }
      ]
    }
  ]
}
```

PR #161, opened 2020-12-17
#154 Evaluator named objects and CONCAT bug fix

CarlosNZ:
Addresses #154 and #145. Was having a quick look last night and both turned out to be super-simple tweaks. Re-factoring the tests, docs and GUI app took 90% of the work!

Once this has been merged I'll re-publish the a new version of the evaluator package and go through and update all the templates and code with the new syntax (#159) -- it's only a very small change, but it's much better. Thanks for the suggestion @andreievg .

So instead of passing in an array of object which need to be referred to by index, you just pass them nested in a single object (e.g. `objects: { user, responses: allResponses } `, and then prefix the property you're fetching with the name of the object (e.g. `responses.Q1.text`). (See docs and tests for more clarity)

- The CONCAT bug (#145) was resolved by just specifying an empty array as the initial value for the accumulator in the `.reduce` method.

- I also wrote documentation for a suggested new "stringSubstitution" operator (#158) so we can assess it. Think it could be useful for dynamic titles, messages, etc. (easier than complex CONCATs)

### DIFF
--- a/documentation/Query-Syntax.md
+++ b/documentation/Query-Syntax.md
@@ -90,7 +90,7 @@ Implements logical OR on the returned values of its child nodes.
 
 Concatenates the values of the child nodes, which can be either strings or arrays.
 
-- Input: 1 or more **string** or **array** nodes (must all be the same)
+- Input: 1 or more **string** or **array** nodes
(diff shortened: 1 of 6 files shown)


andreievg (approved):
looks good thanks, and the stringSubstitution seems like a good operator, I'll comment in related issue

PR #164, opened 2021-01-04
#158 Evaluator -- string substitution operator

CarlosNZ:
Implements #158.

Did this at the start of the break. Now that @nmadruga has made progress on the start/submission pages, this will probably be useful sooner rather than later (as it'll be a lot easier than multiple CONCATs).

Can see implementation by reading and running test suite.

One thing I didn't (yet) implement is escaping when literal `(%1)` values are required in a string. Probably not a priority as it's very unusual to have a percent character before a number (unless writing documentation for a modulus operator, perhaps), so can do it later if required.

Once approved, will update package and make new PRs to upgrade the module in front and back-ends.

### DIFF
--- a/src/modules/expression-evaluator/src/evaluateExpressionTestData.ts
+++ b/src/modules/expression-evaluator/src/evaluateExpressionTestData.ts
@@ -569,13 +569,6 @@ testData.stringSubstitutionParametersNonOrderedAndTooFew = {
   children: ['%2 out of every %3 people are %1', 'stupid', 'Two'],
 }
 
-testData.stringSubstitutionIgnoreLiteral = {
-  operator: 'stringSubstitution',
-  children: [
-    `We don't want \%2 to be replaced but we do want %1 to be replaced, and we want there to be a single \\ here.`,
-    'this',
-  ],
-}
 testData.stringSubstitutionParametersNotSequential = {
   operator: 'stringSubstitution',
   children: [
@@ -585,6 +578,16 @@ testData.stringSubstitutionParametersNotSequential = {
     'parameter',
     'numbers',
   ],
+}
+
+testData.stringSubstitutionNoParameters = {
+  operator: 'stringSubstitution',
+  children: ['This sentence has no replacements.', 'nothing', 'will', 'happen'],
+}
+
+testData.stringSubstitutionNoReplacements = {
+  operator: 'stringSubstitution',
+  children: ['Your name is %2 %1 but we have nothing to replace them with'],
 }
 
 // API operator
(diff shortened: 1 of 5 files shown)


src/modules/expression-evaluator/src/evaluateExpressionTestData.ts
```diff
@@ -527,6 +527,69 @@ testData.singleApplicationProperty_depth2 = {
...
+}
+
+testData.stringSubstitutionNonStringReplacements = {
+  operator: 'stringSubstitution',
+  children: ['We have %1 people listed: %2', 2, { value: ['Boba', 'Mando'] }],
+}
+
+testData.stringSubstitutionTooManyReplacements = {
+  operator: 'stringSubstitution',
+  children: ['The price of milk is %1 per %2', '$2.30', 'liter', 'gallon', '$5.00'],
```
nmadruga, line 554: Should this be using `number` instead of string?

src/modules/expression-evaluator/src/evaluateExpressionTestData.ts
```diff
@@ -527,6 +527,69 @@ testData.singleApplicationProperty_depth2 = {
...
+
+testData.stringSubstitutionParametersNonOrderedAndTooFew = {
+  operator: 'stringSubstitution',
+  children: ['%2 out of every %3 people are %1', 'stupid', 'Two'],
+}
+
+testData.stringSubstitutionParametersNotSequential = {
+  operator: 'stringSubstitution',
+  children: [
+    `It shouldn't matter if %10 are big %100 between %101 %200`,
```
nmadruga, line 575:
```suggestion
    `It shouldn't matter if %10 are big %100 between %101 of %200`,
```

?

nmadruga: Remember to update the test too

nmadruga (commented):
Looks good!

Just 2 comments to have a look : )

src/modules/expression-evaluator/src/evaluateExpressionTestData.ts
CarlosNZ: I don't really understand your suggestion. The expected output is:
`It shouldn't matter if there are big gaps between parameter numbers`, which is what this results in.

src/modules/expression-evaluator/src/evaluateExpressionTestData.ts
CarlosNZ: Use `number` where? 😕

CarlosNZ: Okay, I've updated the earlier "non-string replacements" test to include a float number as well as an integer:
```
{
  operator: 'stringSubstitution',
  children: [
    'We have %1 %2 listed with an average value of %3: %4',
    2,
    'people',
    4.53,
    { value: ['Boba', 'Mando'] },
  ],
}
```

nmadruga (approved):
Thanks!

PR #165, opened 2021-01-04
#216F Add user id to user fetch

nmadruga:
Branch to be used in pair with Front PR of [issue#216](https://github.com/openmsupply/application-manager-web-app/issues/216)

### Changes
- Update dummy users password to be the same
- Include user `id` in the **getUserInfo** response


CarlosNZ (approved):
LGTM!

PR #168, opened 2021-01-06
#232F Start message expression

CarlosNZ:
Update to mock data so changes in [front-end #242](https://github.com/openmsupply/application-manager-web-app/pull/242) can be demonstrated.

~~**NOTE**: This still uses the old array format for the evaluator objects. It will need to be updated after #159 is merged (and again after #158), so hopefully they can be approved before this one and I'll merge them back into here and do those changes before this is merged.~~
(Done)

### DIFF
--- a/database/insertData.js
+++ b/database/insertData.js
@@ -16,7 +16,7 @@ const queries = [
           startMessage: {
             operator: "CONCAT"
             children: [
-              "# Hi User"
+              "# Hi User\\n"
               "## This is the general registration for feature showcase\\n- Proof of identity (Passport, Drivers license)\\n- Proof of your medical certification\\n- Drug ingredient list\\n- Product images\\n- Packging images"
             ]
           }
--- a/database/insertData.js
+++ b/database/insertData.js
@@ -16,8 +16,12 @@ const queries = [
           startMessage: {
             operator: "CONCAT"
             children: [
-              "# Hi User\\n"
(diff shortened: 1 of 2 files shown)


database/insertData.js
```diff
@@ -13,7 +13,28 @@ const queries = [
           code: "TestRego"
           name: "Test -- General Registration"
           isLinear: false
-          startMessage: "## This is the general registration for feature showcase\\n- Proof of identity (Passport, Drivers license)\\n- Proof of your medical certification\\n- Drug ingredient list\\n- Product images\\n- Packging images"
+          startMessage: {
```
nmadruga, line 16: Would this be a good place to use the `substring` evaluator from PR [#158](https://github.com/openmsupply/application-manager-server/pull/164)?

nmadruga (changes requested):
Looks good and it's displaying in the Start page correctly.

Just 1 comment to address please.

database/insertData.js
CarlosNZ: Yes, I was planning to, but can't yet as that update has not been published to package yet. 

I'll need to do a new PR to update the version in package.JSON once the evaluator has been updated, and was planning to change these strings in that PR.

nmadruga: Ok. No problem. I was just pointing out because it seems more redundant using the `substring` type

nmadruga (approved):

PR #170, opened 2021-01-07
#169 Unique username in Database

### LINKED ISSUE #169: Enforce username uniqueness on user table at the Database level
Should use Postgres UNIQUE constraint on that field.

CarlosNZ:
Fixes #169. Just adds UNIQUE constraint to username on user table.


nmadruga (approved):
LGMT

PR #173, opened 2021-01-10
#149F Password hashing

CarlosNZ:
Backend changes for [front end PR #256](https://github.com/openmsupply/application-manager-web-app/pull/256).

Need to run `yarn install` to add `bcrypt` package.

- update Users in insertData.js with hashes instead of password. (Passwords are still all '123456' except "nonRegistered" which remains blank (empty string).)
- implement password hash comparison (using bcrypt) in `/login` endpoint
- add new `/create-hash` endpoint -- not currently used but will be when [password plugin is implemented](https://github.com/openmsupply/application-manager-web-app/issues/253)
- removed the "verifyUser" method from postgresConnect, since hash comparison needs to be done with `bcrypt`. Login method now just fetches the hash and comparison is done in login method.

Note: the Salt is stored as part of the Hash. See [here](https://github.com/kelektiv/node.bcrypt.js#hash-info) for specifics of the hash string format.


database/insertData.js
```diff
@@ -2100,14 +2130,14 @@ const queries = [
...
-`, // Extra user with multiple permissions (apply company rego, review company rego and apply user rego)
+`, // Extra user with multiple permissions (apply company rego, review company rego and apply user rego) -- password is "123456"
   `
 mutation MyMutation {
   createUser(
     input: {
       user: {
         username: "userWithMultiplePermissions"
-        passwordHash: "somehashofpassword"
+        passwordHash: "$2a$10$5R5ruFOLgrjOox5oH0I67.Rez7qGCEwf2a60Pe2TpfmIN99Dr0uW."
```
nmadruga, line 2140: Nice!

src/plugins/action_create_user/src/createUser.js
```diff
@@ -1,78 +0,0 @@
-"use strict";
```
nmadruga, line 1: I didn't understood why this file was removed

nmadruga (changes requested):
Just a minor change.
It's looking great to me, I see that now the query for the user will get the passwordHash, but it won't return in the `userInfo` query, but just as a consideration - for future implementation. Should we note in the documentation about this change?

database/insertData.js
```diff
@@ -2002,12 +2031,13 @@ const queries = [
...
   }`,
   // Non Registered User Permissions
+  // Password is blank
   `mutation {
     createUser(
       input: {
         user: {
           email: ""
-          passwordHash: ""
+          passwordHash: "$2a$10$UIfa3GTUbOS92Ygy/UpqheTngGo3O54Q5UOnJ5CBlra9LYCcr4IGq"
```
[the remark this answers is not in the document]
CarlosNZ: Ha ha, no actually! An empty string still has a hash ;)

> It's looking great to me, I see that now the query for the user will get the passwordHash, but it won't return in the `userInfo` query, but just as a consideration - for future implementation. Should we note in the documentation about this change?

The front-end has no need for the hash value. Can add it later if needed, but I can't see why we would.

src/plugins/action_create_user/src/createUser.js
CarlosNZ: Hmm, probably because I thought I'd have to make changes to the `createUser` action, and so need to delete the .js file first so it doesn't take priority over the .ts file. Guess I forgot to rebuild. 

Re-added it now. Ignore any changes in it if there are any.

CarlosNZ:
> It's looking great to me, I see that now the query for the user will get the passwordHash, but it won't return in the `userInfo` query, but just as a consideration - for future implementation. Should we note in the documentation about this change?

The front-end has no need for the hash value. Can add it later if needed, but I can't see why we would.

database/insertData.js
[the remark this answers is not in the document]
nmadruga: Ok. Cool : ) Sorry about the delay to come back here.

nmadruga (approved):

database/insertData.js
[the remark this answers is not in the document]
CarlosNZ: > Sorry about the delay to come back here.

No probs, I can't really merge this until #174 is merged anyway.

nmadruga:
Just one more comment: We thought about removing the amount of returned property for all mutations. I can see that you made that for the new ones. Can you also remove the returned properties of other mutations as well?

CarlosNZ:
> We thought about removing the amount of returned property for all mutations. I can see that you made that for the new ones. Can you also remove the returned properties of other mutations as well?

I will do when I break up the `insertData` file. However, this won't actually help as much as I thought, because most of the text the scrolls past in the console is not actually the return data, but the mutation info itself. Not sure how to make that less verbose.

src/components/permissions/routes.ts
```diff
@@ -9,17 +12,28 @@ const routeUserPermissions = async (request: any, reply: any) => {
...
+  // Login successful
+  reply.send({
     success: true,
     ...(await getUserInfo(username)),
   })
 }
 
+const routeCreateHash = async (request: any, reply: any) => {
+  const { password } = request.body
+  const hash = await bcrypt.hash(password, saltRounds)
```
andreievg, line 31: I thought that saltRounds should be used in conjunction with `genSalt`to generate salt ? And we would store the salt in db ?

I did a post to `http://localhost:8080/create-hash` with "123456" and got "$2b$10$f8sos4kSywTK8HWFvghacO7BLnAg/464EZvoxlZCy9dziQrfzY4xe" which is different to "passwordHash" in insertData

src/components/permissions/routes.ts
```diff
@@ -9,17 +12,28 @@ const routeUserPermissions = async (request: any, reply: any) => {
...
-  const { username, passwordHash } = request.body || { username: '', passwordHash: '' }
+  const { username, password } = request.body || { username: '', password: '' }
+  const { passwordHash } = (await databaseConnect.getUserDataByUsername(username)) || {}
 
-  if (!(await databaseConnect.verifyUser(username, passwordHash)))
-    return reply.send({ success: false })
+  if (!passwordHash) return reply.send({ success: false })
 
-  return reply.send({
+  if (!(await bcrypt.compare(password, passwordHash))) return reply.send({ success: false })
```
andreievg, line 20: I am a little bit confused by this. Pardon my ignorance/misunderstanding, but I thought:

Salt is prefixed/postfixed to password before hashing for extra protection. So how could bcrypt.compare validate raw password vs it's salted hash, without knowing the hash ?

andreievg (changes requested):
looks pretty good, one change request and one comment (all salt related)

src/components/permissions/routes.ts
andreievg: Ok never mind, it looks like bcrypt output incorporates both salt and hash

src/components/permissions/routes.ts
andreievg: Ok never mind, it looks like bcrypt output incorporates both salt and hash

andreievg (approved):

src/components/permissions/routes.ts
CarlosNZ: Yes, I was confused about that initially too. But yeah, the hash string is actually a concatenation of the random salt and the computed hash, as explained [in the docs](https://github.com/kelektiv/node.bcrypt.js#hash-info).

```
$2b$10$nOUIs5kJ7naTuTFkBy1veuK0kSxUFXfuaOKdOKf9xYT0KKIGSJwFa
 |  |  |                     |
 |  |  |                     hash-value = K0kSxUFXfuaOKdOKf9xYT0KKIGSJwFa
 |  |  |
 |  |  salt = nOUIs5kJ7naTuTFkBy1veu
 |  |
 |  cost-factor => 10 = 2^10 rounds
 |
 hash-algorithm identifier => 2b = BCrypt
```

Pretty nifty. :)

src/components/permissions/routes.ts
andreievg: Not a required change, but maybe we can add a little comment saying bcrypt output is not just hash, but has metadata like salt

CarlosNZ: Sure. Done 33455925

PR #175, opened 2021-01-11
#253F Password entry plugin

CarlosNZ:
Back-end changes (templates only) for Password entry plugin [front-end issue/PR #261](https://github.com/openmsupply/application-manager-web-app/pull/261)

_Only review once #173 is approved._


nmadruga (approved):
LGTM

PR #174, opened 2021-01-10
Permissions first implementation

andreievg:
Fixes: #108, #109

### Brief PR description

This is the initial permissions 'skeleton' PR. Added schema changes, JWT token generation, row level policy generation and added further docs.

#### 02_jwtConfig.sql

Refined to have only the following methods:

* `jwt_get_text`
* `jwt_get_boolean`
* `jwt_get_bigint`

All of the methods check 'session' variable that graphile sets from JWT

#### 13_template_permission.sql

* Added `all_permissions` view to amalgamate all permisison tables

#### 35_row_level_policy_related.sql

* Added a role with login right 'graphile_user' with privileges on all tables
* Enabled row level security for application table

graphile_user can be used in the command line that launches graphile

application table will be restricted by row level policies unlesss user accessing db is postgres

#### insertData.js

* Added rules to permission policies 
   * `oneTimeApply`
   * `basicApply`
   * `basicReview`
* Added a call to 'updateRowPolicies' end point so that row policies will be generated an applied

### API.md

* Added login and userInfo end point description
* Added updateRowPolicies end point description

### package.json

* Added `pg_permissions` script, to launch graphile as `graphile_user`
* Added `json-sql-builder2` dependency

### database helpers

* Changed `getUserTemplatePermissions` to use all_permissions view
* Added `getAllPermission` just calls all_permissions
* Added `getAllGeneratedRowPolicies` which shows all generated policies (from postgres schema table `pg_policies`)

### logic

Reshuffled permission.ts, made a folder with following structure

Screen Shot 2021-01-11 at 12 00 23 PM

#### routes.ts

Just has the three routes, and imports from `loginHelpers` and `rowLevelPolicyHelpers`

#### loginHelpers.ts

Should be unchanged, code extracted from previous permission.ts

#### rowLevelPolicyHelpers

Logic to generate JWT `compileJWT` and to generate and apply row level policies `updateRowPolicies`

#### tests.ts

I tried to extract core logic for both generating JWT and updating row level policies into 'functional' methods that don't interact with DB. These methods can then be easily tested.

They are `compileJWT` and `generateRowLevelPolicies`

#### testdata.ts

Both data sets, test 1 = data from method header comments, test 2 = data from DB

## How to see it run/work ?

If you `database_init` with both graphile (as postgres user) and server running, row level policies should be enabled and installed for application table

You would at this stage need to restart graphile with different user `pg_permissions` script.

### Check with graphile UI

* Try to query for applications, you should see zero records
[...]

Issue #107, opened 2020-11-13
Extend login endpoint to list the companies and login with company


nmadruga:
Extend endpoint created on #84.

PR #180, opened 2021-01-13
#107 #116F Organisation login - back-end changes

CarlosNZ:
Implements #107 and Changes to back-end to complement [front-end issue 116](https://github.com/openmsupply/application-manager-web-app/issues/116).

Requires front-end branch from [PR #266](https://github.com/openmsupply/application-manager-web-app/pull/266)

- Review after #174 and #173

### Summary of changes

- Update `insertData.js` to link some users with orgs
- `/login` endpoint returns a list of user's organisations in payload
- new endpoint `/login-org` which checks that an orgId is linked to a userId. Returns payload and JWT similar to `/login` but with additional selected org info.
- new postGres methods to facilitate the above
- rename `routeUserPermissions` to `routeUserInfo` -- seems more appropriate
- changed `getUsername` method to `getTokenData` and made it a more generic method for extracting data from JWT
- fixed typo in spelling of `organisation_id` in user_organisation table

Please test with Postman and with front-end.

**NOTE**: PR currently shows unrelated changes -- this is because I've merged recent `master` into here, but @andreievg's #174 (which this is branched off via #173) is behind master. Please ignore anything that doesn't seem like it belongs -- these diffs should go once #174 is updated.


src/components/permissions/loginHelpers.ts
```diff
@@ -4,38 +4,37 @@ import { verify, sign } from 'jsonwebtoken'
...
-  let username = 'nonRegistered'
-  if (jwtToken) {
-    try {
-      username = (await verifyPromise(jwtToken, config.jwtSecret)).username
-    } catch (e) {
-      console.log('cannot verify JWT in authorisation header')
-    }
+const getTokenData = async (jwtToken: string) => {
+  try {
+    const data = await verifyPromise(jwtToken, config.jwtSecret)
```
andreievg, line 14: previous implementation didn't actual throw an error, but considered token as invalid (and returned a token for `nonRegistered` user. In this case a 500 will be returned (with no error message). Was that intended ?

src/components/permissions/loginHelpers.ts
```diff
@@ -4,38 +4,37 @@ import { verify, sign } from 'jsonwebtoken'
...
     email,
   } = await databaseConnect.getUserDataByUsername(username)
+  const organisation = orgId ? await databaseConnect.getOrgById(orgId) : undefined
 
   return {
     username,
     templatePermissions: buildTemplatePermissions(templatePermissionRows),
-    JWT: await getSignedJWT(userInfo, templatePermissionRows),
-    user: { id, firstName, lastName, username, dateOfBirth, email },
+    JWT: await getSignedJWT(username, userId, templatePermissionRows, orgId),
```
andreievg, line 36: I think we were planning to not have username in the JWT just the id ?

src/components/permissions/loginHelpers.ts
```diff
@@ -51,8 +50,16 @@ const buildTemplatePermissions = (templatePermissionRows: Array<PermissionRow>)
   return templatePermissions
 }
 
-const getSignedJWT = async (userInfo: UserInfo, templatePermissionRows: Array<PermissionRow>) => {
-  return await signPromise(compileJWT(userInfo, templatePermissionRows), config.jwtSecret)
+const getSignedJWT = async (
+  username: string,
+  userId: number,
```
andreievg, line 55: i kind of like the idea of userId and orgId being in an object, which would remove a few lines of code, and might make it a bit clearer. What do you think ?

src/components/permissions/routes.ts
```diff
@@ -20,12 +27,31 @@ const routeLogin = async (request: any, reply: any) => {
...
+    orgList,
   })
 }
 
+// Authenticates user and checks they belong to requested org (id). Returns:
+// - userInfo (including orgId and orgName)
+// - template permissions
+// - JWT (with orgId included)
+const routeLoginOrg = async (request: any, reply: any) => {
+  const { userId, orgId } = request.body || { userId: '', orgId: '' }
```
andreievg, line 44: do we need userId in the body ?

src/components/postgresConnect.ts
```diff
@@ -397,6 +396,52 @@ class PostgresDB {
     }
   }
 
+  public getUserOrgs = async (userId: number) => {
+    const text = `
+    SELECT organisation.id AS "orgId",
+    user_role as "orgRole",
```
andreievg, line 402: not sure what the intended function of user_role is

src/server.ts
```diff
@@ -44,11 +45,10 @@ const startServer = async () => {
     return reply.sendFile(filename)
   })
 
-  server.get('/userInfo', routeUserPermissions)
-
+  server.get('/user-info', routeUserInfo)
```
andreievg, line 48: Was wondering why we would switch to hyphens (apart from search engine optimisation) ?

andreievg (changes requested):
Looks pretty good, but a few comments and questions.
Also:

Can we consolidate this `const token = (request?.headers?.authorization || '').replace('Bearer ', '')` in a helper method (as part of token validation maybe) ? 

Also was wondering if we can consolidate some of PG queries ?

At the very least getUserOrgs can be used to verifyUserInOrganisation.

Maybe the getUserInfo actually returns everything we need (a left join with user_organisation and organisation) ? (simple filter and map can decide if user belongs to org, gets a list of orgs and gets the name of org from id, the three things we are doing with db helpers). Or would that be pretty messy ?

CarlosNZ:
> Can we consolidate this `const token = (request?.headers?.authorization || '').replace('Bearer ', '')` in a helper method (as part of token validation maybe) ?

a30b6a6 ✅

CarlosNZ:
> Also was wondering if we can consolidate some of PG queries ?
> 

Yes, for sure. Will do. 
> 
> At the very least getUserOrgs can be used to verifyUserInOrganisation.
> 
I can get the three org queries down to one. 
> 
> Maybe the getUserInfo actually returns everything we need (a left join with user_organisation and organisation) ? (simple filter and map can decide if user belongs to org, gets a list of orgs and gets the name of org from id, the three things we are doing with db helpers). Or would that be pretty messy ?

I think it's getting a bit messy. The trouble is, we've got three different endpoints (/login, /login-org, /user-info) which have a lot of overlap, but with only slightly different input and output shapes. They look up mostly the same info from the database, but some take username, some take userId. Some return org, some don't etc. 

I think I can come up with a structure that's not too complicated while minimizing the number of overlapping database calls.

src/components/permissions/loginHelpers.ts
CarlosNZ: Yeah I left it in because in order to use the same `getUserInfo` method it would need the username. But, as per my last comment, I think I can come up with something better that is also more flexible.

src/components/permissions/loginHelpers.ts
CarlosNZ: Yeah, kind of. As we discussed in person, I think the nonRegisterd JWT should only be provided when the front-end specifically requests it, and we already have the login in place for that to happen when it's needed.

src/components/permissions/loginHelpers.ts
CarlosNZ: Yeah my plan is to have an object input for most of these helper methods and they check for which fields are defined for how to proceed. Will update this morning.

src/components/postgresConnect.ts
CarlosNZ: We haven't really discussed how org role would be used, but we will need at least some concept of a company "Owner" to distinguish them from just regular company members. For now it's just a string field and I've put "Owner" in a couple of them, but it's not really doing anything yet.

src/server.ts
CarlosNZ: From what I've read, snake-case is recommended over camelCase for endpoint and browser paths. Strictly speaking, we should make them case-insensitive too, so /login gives the same thing as /Login. 

But I'm not super fussed, was just trying to follow conventions.

src/components/permissions/routes.ts
CarlosNZ: You mean should it just come from the JWT?

Yeah I guess that would be fine. OrgId too then, I guess. So the body can be empty?

CarlosNZ: Ah, not the orgId, it's not in JWT yet. Got it.

CarlosNZ:
Okay, @andreievg, I've simplified it now and consolidated the userOrg queries.

Here's a simplified overview of the endpoints and where they fetch info from:
[Login endpoints.pdf](https://github.com/openmsupply/application-manager-server/files/5817249/Login.endpoints.pdf)

A little bit more simplification could probably be done in making `routeLoginOrg` and `routeUserInfo` have less duplication of functionality, but I don't think it's very much for now, and it's fairly simple to make sense of.

src/components/permissions/loginHelpers.ts
andreievg: Ok, I still think this is not an internal error 500 material, but maybe 401 (which front end can handle)

CarlosNZ: Sure, the actual try/catch here is basically the same as you had it originally:
```
    try {
      username = (await verifyPromise(jwtToken, config.jwtSecret)).username
    } catch (e) {
      console.log('cannot verify JWT in authorisation header')
    }
```
Happy to change it to something else. What should the return content be on error?

src/server.ts
andreievg: ok, I guess it's a bit of a longer discussion, to think about for future projects. I am all good with`kebab-case` or `snake_case` for end points

But just my thoughts, for maybe other projects:

* I've seen plenty of APIs with camelCase (https://developers.google.com/calendar/v3/reference/, DHIS2)
* Is it better to have one case across all of the code base ? Or use recommended conventions from external sources ?
* I am personally leaning toward just using one type of case 'everywhere' for DB for endPoints, files etc.. (previous restriction to using say `camelCase` uniformly across all areas of project where to do with technological barriers, which dont' exists anymore)

CarlosNZ: >     * I've seen plenty of APIs with camelCase (https://developers.google.com/calendar/v3/reference/, DHIS2)

Fair enough, although a lot of URLs with capitals in them are actually case insensitive. e.g. `https://en.wikipedia.org/wiki/Main_Page` also accepts `https://en.wikipedia.org/wiki/main_page` -- it redirects to the canonical one. 

> I am all good with `kebab-case` or `snake_case` for end points

Oh, I thought kebab-case was called snake-case. My bad.

> Is it better to have one case across all of the code base ? Or use recommended conventions from external sources ?

Recommended conventions for me. If I'm using Python, I'm writing in snake_case. Postgres -- snake_case, for reasons previously discussed. Not that it's relevant here, but whenever using capital letters, you need to be 100% sure that case will be preserved throughout your whole workflow, and that's not always the case (Postgres for example, although I know there's ways round that).

CarlosNZ: > Fair enough, although a lot of URLs with capitals in them are actually case insensitive. e.g. `https://en.wikipedia.org/wiki/Main_Page` also accepts `https://en.wikipedia.org/wiki/main_page` -- it redirects to the canonical one.

I guess for endpoints it's not that important, since its intended for access by professionals, but if you're going to have any capitalised URLs that are public-facing, you _must_ make sure that they're case-insensitive -- people would be annoyed if they got a 404 for not capitalising their URLs properly.

CarlosNZ: >     * I've seen plenty of APIs with camelCase (https://developers.google.com/calendar/v3/reference/

Bad example! The camelCase values in all those urls are variable substitutions (e.g. `calendarId` would be a number) - so the actual URLs _would_ be all lower-case 😆

andreievg (commented):
as per our covo, please see below

src/server.ts
andreievg: > Bad example! The camelCase values in all those urls are variable substitutions (e.g. calendarId would be a number) - so the actual URLs would be all lower-case 😆

check out `calendarList` <- i should have been more specific

Thanks for the reply, appreciated. I'll still push the idea of uniform case (probably camel), for our future apps.

> I guess for endpoints it's not that important, since its intended for access by professionals, but if you're going to have any capitalised URLs that are public-facing, you must make sure that they're case-insensitive -- people would be annoyed if they got a 404 for not capitalising their URLs properly.

if pps are playing around with URLs (very low % of users), they should expect to get 404 if they miss type. (interesting enough `-`and `_` are also only a `shift` key away : )

I guess if we are using a router that is case sensitive, we will 100% have to commit to kebab or snake

src/components/permissions/loginHelpers.ts
```diff
@@ -8,34 +8,43 @@ import { compileJWT } from './rowLevelPolicyHelpers'
...
+  userId?: number
+  orgId?: number
+}
+
+const getUserInfo = async (userProperties: any) => {
+  const { username, userId, orgId } = userProperties
+
+  const user =
+    // Don't look up user in DB if we already have it
+    Object.keys(userProperties).length > 3
```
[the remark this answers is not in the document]
CarlosNZ: Okay, @andreievg, have simplified as discussed, and added better error handling for invalid JWT tokens.

Cheers.

src/types.ts
```diff
@@ -146,3 +146,24 @@ export interface TriggerQueueUpdatePayload {
   id: number
   status: TriggerStatus
 }
+
+export interface User {
```
andreievg, line 150: User and Organisation are imported in routes.ts, but not actually used

src/components/permissions/routes.ts
```diff
@@ -1,44 +1,81 @@
...
-  const { username, password } = request.body || { username: '', password: '' }
-  const { passwordHash } = (await databaseConnect.getUserDataByUsername(username)) || {}
+  const { username, password } = request.body
+  if (password === undefined) return reply.send({ success: false })
 
-  if (!passwordHash) return reply.send({ success: false })
+  const userOrgInfo: UserOrg[] = (await databaseConnect.getUserOrgData({ username })) || {}
 
-  if (!(await bcrypt.compare(password, passwordHash))) return reply.send({ success: false })
+  const { userId, firstName, lastName, email, dateOfBirth, passwordHash } = userOrgInfo[0]
```
andreievg, line 33: looks like only the userId and passowrdHash are used in this method

src/components/permissions/loginHelpers.ts
```diff
@@ -4,38 +4,75 @@ import { verify, sign } from 'jsonwebtoken'
 import { promisify } from 'util'
 import { PermissionRow, TemplatePermissions, UserInfo } from './types'
```
andreievg, line 5: looks like UserInfo is not used anymore

src/components/permissions/loginHelpers.ts
andreievg: The way you've done it i the lates changes looks good (it still returns values which is dealt with further down the track, rather then throwing an error. It's good in this case since the error is easily handled)

src/components/postgresConnect.ts
```diff
@@ -376,22 +377,25 @@ class PostgresDB {
...
-      first_name as "firstName",
-      last_name as "lastName",
-      id as "userId",
-      username, date_of_birth as "dateOfBirth",
-      email,
-      password_hash as "passwordHash"
-      FROM "user"
-      WHERE username = $1
-    `
+  public getUserOrgData = async ({ username, userId }: any) => {
```
andreievg, line 380: Since this PR has now gone through quite a few iterations, I am happy to have it this way (with username and userID), so not a required change

Ideally though:

* Keep DB abstraction free from logic (i.e. methods getUserOrgDataByUserId, getuserOrgDataByUsername)
* The view has casting to camelCase alias (i.e. just `SELECT *` when view is used)

src/types.ts
andreievg: do we need these types ?

andreievg (changes requested):
Look Great ! Thanks.

Just a couple of tidy ups, with types and their unused imports

src/types.ts
CarlosNZ: Cool, thanks. Left over from when they were being used.

230858ab

andreievg:
@CarlosNZ, while looking at https://github.com/openmsupply/application-manager-server/pull/181, i ran the test, and i think it's failing due to changes to compileJWT interface

src/components/postgresConnect.ts
CarlosNZ: >     * Keep DB abstraction free from logic (i.e. methods getUserOrgDataByUserId, getuserOrgDataByUsername)

Right, just that the two methods would be basically identical except for the WHERE criteria.

>     * The view has casting to camelCase alias (i.e. just `SELECT *` when view is used)

The thing I don't like about that approach is that you have to remember to quote the camelCase fields everytime you refer to them:
```WHERE "userId" = $1```

Isn't it better to leave everything in postGres as underscore_case and just do the conversions outside?

src/components/permissions/routes.ts
CarlosNZ: > @CarlosNZ, while looking at #181, i ran the test, and i think it's failing due to changes to compileJWT interface

Okay, looking into now. Thx

src/components/postgresConnect.ts
andreievg: Hmm, I can this subject having mixed opinions. I should have worded my comment better, instead stead of `Ideally though` -> `In my opinion`. 

> Right, just that the two methods would be basically identical except for the WHERE criteria.

Just FYI I've done something similar in the permissions PR:

https://github.com/openmsupply/application-manager-server/blob/af2881923936ee76a0b905c6fae9fa1e4de5e9e7/database/buildSchema/13_template_permission.sql#L11-L29

https://github.com/openmsupply/application-manager-server/blob/af2881923936ee76a0b905c6fae9fa1e4de5e9e7/src/components/postgresConnect.ts#L533-L534

https://github.com/openmsupply/application-manager-server/blob/af2881923936ee76a0b905c6fae9fa1e4de5e9e7/src/components/postgresConnect.ts#L544-L545

(I just realised it's kind of weird that the view is names with kebab but the output columns are all camel, I think I should have named the view with camel, which would have helped to indicate that it's fields are also in camel)

>Isn't it better to leave everything in postGres as underscore_case and just do the conversions outside?

This kind of brings on that `case` discussion. A per our previous convo, I am still of the opinion that everything should just be one type of case (even if it requires quotation marks). But I guess I won't be pushing this point forward. 

This PR looks good, just pending test as per https://github.com/openmsupply/application-manager-server/pull/180#issuecomment-761894339

src/components/permissions/routes.ts
CarlosNZ: Okay, should all be passing now. (after a `yarn database_init`)

CarlosNZ:
@andreievg -- don't know if you've got time to look at the front-end partner of this? Cos I won't be able to merge this one until I can merge that at the same time -- this PR will cause breaking changes in front-end otherwise.

andreievg (approved):
Thanks for changes.

> @andreievg -- don't know if you've got time to look at the front-end partner of this? Cos I won't be able to merge this one until I can merge that at the same time -- this PR will cause breaking changes in front-end otherwise.

I'll have to look at it tomorrow or this evening.

Issue #172, opened 2021-01-08
Break-down insertData script


nmadruga:
The function insertData can be the same, but call different ones.

Change the mutation returns the `id` or `name`.

Please, when making this one make sure no one else is editing the `insertData` script

CarlosNZ:
When doing this, need to find an appropriate time when there's not any other branches with changes to `insertData`, otherwise merging those after this will be quite messy. 🤪

PR #181, opened 2021-01-17
#172 Break down insertData.js

CarlosNZ:
Implements #172.

**NOTE**: This is branched from #180 as I wanted to make sure those changes were incorporated into the breakdown, so when testing you'll want to use the front-end branch from [PR #266](https://github.com/openmsupply/application-manager-web-app/pull/266)

Hopefully there's no outstanding changes to `insertData.js` still to be merged (other than the one I'm merging this into) as they'll have to be manually refactored to conform to this breakdown.

- Have split `insertData` into seperate data files, labelled accordingly. The file `insertData.js` itself is purely the logic -- it looks in the `inserData` folder and imports any files it finds in there.

- Have also made a few minor tidy-ups to the `initialise_database.sh` script to make its output more readable.

- regarding the verbose output of from the Postgraphile server, it's possible to turn it off using `disableQueryLog: true` in `.postgraphilerc.js`, but it's all or nothing -- so Errors are suppressed too, so it's probably best to leave that on for now.

Question for @andreievg: wondering what the purpose of this line in `row_level_security_policy_related:
```RAISE NOTICE 'not creating role my_role -- it already exists';```
It prints this message to the console when it runs -- do we need it?

### CODE REVIEW

**Comment by andreievg** [comment]
> Question for @andreievg: wondering what the purpose of this line in `row_level_security_policy_related:
RAISE NOTICE 'not creating role my_role -- it already exists';
It prints this message to the console when it runs -- do we need it?

Don't need the `notice` but need the `exception` ( unless schema drop in `01_createSchema` drops the user )

File: .postgraphilerc.js [hunk 1]
```diff
@@ -10,6 +10,7 @@ module.exports = {
     cors: true,
     dynamicJson: true,
     jwtSecret: 'devsecret',
+    // disableQueryLog: true,
```
andreievg, line 13: is this to indicate that we have this option ?

database/insertData/01_template_C_userRegistration 2.js
```diff
@@ -0,0 +1,216 @@
+/*
```
andreievg, line 1: should the name of this file have a 2 at the end ?

database/insertData_SAFETY.js
```diff
@@ -0,0 +1,2340 @@
+const fetch = require('node-fetch')
```
andreievg, line 1: is this a back up file ?

andreievg (changes requested):
Looks good, just a couple of questions really

.postgraphilerc.js
CarlosNZ: Yeah, pretty much. Can just un-comment this line whenever we want to turn it on.

database/insertData/01_template_C_userRegistration 2.js
CarlosNZ: Nope. Good catch. 3079b6ef

database/insertData_SAFETY.js
CarlosNZ: Yes, this was the original file. Was just keeping it around in case I needed to compare the copied queries with the original. Was planning on deleting before merge.

andreievg:
Thanks for changes. I want to make sure all tests pass before approving, and noticed something (made a comment here: https://github.com/openmsupply/application-manager-server/pull/180#issuecomment-761894339)

CarlosNZ:
> Don't need the `notice` but need the `exception` ( unless schema drop in `01_createSchema` drops the user )

Okay, have added DROP USER to createSchema.

Pulled in changes from #180 now too -- all tests pass :) (after `yarn database_init`)

database/buildSchema/34_row_level_policy_related.sql
```diff
@@ -3,8 +3,8 @@ ALTER TABLE public.application ENABLE ROW LEVEL SECURITY;
 DO $$
 BEGIN
   CREATE ROLE graphile_user WITH NOLOGIN;
```
andreievg, line 5: now that you have drop user, we shouldn't need this condition/exception

CarlosNZ: Okay, does that mean I can replace the whole function block:
```
DO $$
BEGIN
  CREATE ROLE graphile_user WITH NOLOGIN;
  EXCEPTION WHEN DUPLICATE_OBJECT
  THEN RAISE NOTICE 'not creating role graphile_user -- it already exists';
END
$$;
```
with just:
```
CREATE ROLE graphile_user WITH NOLOGIN;
```
??

andreievg: yep, it worked for me when i tested it

andreievg (approved):

PR #182, opened 2021-01-18
#89F Radio buttons in template

CarlosNZ:
Adds radio button elements to Feature Showcase template to demonstrate implementation of [front-end PR#271](https://github.com/openmsupply/application-manager-web-app/pull/271).


nmadruga (commented):
I tested and it seems to be working as expected.

Just have one suggestion: Since `Qradio` is required, it need a error message to display when trying to submit the application without selecting one option. Now it just display the Modal, but no errors show up on the page:
Screen Shot 2021-01-22 at 11 15 59 AM

CarlosNZ:
> Just have one suggestion: Since `Qradio` is required, it need a error message to display when trying to submit the application without selecting one option. Now it just display the Modal, but no errors show up on the page:

That's a good point, but none of the other plugin types like "shortText" show anything if they're required but not yet started in this situation. Isn't this what we want? (If not, then it'll need to change for plugin types -- do we need to discuss this again?)

nmadruga:
> That's a good point, but none of the other plugin types like "shortText" show anything if they're required but not yet started in this situation. Isn't this what we want?

I think I wasn't very clear on **when** I'm saying that this problem is not showing up. It doesn't need to show the error message before the user tries to go to the next page (or for non-linear try to go to the Summary page). I just noticed that **after** I click the **Review and Submit** button with the field unselected I get a message about problems in the application, but nothing showing on the actual field that it is missing something. I thought it was only because it's missing some validation Message and criteria?

CarlosNZ:
> > That's a good point, but none of the other plugin types like "shortText" show anything if they're required but not yet started in this situation. Isn't this what we want?
> 
> I think I wasn't very clear on **when** I'm saying that this problem is not showing up. It doesn't need to show the error message before the user tries to go to the next page (or for non-linear try to go to the Summary page). I just noticed that **after** I click the **Review and Submit** button with the field unselected I get a message about problems in the application, but nothing showing on the actual field that it is missing something. I thought it was only because it's missing some validation Message and criteria?

So as per IRL disucssion -- we can discuss what should happen with required but not filled fields some other time?

nmadruga:
> we can discuss what should happen with required but not filled fields some other time?

Sure. I'll add in the card of things to talk to Debbie about

nmadruga (approved):
Good to go!

Issue #179, opened 2021-01-13
Application needs associated organisation


CarlosNZ:
Just realised, we prob need an additional "organisation_id" field on the application table. The user_id for the application is not sufficient to know which org the user applied as.

CarlosNZ:
The `getApplication` query should probably return the org id and user id too.

PR #186, opened 2021-01-21
#179 Add organisation link to applications

CarlosNZ:
Implements #179 

- add org_id field to application table referencing organisation.id
- add some org_ids to existing applications
- connect one more user to org


nmadruga (changes requested):
Testing to create a new User using the **Create new account** on the login page allows a new user to be created. The back-end have finished running the action to Create user, but when I try to login with that user it's not allowing. I thought it could be because the User won't have a `org_id` associated with this login?

Screen Shot 2021-01-22 at 11 28 27 AM

CarlosNZ:
> Since I'm not sure yet if it's a back-end or a front-end fix, I'll make a new issue and fix it separately. Thanks for picking it up .

Never mind, it was super-quick fix in back-end. Just needed to change the Action in User Registration template to send `response.hash` rather than `response.txt`. Should be all good now :)

nmadruga (approved):

Issue #185, opened 2021-01-20
Application VIEW with all relevant columns for list


CarlosNZ:
Create a VIEW table for Applications that can be queried by the `getApplications` query for the Application list. This will allow more columns to be sortable and will prevent the need to the current additional logic to insert stage/status info into application list.

Will need to update front-end to match new schema, and update sorting/filtering queries if necessary.

Please list here all columns that should be added (only needed for List, not individual applications -- they get queried separately)

- Application id
- Serial
- Application name
- Template name
- Template code
- Username
- User FirstName/LastName (suggest concatenating them into one column -- or is this a bad idea for certain languages?)
- Outcome
- Status
- Stage
- Organisation Name (need to do #179 first)
- Last Active Date (status_history_time)
- ????

nmadruga:
Status date
Expiry date -> need other issue done first
Consolidator first and last name
Consolidation status

(The following are per sections)
Reviewer first and last name
Review status, 
Review status date

PR #187, opened 2021-01-21
#185 Application List View

CarlosNZ:
Implement #185 -- create a new View to return all relevant data for the front-end Application List. 

Will add to this as we implement more filters (by Reviewer, Consolidator, etc.)

Requires [front-end PR #289](https://github.com/openmsupply/application-manager-web-app/pull/289)


nmadruga (approved):
Seems all good, and simple. I'll have a look on Front-end now and how it's using it.

Issue #190, opened 2021-01-25
Evaluator -- new operator: "?" conditional result


CarlosNZ:
This would be quite simple to implement, and I think quite useful. It occurred to me when trying to create an expression to display different text depending on a prior response. Currently, the only way to do it is to create two almost identical text elements, and have their visibility dependent on the result in question. It's a bit clunky.

I think it would nicer to have a "?" operator which works exactly the same as the ternary `?` in javascript. It takes three child elements - the first one is the expression to evaluate, 2nd is the output if expression is true, 3rd if false.

Issue #114, opened 2020-11-16
Create Company Action


nmadruga:
To be included as part of the "Create Company" template. After the template is Approved - `Outcome` the new company should get created using specific fields entered in the template. It would require updating the `company` table to add the basic fields that should be kept for a company.

PR #192, opened 2021-01-26
#190 Evaluator -- Conditional operator

CarlosNZ:
Very small improvement, so I just squeezed this in. Implements #190 which adds a Conditional `?` operator to evaluator. Works exactly like the javascript ternary (`?`) operator.

The actual implementation was two lines of code. Docs and tests slightly more :) 

`yarn test ./src/modules/expression-evaluator/src/evaluateExpression.test.ts` to run evaluator test suite. (On newly initialised database)

### DIFF
--- a/documentation/Query-Syntax.md
+++ b/documentation/Query-Syntax.md
@@ -116,6 +116,20 @@ Adds any number of number nodes together
 - Input: 1 or more nodes of **number** type.
 - Output: **number**
 
+## ? (Conditional)
+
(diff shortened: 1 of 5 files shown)


nmadruga (approved):
Very nice.

Be aware that some tests needs to be updated when we merge with the latest changes on the database (from issue #114 Create Organisation Action) I the test with the previous database version and saw a couple of errors, but all good with this one.

CarlosNZ:
Have re-published and bumped the version in both repos, so please re-run `yarn install` to get latest.

PR #193, opened 2021-01-27
#114 Create Organisation Action

CarlosNZ:
Implements #114.

The actual "createOrg" action is very simple -- it's basically a copy of createUser, but with some fields and values changed.

However, I also:

- hardcoded IDs for many elements into templates and applications (Because this is quite fiddly and hard to keep track of, I only added IDs for entities that have (or are likely to have) other entities with references to them, i.e. template elements, applications, application responses, etc.)
- Added `orgId` to `application_stage_status_all` view (so it can be acquired by `fetchDataFromTrigger`)
- Updated Company Registration template:
   - Changed name to "Organisation Registration"
   - Changed some questions to make them pertinent to Org input data
   - Added Action to create org and change outcome on Review Submit (currently happens regardless, but need to add condition to check for approval once schema for that is settled)

**NOTE**: this PR now includes changes from #194, as we wanted to get that merged for branching off this one.


database/insertData/01_template_C_userRegistration.js
```diff
@@ -16,12 +16,14 @@ exports.queries = [
           templateSectionsUsingId: {
             create: [
               {
+                id: 1004
```
nmadruga, line 19:
```suggestion
                id: 1002
```

Just because previous ones are 1000 and 1001

database/insertData/01_template_D_basicReviewTesting.js
```diff
@@ -15,12 +15,14 @@ exports.queries = [
           templateSectionsUsingId: {
             create: [
               {
+                id: 1005
```
nmadruga, line 18:
```suggestion
                id: 1003
```

database/insertData/03_organisations.js
```diff
@@ -65,115 +65,144 @@ exports.queries = [
...
-          user {
-            username
-          }
-          organisation {
-            name
-          }
+    createUserOrganisation(
+      input: {
+        userOrganisation: {
+          #id: 2
```
nmadruga, line 89: Would this comment be in? Want to do the same with the next userOrg ones?

nmadruga (changes requested):
There are a couple of tests failing, not the tests made for this PR but old ones after the changes were done to the `ids` that needs to be updated. : (

Overall the functions look good to me. I'll try out and leave a comment if anything doesn't go as expected here.

CarlosNZ:
> There are a couple of tests failing, not the tests made for this PR but old ones after the changes were done to the `ids` that needs to be updated. : (

Okay, thanks for catching this. Have updated tests now. f12ff0e

database/insertData/01_template_C_userRegistration.js
CarlosNZ: No, the previous **Section** IDs (in Template B) are 1002, 1003 -- perhaps you got mixed up with the Template or Element IDs? After fresh init, you can see the sequential IDs. :)  🤷‍♂️

Screen Shot 2021-01-29 at 11 35 34 PM

database/insertData/01_template_D_basicReviewTesting.js
CarlosNZ: Same. ;)

database/insertData/03_organisations.js
CarlosNZ: Okay, don't need it. This was leftover from when I was still thinking about putting the IDs in for these records as well, but it was getting unnecessarily complicated. Gone now :) af768c8

database/insertData/01_template_C_userRegistration.js
nmadruga: Yes, sorry about my confusion : D

database/insertData/03_organisations.js
nmadruga: Cool. Thanks!

nmadruga:
All good now with tests. Approving this one now!

nmadruga (approved):

Issue #113, opened 2020-11-13
Create action to join username to company when application to 'Create company' is approved


nmadruga:
Should be done after issue are completed:
- #97 - Create GraphQL for 'Join company' application 
- #104 - Extend action to grant permission for user link to a company

Consider that a user who is Creating a company will be granted **higher** privileges than the user that is joining a company. It doesn't need to be done in this task, but from what is thought in here would be good to create new tasks to consider what other permission this user should have. For example: add other users to this company, remove users from this company. And all that would be done through the application workflow.

PR #198, opened 2021-01-28
#113 Action plugin -- add User to Org

CarlosNZ:
Implements #114. Simple plugin to create an entry in `user_organisation` tables. Takes input `{ user_id, organisation_id }`

Updated "Organisation Registration" template to create this join after creating organisation. Assumes applying user is "Owner" of new Org.

@andreievg -- I know you weren't keen on the need for an "Output" object for sequential Actions. This seems like a clear use case though: the action to join the user to the org needs the newly created ID from the previous "CreateOrg" action. Is there another way this could have been achieved without passing along the output?


andreievg (approved):
Looks good, and precise : )

@CarlosNZ 
> @andreievg -- I know you weren't keen on the need for an "Output" object for sequential Actions. This seems like a clear use case though: the action to join the user to the org needs the newly created ID from the previous "CreateOrg" action. Is there another way this could have been achieved without passing along the output?

I think in this case an ouptut orgId, for the use in sequential `give permissions` action is quite elegant (👍 👍). Non output option is possible but is somewhat (can chat about this one). It sounds like we also need to think through an 'outcome' mechanism to populate application `result` table.

CarlosNZ:
> It sounds like we also need to think through an 'outcome' mechanism to populate application `result` table

Not sure exactly what you mean, but sure.

Issue #196, opened 2021-01-27
Add is_fully_assigned field to Application List View


CarlosNZ:
Based on discussion in [#198](https://github.com/openmsupply/application-manager-server/issues/189#issuecomment-767821740):

Add additional column to `application_list` view to show whether the application is fully assigned for the current stage.

For level 2+, any assignment is fully assigned (whole review), it's only for level 1, with partial assignment, that this is required.

So query for this field needs to:

- Get Sections for Template
- Get current level (by looking at MAX(level) for this application/stage in review_assignment)
- Get all "assigned" records for this level in review_assignment -- just "Available Sections" column
- Compare with initial Template section array and return true or false.

PR #200, opened 2021-01-29
#196 ApplicationList -- is_fully_assigned column

CarlosNZ:
Just wanted to get this PR up before I left. I think it works, but it'll probably need to be thoroughly tested, I've probably missed something.

(You can manually add entries in the review_assignment table and just fiddle with the available_sections field to see if it's working properly.)

Any suggestions as to how to make the query simpler welcome. It basically follows the logic I outlined on the original issue (#196)

Had to move the file to later in the sequence, otherwise it didn't already have review_assignment table created.


database/buildSchema/36_application_list_view.sql
```diff
@@ -0,0 +1,60 @@
...
+    template.name as template_name,
+	username as applicant_username,
+	first_name as applicant_first_name,
+	last_name as applicant_last_name,
+	CONCAT(first_name, ' ', last_name) as applicant,
+	org.name as org_name,
+	stage_status.stage, stage_status.status, outcome,
+	status_history_time_created as last_active_date,
+	(
+	WITH count_unassigned_sections AS (
```
andreievg, line 16: I would extract this into a method or view. (an not sure if the 'WITH' statement is needed, to me the best use of 'WITH' is for recursiveness/loops within a query).

Personally I would do it like this:

```PG

/* make into method 'application_reviewable_questions' */

SELECT count(*) 
FROM template_section
JOIN template_element on template_element.section_id = template_section.id
WHERE
	template_section.template_id = 4 /* from method parameter */ and template_element.category != 'Information'

/* make into method 'application_assigned_questions' */

SELECT count(distinct(template_element.id))
FROM review_assignment
JOIN review_question_assignment on review_question_assignment.review_assignment_id = review_assignment.id
JOIN template_element on template_element.id = review_question_assignment.template_element_id 
WHERE
	review_assignment.status = 'Assigned'
	AND review_assignment.stage_id = 5 /* from parameter */
	AND review_assignment.level = 1 /* from parameter */
	AND review_assignment.application_id = 4000 /* from parameter (application.id) */
	AND template_element.category != 'Information'

/* in application_lise view */

application_assigned_questions(app) >= application_reviewable_questions(app, state_id, 1) as is_fully_assigned
```

Here are two branches which i used to check results of the select statement:

https://github.com/openmsupply/application-manager-server/compare/%23196-is_fully_assigned-to-application-list...%23196-review-assignments-not-fully-assigned?expand=1

and 

https://github.com/openmsupply/application-manager-server/compare/%23196-is_fully_assigned-to-application-list...%23196-review-assignment-fully-assigned?expand=1

andreievg (changes requested):
A few suggestions.

Just fyi we can use existing computer column methods (used for GraphQL), in select statements (just pass as parameter the table name or the alias given to table name in the from query)

database/buildSchema/36_application_list_view.sql
```diff
@@ -0,0 +1,60 @@
...
+			ON ts.template_id = a.template_id
+			WHERE a.id = app.id)
+				
+			EXCEPT
+					
+			(SELECT DISTINCT unnest(available_template_section_ids) AS sections
+			FROM review_assignment
+			WHERE application_id = app.id
+			AND level=(
+				SELECT MAX(level) FROM review_assignment
```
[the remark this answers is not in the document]
CarlosNZ: The max level that is _currently_ in the review_assignment table should be the _current_ level. (i.e. it's not the max level overall, it's the max that is currently in the review_assignment table).

Or were you referring to something else?

CarlosNZ:
Okay, have another look @andreievg . The subquery to get only the current stage unassigned count is a bit gnarly, so le me know if you can think of a nicer way to handle that. Is it worth creating a View with the review_assignment JOINed to current stage and include the unassigned count on that? (So that the application_list query just has a simple Join to that view)?

database/buildSchema/28_review_question_assignment.sql
```diff
@@ -5,3 +5,31 @@ CREATE TABLE public.review_question_assignment (
...
+		AND ra.level = 1 -- currently restrict partial assignment to level 1
+		AND ra.application_id = curr_ra.application_id
+		AND template_element.category = 'Question'
+$$ LANGUAGE sql STABLE;
+
+-- Function to return count of template questions for a given application/template
+CREATE FUNCTION public.template_question_count(application_id int)
+RETURNS bigint AS $$
+	SELECT COUNT(*)
+		FROM application JOIN template ON application.template_id = template.id
```
andreievg, line 29: It seems easier to have this method return question_count for application (and pass application rather the application_id) ?

then you don't need to select application, just app.template_id matches template

(it can even be used in GraphQL, if ever needed))

database/buildSchema/28_review_question_assignment.sql
```diff
@@ -5,3 +5,31 @@ CREATE TABLE public.review_question_assignment (
 	template_element_id integer references public.template_element(id),
 	review_assignment_id integer references public.review_assignment(id)
 );
+
+
+-- Function to return count of assigned questions for current stage/level
+CREATE FUNCTION public.assigned_questions_count(curr_ra public.review_assignment)
```
andreievg, line 11: Hmm. I still think it's better to have this method to accept the following params:
`application`, `stage_id`, `level`. 

It would avoid extra sub queries. And should look neater.

andreievg (changes requested):
> Okay, have another look @andreievg . The subquery to get only the current stage unassigned count is a bit gnarly, so le me know if you can think of a nicer way to handle that. Is it worth creating a View with the review_assignment JOINed to current stage and include the unassigned count on that? (So that the application_list query just has a simple Join to that view)?

I this one more iteration would simplify this a bit more

I like the idea of using methods that takes 'core' table from the query (in this case application) IF they return simple data type (i.e. in this case big int for `application_question_count(application)` and `application_assigned_questions(application, stage_id, level)`.

And if the data type is more complex (say a selection or single record from another table), only then using a join (like you've done with application_stage_status_latest)

andreievg:
I just did a quick check, you can actually use multi parameter methods in Graphile (i.e. application_assigned_questions as per above, would still be usable from GraphQL (if ever needed))

database/buildSchema/28_review_question_assignment.sql
CarlosNZ: Cool, have made suggested changes, seems to be working well now.

I've returned the `is_fully_assigned` column as `is_fully_assigned_level_1` -- you think this is okay?

andreievg (approved):
Thanks for changes @CarlosNZ 

And I think it's probably better to name the column `is_fully_assigned_level_1` as you did

PR #202, opened 2021-02-01
#201 Evaluator -- stringSubstitution bug

CarlosNZ:
Very quick fix. 

Annoying how in javascript an empty string evaluates to `false` -- that's caught me out many times now. Anyway, that's all it was. 

Added additional test to catch this.


nmadruga (approved):
Seems very simple and a bug fix.

All tests are running correctly.

Issue #104, opened 2020-11-13
Extend action to grant permission for user link to a company


nmadruga:
Extend work done in #103 to grant the permission to a user **in** a company. This will be required for permissions that are given to users when they are logged in to a company.

CarlosNZ:
Can I just clarify what this issue is requiring, I'm a bit confused?

Are we extending the existing `grantPermissions` Action so that it can grant a permission name to a `orgId` as well as a `userId`? (Currently just does user)

Or does it mean something else?

PR #203, opened 2021-02-02
#104 Extend grant permissions to organisation

CarlosNZ:
Implements #104. 

List of all changes in this PR:

- grantPermissions now handles an optional `orgName` parameter (main issue)
- schema: changed `permission_join` table to organisation_id field instead of user_org_id, as discussed last week
- Moved checking for existing `permission_join` records to Database functionality (using `ON CONFLICT`), and created relevant UNIQUE constraints required.
- postGres returns IDs of newly created PermissionJoin records (which are passed into Action output)
- UNIQUE constraint on `organisation.name`
- created a new "Permission Name": "reviewJoinCompany" which uses "basicReview" policy
- renumbered files in insertData folder to accommodate above insertion
- fixed spelling of "permission" in 3 .sql filenames
- added "grantPermissions" action to Organisation Rego template
- Misc fixes to Org Rego template
- Additional tests for Action (run `yarn test ./src/plugins/action_grant_permissions/src/grantPermissions.test.ts` )


nmadruga:
@CarlosNZ There are a few conflicts in this PR, could you have a look when you get some time please?

CarlosNZ:
> @CarlosNZ There are a few conflicts in this PR, could you have a look when you get some time please?

Should be good now. :)

src/plugins/action_grant_permissions/src/plugin.json
```diff
@@ -4,8 +4,9 @@
   "name": "Grant Permissions",
   "file": "grantPermissions",
   "function_name": "grantPermissions",
-  "description": "Grants permission to user, creates permission join from user to permission name",
-  "required_parameters": ["username", "permissionNames"],
+  "description": "Grants permission to user/org, creates permission join from user/org to permission name",
```
andreievg, line 7: would be good to say that if orgName is missing, it will grant permission to user and org combination, otherwise just to user.  Also orgName is not required parameter ? Do we need an optional parameter field ?

andreievg (approved):
Looks good, 

just a minor comment

src/plugins/action_grant_permissions/src/plugin.json
CarlosNZ: > would be good to say that if orgName is missing, it will grant permission to user and org combination, otherwise just to user

Okay, done 68d56332

> Also orgName is not required parameter ? Do we need an optional parameter field ?

Ah, you've spotted this! Yeah, it's been bugging me for a bit now, the parameters aren't actually "required", they're just "available". Fortunately, this list is not actually used at all -- it's more for informational purposes, the triggerProcess evaluates whatever parameters that come in from the templateAction. 

This field is stored to the database thought (in action_plugin), with the idea that they would be used in the UI for the TemplateBuilder.

**3 options**:
1. Leave as is for now, worry about it later
2. Change the field name to "available_parameters" (or something similar)
3. Split it into two fields -- "required_parameters" and "optional_parameters"

Options 2 or 3 would require a little change to the schema and the "registerPlugins" method.

CarlosNZ: Will merge this now, but let me know and I'll make an issue to do the above.

Issue #188, opened 2021-01-25
Insert different levels of reviews in different stages of ReviewTest template


nmadruga:

nmadruga:
We do need to add the information of how many level of review are required in each stage of the template.
Quick discussing with @CarlosNZ we see the need of adding one `review_decision` per review, regardless if it is a consolidation review or not. In the case that there is only one level of review => When the reviewer submits the review, the `review_decision` should also get created (in the same mutation) and since there is only one level of review - this should be used to change the application stage (in case of approval). It seems reasonable to me that, in order to change the application stage, this reviewer needs to have consolidator permission on that stage+template. 
In the UI we can deal as if - review is completed as usual. Then if the user has `consolidation` permission & is the final review level & all reviews have been finished (there may be a case that review is done by 2 reviewers), we would shouw the user the **Consolidation step** where they confirm the `review_decision` to be send back to the Applicant.

CarlosNZ:
Also I would add that with each review level (in the template) we should also have a "name". So names like "review", "initial review", "consolidate", etc. would be template-specific and not hard-coded into the system, similar to what we do with Stages. 

That way, the UI doesn't have to do any logic as to what heading/page names to show, it would just get `currentReviewLevel.name`, or something like that.

CarlosNZ:
Andrei's response in Telegram chat for example of how this is represented: 

> With current thinking:
>
> Stage 1:
> 
> template_permission (restrictions: `{ stage: 1, lvl: 1, canMakeOverallDecision: true }`) -> permission_name (name: "Screener") -> permission_policy (type: "Review, name: "BaseReviewer")
> 
> Stage 2:
> 
> template_permission (restrictions: `{ stage: 2, lvl: 1, section: 1 }`) -> permission_name (name: "Assessor of First Section") -> permission_policy (type: "Review", name: "BaseReviewer")
> 
> template_permission (restrictions: `{ stage: 2, lvl: 1, section: 2 }`) -> permission_name (name: "Assessoor of Second Section") -> permission_policy (type: "Review", name: "BaseReviewer")
> 
> template_permission (restrictions: `{ stage: 2, lvl: 2 }`) -> permission_name (name: "Senior Assessors") -> permission_policy (type: "Review", name: "Consolidator") <- this policy will prolly have slightly different rules to "BaseReviewer"
> 
> Stage 3 (needs a bit more thinking about):
> 
[...]

Issue #183, opened 2021-01-18
Add new Reviews for ReviewTest application with different status


nmadruga:

CarlosNZ:
To clarify -- did you mean the "ReviewTest" application? That's the one we're using for testing review stuff. "TestRego" is the Feature Showcase (for showing off plugins/functionality).

PR #204, opened 2021-02-03
#188 Review levels in ReviewTest template

CarlosNZ:
Had to refactor quite a bit to get this to work -- since most of the permissions were being created as part of the `createUser` section, they weren't available to add to templates that were inserted earlier.

So I've moved all policy creation and permissionName creation to their own files right at the beginning and the createUser mutations just link to existing permission names.

Added a new permission name "reviewReviewTest" with basicReview policy, linked to ReviewTest template and applies to all review stages and levels.

(Currently the new files are numbered 00A and 00B so I didn't have to change the names of other files. We'll renumber correctly at some point)

### DIFF
--- a/database/insertData/02_users.js
+++ b/database/insertData/02_users.js
@@ -122,7 +122,7 @@ exports.queries = [
     }
   }`,
   //
-  // Non Registered User with Permissions
+  // Non Registered User with Permissions for UserRegistration only
   // Password is blank
   `mutation {
     createUser(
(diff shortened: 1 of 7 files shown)


database/insertData/02_users.js
```diff
@@ -161,47 +141,57 @@ exports.queries = [
...
+      input: { permissionJoin: { permissionNameId: 3000, userId: 1 } }
+    ) {
+      permissionName {
+        name
+      }
+    }
+  }`,
+  `mutation userToPermissionNameJoin {
+    createPermissionJoin(
+      input: { permissionJoin: { permissionNameId: 3000, userId: 2 } }
```
nmadruga, line 191:
```suggestion
`mutation joinUsersToPermissionName {
    updatePermissionName(
      input: {
        patch: {
          permissionJoinsUsingId: {
            create: [{ userId: 1 }, { userId: 2 }, { userId: 3 }, { userId: 4 }]
          }
        }
        id: 2000
      }
    ) {
      permissionName {
        name
      }
    }
  }`,
  `mutation joinUsersToPermissionName {
    updatePermissionName(
      input: {
        patch: {
          permissionJoinsUsingId: {
            create: [{ userId: 1 }, { userId: 2 }]
          }
        }
        id: 3000
      }
```

nmadruga (commented):
Nice work!

I just added one suggestion to it. I will start working on top (for issue #183) of this one to include some Reviews and Consolidations preset to help us viewing the Review/Consolidator pages.

CarlosNZ:
> I just added one suggestion to it.

Great suggestion, I didn't realise you can do that :)

nmadruga:
> Great suggestion, I didn't realise you can do that :)

I think there are too many ways of doing similar things. But it's nice to have an array passed for things like this, so everytime you have a `create` inside another table you can pass an array instead of a single object fields.

CarlosNZ:
> > 
> 
> I think there are too many ways of doing similar things. But it's nice to have an array passed for things like this, so everytime you have a `create` inside another table you can pass an array instead of a single object fields.

Do you know if you can do the same thing at the top level? Like, creating a bunch of new users in one mutation rather than several single mutations? I've tried, but not been able to get it to work.

Issue #199, opened 2021-01-28
Action to set review_assignments to "Not Available" when reviewer is assigned (self assigment)


nmadruga:
Trigger: OnAssignment

- needs to check all sections are assigned, and only set "not_available" if all sections are.

CarlosNZ:
I think this is a duplicate of #195. Didn't we have something else that needed to happen on Review submit that we were going to include in this issue?

nmadruga:
The self-assignment should change other review_assigment statuses to NOT AVAILABLE. And for review level 1, although there might be more than one review_assigment (for the same sections/application) we don't change the status, so after the section is assigned to the reviewer there should be one ASSIGNED and others are still AVAILABLE. 

That is to simplify the re-assignment for now.

PR #215, opened 2021-02-10
#199 Set assignment status on assignment

CarlosNZ:
Implements #199. Branched from PR #213 -- please review that first.

**NOTE: This post has been UPDATED based on discussions below (and IRL)**

Action to set the `status` field for other assignments when one assignment becomes `assigned`.

Currently this works for ALL levels ~~(checks for fully assigned on level 1), but this can be changed within the `getNewReviewAssignmentStatus` helper function if we need to~~.

### How to see it working:

First make sure you can browse the `review_assignment` table.

I've added an additional application and an additional available reviewer for it, so we can see three reviewers for Stage 1.
These assignments are not yet created though, so trigger them with:
```UPDATE application SET trigger = 'onApplicationSubmit' WHERE id = 4003```
You should see three new review_assignment records (generated with the Action from #213), with status "Available"

Now "Assign" one of these records using:
```UPDATE review_assignment SET status = 'Assigned', trigger='onReviewAssign' WHERE id = 1```

Because this was "assigned" rather than "self-assigned" (based on trigger), the other records remain "Available".

~~Now see what happens when a record is assigned and it IS fully assigned:~~
~~First, let's pretend records 1001, 1002 are not assigned and reset them:~~
~~``` UPDATE review_assignment SET status = 'Available' WHERE id in (1001, 1002)```~~

~~Now Assign one of them:~~
~~```UPDATE review_assignment SET status = 'Assigned', trigger='onReviewAssign' WHERE id = 1001```~~

~~Now the other one (1002) shows as "Not Available" as all the questions are assigned (Note: we're faking this, since strictly some of the questions are assigned to 1002, but we're just showing how the action behaves even though it makes the database inconsistent)~~

And now see what happens for self-assignment (using records 1003, 1004 which initially have status "Available for self-assignment"):
```UPDATE review_assignment SET status = 'Assigned', trigger='onReviewSelfAssign' WHERE id = 1003```
The other one (1004) has been set to "Not available"

### CODE REVIEW

**Comment by andreievg** [comment]
@CarlosNZ sorry to be a pain on this one.

I think it should only set to not available when self assigning. So we would have another value for `trigger` enum `onReviewSelfAssigned`, and do basic logic of setting all other `review_assignmnets` for stage/level to `Not Available`.

Just to sum from our previous group discussions, for `review_assignments` that are assignable by assigner, there is no reason to set them as `Not Available` as the reviewers are still available (for reassignment)

Lastly, I was thinking if we need similar logic to review/assignment, where action actually changes the status and mutation just changes the `trigger`. I am not suggesting that we should implement it now, but wondering what you think about it.

**Comment by CarlosNZ** [comment]
> Just to sum from our previous group discussions, for `review_assignments` that are assignable by assigner, there is no reason to set them as `Not Available` as the reviewers are still available (for reassignment)

Yeah, but technically the higher-level reviewers are all "available for re-assignment" too, we just don't have any immediate plans to implement that functionality, but they still could be. One final alternative: how about we add a new value to the enum: "Available for re-assignment" which gets set for the L1 reviewers? That way we can still distinguish them from true "Available" (i.e. the review is not fully assigned)

If you don't think that's worth it, that's okay I'll stop banging on about it, I do think it's slightly mis-leading (and inconsistent) though to leave it as just "available".

> So we would have another value for `trigger` enum `onReviewSelfAssigned`, and do basic logic of setting all other `review_assignmnets` for stage/level to `Not Availabl`

We shouldn't need a different trigger value. This action already makes the distinction between level 1 reviewers and above -- it's just a matter of choosing what they get set to. (It's all contained within the `getNewReviewAssignmentStatus` function)

> Lastly, I was thinking if we need similar logic to review/assignment, where action actually changes the status and mutation just changes the `trigger`. I am not suggesting that we should implement it now, but wondering what you think about it.

Yeah maybe, although the review/application submission mutation actually updates all the response data along with the trigger, so it's doing "something", whereas with Assignment, changing the assignment "status" is literally all it needs to do, otherwise it would *only* be setting the Trigger, which seems a bit weird. Plus, the mutation needs to specify *what* the assignment status is being set to -- for now it's only "Assigned", but it could be changing it to "Not available" or something else at some point, in which case just changing the trigger wouldn't be enough to specify what it should be -- if that makes sense.

**Comment by andreievg** [comment]
> Yeah, but technically the higher-level reviewers are all "available for re-assignment" too, we just don't have any immediate plans to implement that functionality, but they still could be. One final alternative: how about we add a new value to the enum: "Available for re-assignment" which gets set for the L1 reviewers? That way we can still distinguish them from true "Available" (i.e. the review is not fully assigned)

I think the original point of 'Not Available' was to indicate that review is not self assignable (someone else  self assigned it already), this was the alternative to reviews auto created and being marked as locked when assigned was done. I don't remember us needing a way to identifying if review_assignment (for non self-assignable reviews) is 'not the one that was assigned when all sections/questions of application were assigned'. 

> "Available for re-assignment"

This would require front end querying either for 'available' or 'available for re-assignment' and would require a non essential logic to figure out which review_assignment should be set to what status (i can give a scenario or two).

> We shouldn't need a different trigger value. This action already makes the distinction between level 1 reviewers and above -- it's just a matter of choosing what they get set to. (It's all contained within the getNewReviewAssignmentStatus function)

Level 1 can also be self-assignable, I think it's best to indicate it with a trigger vs deducing from level in this instance.

> If you don't think that's worth it, that's okay I'll stop banging on about it, I do think it's slightly mis-leading (and inconsistent) though to leave it as just "available".

Let's just go back to the origin of why we wanted 'Not Available' status (as per above), it was to identify that reviewer cannot start a review because another reviewer already started it (only use case was for self assign scenario). Can we reach a compromise, and have a status that reads something like: 'Already self assigned' 

> Yeah maybe, although the review/application submission mutation actually updates all the response data along with the trigger, so it's doing "something", whereas with Assignment, changing the assignment "status" is literally all it needs to do, otherwise it would only be setting the Trigger, which seems a bit weird. Plus, the mutation needs to specify what the assignment status is being set to -- for now it's only "Assigned", but it could be changing it to "Not available" or something else at some point, in which case just changing the trigger wouldn't be enough to specify what it should be -- if that makes sense

Cool, let's can it for now

**Comment by CarlosNZ** [comment]
Okay, no problem, @andreievg. So to summarize the remaining changes needed:

- use a different trigger to differentiate between Assign and Self-Assign
- use that trigger to differentiate between whether other statuses get set to "Not Available" or left as-is
- (Maybe?) change "Not available" status to "Already self-assigned". (Probably better as it is TBH)

That sound about right?

I guess the one thing I'm not entirely clear about is we seem to closing the door to the possibility of re-assigning a L2+ review? I know we don't want to implement that option at this time, but wouldn't we still want to make that easy to do later? I just thought that, for the most part, we'd want to treat the process of "assignment" to being essentially the same regardless of whether it's done by "self" or a supervisor (obviously the UI would be different). But what you're suggesting (if I'm understanding it correctly) is that L2+ reviews *can't* be re-assigned and it would take quite a bit of re-design to make it possible. That true?

**Comment by andreievg** [comment]
> use a different trigger to differentiate between Assign and Self-Assign
use that trigger to differentiate between whether other statuses get set to "Not Available" or left as-is
(Maybe?) change "Not available" status to "Already self-assigned". (Probably better as it is TBH)

Yep 

> I guess the one thing I'm not entirely clear about is we seem to closing the door to the possibility of re-assigning a L2+ review? I know we don't want to implement that option at this time, but wouldn't we still want to make that easy to do later? I just thought that, for the most part, we'd want to treat the process of "assignment" to being essentially the same regardless of whether it's done by "self" or a supervisor (obviously the UI would be different). But what you're suggesting (if I'm understanding it correctly) is that L2+ reviews can't be re-assigned and it would take quite a bit of re-design to make it possible. That true?

> I guess the one thing I'm not entirely clear about is we seem to closing the door to the possibility of re-assigning a L2+ review? 

I would image the only thing would be left to do is change `Not Available`/`Already self-assigned` to `Available for self-assignment` (for all review back to available and lock the consolidator review). Who will be able to do that ? Not sure what role tbh. 

Come to think of it, I like `Already self-assigned`, it's more descriptive

**Comment by CarlosNZ** [comment]
> Come to think of it, I like `Already self-assigned`, it's more descriptive

I think it's slightly ambiguous in that "already self-assigned" could imply self-assigned to _this_ reviewer if read in isolation. How about "Self-assigned by Other" ? (Or something that makes it clear that its assigned to _another_ reviewer)

**Comment by CarlosNZ** [comment]
> > We shouldn't need a different trigger value. This action already makes the distinction between level 1 reviewers and above -- it's just a matter of choosing what they get set to. (It's all contained within the getNewReviewAssignmentStatus function)
> 
> Level 1 can also be self-assignable, I think it's best to indicate it with a trigger vs deducing from level in this instance.

Hey, just realised -- we can distinguish between assignment and self-assignment just by comparing the user_id of the current user to the reviewer_id on the review_assignment record, so we shouldn't need to use a different trigger.

I'd prefer not to make a new trigger if possible, as we'd need to write two different actions for every assignment in each template

**Comment by CarlosNZ** [comment]
>     * use a different trigger to differentiate between Assign and Self-Assign
> 
>     * use that trigger to differentiate between whether other statuses get set to "Not Available" or left as-is
> 
>     * (Maybe?) change "Not available" status to "Already self-assigned". (Probably better as it is TBH)

Okay, have done those changes now. (used "Self-assigned to another" for the "Not Available" status)

I'm slightly bothered by the fact that we now seem to be using different rules for the "create review_assignment records" action and this one. The other one used L1 vs L2+ to determine whether to set "Available" or "Available for self-assignment", but now we're using the different trigger because, as you said, "Level 1 could be self-assigned too". But we didn't allow for that possibility in the previous Action. Maybe we need to have a quick offline discussion just to get 100% clarity around this?

File: src/components/postgresConnect.ts [hunk 1]
```diff
@@ -772,6 +776,24 @@ class PostgresDB {
     }
     return reviewAssignmentIds
   }
+
+  public isFullyAssignedLevel1 = async (applicationId: number) => {
```
andreievg, line 780: is this one actually used ? (I wonder if there is a use case for it in the futute)

src/plugins/action_update_review_assignments/src/databaseMethods.ts
```diff
@@ -0,0 +1,68 @@
...
+    }
+  },
+  updateReviewAssignments: async (reviewAssignments: any) => {
+    const reviewAssignmentUpdateResults = []
+    for (const reviewAssignment of reviewAssignments) {
+      const { id, status } = reviewAssignment
+      const text = `
+      UPDATE review_assignment
+      SET status = $2
+      WHERE id = $1
```
andreievg, line 49: maybe 'any' once again ?

src/plugins/action_update_review_assignments/src/updateReviewAssignmentsStatus.ts
```diff
@@ -0,0 +1,52 @@
...
+      stageNumber,
+      reviewLevel
+    )
+
+    const reviewAssignmentUpdates = await Promise.all(
+      otherReviewAssignments.map(async (reviewAssignment: any) => {
+        const { id, status } = reviewAssignment
+        return {
+          id,
+          status: trigger === 'onReviewSelfAssign' ? AssignmentStatus.SELF_ASSIGNED_OTHER : status,
```
andreievg, line 29: Is this the first time we are checking 'trigger' inside an action ? Does it mean this action is specific to that trigger ?

andreievg (approved):
A few comments, nothing that stops approval though

src/components/postgresConnect.ts
CarlosNZ: Yeah, it's no longer used but I just left it here as I thought it might come up again.

src/plugins/action_update_review_assignments/src/databaseMethods.ts
CarlosNZ: Actually, doesn't quite work as easily, since `reviewAssignments` is an array of objects with {id, status}, and each of the records to be updated gets a (possibly) different status, so we need to loop over them one by one.

src/plugins/action_update_review_assignments/src/updateReviewAssignmentsStatus.ts
CarlosNZ: > Is this the first time we are checking 'trigger' inside an action ?

Yes, I think it is, which is why I was hoping to avoid it, but there doesn't seem to be any other way to distinguish between an assignment and a self-assignment.


> Does it mean this action is specific to that trigger ?

Technically, no -- but in practice yes, since it wouldn't make sense to use it in other contexts.

PR #224, opened 2021-02-18
Deep comparison module

CarlosNZ:
I moved this utility function to its own package so it could be used in #219. This PR just has those changes done to move the function out of "utilityFunctions" into the 'modules' folder, and update any references to it to use the new package (@openmsupply/deep-comparison)


andreievg (commented):
sorry man, i like what' you've done with this, but I am resisting the urge to traverse the logic and think of all of the use cases of the code, when there is a package with (12 million weekly downloads) https://www.npmjs.com/package/deep-equal

CarlosNZ:
No problem. 👍

PR #227, opened 2021-02-26
add review status to application list

andreievg:
Fixes ? (not sure which issue it fixes

needs frontend branch: https://github.com/openmsupply/application-manager-web-app/pull/382

## Changes

Added new view "review_stats_list" which is used in application_list now. Have to filter application list by reviewer_id (even for applicant). See explanation in 36_revies_status_list.sql


database/buildSchema/36_review_status_list.sql
```diff
@@ -0,0 +1,27 @@
...
+CREATE OR REPLACE VIEW review_status_list AS
+SELECT
+    COUNT(*) FILTER (WHERE review_assignment.status = 'Available for self-assignment') AS number_of_self_assignable_reviews,
+    COUNT(*) FILTER (WHERE review_assignment.status = 'Assigned') AS number_of_assigned_reviews,
+    COUNT(*) FILTER (WHERE review_assignment.status = 'Assigned'
+        AND review.id IS NULL) AS number_of_assigned_not_started_reviews,
+    COUNT(*) FILTER (WHERE review_assignment.status = 'Self-assigned by another') AS number_of_reviews_self_assigned_by_someone_else,
+    COUNT(*) FILTER (WHERE review_status_history.status = 'Draft') AS number_of_draft_reviews,
+    COUNT(*) FILTER (WHERE review_status_history.status = 'Submitted') AS number_of_submitted_reviews,
+    COUNT(*) FILTER (WHERE review_status_history.status = 'Changes Requested') AS number_of_changes_requested_reviews,
```
nmadruga, line 14: Would be worth to have on already for `Pending` status?

database/buildSchema/37_application_list_view.sql
```diff
@@ -1,5 +1,9 @@
--- Aggregated VIEW table of all data required for application list page
 
+
+
+-- Aggregated VIEW table of all data required for application list page
+-- Now review_status_list is joined (see comment in declaration), it needs to be filtered by reviewer_id (current user_id)
+-- every time it's quired by front end to avoid duplicates and allow relevant review status information to be shown
```
nmadruga, line 6:
```suggestion
-- every time it's queried by front end to avoid duplicates and allow relevant review status information to be shown
```

database/buildSchema/37_application_list_view.sql
```diff
@@ -15,13 +19,15 @@ CREATE OR REPLACE VIEW "public".application_list AS
 	-- 	template_questions_count(app),
 	-- 	assigned_questions_count(app, stage_status.stage_id, 1),
 		assigned_questions_count(app, stage_status.stage_id, 1) >= template_questions_count(app)
-		AS is_fully_assigned_level_1
+		AS is_fully_assigned_level_1,
```
nmadruga, line 22: Not from this PR, sorry I just have investigated if this `is_fully_assigned` would return what is expected, but I don't this it is considering questions that aren't visible therefore not assigned in the review... I think I'll create a separated issue for that. Just pointing out here so we can discuss - maybe I'm just not seeing something

nmadruga (commented):
It was a great idea to add a view, simplifies so much the result and how it needs to be deduced by front-end (I'm going to have a look on the front-end issue now). But as a reminder of what @CarlosNZ was saying on today's meeting: what would be done for users that need to view other reviews/consolidator. Also if we consider the consolidator we might need to know the number of review **submitted**/**pending**/**changes required** from a level below.

database/buildSchema/37_application_list_view.sql
CarlosNZ: I'd have thought that when you assign a Section (or a whole application), they'd be assigned regardless of whether or not question is visible. In fact, back-end has no concept of "is_visible"  as it's only ever calculated on-the-fly in the front end (and changes as responses change). You assign reviews based on template_elements, don't you, rather than specific applications responses? If not, they probably should be. You don't want a situation where a reviewer is assigned a section, and then the applicant makes changes which makes another question visible but no one is assigned to it. Surely we assigned a selection of "template elements" which are tied to responses, not responses themselves?

database/buildSchema/37_application_list_view.sql
```diff
@@ -15,13 +19,15 @@ CREATE OR REPLACE VIEW "public".application_list AS
...
+		review_status_list.*
 	FROM application app LEFT JOIN
 	template ON app.template_id = template.id LEFT JOIN
 	"user" ON user_id = "user".id LEFT JOIN
 	application_stage_status_latest AS stage_status
 	on app.id = stage_status.application_id
-	LEFT JOIN organisation org ON app.org_id = org.id;
+	LEFT JOIN organisation org ON app.org_id = org.id
+	LEFT JOIN review_status_list on review_status_list.application_id = app.id
```
CarlosNZ, line 30: I wonder if this View should also return one set of applications that are not joined to a reviewer_id (i.e. all the review columns would be NULL). That way it would still be possible to query without a review_id (or reviewer_id = NULL) and get the full set of applications for a list that doesn't need to care about reviewer id.

Cos what happens to the application list for applicants. If they have a bunch of applications in progress, they'd want to be able to see them all without caring who the reviewer is, right?

You could maybe do a UNION of this query with the original query, perhaps?

CarlosNZ (commented):
Just made a couple of comments. Will check out the front-end now to see how it works together.

andreievg:
this needs to be re-thought, and I am glad we have time to do it (it's not immediately required).

I was also worried about the join being quite big (as per comments on front end pr).

I think a better way to do it would be to make a function, something like this:

(this is just a simple demo example, the second function will return everything application_list view does now)
```SQL
create function review_stuff(reviewerid int) 
  returns table(assigned bigint, available bigint, self_assigned bigint, application_id int) as $$
  	select 
  		count(*) filter (where status = 'Assigned') as assigned,
  		count(*) filter (where status = 'Available') as available,
 		count(*) filter (where status = 'Available for self-assignment') as self_assigned,
 		application_id
  	from review_assignment where reviewer_id = $1 group by application_id ;
$$ language sql stable;


create function application_list_check(reviewerid int)
returns table(application_id int, assigned bigint, available bigint, self_assigned bigint) as $$
  select id, rs.assigned, rs.available, rs.self_assigned from application left join review_stuff($1) as rs on application.id = rs.application_id 
$$ language sql stable;
```

then we can do this in graphQL (making sure to always specify reviewer_id and similar for assignment).


Screen Shot 2021-03-02 at 8 02 22 PM

andreievg:
Comments will be transferred to: https://github.com/openmsupply/application-manager-server/pull/266

PR #229, opened 2021-03-05
#378F Review_status_history to expose the time_created of status in review table

nmadruga:
To be used combined with Front-end branch [PR#411](https://github.com/openmsupply/application-manager-web-app/pull/411)

### DIFF
--- a/database/buildSchema/31_review_status_history.sql
+++ b/database/buildSchema/31_review_status_history.sql
@@ -35,3 +35,10 @@ RETURNS public.review_status AS $$
 	SELECT "status" FROM review_status_history 
 	WHERE review_id = app.id and is_current = true
 $$ LANGUAGE sql STABLE;
+
+-- Function to expose time_create field on review table in GraphQL
+CREATE FUNCTION public.review_time_created(app public.review)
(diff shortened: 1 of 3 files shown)


database/buildSchema/31_review_status_history.sql
```diff
@@ -35,3 +35,10 @@ RETURNS public.review_status AS $$
 	SELECT "status" FROM review_status_history 
 	WHERE review_id = app.id and is_current = true
 $$ LANGUAGE sql STABLE;
+
+-- Function to expose time_create field on review table in GraphQL
+CREATE FUNCTION public.review_time_created(app public.review)
+RETURNS timestamptz AS $$
+	SELECT time_created FROM review_status_history
```
andreievg, line 42: hmm maybe we should change it to time_updated ? (similar to what is being done with review_responses), but happy to merge as is (and make the changes in future PRs

andreievg (approved):

database/buildSchema/31_review_status_history.sql
andreievg: never mind, this is fine since review_status_history should just be created (and never changed)

Issue #184, opened 2021-01-19
Default Actions -- create GraphQL fragment?


CarlosNZ:
First step: Make a list (create a diagram) of all Actions that happen for every application type.

Second step: Write createAction mutations for the above as GraphQL fragment.

Third step: Add above fragment ton existing createTemplate mutations

---

### onApplicationCreate
- IncrementStage (also updates status)

### onApplicationSubmit
- Change Status (to Submitted)
- ReviewAssignmentGenerate (still to be worked out how it will work -- might be "global" rather than template-based)

### onReviewAssign
- TBD

### onReviewCreate
- Change (Review) Status

### onReviewSubmit
- Change (Review) Status
- ReviewAssignmentGenerate (TBD, as above)
- Check if approved and on Final review stage, if so ChangeOutcome, etc...

CarlosNZ:
Need to wait until OnReview triggers/actions have been implemented (and maybe Consolidate?)

andreievg:
actually, this would require a bit more thought.

* dont' think you can use fragments for mutations
* if you could we would need to concat the fragment definition to the end of each query (apollo client does this automatically)
* would be a bit harder if we added custom actions to know next sequence number (but I guess we can start them at like 101)

CarlosNZ:
>     * dont' think you can use fragments for mutations

You can -- currently have some in front-end, `createApplication.mutation.ts` for example.

>     * would be a bit harder if we added custom actions to know next sequence number (but I guess we can start them at like 101)

Yeah, exactly, just start at a high number for later ones, or make them asynchronous (no sequence number) where possible.

andreievg:
> You can -- currently have some in front-end, createApplication.mutation.ts for example.

ahh ok, i see what you mean (it's not quite a fragment though).

So we would define a new query for mutation (with template id). And then concat it to the end of each of the template generation queries (replacing template id with the new template id), right ?

Would still like to have core action template (that can be copied when creating new template on front end (otherwise would need to share this core mutation with front end code, or try do this mutation on back end when new template is generated)

CarlosNZ:
> Yeah, exactly, just start at a high number for later ones, or make them asynchronous (no sequence number) where possible.

Or, we can make the Fragment sequence values use negative numbers, so any additional Actions can start at 1 with no problems ;)

CarlosNZ:
Will do this quickly now. As discussed with @andreievg will do:
[...]

PR #234, opened 2021-03-08
#228 Update timestamps when trimming duplicate responses

CarlosNZ:
Implements #228.

Changed the field name on application_response and review_response from `time_created` to `time_updated`

Also updated tests. Can run just these tests (as opposed to whole test suite) by running `yarn test` from within the plugin folder.

Front-end PR: https://github.com/openmsupply/application-manager-web-app/pull/422
~~Will need some front-end adjustments for the new field name, will add that in front-end PR shortly.~~


src/plugins/action_trim_responses/src/databaseMethods.ts
```diff
@@ -64,6 +64,38 @@ const databaseMethods = (DBConnect: any) => ({
       throw err
     }
   },
+  updateApplicationResponseTimestamps: async (responsesToUpdate: number[]) => {
+    const text = `UPDATE application_response
+      SET time_updated = current_timestamp
```
andreievg, line 69: I think we discussed for this to be matched to latest application/review status history ?

andreievg (changes requested):
Ohh man, i've thought of some use cases where this would need to change (and would probably need different actions for application_response trimming and review_response trimming)

I think the logic should be

**Application Response** (only changes I think is the deletion of empty response)
* get application_responses for application
* order by timestamp
* group in for each (like you've done) by template_element_id
* compare two of the latest (if match remove the latest)
* if latest is empty response remove it <- i think this is new

**Review Response**
* get all review_response for review
* order by timestamp
* group by either application_response_id or review_response_id (whichever exists for review_response) <- this is new
* compare two of the latest (if match remove the latest)
* if latest is empty response remove  it <- i think this is new

src/plugins/action_trim_responses/src/databaseMethods.ts
CarlosNZ: Oh, I don't remember discussing that, sorry. That assumes the update status Action (to create a new status_history) would be ahead of the `trimResponses` Action. Is that a reasonable assumption? 
How about the changeStatus Action pass along the new status timestamp in its output, then we can make this Action use that timestamp if it exists, but write the current_timestamp if it doesn't?

andreievg: sure, or query for it in the update statement. up to you.

> That assumes the update status Action (to create a new status_history) would be ahead of the trimResponses

in the set `is visible` PR I'll make one empty template with 'core actions' (we are getting a few of them now, want to keep track of them)

CarlosNZ: Yeah, I've been trying to keep a list of the core/default actions in #184

CarlosNZ: Okay, changes to reflect the above: 65197f97

Haven't done extensive testing yet, but it passes the existing tests (which only deals with Level 1 reviews)

src/plugins/action_trim_responses/src/databaseMethods.ts
```diff
@@ -64,6 +66,38 @@ const databaseMethods = (DBConnect: any) => ({
...
+    } catch (err) {
+      console.log(err.message)
+      throw err
+    }
+  },
+  updateReviewResponseTimestamps: async (responsesToUpdate: number[], timestamp: string) => {
+    const text = `UPDATE review_response
+      SET time_updated = $1
+      WHERE id = ANY ($2)
+      RETURNING (SELECT code FROM
```
andreievg, line 88: hmm why do we need 'code' from application_response ? (maybe from template_element ?)

src/plugins/action_trim_responses/src/trimResponses.ts
```diff
@@ -1,63 +1,77 @@
...
+      ? 'review_response_link_id'
+      : 'application_response_id'
+
+    const responsesById = groupResponses(responses, groupByField)
     const responsesToDelete: number[] = []
+    const responsesToUpdate: number[] = []
 
-    Object.values(responsesByCode).forEach((responseArray) => {
+    Object.values(responsesById).forEach((responseArray) => {
       if (responseArray.length === 1) return
```
andreievg, line 47: can you add an 'undefined' review_response to the test and see if it's deleted please ?

use case -> reviewer starts review, marks one question as declined, all others should be deleted

src/plugins/action_trim_responses/src/trimResponses.ts
```diff
@@ -1,63 +1,77 @@
...
-      if (!(code in responsesByCode)) responsesByCode[code] = [response]
-      else responsesByCode[code].push(response)
-    })
+    const reviewLevel = responses?.[0]?.level
+
+    // Create object of indexed responses, with array of response
+    // objects for each indexed id
 
+    const groupByField = !reviewId
+      ? 'template_element_id'
```
andreievg, line 37: this is stylistic (so not required change, I've suggested this to Nicole a few times), but more nested in-line condition statement are much harder to read then just if statement.

database/insertData/03_template_D_basicReviewTesting.js
```diff
@@ -359,14 +359,30 @@ exports.queries = [
...
+                trigger: ON_REVIEW_SUBMIT
+                sequence: 2
+                parameterQueries: {
+                  reviewId: {
+                    operator: "objectProperties"
+                    children: ["applicationData.record_id"]
+                  }
+                  timestamp: {
+                    operator: "objectProperties"
+                    children: ["output.reviewStatusHistoryTimestamp"]
```
andreievg, line 384: ideally we would also do this for ON_APPLICATION_SUBMIT

andreievg (changes requested):
Look good, thanks Carl !

One remaining question (as per below). Would it delete undefined review_response ?

src/plugins/action_trim_responses/src/databaseMethods.ts
CarlosNZ: It's `code` from application_response JOIN template_element -- so it is the template_element code ;)

src/plugins/action_trim_responses/src/trimResponses.ts
CarlosNZ: Yeah, I agree in general, although when the expressions are short I quite like the way Prettier formats them to be quite logical. But yeah, definitely when the expressions are long -- like chunks of JSX! ;)

database/insertData/03_template_D_basicReviewTesting.js
CarlosNZ: ef4d5dc0

src/plugins/action_trim_responses/src/trimResponses.ts
CarlosNZ: Okay, that should be deleting NULL responses now: ef4d5dc0

andreievg: I think this will still not quite work if there is one review_response

andreievg (changes requested):

src/plugins/action_trim_responses/src/trimResponses.ts
CarlosNZ: Yeah good point. Think it should be correct now: d0aad32f

Just to clarify -- if a response has no duplicate (i.e. it's perId array will have one element), that element's timestamp should be updated too.

andreievg (approved):
thanks, look good!

PR #236, opened 2021-03-08
#231 add trigger to associate original response

### LINKED ISSUE #231: Original response to be associated to every review_response on creation
Part of epic: https://github.com/openmsupply/application-manager-web-app/issues/94

review_response to have a new field, original_reponse_id (which link to the first reivew_response in chain, the one that links to applicaiton_response).

a trigger would work nicely (in the create trigger, do we have id of newly available ? )

has linked review_response ? then use original_response_id from that one
doesn't have linked review response ? then set original_response_id = new.id

See diagram `LOQ proposal`, under `Review Response Relation`

andreievg:
Fixes #231 

new field and trigger

### Test

"select * from review_response", you should see original_review_response is populated

### DIFF
--- a/database/buildSchema/29_review_response.sql
+++ b/database/buildSchema/29_review_response.sql
@@ -11,7 +11,28 @@ CREATE TABLE public.review_response (
 	review_question_assignment_id integer references public.review_question_assignment(id),
 	application_response_id integer references public.application_response(id),
 	review_response_link_id integer references public.review_response(id),
+	original_response_id integer references public.review_response(id),
 	review_id integer references public.review(id),
     time_created timestamptz default current_timestamp,
 	status public.review_response_status default 'Draft'
 );
+
+-- set review response original_response_id (the response that links to application id should be available for all reaponses)
+
+CREATE OR REPLACE FUNCTION set_original_response()
+RETURNS trigger AS $$
+BEGIN
+	IF NEW.application_response_id IS NOT NULL THEN
+	  NEW.original_response_id = NEW.id;
+	END IF;
+	IF NEW.review_response_link_id IS NOT NULL THEN
+	   NEW.original_response_id = (SELECT original_response_id FROM review_response where id = NEW.review_response_link_id);
+	END IF;
+  RETURN NEW;
+END;
+$$ LANGUAGE plpgsql;
+
+
+CREATE TRIGGER set_original_response_trigger BEFORE INSERT ON public.review_response
+FOR EACH ROW
+EXECUTE FUNCTION set_original_response()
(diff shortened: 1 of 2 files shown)


database/buildSchema/29_review_response.sql
```diff
@@ -11,7 +11,28 @@ CREATE TABLE public.review_response (
...
 	review_response_link_id integer references public.review_response(id),
+	original_response_id integer references public.review_response(id),
 	review_id integer references public.review(id),
     time_created timestamptz default current_timestamp,
 	status public.review_response_status default 'Draft'
-);
\ No newline at end of file
+);
+
+-- set review response original_response_id (the response that links to application id should be available for all reaponses)
```
CarlosNZ, line 20:
```suggestion
-- set review response original_response_id (the response that links to application id should be available for all responses)
```

database/buildSchema/29_review_response.sql
```diff
@@ -11,7 +11,28 @@ CREATE TABLE public.review_response (
...
+
+CREATE OR REPLACE FUNCTION set_original_response()
+RETURNS trigger AS $$
+BEGIN
+	IF NEW.application_response_id IS NOT NULL THEN
+	  NEW.original_response_id = NEW.id;
+	END IF;
+	IF NEW.review_response_link_id IS NOT NULL THEN
+	   NEW.original_response_id = (SELECT original_response_id FROM review_response where id = NEW.review_response_link_id);
+	END IF;
```
CarlosNZ, line 30: It shouldn't be possible for application_response_id to be NULL, so maybe it's easier to read/understand like:
```
IF NEW.review_response_link_id IS NOT NULL THEN
	   NEW.original_response_id = (SELECT original_response_id FROM review_response where id = NEW.review_response_link_id)
ELSE
          NEW.original_response_id = NEW.id
END IF;
```
Apologies if syntax not exactly right, didn't test it, just a possible suggestion.

CarlosNZ (commented):
One other suggestion: do you think the field should perhaps be called `original_review_response_id`? It's a bit more verbose, but I think it's less ambiguous. Up to you.

database/buildSchema/29_review_response.sql
andreievg: i think we dont' set application_response_id on lvl above 1

andreievg:
> One other suggestion: do you think the field should perhaps be called original_review_response_id

sounds good, i'll change

database/buildSchema/29_review_response.sql
andreievg: agree to both:
* application_response_id to be on all review_responses 

check out the next PR up where template_element_id is added, we can change logic there if that's ok

CarlosNZ (approved):

database/buildSchema/29_review_response.sql
CarlosNZ: Cool.

andreievg:
@CarlosNZ thanks for approving, i've set it here:

https://github.com/openmsupply/application-manager-server/pull/237#issuecomment-793378453

PR #237, opened 2021-03-08
#421fe is visible to applicant logic

andreievg:
branched from #231

add schema changes and add template_element_id to review_response (at all levels), noticed that this would be very useful when building structure etc.. (and hook useFullReviewStructure currently assumes it's there)

### DIFF
--- a/database/buildSchema/29_review_response.sql
+++ b/database/buildSchema/29_review_response.sql
@@ -1,8 +1,10 @@
 -- review response
 
-CREATE TYPE public.review_response_decision as ENUM ('Approve', 'Decline');
+CREATE TYPE public.review_response_decision as ENUM ('Approve', 'Decline', 'Agree', 'Disagree');
 
 CREATE TYPE public.review_response_status as ENUM ('Draft', 'Submitted');
+
+CREATE TYPE public.review_response_recommended_applicant_visibility as ENUM ('Original Response Visible to Applicant', 'Original Response Not Visible to Applicant');
 
 CREATE TABLE public.review_response (
 	id serial primary key,
@@ -14,6 +16,8 @@ CREATE TABLE public.review_response (
 	original_response_id integer references public.review_response(id),
 	review_id integer references public.review(id),
     time_created timestamptz default current_timestamp,
+	is_visible_to_applicant boolean default false,
+	recommended_applicant_visibility public.review_response_recommended_applicant_visibility default 'Original Response Not Visible to Applicant',
 	status public.review_response_status default 'Draft'
 );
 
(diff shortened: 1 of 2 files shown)


database/buildSchema/29_review_response.sql
```diff
@@ -1,9 +1,11 @@
 -- review response
 
-CREATE TYPE public.review_response_decision as ENUM ('Approve', 'Decline');
+CREATE TYPE public.review_response_decision as ENUM ('Approve', 'Decline', 'Agree', 'Disagree');
 
 CREATE TYPE public.review_response_status as ENUM ('Draft', 'Submitted');
 
+CREATE TYPE public.review_response_recommended_applicant_visibility as ENUM ('Original Response Visible to Applicant', 'Original Response Not Visible to Applicant');
```
CarlosNZ, line 7: Why not just `true/false`? There's a lot of overlap in the wording of the field name and the values, so we can probably remove some redundancy?

database/buildSchema/29_review_response.sql
```diff
@@ -14,25 +16,32 @@ CREATE TABLE public.review_response (
 	original_review_response_id integer references public.review_response(id),
 	review_id integer references public.review(id),
     time_created timestamptz default current_timestamp,
+	is_visible_to_applicant boolean default false,
+	template_element_id integer references public.template_element,
+	recommended_applicant_visibility public.review_response_recommended_applicant_visibility default 'Original Response Not Visible to Applicant',
 	status public.review_response_status default 'Draft'
```
CarlosNZ, line 22: If the field was named 'is_recommended_visible_to_applicant` then `true`/`false` would be okay, wouldn't it? (sorry, booleans are just way easier to work with in the front-end, but no worries if you'd prefer this)

CarlosNZ (commented):
Cool, just minor comment. 

And have you tested these with data for reviews higher than level 1?

andreievg:
@CarlosNZ as per our discussion, would prefer keeping recommended field naming.

also: 610741926686e0926232908115ddb2808ce4c457, to deal with previous PR comments: https://github.com/openmsupply/application-manager-server/pull/236#discussion_r589928575

database/buildSchema/29_review_response.sql
```diff
@@ -22,22 +22,21 @@ CREATE TABLE public.review_response (
...
 );
 
--- set review response original_response_id (the response that links to application id should be available for all reaponses)
--- also flatten out review response chain by providing template_element_id in review_response
+-- set review response original_review_response_id (the response that links to application id should be available for all reaponses)
```
CarlosNZ, line 25: You already changed this on the earlier PR, so maybe merge that in here now?

andreievg: i think it's in there now (if you look at full diff)

CarlosNZ (approved):

Issue #232, opened 2021-03-08
is visible to applicant logic on back end


andreievg:
Part of epic https://github.com/openmsupply/application-manager-web-app/issues/94

as per diagram LOQ proposal diagram.

needs this before start:  https://github.com/openmsupply/application-manager-web-app/issues/421

Action:
<img width="827" alt="Screen Shot 2021-03-08 at 1 06 26 PM" src="https://user-images.githubusercontent.com/26194949/110259961-1e879980-800f-11eb-864b-ea771b17de97.png">

PR #238, opened 2021-03-09
#232 is visible to applicant logic on back end [Action]

andreievg:
Branched from #421FR-is-visislbe-to-applicant-logic
Fixes: #232

### Changes

**review_decision schema**

* changed form `review_decision` to `decision` in as per upstream
* add latest decision helper getter

**insertData.js**

* now throws error on graphql error

** template insertion **

* added updateReviewVisibility to basicReviewTesting
* added coreAction template

** DB helpers and trigger-actions **

* added reviewData to applicationData (accessible for action condition/parameters

** added updateReviewVisibility action **

### Testing

apart from running jest test, try the following mutation (see appropriate console log for server, with different permuatation of `review_decision_decision`, `is_last_level` property on review, different review_responses having `ORIGINAL_RESPONSE_VISIBLE_TO_APPLICANT` toggled

```
mutation MyMutation {
  __typename
  updateReview(
    input: {
      patch: {
        trigger: ON_REVIEW_SUBMIT
        reviewDecisionsUsingId: { create: { decision: LIST_OF_QUESTIONS } }
        reviewResponsesUsingId: {
          updateById: [
            {
              patch: {
                recommendedApplicantVisibility: ORIGINAL_RESPONSE_VISIBLE_TO_APPLICANT
              }
              id: 38
            }
            {
              patch: {
                recommendedApplicantVisibility: ORIGINAL_RESPONSE_VISIBLE_TO_APPLICANT
              }
              id: 37
            }
          ]
        }
      }
      id: 6
    }
  ) {
    clientMutationId
  }
}

```


src/components/graphQLConnect.ts
```diff
@@ -28,6 +28,27 @@ class GraphQLdb {
     return data.data
   }
 
+  public getReviewData = async (reviewId: number) => {
```
CarlosNZ, line 31: Finally, we use a GraphQL query in the back-end! I knew the DBConnect layer would come in handy eventually ;)

database/insertData/03_template_D_basicReviewTesting.js
```diff
@@ -369,6 +369,38 @@ exports.queries = [
...
+                condition: {
+                  operator: "AND"
+                  children: [
+                    {
+                      operator: "="
+                      children: [
+                        {
+                          operator: "objectProperties"
+                          children: [
+                            "applicationData.reviewData.latestDecision.decision"
```
CarlosNZ, line 385: This throws an error ("Can't resolve object") when latestDecision is NULL (as it's trying to find a property that doesn't exist), which it will be if review is Level 1. I've made a new issue (#240) to make the evaluator return Null in cases like this, which will probably help. But is there anything you can do differently in the meantime to avoid this error?

src/components/graphQLConnect.ts
```diff
@@ -28,6 +28,27 @@ class GraphQLdb {
...
+    const data = await this.gqlQuery(
+      `
+      query getReviewData($id: Int!) {
+        review(id: $id) {
+          level
+          isLastLevel
+          status
+          trigger
```
CarlosNZ, line 39: Don't really need the trigger here do we? (It should always be Processing when an Action is running)

src/components/graphQLConnect.ts
```diff
@@ -28,6 +28,27 @@ class GraphQLdb {
...
   }
 
+  public getReviewData = async (reviewId: number) => {
+    const data = await this.gqlQuery(
+      `
+      query getReviewData($id: Int!) {
+        review(id: $id) {
+          level
+          isLastLevel
+          status
```
CarlosNZ, line 38: Hmm, the only problem with using a GraphQL query here rather than SQL is that the status comes through as "CHANGES_REQUESTED", rather than "Changes Requested". I guess it's fine, we just need to be careful with this kind of thing -- they're not completely interchangeable and might run into a problem when updating Review status later on.

database/insertData/03_template_E_coreActions.js
```diff
@@ -0,0 +1,211 @@
+/*
+TEMPLATE E - Core Actions
+  - empty template that lists all of the core actions
```
CarlosNZ, line 3: Is this just intended as a reference for now?
We had the idea to save all the core actions in a GraphQL fragment (#184) and then every template could just slug this Fragment into it's mutation. It's probably a good idea to do that now, right, or at least soon after this issue -- then we can remove the need for this file.

CarlosNZ (changes requested):
Be good to get this one merged soon as it's got some useful stuff for other issues. Most of my comments just questions or observations.

database/insertData/03_template_D_basicReviewTesting.js
CarlosNZ: I'm about to make this small fix on #240. Do you think handling unresolved objects should be an optional parameter, or should it always return `null`?

src/plugins/action_review_visibility/src/plugin.json
```diff
@@ -0,0 +1,25 @@
+{
+  "type": "action_plugin",
+  "code": "updateReviewVisibility",
+  "name": "Update Applicant's Review Visibility",
+  "file": "updateReviewVisibility",
+  "function_name": "updateReviewVisibility",
+  "description": "Updates review records to be visible to Applicant",
```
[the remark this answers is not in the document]
andreievg: 815e4eafac87a622920e6844f40fdc3eb444b57a

src/plugins/action_review_visibility/src/databaseMethods.ts
```diff
@@ -0,0 +1,24 @@
+const databaseMethods = (DBConnect: any) => ({
+  updateReviewResponseVisibility: async (reviewId: number) => {
+    const text = `
+    UPDATE review_response 
+    SET is_visible_to_applicant = true 
+    WHERE id IN (
+      SELECT original_response.id
```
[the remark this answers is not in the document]
andreievg: oops, thanks for noticing: 0aa6590442df6a5fbd054c227e574394e5c0cdd9

src/components/databaseConnect.ts
```diff
@@ -121,6 +121,8 @@ class DBConnect {
 
   public isFullyAssignedLevel1 = PostgresDB.isFullyAssignedLevel1
 
+  public getReviewData = GraphQLdb.getReviewData
```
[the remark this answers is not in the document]
andreievg: 79a079b5677d8a55129c651a29ce6c68172389b7

src/components/graphQLConnect.ts
CarlosNZ: Yeah, neither was I. But I think the reasoning was that you'd be able to swap an SQL query for a GraphQL one without having to touch any of the Actions that call it.

src/components/graphQLConnect.ts
andreievg: yeah. nicely spotted, as per convo, lets replace postgresql ones with capital and underscore

src/components/graphQLConnect.ts
andreievg: 9c984f37500ebb53be2fce20bd71a82a58205398

database/insertData/03_template_E_coreActions.js
andreievg: ok good point, i'll do that now

src/components/graphQLConnect.ts
CarlosNZ: #241

database/insertData/03_template_E_coreActions.js
andreievg: actually, this would require a bit more thought.

* dont' think you can use fragments for mutations
* if you could we would need to concat the fragment definition to the end of each query (apollo client does this automatically)
* would be a bit harder if we added custom actions to know next sequence number (but I guess we can start them at like 101)

andreievg: discussion to continue on that issue if it's ok

CarlosNZ (approved):
👍 Thanks.

andreievg:
>👍 Thanks.

Thanks for review!

Issue #240, opened 2021-03-09
Evaluator: Return Null if "Can't resolve object" error


CarlosNZ:
It currently throws the error, but it'd be good if it would gracefully return nothing when it can't find the object reference in question.

(And update tests to demonstrate it working)

PR #242, opened 2021-03-10
#240 Evaluator -- handle unresolved objects

CarlosNZ:
Small improvement so that unresolved objects return a fallback value rather than throwing an error. Default fallback is "Can't resolve object", but can be useful to set to `null` in some case if you're not sure that the property will exist (Like for an API call, say).

Since GraphQL and API operators also use the same `extractProperty` method, this will work for those operators too -- I've added a specific fallback for them, mainly for debugging purposesd.

Also updated documentation, and added a couple of additional tests.

Bear in mind that whatever is returned by the node will be passed up back up the tree, so will be treated exactly like it would in normal javascript (e.g. "string" + null => "stringnull"). If you need to test if a property exists before returning something, you can use the `?` ternary operator. Use the expression-evaluator-gui if you want to see what happens with various expressions.

### DIFF
--- a/documentation/Query-Syntax.md
+++ b/documentation/Query-Syntax.md
@@ -143,7 +143,11 @@ For extracting values from local state objects (e.g. `user`, `organisation` or `
 
 The `evaluateExpression` function expects each expression to be passed along with _all_ objects that are required for evaluation in any descendant nodes. See **Usage** below for detailed overview of arguments.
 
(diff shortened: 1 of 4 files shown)


CarlosNZ:
Note: once approved I'll re-build the package, update the version in package.json and push bumped version straight to master to save another round of PR just for that.

andreievg (approved):
Looks good !

PR #243, opened 2021-03-10
#184 Core Actions GraphQL fragment

CarlosNZ:
Implements #184 as per [this comment](https://github.com/openmsupply/application-manager-server/issues/184#issuecomment-794823010)

From now one, common Actions can be added to/edited in `core_actions.js` and will be applied to all templates (except User Registration)


andreievg (approved):
👍

PR #244, opened 2021-03-10
add deep equal

andreievg:
otherwise fails to load trimResponses plugin

### DIFF
--- a/package.json
+++ b/package.json
@@ -51,6 +51,7 @@
     "@types/jsonwebtoken": "^8.5.0",
     "bcrypt": "^5.0.0",
     "concurrently": "^5.3.0",
+    "deep-equal": "^2.0.5",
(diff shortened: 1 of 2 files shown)


CarlosNZ:
FYI -- this is currently added to the package.json of the plugin folder. So once you run `yarn install` from within the plugin folder, it'll work.

At some point, I need to replace the custom deepEquality function used in "registerPlugins.ts" with this same npm package, at which point it won't be necessary to install it locally for the plugin.

So this issue is basically:
- remove "deepEquality" function from utilityFunctions
- install [deep-equal package](https://www.npmjs.com/package/deep-equal]) for the overall repo
- replace deepEquality in "registerPlugins" with the installed package

andreievg:
ok i am closing this, I'll add a comment on telegram

Issue #235, opened 2021-03-08
action to change review to appropriate status when application is re-submitted


andreievg:
Part of epic: https://github.com/openmsupply/application-manager-web-app/issues/94

as per `Review Life Cycle 28-01-21` diagram

![Screen Shot 2021-03-08 at 2 48 51 PM](https://user-images.githubusercontent.com/26194949/110264144-74633e00-801d-11eb-8673-db5a0c7f720c.png)

On application submitted, after the review_assignment action:

* check all review_assignments for application (lvl 1).
* if status is not submitted, change to `pending`
* if status is submitted, check for new changes (i.e. template elements, that are assigned, have changed), if changes change status to 'pending'

PR #247, opened 2021-03-11
#235 Action - change review statuses on Application submit

CarlosNZ:
Implements #235, following the logic shown in the diagram on that issue, and from discussion with @andreievg.

There is a single test case that demonstrates it in action, but we need to get further along with front-end to really put it through its paces with all conditions. Also, if you duplicate the responses for Application ID 4000, then change a couple, then artificially trigger `onApplicationSubmit` for that application, you should see the result of the sequence of core actions, the last of which is this one.

Note about tests: If you run individual tests on a fresh database init, they should all pass. However, you just run `yarn test` from the root, unfortunately the `trimResponses` one fails as it runs _after_ the `updateReviews` one, and the IDs have changed in the database. Need to sort this at some point, but for now just run tests on individual plugins.

Also:
 - updated "coreActions" to include this in the application submit sequence
 - made some tweaks to `trimResponses` to output values in a format useful to this Action
 - made tweaks to changeStatus to also output the new status timestamp


src/plugins/action_update_review_status/src/databaseMethods.ts
```diff
@@ -0,0 +1,124 @@
...
+    `
+    try {
+      const result = await DBConnect.query({ text, values: [reviewAssignmentId] })
+      return result.rows.map(({ templateElementId }: any) => templateElementId)
+    } catch (err) {
+      console.log(err.message)
+      throw err
+    }
+  },
+  getAllReviewResponses: async (reviewId: number) => {
```
andreievg, line 40: i think from here downwards, duplicates form postresConnect

src/plugins/action_update_review_status/src/updateReviews.ts
```diff
@@ -0,0 +1,70 @@
...
+  const { applicationId, changedApplicationResponses = [] } = input
+
+  const reviewsToUpdate = []
+
+  console.log('changedApplicationResponses', changedApplicationResponses)
+
+  try {
+    // Get reviews/review assignments (with status) matching application_id
+    const reviews = (await db.getAssociatedReviews(applicationId)).filter((review: Review) =>
+      ['Submitted', 'Locked', 'Draft'].includes(review.reviewStatus)
```
andreievg, line 27: kinda like the idea of filtering with postgresql, but this looks fine (don't need to change).

src/plugins/action_update_review_status/src/updateReviews.ts
```diff
@@ -0,0 +1,70 @@
...
+    )
+    // Deduce which ones should be updated
+    for (const review of reviews) {
+      const { reviewAssignmentId, level, reviewStatus } = review
+      if (level > 1) reviewsToUpdate.push({ ...review, reviewStatus: 'Pending' })
+      else if (await haveAssignedResponsesChanged(reviewAssignmentId))
+        reviewsToUpdate.push({ ...review, reviewStatus: 'Pending' })
+      else if (reviewStatus === 'Pending' || reviewStatus === 'Locked')
+        reviewsToUpdate.push({ ...review })
```
andreievg, line 36: should this have reviewStatus ?

also reviewStatus === 'Pending' would be filtered out (so not sure if you want to have it here)

src/plugins/action_update_review_status/src/databaseMethods.ts
```diff
@@ -0,0 +1,124 @@
...
+    `
+    try {
+      const result = await DBConnect.query({ text, values: [applicationId] })
+      return result.rows
+    } catch (err) {
+      console.log(err.message)
+      throw err
+    }
+  },
+  getReviewQuestionAssignments: async (reviewAssignmentId: number) => {
```
andreievg, line 26: getReviewAssignedElementIds ?

src/plugins/action_update_review_status/src/updateReviews.ts
```diff
@@ -0,0 +1,70 @@
...
+    console.log(error.message)
+    return {
+      status: 'Fail',
+      error_log: 'There was a problem updating review statuses.',
+    }
+  }
+
+  async function haveAssignedResponsesChanged(reviewAssignmentId: number) {
+    const questionAssignments = await db.getReviewQuestionAssignments(reviewAssignmentId)
+    return changedApplicationResponses.reduce(
```
andreievg, line 63: some would be better here i think (jut match the first one).

another place where I think postgresql would be useful (this is solid thought dont' need to change).

select count(*) from review_assignment where template_element_id = any ($1) and assignment_id = ($2)

andreievg (changes requested):
looks pretty solid thanks. Will do another round of review and testing.

main change (not commented below), we want to make sure we also match the **stage**

CarlosNZ:
> main change (not commented below), we want to make sure we also match the **stage**

Okay, yes, of course. That should be done now bfdd956 and 6388e34.

I also update the triggers and actions processing to always pass the fully "applicationData" object as part of the input (as we discussed the other day), so you don't have to explicitly specify an expression for "stageId" in the action definition in the template. This will simplify the input requirements for other actions too, although will do those incrementally as required.

src/plugins/action_update_review_status/src/databaseMethods.ts
CarlosNZ: Ah yes, I duplicated the previous Action's databaseMethods and modified -- guess I left a few behind 😊
764de8e

src/plugins/action_update_review_status/src/updateReviews.ts
CarlosNZ: Yeah, I think I was thinking that the database method might be more re-usable if if didn't have additional filtering -- but it's associated with this Action only, so doesn't really matter.

src/plugins/action_update_review_status/src/updateReviews.ts
CarlosNZ: > should this have reviewStatus ?

Yeah, if we want "Locked" ones to now become "Pending". Have updated now 0bf1dc0.

> also reviewStatus === 'Pending' would be filtered out (so not sure if you want to have it here)

Yeah, that's true. The thought was that "Pending" reviews should have their status_history timestamp updated anyway, but I guess that doesn't need to happen does it? 

Have now removed, already "Pending" review will be untouched. 37c8b3e

src/plugins/action_update_review_status/src/databaseMethods.ts
CarlosNZ: Yeah sure. b4c6f8c

src/plugins/action_update_review_status/src/updateReviews.ts
CarlosNZ: > some would be better here i think (jut match the first one).

You saying that `.some()` would stop checking as soon as it matches the first inclusion? Good to know. I was planning on using a traditional `for.. of` loop and using `break` as soon as it gets a match, but I figured the list of items would be so small that it doesn't really matter.

andreievg (changes requested):
looks pretty good.

Sorry but there is one more change (pass through level as input, for onApplicationSubmit it would be 0, for onReviewSubmit it would be current review level. Use level + 1 in `getAssociatedReviews`

The diagram was drawn to show all the logic (as if it was one action), and there was a bit there that level is passed through that is easy to miss.

i.e. on Application submit only update level 1, on review submit, only update level + 1

CarlosNZ:
> Sorry but there is one more change (pass through level as input, for onApplicationSubmit it would be 0, for onReviewSubmit it would be current review level. Use level + 1 in `getAssociatedReviews`
> 
> The diagram was drawn to show all the logic (as if it was one action), and there was a bit there that level is passed through that is easy to miss.
> 
> i.e. on Application submit only update level 1, on review submit, only update level + 1

Right, sorry. Should be ok in b4c6f8c.

I'm not 100% sure this action is working completely correctly for Review Submissions, as I don't really have enough data to test with. But I've made sure the review level is being taken into account and it still works with the "onApplicationSubmit" example in the tests file. I've assigned level thusly:
```
const level = input?.level || applicationData?.reviewData?.level || 1
```
So you can pass it in as parameter, but most likely will fall back to second expression and use existing review level. If none exists, it means we're dealing with "ApplicationSubmit", therefore fall back to level 1. Sound about right?

database/insertData/core_actions.js
```diff
@@ -115,9 +130,24 @@ exports.coreActions = `
...
     }
     {
-        actionCode: "generateReviewAssignments"
+        actionCode: "updateReviews"
         trigger: ON_REVIEW_SUBMIT
         sequence: 3
+        parameterQueries: {
+          applicationId: {
+            operator: "objectProperties"
+            children: ["applicationData.applicationId:"]
```
andreievg, line 139: i think you want to remove `:`

src/plugins/action_update_review_status/src/updateReviews.ts
```diff
@@ -0,0 +1,72 @@
...
+}
+
+module.exports['updateReviews'] = async function (input: any, DBConnect: any) {
+  const db = databaseMethods(DBConnect)
+
+  console.log('Updating reviews status...')
+
+  const { applicationId, applicationData, changedApplicationResponses = [] } = input
+  const stageId = input?.stageId || applicationData.stageId
+  const level = input?.level || applicationData?.reviewData?.level || 1
```
andreievg, line 20: hmm, is there a way to do additions in evaluator (as the level to check is + 1).

Still requires a little tweak (might have been better done in another issue, so we can merge this one. Maybe we shouldn't rely at all on the `input`, since it's hard to name the level input properly (since it's + 1 of the related review).

something like:
```
const reviewLevel = applicationData?.reviewData?.level || 0
const level = reviewLevel + 1
```

andreievg (changes requested):

database/insertData/core_actions.js
CarlosNZ: Oops 😬 1cbddb7

src/plugins/action_update_review_status/src/updateReviews.ts
CarlosNZ: Yeah, we can ignore the input, sure, I'll just remove it from this expression.

I'm slightly confused about the other bit -- if an application is submitted, we want to update all the level 1 reviews, correct? And if a level 1 review is submitted, we also want to update the other level 1 reviews, don't we? And if a level 2 review is submitted, we want to update the other L2 reviews? Or is it update the current level reviews AND the one above?

andreievg: >  And if a level 1 review is submitted, we also want to update the other level 1 reviews, don't we?
 
 In this action, we just want to update lvl + 1 (or lvl 1 review) <- this action should really be about notifying reviewers that something below their level has happened that needs their attention.

I would imagine there would be another action that would set status of reviews to locked ?

andreievg: I am going to approve this for now, and we can change it when we get to consolidation level

src/plugins/action_update_review_status/src/updateReviews.ts
andreievg: yeah, I find it much easier to see then the reduce logic (frankly I find reduce harder to follow then forEach)

andreievg (approved):

src/plugins/action_update_review_status/src/updateReviews.ts
CarlosNZ: > I am going to approve this for now, and we can change it when we get to consolidation level

Yeah cool, I think it will be easier to make small tweaks to it once we're working with real applications in the UI. 

Cheers.

Issue #246, opened 2021-03-11
Extend /upload and /file endpoints


CarlosNZ:
From File upload management [Epic #41](https://github.com/openmsupply/application-manager-web-app/issues/430)

See implementation spec: https://github.com/openmsupply/application-manager-web-app/issues/301

Additional thoughts:
- For files that are not images or docs, we should return a generic thumbnail image.

Issue #245, opened 2021-03-11
File management - schema updates


CarlosNZ:
From File upload managment [Epic #41](https://github.com/openmsupply/application-manager-web-app/issues/430)

See spec: https://github.com/openmsupply/application-manager-web-app/issues/301

PR #250, opened 2021-03-14
#246 #245 Extend file upload/download endpoints

CarlosNZ:
Implements #245 and #246 as per [design spec](https://github.com/openmsupply/application-manager-web-app/issues/301) (in hindsight, #245 was too trivial to warrant its own issue).

Requires fresh `database_init` plus a `yarn install` to grab the thumbnail and unique_id generation packages.

To test, here's a Postman collection configured for file upload endpoint: [IRIMS file endpoints.postman_collection.json.zip](https://github.com/openmsupply/application-manager-server/files/6137972/IRIMS.file.endpoints.postman_collection.json.zip)

Go the "Body" tab, and click "Select files" to upload one or more files. After successfully sending, you should get response like:
```
{
    "success": true,
    "fileData": [
        {
            "filename": "IMG_2605.jpeg",
            "fileUrl": "/file?uid=h2HgjWuBDMFsGbEJconoQ",
            "thumbnailUrl": "/file?uid=h2HgjWuBDMFsGbEJconoQ&thumbnail=true",
            "mimetype": "image/jpeg"
        }
    ]
}
```

You should then be able to click directly on the `fileUrl` or the `thumbnailUrl`, and they'll open in a new "GET" tab, which will then fetch the file.

### Some comments

- I've used the [sharp](https://www.npmjs.com/package/sharp) package for processing image files to thumbnails. I have NOT been able to find a decent package to turn PDFs into image thumbnails in the back. Suggestions welcome, but I've looked at most of the obvious options. The packages I tried were all quite out of date, required additional (non-package) software to be installed (i.e. graphicsmagick or imagemagick), and/or had very inadequate results. We might want to just to thumbnails for PDFs in the front-end (using the ReactPDF package Andrei was looking at), but I thought I should talk to you guys about ti first. (Let alone how to thumbnail Word docs or other common formats)
- I've added some generic thumbnail images for document/image/pdfs as fallbacks. These are just from https://thenounproject.com/ and we won't be able to use them without attribution, so we'll need to look at deciding on what we use for production. These should do for the demo though, I'd say. (We can probably use FontAwesome icons, perhaps?)
- ~~Uploaded files are currently all going into the root "files" folder. I thought perhaps they should be put into subfolders named with the Application Serial~~ -- any thoughts on folder structure? Edit: They're now going to a subfolder from application_serial, or just to the root if no application_serial provided -- we can discuss if this should be different
- Do you think the `/file` endpoint (download) should return files renamed to their original filename? I think they should, but I couldn't see a simple way to do this. There doesn't seem to be a way for the `sendFile` method to change the served filename, so would have to do something kinda clunky like: make a copy of the file with the original name, then serve that, then delete the copy after download. Any better suggestions?
- the `createThumbail` module is a bit messy because it's using some combination of mime type, subtype AND the file extension  trying to decide the best generic thumbnail to serve. This can be cleaned up somewhat once we know exactly what generic thumbnails we want to use for which file types.


andreievg:
> I've used the sharp package for processing image files to thumbnails. I have NOT been able to find a decent package to turn PDFs into image thumbnails in the back. Suggestions welcome, but I've looked at most of the obvious options. The packages I tried were all quite out of date, required additional (non-package) software to be installed (i.e. graphicsmagick or imagemagick), and/or had very inadequate results. We might want to just to thumbnails for PDFs in the front-end (using the ReactPDF package Andrei was looking at), but I thought I should talk to you guys about ti first. (Let alone how to thumbnail Word docs or other common formats)

I think it's fine to just show the default thumbnail for pdfs

> I've added some generic thumbnail images for document/image/pdfs as fallbacks. These are just from https://thenounproject.com/ and we won't be able to use them without attribution, so we'll need to look at deciding on what we use for production. These should do for the demo though, I'd say. (We can probably use FontAwesome icons, perhaps?)

They look pretty good, we can buy them if we choose to use them (should be fine without attribution for dev though)

> Do you think the /file endpoint (download) should return files renamed to their original filename? I think they should, but I couldn't see a simple way to do this. There doesn't seem to be a way for the sendFile method to change the served filename, so would have to do something kinda clunky like: make a copy of the file with the original name, then serve that, then delete the copy after download. Any better suggestions?

Hmm, i think original name would be good to return at some point, but with some extra bits added (maybe application serial -> elementCode, and revision number ? (but not sure how this would work yet, considering multi file upload etc..). We can make an issue fo this to be done in the future (if we have time)

src/components/files/fileHandler.ts
```diff
@@ -0,0 +1,119 @@
...
+}
+
+export function createFilesFolder() {
+  try {
+    fs.mkdirSync(path.join(getAppRootDir(), filesFolderName))
+  } catch {
+    // Folder already exists
+  }
+  // Move generic thumbnails to files root
+  fs.readdir(path.join(getAppRootDir(), 'images', 'generic_file_thumbnails'), (_, files) => {
```
andreievg, line 26: we can just create that folder on server (can add git ignore inside those folders, to git ignore anything that's added and to make sure folders structure is committed)

src/components/files/createThumbnails.ts
```diff
@@ -0,0 +1,77 @@
...
+const createThumbnail = async ({
+  filesPath,
+  unique_id,
+  basename,
+  ext,
+  mimetype,
+}: ThumbnailInput) => {
+  const { type, subtype } = splitMimetype(mimetype)
+
+  const origFilePath = path.join(filesPath, `${basename}_${unique_id}${ext}`)
```
andreievg, line 39: I am not sure you've tested this on the latest branch, this should include serial

andreievg, line 39: I think you might want to retest this

andreievg (changes requested):
A few comments, reviewed up to saveFiles

src/components/files/fileHandler.ts
```diff
@@ -0,0 +1,119 @@
...
+  if (!result) throw new Error()
+  return result
+}
+
+const pump = util.promisify(pipeline)
+
+export async function saveFiles(data: any, queryParams: HttpQueryParameters) {
+  const filesInfo = []
+  try {
+    for await (const file of data) {
```
andreievg, line 48: I am not entirely sure why await is necessary here, is it to await fo something inside for loop to finish (i think that would automatically await), is it to await for stream to be opened ? What happens if it's removed ?

src/components/files/fileHandler.ts
```diff
@@ -0,0 +1,119 @@
...
+
+      // Process file data/url queries
+      const parameters = { ...queryParams }
+      ;['user_id', 'application_serial', 'application_response_id'].forEach((field) => {
+        const queryFieldValue = queryParams[field]
+        const bodyFieldValue = file.fields[field] ? file.fields[field].value : undefined
+        parameters[field] = queryFieldValue ? queryFieldValue : bodyFieldValue
+      })
+
+      const subfolder = parameters?.application_serial ? parameters.application_serial : ''
```
andreievg, line 61: parameters?.application_serial || ''

src/components/files/fileHandler.ts
```diff
@@ -0,0 +1,119 @@
...
+    for await (const file of data) {
+      const ext = path.extname(file.filename)
+      const basename = path.basename(file.filename, ext)
+      const unique_id = nanoid()
+
+      // Process file data/url queries
+      const parameters = { ...queryParams }
+      ;['user_id', 'application_serial', 'application_response_id'].forEach((field) => {
+        const queryFieldValue = queryParams[field]
+        const bodyFieldValue = file.fields[field] ? file.fields[field].value : undefined
```
andreievg, line 57: would file.fields have `'user_id', 'application_serial', 'application_response_id'` ?

src/components/files/fileHandler.ts
andreievg, line 61: should you also create subfolder path ? (that could be passed to createThumbnail)

andreievg (changes requested):
I think that saveFile can be improved with:
* storing all results of path in variables (and re-using)
* getting rid of `parameters`, I am not really sure why that's needed.

src/components/files/createThumbnails.ts
CarlosNZ: Yeah, my bad. Changed it to save to subfolder and hadn't updated the thumbnail path. Fixed in 24aaeec

CarlosNZ: See prev comment

src/components/files/createThumbnails.ts
```diff
@@ -0,0 +1,77 @@
...
+  pdf: 'noun_PDF_3283219.png',
+  text: 'noun_word_3515287.png',
+  msword: 'noun_word_3515287.png',
+  file: 'noun_File_3764922.png',
+}
+
+import * as config from '../../config.json'
+import path from 'path'
+import fs from 'fs'
+const sharp = require('sharp')
```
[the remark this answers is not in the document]
CarlosNZ: Yeah, good point. Added @types and changed to `import` 👍 16b4980

src/components/files/createThumbnails.ts
```diff
@@ -0,0 +1,77 @@
...
+      await fs.rename(thumbnailFilePath, `${thumbnailFilePath}.${format}`, () => {})
+      return `${basename}_${unique_id}_thumb.${format}`
+    } catch {
+      return getGenericThumbnailPath(type)
+    }
+  } else if (mimetype === 'application/pdf') {
+    // TO-DO Generate PDF previews/thumbnails
+    return getGenericThumbnailPath('pdf')
+  } else if (['.pdf', '.doc', '.docx'].includes(ext))
+    return getGenericThumbnailPath(`_/${ext.replace('.', '')}`)
```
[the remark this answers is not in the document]
CarlosNZ: Hmmm, I was doing that so that the file extension was the second part of the "mimetype" string (a bit hacky I know), but it looks like I didn't end up needing it like that anyway. Simplified it now: cff046b

src/components/files/createThumbnails.ts
```diff
@@ -0,0 +1,77 @@
...
+        .toFile(thumbnailFilePath)
+      await fs.rename(thumbnailFilePath, `${thumbnailFilePath}.${format}`, () => {})
+      return `${basename}_${unique_id}_thumb.${format}`
+    } catch {
+      return getGenericThumbnailPath(type)
+    }
+  } else if (mimetype === 'application/pdf') {
+    // TO-DO Generate PDF previews/thumbnails
+    return getGenericThumbnailPath('pdf')
+  } else if (['.pdf', '.doc', '.docx'].includes(ext))
```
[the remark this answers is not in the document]
CarlosNZ: Yeah, it's not ideal, but the problem is there is not a lot of consistency. For example, all "image" file types have "image" in the first part of their mimetype, so it makes a lot more sense to match by type rather than think of all possible image extensions. And on the other hand, there's no generic mimetype that will capture all the possible types of file that should get the generic "DOC" thumbnail, so you need to really match by extension. I think once we have a clearer idea of what our final set of generic thumbnails will be and what the possible types and extensions are, we can make a simpler, definitive mapping. Current version was just to make sure it works for a few basic, common format we're likely to use.

src/components/files/createThumbnails.ts
```diff
@@ -0,0 +1,77 @@
...
+const splitMimetype = (mimetype: string) => {
+  // To-do: parse correctly if mimetype has parameter
+  // (e.g. type/subtype;parameter=value)
+  const [type, subtype] = mimetype.split('/')
+  return { type, subtype }
+}
+
+const getGenericThumbnailPath = (mimetype: string) => {
+  const { type, subtype } = splitMimetype(mimetype)
+  if (subtype in genericThumbnails) return genericThumbnails[subtype]
```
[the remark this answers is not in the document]
CarlosNZ: Yeah fair enough, that's a bit nicer: cff046b

src/components/files/fileHandler.ts
CarlosNZ: Yeah, but I've made the "root files folder" a configurable value (in .config). It's currently set to "files", but it could be changed, which is why I didn't include "files" in the repo.

src/components/files/fileHandler.ts
CarlosNZ: I'm don't fully understand it, but it doesn't work without it. 🤷‍♂️

There's some info on `for await...of` here : https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/for-await...of

Without the await, the files don't even get written, so it must be something to do with what happens with the `await pump...... createFileStream`, I don't know, I just got the boilerplate from the official repo, and they use the `await`: https://github.com/fastify/fastify-multipart#handle-multiple-file-streams

CarlosNZ: Ah, I think it's because `data` is an async Promise, so the await makes the loop wait for `data` to resolve before iterating over it.

CarlosNZ:
>     * storing all results of path in variables (and re-using)

There's probably some more improvements I could make here, but the complication is that the download endpoint requires the URL request to be relative to the fileFolder root (which is "/files"), but all the local file manipulation is relative to the project root, hence it gets a little more complicated than I'd like. I'll have another look and see if there's any additional simplifications I can make.
 
>     * getting rid of `parameters`, I am not really sure why that's needed.

Yeah, this doesn't do what it used to. Originally we had it so that the parameters could be passed in as either "query" parameters (in URL), OR as key-value fields in the request body (where "files" is posted), so the "parameters" variable was for storing the result of checking either source. However, the way I've restructured this now (I've made the file saving go _after_ the query and thumbnail processing, it used to be the first thing it did), the body parameters are no longer available (they get resolved in the `await pump...` and I don't know how to separate that from the createWriteStream).

Anyway, long story short -- I've got rid of all the extra parameters and it now only accepts URL query parameters. 130e5b8

Have also updated the docs to reflect this change. e8af4c7

CarlosNZ:
> There's probably some more improvements I could make here, but the complication is that the download endpoint requires the URL request to be relative to the fileFolder root (which is "/files"), but all the local file manipulation is relative to the project root, hence it gets a little more complicated than I'd like. I'll have another look and see if there's any additional simplifications I can make.

Also -- another reason why I passed the path/filename elements separately into createThumbnail is that I needed to keep the file extension separate. I was imagining that the file format of the thumbnail might be different to the original file, (e.g. if we generated PDF thumbnails, they'd be .png thumbnails), so needed to build the thumbnail filename separately and use the `sharp` output "format" field as its extension.

src/components/files/fileHandler.ts
andreievg: sweet

src/components/files/createThumbnails.ts
andreievg: oops didn't mean to double comment

src/components/files/fileHandler.ts
CarlosNZ: It used to -- but see [this comment](https://github.com/openmsupply/application-manager-server/pull/250#issuecomment-799074556) Gone now.

src/components/files/fileHandler.ts
andreievg: cool, makes sense, would you mind adding a comment ?

CarlosNZ:
Okay, have also changed the thumbnail generation to always convert to .jpg.

src/components/files/fileHandler.ts
CarlosNZ: Sure. 6643e49

andreievg (approved):
Looks solid as, thanks for changes

Issue #256, opened 2021-03-16
trimResponse timestamp dont' quite match


andreievg:
the updated timestamp dont' quite match to review and application latest status history (must be to do with transition in pg node between db, node and back to db). 

recommend to just set directly from application/review status history

PR #257, opened 2021-03-16
#256-set-timestamp-in-trimResponses-from-db-records

andreievg:
fixesd: #256


CarlosNZ (approved):
Great, thanks for this. Assume you've tested and it works?

andreievg:
> Great, thanks for this. Assume you've tested and it works?

tests worked, and i am using it in a front end branch i am working on. btw trim functionality works great (on submission of application). but found a bug in front end where more then the value is submitted for application (will be fixed in my front end branch)

Issue #248, opened 2021-03-12
Action to process files on Application Submit


CarlosNZ:
On Application Submit, trigger an Action that checks all files associated with applicationID. Delete any that are not in their associated applicationResponse, and set the rest to submitted=true

PR #259, opened 2021-03-17
#248 Action: clean up deleted files

CarlosNZ:
Implements #248 -- action to clean-up deleted files, to be run on application submit.

To test: use front-end branch from [PR #454](https://github.com/openmsupply/application-manager-web-app/pull/454)

Complete a "Feature Showcase" application, at the "File Upload" questions (Section 2), upload a bunch of files, then delete most of them.
When you Submit the application, the Action will run and only the files still present in responses will be kept, and set to "submitted = true".

You can also test the Action without completing the whole application by triggering the "devTest" trigger on it (see below). 

### Also included here:

- Added a new "devTest" trigger (as discussed) to be used for development. There is also a "dev_actions.js" file that is inserted into each template (same as "core_actions"). The intention is that we can chuck Actions into "dev_actions" for testing purposes, using the "devTest" trigger, and run them at will by manually setting the "trigger" field to "devTest". As long as all actions within "dev_actions" are triggered by "devTest", they'll never be executed "for real".
- Added `uniqueId` to the file upload endpoint response. Not needed for front-end, but needed in this Action (and probably elsewhere eventually), and it seemed a bit hacky to try and re-parse the UID from a URL string.
- Added an additional "environmentData" field to "applicationData" so Action plugins can have access to configuration values that can't easily be passed in as parameters (currently only includes `appRootFolder` and `filesFolderName`)
- include "applicationSerial" in "applicationData" object
- Empty string text fallback for some `objectProperties` expressions (so doesn't print "Can't resolve object" to UI)


src/plugins/action_files_cleanup/src/cleanupFiles.ts
```diff
@@ -0,0 +1,93 @@
...
+    return {
+      status: 'Fail',
+      error_log: 'Problem cleaning up files: ' + error.message,
+    }
+  }
+
+  async function deleteFile(file: FileInfo) {
+    try {
+      // Delete the file
+      await fs.unlink(path.join(appRootFolder, filesFolderName, file.filePath), () => {})
```
andreievg, line 81: i think you want to promisify fs.unlink to make await work properly (not sure if it's awaiting for the unlink atm)

src/plugins/action_files_cleanup/src/plugin.json
```diff
@@ -0,0 +1,25 @@
+{
+  "type": "action_plugin",
+  "code": "cleanupFiles",
+  "name": "Clean up application files",
+  "file": "cleanupFiles",
+  "function_name": "cleanupFiles",
+  "description": "Checks all the uploaded files associated with an application, and deletes any that are not part of an application_response",
+  "required_parameters": ["applicationId"],
```
andreievg, line 8: i think these are optional parameters (also has applicationSerial).

Btw I would suggest not using 'serial' (apart from some UI and URL in front end) <- I still think we should have just used the id of application in URL, and use serial for display purposes only.

The serial to id change would require quite a bit of juggling, so not a required change, just something to think about for future cases.

src/plugins/action_files_cleanup/src/cleanupFiles.ts
```diff
@@ -0,0 +1,93 @@
...
+  uniqueId: string
+  applicationResponseId: number
+  filePath: string
+  thumbnailPath: string
+}
+
+interface ResponseValue {
+  text?: string
+  files?: {
+    fileUrl: string
```
andreievg, line 17: only one of these is used, do we envision others being used ?

andreievg (changes requested):
Looks pretty solid, can approve as is tbh, just a couple of small tidy ups

src/plugins/action_files_cleanup/src/cleanupFiles.ts
CarlosNZ: Oh, it was my understanding that this was a proper async method. From what I can the fs module has Sync and Async versions of most of its functions. The sync version of this is called `unlinkSync` (I think).

src/plugins/action_files_cleanup/src/plugin.json
CarlosNZ: Yeah, I think it's useful to have "serial" available in `applicationData` for some situations. Ideally Actions could take either and still work fine. But sure, ID as the main one to use for back-end makes sense (except for when it doesn't lol).

src/plugins/action_files_cleanup/src/cleanupFiles.ts
CarlosNZ: No, where not using them here, I was just being strict about defining what is actually going to be returned from the database (that's the shape of the JSON object in "value" for file responses).

src/plugins/action_files_cleanup/src/cleanupFiles.ts
andreievg: it's old school call back style, it would be something like this (i'll use timeout as example of operation)
```
const unlinkSync = (file, callback) => {
// do operation

setTimeout(() => {

callback('i am done')

}
 , operationTime)

}
```

promisify would wrap it in a promise:

```
(file) => new Promise(resolve => unlinkSync(file, resolve))
```

To test this you can specify a wrong file name and there won't be error thrown

CarlosNZ: Cool, have used the fs.promises method now. d5d523c0

andreievg (approved):

Issue #260, opened 2021-03-17
Slight issue with updates of reviewResponse in trimResponse


andreievg:
Slight caveat in trimResponse logic (my bad, i didn't quite think this bit fully through when communicating to Carl).

We want to trim reviewResponse grouped by application response (this is for review change request, when reviewer is making changes to requests from hight up reviewers/consolidators), which is happening correctly now.

But for updating timestamp of reviewResponse to match latest review status change, we want to group by templateElementID (to make sure we don't update previous reviewResponses)

I've noticed this and fixed while working on #443

PR #261, opened 2021-03-17
#443f and #260 restart status for review, and fix update of reviewResponse in trimResponses

andreievg:
Fixes: #269

### Changes

* added re-start review status and action (change review to draft)
* fixed update of reviewResponses

### DIFF
--- a/src/plugins/action_trim_responses/src/trimResponses.ts
+++ b/src/plugins/action_trim_responses/src/trimResponses.ts
@@ -42,8 +42,8 @@ module.exports['trimResponses'] = async function (input: any, DBConnect: any) {
 
     const responsesById = groupResponses(responses, groupByField)
     const responsesToDelete: number[] = []
-    let responsesToUpdate: number[] = []
 
+    // Calculate review and application responses that should be deleted
     Object.values(responsesById).forEach((responseArray) => {
       const latestResponse = responseArray[responseArray.length - 1]
       const previousResponse =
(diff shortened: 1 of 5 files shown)


src/plugins/action_trim_responses/src/trimResponses.ts
```diff
@@ -54,9 +54,19 @@ module.exports['trimResponses'] = async function (input: any, DBConnect: any) {
...
+      else if (!reviewId) responsesToUpdate.push(latestResponse.id)
     })
 
+    // For level 1 reviewer:
+    // We want to update latest review responses if they are not null or duplicates
+    // If we update the latest review response for template element it should be
+    // review response that was changed in this review (not changed or null would be deleted in responsesToDelete)
+    if (reviewId && reviewLevel === 1)
+      responsesToUpdate = Object.values(groupResponses(responses, 'template_element_id')).map(
+        (responseArray) => responseArray[responseArray.length - 1].id
```
CarlosNZ, line 67: From what I can tell, the `responsesToUpdate` array should be empty by here if it's a review (the above loop only pushes application responses to it). So why not just push on to the array here, then you go back to `const responsesToUpdate: number[] = []` up the top?
```
      responsesToUpdate.push(...Object.values(groupResponses(responses, 'template_element_id')).map(
        (responseArray) => responseArray[responseArray.length - 1].id
      ))
```

And you can chuck a `responsesToUpdate.length = 0` before just to be sure.

Sorry, that's super nit picky, I just try and avoid reassigning objects and arrays (always `const`), but its no real issue in this case.

CarlosNZ (approved):
Cool. I just made one tiny comment, feel free to ignore it. Just being picky about `let` vs `const`. 😉

src/plugins/action_trim_responses/src/trimResponses.ts
andreievg: good point, come to think of it, this might be simpler ? c68fd1ddd34b0e97874e82e93df6e22533d3d2e3
I'll wait for your comment before mergin

andreievg: tests passed

andreievg: i just looked at that last commit and changed the name of variable: 315b1b9b262ec37fa2cda4ecc26ee2e798831a3f

CarlosNZ: Yeah, LGTM :) Seems nicely logically broken up now.

I haven't checked, but did you confirm the tests still pass?

andreievg: yep, just re-ran again to be sure and did a couple of review cycles. cheers for review !

Issue #264, opened 2021-03-18
Changes to schema and documentation for Assigner using join table user, org and template type which should allow Assigner user being able to see assignable applications in the list


CarlosNZ:
Plus update "generateReviewAssignments" action to create records on this table

PR #265, opened 2021-03-18
#264 Assigner/ReviewAssignment joins

CarlosNZ:
Implements #264.

## Summary of changes:

- new table: `review_assignment_assigner_join`
- Updates to "all_permissions" VIEW:
   - renamed to "permissions_all" for consistency with other Views
   - added some additional fields: permissionName, stageNum, Level, orgName/Id
- create `basicAssign` permission policy
- create "generalAssign" permission name, and apply it to ReviewTest template
- create/update "Assigner" users, and give them "generalAssign" permissions
- Update `generateReviewAssignment` Action:
   - Move database methods to package "databaseMethods" file
   - change existing "getReviewers..." method to query "permissions_all" view rather than doing its own joins
   - renamed `getReviewers..." method to `getPersonnelForApplicationStageLevel` and generalised it to look up either Assigners or Reviewers
   - after generating review_assignments, continue and generate review_assignment_assigner_join records for the recently updated/created review_assignments (i.e. a list of assigners available to assign those assignments)
-  update tests

To quickly test: manually change the trigger on Application 4001 (Vitamin B) to "onApplicationSubmit" -- new review_assignments will be generated/updated, and 6 new review_assignment_assigner_join records created.


src/plugins/action_generate_review_assignment_records/src/databaseMethods.ts
```diff
@@ -0,0 +1,131 @@
...
+        text,
+        values: [templateId, stageNumber, reviewLevel],
+      })
+      return result.rows
+    } catch (err) {
+      console.log(err.message)
+      throw err
+    }
+  },
+  addReviewAssignments: async (reviewAssignments: any) => {
```
andreievg, line 29: would assume nothing changed in this method ?

src/plugins/action_generate_review_assignment_records/src/databaseMethods.ts
```diff
@@ -0,0 +1,131 @@
...
+          ${orgId ? ', organisation_id' : ''} 
+        )
+          WHERE organisation_id IS ${orgId ? 'NOT ' : ''}NULL
+        DO UPDATE SET organisation_id = ${orgId ? 'orgId' : 'NULL'}
+        RETURNING id
+      `
+      try {
+        const result = await DBConnect.query({
+          text,
+          values: [assignerId, reviewAssignmentId, ...(orgId ? orgId : [])],
```
andreievg, line 114: I think this needs brackets around orgId
`...(orgId ? [orgId]: [])`

also I think you can have orgId in the values list and have `VALUES ($1, $2, $3)`

src/plugins/action_generate_review_assignment_records/src/databaseMethods.ts
```diff
@@ -0,0 +1,131 @@
...
+      const { assignerId, orgId, reviewAssignmentId } = reviewAssignmentAssignerJoin
+      // Needs a slightly different query with different CONFLICT restrictions
+      // depending on whether orgId exists or not.
+      const text = `
+        INSERT INTO review_assignment_assigner_join (
+          assigner_id,
+          review_assignment_id
+          ${orgId ? ', organisation_id' : ''}          
+          )
+        VALUES ($1, $2${orgId ? ', $9' : ''})
```
andreievg, line 102: $3

src/plugins/action_generate_review_assignment_records/src/generateReviewAssignments.ts
```diff
@@ -61,19 +65,44 @@ module.exports['generateReviewAssignments'] = async function (input: any, DBConn
           isLastLevel: nextReviewLevel === numReviewLevels,
         }
     })
+    console.log('ReviewAssignments', reviewAssignments)
```
andreievg, line 68: I think this one is for dev ?

src/plugins/action_generate_review_assignment_records/src/generateReviewAssignments.ts
```diff
@@ -61,19 +65,44 @@ module.exports['generateReviewAssignments'] = async function (input: any, DBConn
...
+    // For now we assume that assigners have no Section restrictions
+    console.log('Generating review_assignment_assigner_join records...')
+    const availableAssigners = await db.getPersonnelForApplicationStageLevel(
+      templateId,
+      stageNumber,
+      nextReviewLevel,
+      'Assign'
+    )
+    const reviewAssignmentAssignerJoins = []
+    for (const reviewAssignmentId of Object.values(reviewAssignmentIds)) {
```
andreievg, line 83: Hmm i thought `reviewAssignmentIds` is an array of ids (or possibly an error message) ? Would be good to change even thought Object.values work (also would this break if there was an error message as one of reviewAssignemntIds) ?

Screen Shot 2021-03-19 at 3 34 34 PM

src/plugins/action_generate_review_assignment_records/src/generateReviewAssignments.ts
andreievg: I think it's ok to log reviewAssignmentIds returned by `addReviewAssignmentAssignerJoins`, though

andreievg (changes requested):
looks pretty good @CarlosNZ, a couple of request though

src/plugins/action_generate_review_assignment_records/src/databaseMethods.ts
CarlosNZ: Nope, just copy-pasted.

src/plugins/action_generate_review_assignment_records/src/databaseMethods.ts
CarlosNZ: > I think this needs brackets around orgId
> `...(orgId ? [orgId]: [])`

Yeah I think you're right. Which is weird cos it's copied and pasted from the previous one, so I guess that's been wrong too. Guess we've had no users with orgs being added. Probably should write some tests for that...

> also I think you can have orgId in the values list and have `VALUES ($1, $2, $3)`

Oh yeah, that works. Cool. Still have to use the `orgId ? ' organisation_id,' : ''` for the conflict resolution though.

Have changed both methods now: 467704d9

src/plugins/action_generate_review_assignment_records/src/databaseMethods.ts
CarlosNZ: 🤦‍♂️

src/plugins/action_generate_review_assignment_records/src/generateReviewAssignments.ts
CarlosNZ: Okay, I'll just comment out the first one.

src/plugins/action_generate_review_assignment_records/src/generateReviewAssignments.ts
CarlosNZ: Oh yeah, good spotting. I was getting mixed with `reviewAssignments`, which is an object. Cheers.

6a920982

src/plugins/action_generate_review_assignment_records/src/generateReviewAssignments.ts
```diff
@@ -61,19 +65,44 @@ module.exports['generateReviewAssignments'] = async function (input: any, DBConn
...
+      nextReviewLevel,
+      'Assign'
+    )
+    const reviewAssignmentAssignerJoins = []
+    for (const reviewAssignmentId of reviewAssignmentIds) {
+      for (const assigner of availableAssigners) {
+        reviewAssignmentAssignerJoins.push({
+          assignerId: assigner.userId,
+          orgId: assigner.orgId,
+          reviewAssignmentId,
```
andreievg, line 88: still not sure how this would work if there is an error (instead of id) as reviewAssignmentId.

I forgot what happens if an action thrown uncaught error (if it's gracefully handles further up or not)

andreievg (approved):
Nice! Thanks for changes @CarlosNZ, have a question still about reviewAssignmentId as error vs id. But that shouldn't stop the merge

src/plugins/action_generate_review_assignment_records/src/generateReviewAssignments.ts
CarlosNZ: The whole Action is in a `try{ }` block, and the `catch` stores the error message in its output log, which gets written to the action_queue along with Success/Fail etc.

So the Action should throw the error on the first error and never reach here if there was a problem with generating reviewAssignmentIds.

CarlosNZ:
```
 } catch (error) {
    console.log(error.message)
    return {
      status: 'Fail',
      error_log: 'Problem creating review_assignment records: ' + error.message,
    }
  }
```
Every Action follows this same template, FYI.

PR #266, opened 2021-03-18
#227 Add reviewer list and use function for application list

andreievg:
Needed to solve this issue: https://github.com/openmsupply/application-manager-web-app/issues/227
Previous PR attempt: https://github.com/openmsupply/application-manager-server/pull/227
Needs to merged together with this front end branch: https://github.com/openmsupply/application-manager-web-app/pull/469

Ended up implementing something similar to this:https://github.com/openmsupply/application-manager-server/pull/227#issuecomment-788670002

### Changes

I thought it would be a bit simpler, and things were looking bring, until i realised that custom queries (functions in pg that return sets of, that are exposed by graphile as GraphQL queries) I've found something that fixes it `https://github.com/graphile/graphile-engine/pull/378`, but requires a set of 'table', which means a table with the shape of the output needs to be created. Anyways it seems to be working now, and the application_list_shape table is generated into types in front end.


andreievg:
### Comments from previous PR:

@nmadruga :

> And also we need some solution to the problem - when review lvl > 1 needs to see how many reviews (from other users) have been submitted/re-submitted so they can take action to start a new consolidation.

We can probably do a query for lvl -1 assignments. For now an action creates review_assignment for consolidator when previous lvl -1 submits a PR (and if review is already created for that review_assignment, we will change the status of that review to 'pending'). So that should allow for actions to show for consolidator

> I think this field name could be simplified with something as numberOfAssignedToAnother

I've changed them all `review...Count` <- that kind of indicates what it's for and what it is, is that ok ?

> It was a great idea to add a view, simplifies so much the result and how it needs to be deduced by front-end (I'm going to have a look on the front-end issue now). But as a reminder of what @CarlosNZ was saying on today's meeting: what would be done for users that need to view other reviews/consolidator. Also if we consider the consolidator we might need to know the number of review submitted/pending/changes required from a level below.

I kind of think that this information should be available through review home ?

> Would be worth to have on already for Pending status?

Added

@CarlosNZ 

> This is just a error-protection because graphQL can't accept empty filter objects. It was meant to return everything if no filter object was supplied, but with isNull: true it'll return nothing. Since the next line adds a reviewerId to the filter object, we probably don't need this line (maybe comment it out for now?)

not touching any of that logic anymore (in the front end pr) : ) 

https://github.com/openmsupply/application-manager-web-app/pull/382#issuecomment-788670914

> Just a personal pref, but I think these two columns should be first in the column order -- so we can easily see what the counts are referring to. (Seems like a standard convention and it makes browsing the table directly a lot clearer)

this was in regards to application_id column, which is the top in the query now (and user_id one is not needed)

> wonder if this View should also return one set of applications that are not joined to a reviewer_id (i.e. all the review columns would be NULL). That way it would still be possible to query without a review_id (or reviewer_id = NULL) and get the full set of applications for a list that doesn't need to care about reviewer id.

I think review home can show more fine grained information about review. if we want to see total actions from all of the reviewers, we can set up a mechanism for passing through say null as reviewerID, if if it's null all off the review assignment can be returned

> Cos what happens to the application list for applicants. If they have a bunch of applications in progress, they'd want to be able to see them all without caring who the reviewer is, right?

> You could maybe do a UNION of this query with the original query, perhaps?

I think the above two comments were answered here: https://github.com/openmsupply/application-manager-server/pull/227#issuecomment-788670002

@nmadruga

> Not from this PR, sorry I just have investigated if this is_fully_assigned would return what is expected, but I don't this it is considering questions that aren't visible therefore not assigned in the review... I think I'll create a separated issue for that. Just pointing out here so we can discuss - maybe I'm just not seeing something

@CarlosNZ 

> I'd have thought that when you assign a Section (or a whole application), they'd be assigned regardless of whether or not question is visible. In fact, back-end has no concept of "is_visible" as it's only ever calculated on-the-fly in the front end (and changes as responses change). You assign reviews based on template_elements, don't you, rather than specific applications responses? If not, they probably should be. You don't want a situation where a reviewer is assigned a section, and then the applicant makes changes which makes another question visible but no one is assigned to it. Surely we assigned a selection of "template elements" which are tied to responses, not responses themselves?

Yeah i agree, assign the whole thing

nmadruga (commented):
Hmm files are jumping a number. Better to renamed as:
`36_review_list.sql `
`37_application_list_view.sql`

database/buildSchema/37_review_list.sql
```diff
@@ -0,0 +1,37 @@
+-- Aggregated VIEW method of all data review data, use in application_list for application list page
```
nmadruga, line 1:
```suggestion
-- Aggregated VIEW method of all related review data to each application on application list page
```

database/buildSchema/37_review_list.sql
```diff
@@ -0,0 +1,37 @@
...
+        review_draft_count bigint,
+        review_submitted_count bigint,
+        review_change_request_count bigint,
+        review_pending_count bigint
+    )
+    AS $$
+    SELECT
+        review_assignment.application_id AS application_id,
+        COUNT(*) FILTER (WHERE review_assignment.status = 'Available for self-assignment') AS review_available_for_self_assignment_count,
+        COUNT(*) FILTER (WHERE review_assignment.status = 'Assigned') AS review_assigned_count,
```
nmadruga, line 17: Maybe better `review_assigned_started_count` instead? If we need the total we can sum up both

database/buildSchema/37_review_list.sql
```diff
@@ -0,0 +1,37 @@
...
+        COUNT(*) FILTER (WHERE review_assignment.status = 'Assigned'
+            AND review.id IS NULL) AS review_assigned_not_started_count,
+        COUNT(*) FILTER (WHERE review_status_history.status = 'Draft') AS review_draft_count,
+        COUNT(*) FILTER (WHERE review_status_history.status = 'Submitted') AS review_submitted_count,
+        COUNT(*) FILTER (WHERE review_status_history.status = 'Changes Requested') AS review_change_request_count,
+        COUNT(*) FILTER (WHERE review_status_history.status = 'Changes Requested') AS review_pending_count
```
nmadruga, line 23:
```suggestion
        COUNT(*) FILTER (WHERE review_status_history.status = 'Pending') AS review_pending_count
```

andreievg:
> Hmm files are jumping a number. Better to renamed as:
36_review_list.sql
37_application_list_view.sql

Oops, I think i left one for assigner actions (not to move them again), added a placeholder for it: 73d39f7d6924cdf6e3a9454ff33c5dcd53cab77f

database/buildSchema/37_review_list.sql
andreievg: thanks, did this as separate commit as my commit before that conflicted 0af288a113c0c0af4aa6c1a329601b2cc43d9a1f

database/buildSchema/37_review_list.sql
andreievg: would that be review_draft_count ?

That would make the following actions:

Start -> review_assigned_not_started_count
Re-Review-> review_pending_count
Update -> review_change_request_count
Continue -> review_draft_count
Self-Assign -> review_available_for_self_assignment_count

review_assigned_count -> not really sure what to indicate as action, but the idea is it can be filtered by it

Re-Review and Update would change to 'Continue' in application list (but would be correctly represented in review_home

database/buildSchema/37_review_list.sql
andreievg: oops, thanks

database/buildSchema/37_review_list.sql
nmadruga: Got it. Thanks for the explanation : ) I think this would be very useful to add to our documentation about showing actions (if we already have one in progress otherwise to be done as part of separated issue) #474

nmadruga (approved):
LGTM

PR #267, opened 2021-03-19
#442f-add-status-to-application-response

andreievg:
Needed for front end: https://github.com/openmsupply/application-manager-web-app/pull/472

Adds status to application response (default draft, so that front end only needs to set it on application submission)


database/buildSchema/22_application_response.sql
```diff
@@ -1,10 +1,16 @@
 -- application response
+CREATE TYPE public.application_response_status AS ENUM (
+    'Draft',
+    'Submitted'
+);
```
CarlosNZ, line 5: Since we talked about keeping the Enum strings the same as postGraphile displays them (i.e. "DRAFT", "SUBMITTED"), do you want to start doing that with these new ones? (One less to change later)

database/buildSchema/22_application_response.sql
```diff
@@ -1,10 +1,16 @@
...
 
 CREATE TABLE public.application_response (
-    id serial primary key,
-    template_element_id integer references public.template_element(id),
-    application_id integer references public.application(id),
+    id serial PRIMARY KEY,
+    template_element_id integer REFERENCES public.template_element (id),
+    application_id integer REFERENCES public.application (id),
+    "status" public.application_response_status DEFAULT 'Draft',
```
CarlosNZ, line 11: Pretty sure "status" doesn't need the quotes (it doesn't have it in "review_assignment" for example).

database/insertData/06_applications_A_featureShowcase.js
```diff
@@ -17,36 +17,43 @@ exports.queries = [
...
               }
               {
                 id: 1005
+                status: DRAFT 
                 value: { optionIndex: 0, text: "Manufacturer" }
                 templateElementToTemplateElementId: { connectById: { id: 1007 } }
               }
               {
                 id: 1006
+                status: DRAFT
```
CarlosNZ, line 56: Doesn't really matter, but since you added a default value for this, we don't really need to specify "DRAFT".

database/buildSchema/22_application_response.sql
andreievg: good point: 2810c55830694ef2560b2d621bcb001f20f2009d

database/insertData/06_applications_A_featureShowcase.js
andreievg: 2810c55830694ef2560b2d621bcb001f20f2009d

database/buildSchema/22_application_response.sql
andreievg: 9647e30ce59ee7b0568b2a56f41e4d778b932548

database/buildSchema/22_application_response.sql
andreievg: btw front end types didn't change (just FYI, global change should be easy from front end point of view)

database/buildSchema/22_application_response.sql
```diff
@@ -1,14 +1,14 @@
...
+    'DRAFT',
+    'SUBMITTED'
 );
 
 CREATE TABLE public.application_response (
     id serial PRIMARY KEY,
     template_element_id integer REFERENCES public.template_element (id),
     application_id integer REFERENCES public.application (id),
-    "status" public.application_response_status DEFAULT 'Draft',
+    status public.application_response_status DEFAULT 'SUBMITTED',
```
CarlosNZ, line 11: Why did you change the default to "Submitted"?

andreievg: sorry, thanks for noticing

CarlosNZ (approved):

nmadruga:
I though this one was in already. Are you merging this one today @andreievg ?

nmadruga:
Ok, I'll have a look on the front-end PR, I think that's what is missing

PR #269, opened 2021-03-21
#476 fix question assignments and missing data

nmadruga:
Fixes to show progress on serial 12345 and first name for user


nmadruga:
Move to new PR based on `master`

PR #272, opened 2021-03-22
#270 Drug registration break-down of pages for grouping

### LINKED ISSUE #270: Create Drug Registration template more basic and using the upload document plugins

nmadruga:
Fixes #270 

### Merging to demo branch
`demo-22-03` branch

### Changes
- Fixes order of elements (some indexes were out of order during demo)
- Remove specific `id` of template_element (Drug Registration and Join company template) - I had a lot of hassle with these, and they aren't needed to be specified if we are not creating review_question_assignment manually anymore.
- Added more pages to Section 1 and titles to each page


CarlosNZ:
Don't we want to extract this FROM the demo branch (plus your changes) and merge it to master? I don't think we're going to be merging the demo branch are we? I thought that was just a temporary hack from which we could extract useful stuff as needed.

nmadruga:
I think it would be more work to do that - can we have a chat about it after break? @andreievg @CarlosNZ 
I don't think we need to spend too long on that, just a brief conversation to give us some time to do development this week : )

nmadruga:
New branch based on `master` created

PR #278, opened 2021-03-23
#273 fix review assignment generation in different stages

### LINKED ISSUE #273: Add incrementStage action to coreActions to be triggered when REVIEW_SUBMITTED for CONFORM decision and review is last level

nmadruga:
Fixes #273 

### Branched from
`demo-22-03`
PR #272 

### Changes
- Fix problem to create review Assignments in next `stage` after review is submitted for last level in stage
- Changed `generateReviewAssignment` to call method `generateNextReviewAssignments` and deal differently in 3 cases:
  - No `reviewId` received => First stage and first level review
  - `reviewId` received and `review.stageNumber` != `application.stageNumber` => New stage first level review
  - `reviewId` received and `review.stageNumber` == `application.stageNumber` => Same stage next level review

### Tests
- Have updated tests to not be passing `stageNumber`  and `stageId` anymore, but haven't tried to run


src/components/postgresConnect.ts
```diff
@@ -472,7 +472,7 @@ class PostgresDB {
 
   public getCurrentStageHistory = async (applicationId: number) => {
     const text =
-      'SELECT stage_id, stage_number, stage, stage_history_id, status_history_id, status FROM application_stage_status_all WHERE application_id = $1 AND stage_is_current = true'
+      'SELECT stage_id, stage_number, stage, stage_history_id, status_history_id, status FROM application_stage_status_all WHERE application_id = $1 AND stage_is_current = true AND status_is_current = true'
```
CarlosNZ, line 475: Best to just use the `application_stage_status_latest` View then you don't need the `current=true`.

src/plugins/action_generate_review_assignment_records/src/types.ts
```diff
@@ -5,6 +5,13 @@ export enum AssignmentStatus {
   SELF_ASSIGN = 'Available for self-assignment',
 }
 
+export interface ApplicationData {
+  applicationId: number
+  templateId: number
+  stageId: number
+  stageNumber: number
+}
```
CarlosNZ, line 13: Do we need to fully define this type? Cos there's a lot more in it than just that. Just wondering, maybe it's best to leave it as `any` in the Actions if we're not going to fully define it, otherwise it could be misleading.

CarlosNZ (commented):
Sorry, haven't looked at this too thoroughly yet. But don't you need to pull out the relevant bits to merge back into master? Not sure why it's merging back to demo branch.

src/plugins/action_generate_review_assignment_records/src/types.ts
nmadruga: Well, I was going to define, but when I tried adding ApplicationStatus and realise it doesn't have access to generated GraphQL types (as I saw there is a duplication of `AssignmentStatus`) I gave up, I think it would be best if we would be importing those types from there instead of duplicating otherwise if we end up changing this will be source of problems

CarlosNZ: Except the Actions need to be "self-contained" and not import stuff from higher up in the application structure. Can we pass in the GraphQL types somehow?

src/components/postgresConnect.ts
nmadruga: I'll make changes on new PR

src/plugins/action_generate_review_assignment_records/src/types.ts
nmadruga: I think we need a why to use the GraphQL types on actions too - instead of recreating types which is dangerous.
Not really part of this PR though, but I suggest simple adding the same package and config to generate GraphQL types on each plugin

nmadruga:
Have recreated PR based on `master`

PR #280, opened 2021-03-24
#273 Fix reviewAssignment creation after stage change + incrementStage as coreAction

### LINKED ISSUE #273: Add incrementStage action to coreActions to be triggered when REVIEW_SUBMITTED for CONFORM decision and review is last level

nmadruga:
Fixes #273

### Branch of
#279 

### Changes
- Fix problem to create review Assignments in next `stage` after review is submitted for last level in stage
- Changed `generateReviewAssignment` to call method `generateNextReviewAssignments` and deal differently in 3 cases:
  - No `reviewId` received => First stage and first level review
  - `reviewId` received and `review.stageNumber` != `application.stageNumber` => New stage first level review
  - `reviewId` received and `review.stageNumber` == `application.stageNumber` => Same stage next level review

### Tests
- Have updated tests to not be passing `stageNumber`  and `stageId` anymore, but haven't tried to run

### Problem
- Having been able to test assignment working on master yet


src/components/postgresConnect.ts
```diff
@@ -472,7 +472,7 @@ class PostgresDB {
 
   public getCurrentStageHistory = async (applicationId: number) => {
     const text =
-      'SELECT stage_id, stage_number, stage, stage_history_id, status_history_id, status FROM application_stage_status_all WHERE application_id = $1 AND stage_is_current = true'
+      'SELECT stage_id, stage_number, stage, stage_history_id, status_history_id, status FROM application_stage_status_all WHERE application_id = $1 AND stage_is_current = true AND status_is_current = true'
```
CarlosNZ, line 475: You can just query the "application_stage_status_latest" View, then you don't need `AND status_is_current = true`.

src/components/postgresConnect.ts
```diff
@@ -682,11 +682,17 @@ class PostgresDB {
     }
   }
 
+  // TODO: Rename to getReviewStageAndLevel
```
CarlosNZ, line 685: I'd suggest moving this to the Action's databaseMethods file, then you can make the name change.

The general rule (since I've been caught out a couple of times) is that any methods in `postGresConnect` should be assumed to be used elsewhere, so we should be very cautious about editing them. And if they're exclusive to a particular Action, then they should stay in the Action package (databaseMethods), then we know they can safely be edited without affecting anything else.

src/plugins/action_generate_review_assignment_records/src/generateReviewAssignments.ts
```diff
@@ -1,112 +1,105 @@
...
   const db = databaseMethods(DBConnect)
 
   console.log('Generating review assignment records...')
   try {
-    const { applicationId, reviewId, templateId, stageId, stageNumber } = input
-    // NB: reviewId comes from record_id on TriggerPayload when triggered
-    // from review table
+    const { applicationId, reviewId } = input
+    // Get template information and current stage for application
+    const applicationData: ApplicationData = await DBConnect.getApplicationData(applicationId)
```
CarlosNZ, line 11: No need to query this, it's already passed through as part of input. So you just need to add `applicationData` to the destructured field above:
`const { applicationId, reviewId, applicationData } = input`

src/plugins/action_generate_review_assignment_records/src/generateReviewAssignments.ts
```diff
@@ -1,112 +1,105 @@
...
-        reviewAssignments[userOrgKey] = {
-          reviewerId: userId,
-          orgId,
-          stageId,
-          stageNumber,
-          // TO-DO: allow STATUS to be configurable in template
-          status: nextReviewLevel === 1 ? AssignmentStatus.AVAILABLE : AssignmentStatus.SELF_ASSIGN,
+    // For level 1+ or next stages review assignment
+    // TODO - #440: Needs to lock same level reviews
+    // TODO: Unlock lower level review when consolidator has submitted "Change requested"
```
CarlosNZ, line 45: Done in another Action

src/plugins/action_generate_review_assignment_records/src/generateReviewAssignments.ts
```diff
@@ -116,6 +109,111 @@ module.exports['generateReviewAssignments'] = async function (input: any, DBConn
...
+    output: {
+      reviewAssignments: Object.values(reviewAssignments),
+      reviewAssignmentIds,
+      reviewAssignmentAssignerJoins,
+      reviewAssignmentAssignerJoinIds,
+      nextStageNumber,
+      nextReviewLevel,
+    },
+  }
+}
```
CarlosNZ, line 215: Is this all the same as it was, just extracted into a method?

src/plugins/action_generate_review_assignment_records/src/generateReviewAssignments.ts
```diff
@@ -1,112 +1,105 @@
...
 
   console.log('Generating review assignment records...')
   try {
-    const { applicationId, reviewId, templateId, stageId, stageNumber } = input
-    // NB: reviewId comes from record_id on TriggerPayload when triggered
-    // from review table
+    const { applicationId, reviewId } = input
+    // Get template information and current stage for application
+    const applicationData: ApplicationData = await DBConnect.getApplicationData(applicationId)
+    const { templateId, stageNumber, stageId } = applicationData
```
CarlosNZ, line 12: I thought you were querying `stageNumber` using the amended `getCurrentReviewLevel` method (above)? This looks like you're still getting if from `applicationData`, which will be the stageNum that it was at at the start of this sequence of Actions -- which will cause the issue we discussed.

src/plugins/action_generate_review_assignment_records/src/generateReviewAssignments.ts
```diff
@@ -1,112 +1,105 @@
...
+      const isLastLevel = nextReviewLevel === numReviewLevels
+      return generateNextReviewAssignments({
+        db,
+        applicationId,
+        templateId,
+        nextReviewLevel,
+        nextStageNumber: stageNumber,
+        nextStageId: stageId,
+        isLastLevel,
+      })
```
CarlosNZ, line 41: This block is basically the same as in the `else` block below, so probably want to move it outside the `if/else`.

src/plugins/action_generate_review_assignment_records/src/generateReviewAssignments.ts
```diff
@@ -1,112 +1,105 @@
...
           applicationId,
-          templateSectionRestrictions,
-          level: nextReviewLevel,
-          isLastLevel: nextReviewLevel === numReviewLevels,
+          templateId,
+          nextReviewLevel,
+          nextStageNumber: stageNumber,
+          nextStageId: stageId,
+          isLastLevel,
+        })
```
CarlosNZ, line 75: Same here -- can the return block go outside all the conditionals, since it's the same for all three?

src/plugins/action_generate_review_assignment_records/src/generateReviewAssignments.ts
```diff
@@ -116,6 +109,111 @@ module.exports['generateReviewAssignments'] = async function (input: any, DBConn
...
+  db,
+  applicationId,
+  templateId,
+  nextReviewLevel,
+  nextStageNumber,
+  nextStageId,
+  isLastLevel,
+}: GenerateNextReviewAssignmentsProps) => {
+  // TODO: Needs to check if there are already reviews created - if partial review
+  // for level - 1 are submitted separetedily will create 2 reviewAssignments for level 1++
```
CarlosNZ, line 132: I'm not sure I follow what you're saying here. It won't create additional reviewAssignments, it'll just update existing ones.

CarlosNZ (changes requested):
Cool, haven't run the tests or done any manual testing yet, but I'll revisit this again on next pass.

src/components/postgresConnect.ts
nmadruga: In this case should I move `getCurrentStageHistory` to incrementStage since it's only used in there?

nmadruga: I don't know because how do you know there is one already if we keep spreading things on this separated files. I like having everything in the same file to allow for reusability

CarlosNZ: Don't worry about moving this one specifically. We're slowly migrating them over where appropriate.

> I like having everything in the same file to allow for reusability

Yes, same, but it turned out that hardly any of these were re-usable since they're so specific to their use case. So the postGresConnect file was just becoming a bloated mess and very hard to work with. And the big danger is that it's really tricky to edit an existing query in there as it might mess up another Action that uses it. (This happened a couple of times, which is why we went in this direction). If we find we need a query that already exists for another Action, then it's no trouble to move it back to the shared one. But it makes life sooo much easier when working on actions not constantly scrolling and searching through a giant file.

nmadruga: > I don't know because how do you know there is one already if we keep spreading things on this separated files. I like having everything in the same file to allow for reusability

I don't think that will happen because it's not very common to go looking on other actions when you are developing a new one. But, you know, it really doesn't matter. Is just a preference in this case. It's nicer to have thing that are related closer or at least easy to find.

src/plugins/action_generate_review_assignment_records/src/generateReviewAssignments.ts
nmadruga: Cool. I'll remove this comment

src/plugins/action_generate_review_assignment_records/src/generateReviewAssignments.ts
nmadruga: Hmm I'll see if I can do this - was just trying to get this working properly first. Now it's tricky to test on master. But let's see

src/plugins/action_generate_review_assignment_records/src/generateReviewAssignments.ts
nmadruga: Yes

src/components/postgresConnect.ts
nmadruga: Done: e0f2c49

src/plugins/action_generate_review_assignment_records/src/generateReviewAssignments.ts
nmadruga: Nice!

src/plugins/action_generate_review_assignment_records/src/generateReviewAssignments.ts
nmadruga: Ah, ok. So this block needs to be separated so it knows to use this `stageNumber` and don't do the checking to check `getReviewStageAndLevel` is in the same stage as application or not. This can be simplified for sure. I just don't want to focus on that, to be able to do other stuff (it already took half day for me to find this solution and test it properly). I think the 2 blocks look similar but the main task is done by the `generateNextReviewAssignments` so that's fine by me.

src/plugins/action_generate_review_assignment_records/src/generateReviewAssignments.ts
nmadruga: Happy to see suggestions for how to move this block, but it looks ok to me like this for now

src/plugins/action_generate_review_assignment_records/src/generateReviewAssignments.ts
nmadruga: Isn't the DBConnect going to get this from the DB instead of from the `applicationData` on send through the action?

And if that's is correct, I think I need to revert the change, since this Should NOT be the same as getting from `applicationData` received from the input

nmadruga:
I need to get going. Will continue looking into this tomorrow. I have tested a lot this (on demo-2203)  and after querying the applicationId it was getting the correct stageNumber, but after changing it back to use the applicationData from the input I guess it will go back to fail. Please have a try if you have time today.

src/plugins/action_generate_review_assignment_records/src/generateReviewAssignments.ts
CarlosNZ: If you want the current (i.e. new) stage details, just get it from `cumulativeOutput`, no need to make additional queries:
`const { applicationId, reviewId, applicationData, cumulativeOutput } = input`

Then you get templateId from applicationData and stageNumber and stageId from cumulativeOutput. Wiki currently describes what fields to expect from what Actions: https://github.com/openmsupply/application-manager-server/wiki/List-of-Action-plugins

src/plugins/action_generate_review_assignment_records/src/generateReviewAssignments.ts
```diff
@@ -1,112 +1,105 @@
...
-          // TO-DO: allow STATUS to be configurable in template
-          status: nextReviewLevel === 1 ? AssignmentStatus.AVAILABLE : AssignmentStatus.SELF_ASSIGN,
+    // For level 1+ or next stages review assignment
+    // TODO - #440: Needs to lock same level reviews
+    // TODO: Unlock lower level review when consolidator has submitted "Change requested"
+    else {
+      const {
+        stageNumber: previousStage,
+        level: previousLevel,
+      } = await DBConnect.getReviewStageAndLevel(reviewId)
```
CarlosNZ, line 50: These values should be available in applicationData, so no need to do another DB query. (That is, the applicationData that comes in on input, not a new one)

CarlosNZ:
Okay, based on our IRL discussion, I've updated the issue for overall improvements to the Action system: #282 

So, for now, as long as the test cases work, this should be good to merge and I'll come and re-visit it again after #282.

Cheers.

src/plugins/action_generate_review_assignment_records/src/generateReviewAssignments.ts
nmadruga: Ok. I think yesterday was just a bad day for me - realised it can be simplified a lot by this suggestion. Thanks : ) Done: 1cc6108

CarlosNZ (approved):
Thanks, looks good. Will have a look at the whole test suite after the next round of merging and see what needs fixing.

PR #281, opened 2021-03-24
#476 - Fix question assignments on existing application review

nmadruga:
Fixes to show progress on serial 12345 and first name for user


andreievg (approved):
looks good to me

PR #284, opened 2021-03-25
Add template_stage_review_level table

CarlosNZ:
Implement #495 and added review levels to templates

Not thoroughly tested yet, but I've seen no immediate problems.


database/buildSchema/26A_review_assignment.sql
```diff
@@ -15,6 +15,7 @@ CREATE TABLE public.review_assignment (
 	trigger public.trigger,
 	time_created timestamptz default current_timestamp,
 	level integer,
+-- 	level_id integer references public.template_stage_review_level(id),
```
CarlosNZ, line 18: This is very bizarre. I've left it commented out for now, but if I try and make this foreign key reference, it creates a GraphQL schema problem, and the `templateActionsUsingId` and `templatePermissionsUsingId` create nodes are no longer available and the create template mutations fail. No idea why, since this shouldn't affect those nodes at all. Maybe it creates a circular reference somewhere?

Anyway, I don't think we really need it here. We only need to know the level number most of the time, a reference to the original record is not really necessary.

nmadruga: I think it might be conflicting with the field above `level` because of the types which get generated by GraphQL. Can you try renaming the one above to `level_number` or just removing it?

database/insertData/03_template_E_drugRegistration.js
```diff
@@ -556,11 +566,11 @@ exports.queries = [
...
               { permissionNameId: 10100 }
               # Review Drug Registration Stage 1
-              { permissionNameId: 10300, restrictions: {canSelfAssign: true}, level: 1, stageNumber: 1 }
+              { permissionNameId: 10300, restrictions: {canSelfAssign: true}, levelNumber: 1, stageNumber: 1 }
               # Assign Drug Registration Stage 2
-              { permissionNameId: 10200, level: 1, stageNumber: 2 }
+              { permissionNameId: 10200, levelNumber: 1, stageNumber: 2 }
               # Review Drug Registration Stage 2 
-              { permissionNameId: 10400, level: 1, stageNumber: 2 }
+              { permissionNameId: 10400, levelNumber: 1, stageNumber: 2 }
```
nmadruga, line 573: Maybe should remove this while we don't have a 2nd level review for `stageNumber`  2 yet - I'll add this to be done as part of changes required to finish up the Drug Registration template.

nmadruga (changes requested):

database/buildSchema/26A_review_assignment.sql
CarlosNZ: Okay, I've renamed "level" to "level_number" and the level_id now seems to work fine. Thank.

However, changing that field name required quite a bit of updating to other tables/data/actions, so the number of files changed here is now quite a bit more. I think I got them all. I'm planning to thoroughly check all the Actions are working properly as part of the "create company/logo" PR that I've almost finished, so I'll pull these changes into there before publishing once this is approved.

b4f0da5a

database/insertData/03_template_E_drugRegistration.js
CarlosNZ: Okay, have commented out this line for now: e17d0baf

nmadruga (approved):
Wow... That's a lot of changes indeed. I had a quick look thought, will approve but not 100% sure if all is good. It seems ok, but let's keep an eye.

CarlosNZ:
> Wow... That's a lot of changes indeed. I had a quick look thought, will approve but not 100% sure if all is good. It seems ok, but let's keep an eye.

Yes, for sure! I will do thorough testing as part of https://github.com/openmsupply/application-manager-web-app/issues/583

Thanks!

CarlosNZ:
> Wow... That's a lot of changes indeed. I had a quick look thought, will approve but not 100% sure if all is good. It seems ok, but let's keep an eye.

FYI, instead of merging this straight to master, I'll merge into my branch for [Final tidy-ups](https://github.com/openmsupply/application-manager-web-app/issues/583) so I can thoroughly test all the Actions are still working properly before we merge it to master.

CarlosNZ:
I'm going to close this one as I've pulled this branch into #303, and it's getting a bit messy to resolve and double-check all the changes from Andrei's latest big plugin merge. I'll do all the conflict resolution in #303.

Issue #285, opened 2021-03-25
Lookup table: Create endpoints Import CSV data into lookup tables 


sworup:
Part of epic: https://github.com/openmsupply/application-manager-web-app/issues/55

sworup:
#294 and #285 PRs has been merged to master. Any issues outstanding is being handled wtih #328 PR.

PR #286, opened 2021-03-25
#285 Lookup Table: Create endpoints to import csv data into lookup tables

sworup:


andreievg (changes requested):
Hi Sworup,

Generally speaking code looks solid, there are some suggestion about renaming of things.

I like the idea of using a service and modal (although it deviates slightly from other components). Honestly thought I found it hard to follow the code, it seems the functionality is spreading too much between services and controllers. 
In my view services should encapsulate all of the logic it's trying to abstract (controller shouldn't be holding state i.e. `lookupTableDbStructure`), and also having two services with crossed functionality is confusing (i.e lookupTableStructure storing csv rows).

I think a single service and single modal would simplify things quite a bit (and use the benefits of simplicity added by abstraction via service). 

i.e. 
```JS
// importCsv
const lookupTableService = LookupTableService({tableName})
// updateCsv
const lookupTableService = LookupTableService({tableId})
// both
await parseStream(data.file, {
    headers: lookupTableService.parseCsvHeaders,
  })
// not needed
//   .on('headers', (_) => {
//     fieldMaps = lookupTableStructureService.csvFieldMap
//    })
    .on('data', lookupTableService.addRow)
    .on('end', async() => {
      const result = await lookupTableService.process()
      reply.send({ status: 'success', message: result })
    })
    .on('error', (error) => {
      reply.status(422).send({ status: 'error', name: error.name, message: error.message })
    })
}
```
`LookupTableService({tableName})`
* just store tableName in internal variable

`LookupTableService({tableId})` 
* query for tableName (from tableId) and fieldMap

`lookupTableService.parseCsvHeaders`
* same as now but also adds 'newColumns' to internal array of newColumns (based on fieldMap)

`lookupTableService.process()`
* add table if id is not provided (with just the serial id)
* alter table and add `newColumns`
* iterate over each row, check if id exists, if so update, otherwise insert


Lastly:

Not sure about your preference for naming variables with underscore, I can't quite see consistency and not sure why it's needed, it certaintly makes the code less readable. (i.e. in `LookupTableService`, `_`lookupTableModal, in `LookupTableStructureService`,  lookupTableStructureModel, with no underscore).

src/lookup-table/controllers/ImportCsvUpdate.controller.ts
```diff
@@ -0,0 +1,47 @@
...
+      .on('headers', (_) => {
+        lookupTableStructure.compareFieldMaps(
+          lookupTableDbStructure.fieldMap,
+          lookupTableStructure.csvFieldMap
+        )
+      })
+      .on('data', async (row: any) => {
+        stream.pause()
+        // TODO: Row ID validation, if row has id and that does not exist in database
+        await lookupTableStructure.addToCsvRows(row, lookupTableDbStructure)
```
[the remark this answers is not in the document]
sworup: > Not sure about your preference for naming variables with underscore, I can't quite see consistency and not sure why it's needed, it certaintly makes the code less readable. (i.e. in `LookupTableService`, `_`lookupTableModal, in `LookupTableStructureService`, lookupTableStructureModel, with no underscore).

@andreievg the underscore is to denote a private variable in the function. Yes, they can be used without underscore and still will be private if we do not return them, but I thought it would be a bit easier for the developers to understand that they are private.

sworup:
> Not sure about your preference for naming variables with underscore, I can't quite see consistency and not sure why it's needed, it certaintly makes the code less readable. (i.e. in `LookupTableService`, `_`lookupTableModal, in `LookupTableStructureService`, lookupTableStructureModel, with no underscore).

@andreievg the underscore is to denote a private variable in the function. Yes, they can be used without underscore and still will be private if we do not return them, but I thought it would be a bit easier for the developers to understand that they are private.

sworup:
@andreievg thank you for the feedback.

My initial plan was to build a single service for structure and table but I was wary that the single file would grow to be too big. So I split it but later realized that the services became tightly coupled increasing the complexity of the code. Also I did not like that I had to pass in tablename for each service method and I wanted to inject dependency (as you have suggested to pass in tablename and/or id in constructor), but I decided to complete the code for now, get things working, and refactor later. 

I will refactor the code, merge the two services, inject tablename/id and make other changes you requrested, and will report back soon.

andreievg:
@sworup 

> @andreievg the underscore is to denote a private variable in the function. Yes, they can be used without underscore and still will be private if we do not return them (or returned using getter and setter methods), but I thought it would be a bit easier for the developers to understand that they are private.

Sure, I saw that after the comment I made, but I also saw _ naming in LookupTableService without return, so I wasn't sure. I agree, to use alongside getters (when private variable name = getter name) it makes sense.

> My initial plan was to build a single service for structure and table but I was wary that the single file would grow to be too big. So I split it but later realized that the services became tightly coupled increasing the complexity of the code. 
Also I did not like that I had to pass in tablename for each service method and I wanted to inject dependency (as you have suggested to pass in tablename and/or id in constructor), but I decided to complete the code for now, get things working, and refactor later.

I see. Perhaps some logic can go into helper methods, like you are doing here, rather then seperate services: https://github.com/openmsupply/application-manager-server/pull/286/commits/d02909ef27ea28d37ce299dea3d57cde07a8b5a4

src/lookup-table/models/LookupTable.model.ts
```diff
@@ -0,0 +1,107 @@
...
+  }
+
+  const createRow = async ({ tableName, row }: { tableName: string; row: any }) => {
+    const text = `INSERT INTO lookup_table_${tableName}(${Object.keys(row)}) VALUES (
+          ${Object.keys(row)
+            .map((key, index) => {
+              return '$' + String(index + 1)
+            })
+            .join(', ')}) RETURNING id`
+    const result = await DBConnect.query({ text, values: [...Object.values(row)] })
```
andreievg, line 55: do we need to destructure Object.values here ?

src/lookup-table/models/LookupTable.model.ts
```diff
@@ -0,0 +1,107 @@
...
+        return `${key} = $${currentIndex}`
+      })
+      .filter(Boolean)
+      .join(', ')
+
+    const text = `UPDATE lookup_table_${tableName} SET
+    ${setText} WHERE id = $${primaryKeyIndex}`
+
+    const result = await DBConnect.query({ text, values: [...Object.values(row)] })
+    return true
```
andreievg, line 78: should this return similar to `createRow`

andreievg (changes requested):
Thanks Sworup, overall it looks great. I find it much easer to follow.

I've noticed that ID is required for import, and if ID exists it tried to update (i think previously you did a lookup query to decide if update or insert, but it was done while parsing csv, which I wasn't a fan of). 

To keep it simple and get the PR merged quickly can we do this:
* for import, fail validation if `ID` exists with message "ID is auto generated when importing new lookup table, please rename or remove this column"
* for update, has to have `ID` field (as it is now)
  * if `ID` is blank insert (as it is now)
  * Otherwise update, 'returning ID', if no rows are returned insert

I've resolved most of the comments made in the first PR, but there are still a couple there. 

Thanks

src/lookup-table/controllers/ImportCsvUpdate.controller.ts
```diff
@@ -0,0 +1,27 @@
...
+  const lookupTableId = Number(request.params.lookupTableId)
+  const data = await request.file()
+
+  const lookupTableService = LookupTableService({ tableId: lookupTableId })
+
+  let stream = await parseStream(data.file, {
+    headers: lookupTableService.parseCsvHeaders,
+  })
+    .on('data', async (row: any) => {
+      stream.pause()
```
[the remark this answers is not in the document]
sworup: @andreievg actually when I did not use pause earlier, I found that since the calls are housed inside event listeners. In some cases, the rows were getting pushed onto the array out of order. I used `pause` and `resume` calls to be safe.

src/lookup-table/models/LookupTable.model.ts
sworup: Ah, it seems like we do not as an array is returned from Object.values. I will make the changes.

src/lookup-table/controllers/ImportCsvUpdate.controller.ts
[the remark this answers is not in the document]
andreievg: Was this before refactor ? In the previous version there was a DB call (which would cause order mismatch)
I dont' think order should be important, user may decide to sort csv before re-importing as update

src/lookup-table/services/LookupTable.service.ts
```diff
@@ -89,6 +145,25 @@ const LookupTableService = (structure: LookupTableServiceProps) => {
     fieldMaps = [...dbFieldMap, ...fieldMaps].filter(
       ((set) => (obj: any) => (set.has(obj.label) ? false : set.add(obj.label)))(new Set())
     )
+
+    return {
+      onlyInDb: dbFieldMap.filter(_comparer(fieldMaps)),
```
andreievg, line 150: I've suggested a simpler way for fieldMaps to be compared (https://github.com/openmsupply/application-manager-server/pull/286#discussion_r612046890)< iI really want to improve readability of our code vs code size.

src/lookup-table/services/LookupTable.service.ts
```diff
@@ -89,6 +145,25 @@ const LookupTableService = (structure: LookupTableServiceProps) => {
...
+      onlyInCsv: fieldMaps.filter(_comparer(dbFieldMap)),
+      inBoth: dbFieldMap.filter((current: any) =>
+        fieldMaps.filter((csvMap: any) => csvMap.label === current.label)
```
andreievg, line 153: I am not sure if it's useful to show user what's been changed without confirming with them first. This seems like a new spec, would be good if you can message me if there is a change like that first, so we can brainstorm how it works.

What should happen, imo, is before we do the update, we would have another end point that checks the diff and tells the user what will be changed (but that's out of scope for now))

andreievg (changes requested):
Hi Sworup, I am really confused by the last commit, I think this sort of changes would require us agreeing on the solution, see comment.

This PR is very close to being good to go, just a few comments to be dealt with as per previous reviews

src/lookup-table/models/LookupTable.model.ts
```diff
@@ -47,69 +63,107 @@ const LookupTableModel = () => {
...
+  }: {
+    tableName: string
+    row: any
+  }): Promise<boolean> => {
+    try {
+      let primaryKeyIndex = 0
+      const setText = Object.keys(row)
+        .map((key, index) => {
+          const currentIndex = index + 1
+          if (key === 'id') {
```
andreievg, line 128: why is this needed ? you can have id in where and set clause
```
const keys = Object.keys(row)
const primaryKey = keys.indexOf('id') + 1
const setText = keys.map((key, index) => `key=${index+1}`)
```

src/lookup-table/types.ts
```diff
@@ -1,14 +1,24 @@
...
+type LookupTableStructureType = {
+  id?: number
+  name: string
   label?: string
   fieldMap: FieldMapType[]
 }
 
+// Removes 'optional' attributes from a type's properties
+type Concrete<Type> = {
+  [Property in keyof Type]-?: Type[Property]
```
andreievg, line 10: What is this for ? I mean what purpose does it server in terms of checking that code is strongly typed ? Why not have two interfaces (one inheriting another ?, one without IDs one with IDs)

andreievg (changes requested):

andreievg:
@sworup I've updated from master and going to merge this as it's needed for demo. Also PR change request seems to be ignored (and other commits added), so would like a more strict way of dealing with them. By the way for the future can you put commit hash in reply to the change request (so it's easy to see what comment have been dealt with or not).
https://github.com/openmsupply/application-manager-server/issues/328

andreievg (approved):

sworup:
> Sure, I saw that after the comment I made, but I also saw _ naming in LookupTableService without return, so I wasn't sure. I agree, to use alongside getters (when private variable name = getter name) it makes sense.

@andreievg resolved with 0507d9b

src/lookup-table/models/LookupTable.model.ts
sworup: Resolved with 48af4f5

src/lookup-table/models/LookupTable.model.ts
sworup: Resolved with 48af4f5

src/lookup-table/controllers/ImportCsvUpdate.controller.ts
```diff
@@ -7,17 +7,21 @@ const ImportCsvUpdateController = async (request: any, reply: any) => {
...
   })
     .on('data', async (row: any) => {
-      stream.pause()
       await lookupTableService.addRow(row)
-      stream.resume()
     })
     .on('end', async (rowCount: any) => {
-      await lookupTableService.updateTable()
-      reply.send({ status: 'success', message: 'Lookup table successfully updated' })
+      await lookupTableService
```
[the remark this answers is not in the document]
sworup: Resolved with c7c0bc2

src/lookup-table/controllers/ImportCsv.controller.ts
```diff
@@ -13,8 +13,14 @@ const ImportCsvController = async (request: any, reply: any) => {
       lookupTableService.addRow(row)
     })
     .on('end', async (rowCount: any) => {
-      await lookupTableService.createTable()
-      reply.send({ status: 'success', message: rowCount })
+      await lookupTableService
+        .createTable()
+        .catch((error: any) =>
```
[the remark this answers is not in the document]
sworup: Resolved with c7c0bc2

src/lookup-table/models/LookupTable.model.ts
```diff
@@ -35,6 +35,24 @@ const LookupTableModel = () => {
     }
   }
 
+  const countStructureRowsByTableName = async (lookupTableName: string) => {
```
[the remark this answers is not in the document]
sworup: That worked @andreievg 
Screenshot 2021-04-29 at 12 28 30 pm
Resolved with 817baa7

src/lookup-table/services/LookupTable.service.ts
```diff
@@ -21,6 +21,18 @@ const LookupTableService = (structure: LookupTableServiceProps) => {
   const createTable = async () => {
     tableName = toSnakeCase(tableNameLabel)
 
+    const lookupTableNameValidator: ILookupTableNameValidator = new LookupTableNameValidator({
```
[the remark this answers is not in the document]
sworup: Resolved with 817baa7

src/lookup-table/models/LookupTable.model.ts
sworup: @andreievg I also wanted to avoid updating the primary key but the code you added simplified things a lot. Implementing.

sworup: Resolved with 0c77bb6

src/lookup-table/types.ts
sworup: I wanted to set all the properties of the type required,  and I did not want to create a new type while doing it. 
I just realized that there is a `Required<T>` utility type provided by typescript off the box. I am going to use that and remove the `Concrete<T>` type I created.

sworup: Resolved with 2f70fb4

src/lookup-table/services/LookupTableStructure.service.ts
```diff
@@ -0,0 +1,146 @@
...
+
+      _csvFieldMap.push(fieldMap)
+
+      return fieldMap.fieldname
+    })
+  }
+
+  const compareFieldMaps = (dbFieldMaps: any, csvFieldMaps: any) => {
+    _finalFieldMaps = [...dbFieldMaps, ...csvFieldMaps].filter(
+      ((set) => (obj: any) => (set.has(obj.label) ? false : set.add(obj.label)))(new Set())
```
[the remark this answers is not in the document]
sworup: Resolved with cc04c69

src/lookup-table/services/LookupTableStructure.service.ts
```diff
@@ -0,0 +1,146 @@
...
+        dataType: fieldName == 'id' ? 'serial PRIMARY KEY' : 'varchar',
+      }
+
+      _csvFieldMap.push(fieldMap)
+
+      return fieldMap.fieldname
+    })
+  }
+
+  const compareFieldMaps = (dbFieldMaps: any, csvFieldMaps: any) => {
```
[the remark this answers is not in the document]
sworup: Resolved with cc04c69

src/lookup-table/services/LookupTable.service.ts
sworup: I used the same method but added sticter types that will give code completion and extracted the comparer method to a helper function.

Resolved with cc04c69

src/lookup-table/services/LookupTable.service.ts
sworup: I am keeping the method since as of your description we might have a use for this later.

PR #287, opened 2021-03-30
Snapshot

andreievg:
TODO - this requires an issue, and requires documentation. Wanted to make a PR to show the functionality (and if there is an immediate need to use it we can marge and I can do docs and issue soon after) 

### The idea

Able to take data snapshot and re-load.

### How to use

`yarn take_snapshot` -> can pass through snapshot name (otherwise would use whatever the current snapshot is set at in `./databse/snapshots.currentSnapshot.txt` (defaults to current)
`yarn use_snapshot` -> similar to the above regarding snapshot name

### Changes

* re-structure initialise_databse.sh (so that parts of it can be used in useSnapshot)
* add graphQLdefinition: 
```
[
 {
   table: 'tableName',
   generatedColumns: ['column'],
   generateFileName: (record) => return 'fileNameOfRecord',
   skip: true/false/undefined
 }
]
```
* added takeSnapshot and useSnapshot scripts
* added resetSerial.sql (to reset serial number after data insert)


database/graphQLdefinitions.js
```diff
@@ -0,0 +1,203 @@
...
+    skip: true,
+  },
+  {
+    table: 'notification',
+    skip: true,
+  },
+  {
+    table: 'notification',
+    skip: true,
+  },
```
CarlosNZ, line 195: Duplicate!

package.json
```diff
@@ -63,6 +65,7 @@
     "pg": "^8.4.1",
     "postgraphile": "^4.8.0",
     "postgraphile-plugin-connection-filter": "^2.0.0",
+    "prettier": "^2.2.1",
```
CarlosNZ, line 68: Hmmm, don't we already have Prettier installed globally? We're definitely already using it in our project (we have a .prettierrc file). Is this going to cause a problem, or do we need it?
Or should this local version be the one have active for formatting?

CarlosNZ (changes requested):
This is really cool, and pretty complicated! It'd be good to merge this ASAP so we can use it for posting snapshots with our PRs -- that'd be very handy!

I have a couple of suggestions, though nothing major:

- I'm getting this error at the end of "use_snapshot":  
`UnhandledPromiseRejectionWarning: Error: ENOENT: no such file or directory, scandir './database/snapshots/test/reviewAssignmentAssignerJoin'`  
Not sure why, but can confirm there is no "reviewAssignmentAssignerJoin" folder in the snapshot folder. Perhaps because the table is empty it doesn't create the folder when saving the snapshot?
- Would be good if we didn't see every mutation go whizzing by in full in the console, cos it quickly fills the buffer and anything near the top is lost, not to mention the summary logs aren't really readable. I managed to avoid this (somehow, can't remember now) when running the insertData.js script, so should be do-able.
- Bonus feature if time: it'd be good if a .zip file of each snapshot was automatically created for quick-and-easy uploading to github.

CarlosNZ:
>     * Would be good if we didn't see every mutation go whizzing by in full in the console, cos it quickly fills the buffer and anything near the top is lost, not to mention the summary logs aren't really readable. I managed to avoid this (somehow, can't remember now) when running the insertData.js script, so should be do-able.

Ah, you just need to remove (or comment out) the `console.log(content)` -- line 45 of useSnapshot.js. Much nicer :)

CarlosNZ:
One more comment -- should probably put the "snapshots" folder into .gitignore. (Can add a line to create an empty snapshots folder locally in `initialise_database.sh`)

andreievg:
>One more comment -- should probably put the "snapshots" folder into .gitignore. (Can add a line to create an empty snapshots folder locally in initialise_database.sh)

2e61a93411fb9a34229862a44264e464c3a241da

database/graphQLdefinitions.js
andreievg: 7894dd1a99390980b65934a3e9dd7664f0e73309

package.json
andreievg: I think it comes with extension, I've changed to use npx:

6f9de176bb56d9e96e84dc132a679a3a6a08f9d8

CarlosNZ:
> > One more comment -- should probably put the "snapshots" folder into .gitignore. (Can add a line to create an empty snapshots folder locally in initialise_database.sh)
> 
> [2e61a93](https://github.com/openmsupply/application-manager-server/commit/2e61a93411fb9a34229862a44264e464c3a241da)

Cool, thanks for the changes -- although this comment doesn't actually address the .gitignore issue. Don't you think we shouldn't be saving snapshots to the repo, and just post them directly in PRs if we want to share them?

andreievg:
> Cool, thanks for the changes -- although this commit doesn't actually address the .gitignore issue. Don't you think we shouldn't be saving snapshots to the repo, and just post them directly in PRs if we want to share them?

Oops that was meant to address the console log comment

> One more comment -- should probably put the "snapshots" folder into .gitignore. (Can add a line to create an empty snapshots folder locally in initialise_database.sh)

I got tripped a couple of times with huge diff showing in github desktop, accidentally discarded some files that were not snapshot (while discarding snapshots), thankfully git discards go to 'trash' (recycle bin).

Anyways, the idea for having snapshots showing in git, was to:
* show diff after certain operation (I thought i was quite useful to test etc..)
* commit snapshot alongside the PR

I still like above reason, and thought of this option (one file snapshot, very easy to see full diff now i think, and should be easier to either zip or commit):
ae5f7956514efe477b59390b26c683c784393aad

The idea is to give an option to either committing (just to local branch for diff between operation, or for sharing in test branch), or zipping (commit doesn't really address zipping or .gitignore, but makes git diff manageable in UI git clients). I would personally find it easier to vie a branch (or even commit to a PR if snapshot is re-usable). I am kind of stuck on the idea of seeing a 'diff' of an operation (found it pretty useful), and it seems the easiest way to do this is to commit a snapshot to say a temporary branch then take another snapshot right after the operation

CarlosNZ (approved):
Awesome, will be very handy!

Just do that little .gitignore change we discussed and then merge it I reckon :)

CarlosNZ (approved):

andreievg:
> Awesome, will be very handy!

Thanks

> Just do that little .gitignore change we discussed and then merge it I reckon :)

I've also added documentation

PR #289, opened 2021-03-31
#488F "Other" option for radio selector

CarlosNZ:
Small change to template to demonstrate the "Other" option in the RadioButton element plugin. See [front-end PR #554](https://github.com/openmsupply/application-manager-web-app/pull/554)


database/insertData/03_template_A_featureShowcase.js
```diff
@@ -456,6 +456,8 @@ exports.queries = [
                         label: "This Radio button group has no default"
                         options: ["Option A", "Option B", "Option C"]
                         # Testing no default
+                        hasOther: true
```
nmadruga, line 459: Oh! So we have this `hasOther` I think I made a comment to remove this, so please ignore my comment - I think you might want to move that documentation change to this PR?

nmadruga (approved):

PR #290, opened 2021-03-31
#479F Long text plugin

CarlosNZ:
Update Feature Showcase template to include a couple of LongText element plugins. To use with [front-end PR #555](https://github.com/openmsupply/application-manager-web-app/pull/555)


nmadruga (approved):
LGTM

I was wondering if we should start removing those `id` fields from each question now that we don't use that anymore - since we have Assignment in place. But you decide if it's worth or not doing that (for Feature showcase only) as part of this PR.

CarlosNZ:
> I was wondering if we should start removing those `id` fields from each question now that we don't use that anymore - since we have Assignment in place. But you decide if it's worth or not doing that (for Feature showcase only) as part of this PR.

Yeah, we should, but I'd rather do it in a more systematic way than trying to squeeze it in here, cos there might be unintended consequences that make it more complicated than we think.

nmadruga:
> Yeah, we should, but I'd rather do it in a more systematic way than trying to squeeze it in here, cos there might be unintended consequences that make it more complicated than we think.

Sure. No problem. Then probably just need to add a issue related to the current Epic - so we don't forget about this?

CarlosNZ:
> Then probably just need to add a issue related to the current Epic - so we don't forget about this?

Okay, made an issue for "final tidy-ups": https://github.com/openmsupply/application-manager-web-app/issues/583

Feel free to add to it if you think of anything

nmadruga:
Thanks!

nmadruga (approved):

PR #292, opened 2021-04-01
#458F applicationData for front-end expressions

CarlosNZ:
Back-end changes for [front-end PR #557](https://github.com/openmsupply/application-manager-web-app/pull/557)

This just updates FeatureShowcase template and adds a (basic, incomplete) "JoinCompany" template -- they now use "applicationData" expressions rather than "currentUser" in question elements, so that the dynamic value will show the correct (i.e. applicant) user during reviews, etc.

Don't look at the joinCompany template too carefully, as I'll be pushing a new "proper" version out in the next few days.


nmadruga:
> Don't look at the joinCompany template too carefully, as I'll be pushing a new "proper" version out in the next few days. 

👍

nmadruga (approved):
LGTM

PR #295, opened 2021-04-08
#294 Lookup table: Create endpoints to Export database table's data into CSV 

sworup:


src/lookup-table/models/LookupTable.model.ts
```diff
@@ -2,6 +2,12 @@ import { LookupTableStructurePropType } from '../types'
 import DBConnect from '../../components/databaseConnect'
 
 const LookupTableModel = () => {
+  const getAll = async ({ tableName }: any) => {
+    const text = `SELECT * FROM lookup_table_${tableName}`
```
andreievg, line 6: I think it's neater to use fieldMap here instead of in controller:
```JS
fields = fieldMap.map(({label, fieldname}) => `"${fielname}" as "${label}"`).join(',')
`SELECT ${fields} FROM lookup_table_${tableName}`
```

andreievg (changes requested):
Just a small change (after it's update from parent branch)

I think we need a new issue for this (if you agree):
Extend field map to include "originalFieldname" ?

andreievg (approved):

src/lookup-table/models/LookupTable.model.ts
sworup: This has already been done by @andreievg and is in `master`

PR #302, opened 2021-04-13
#311F Add assigner_list fields to application_list_view

nmadruga:
### Front-end branch
TODO

### Changes
- Copying changes made for `demo-22-03` regarding listing assigner actions


CarlosNZ (changes requested):
A bit out of scope, but it's quite an important quick fix, if you could squeeze it in here:
Was just talking to Andrei -- I wasn't up to date with the "application_list" being a function rather than a View. I noticed that there is one Action that makes a call to "application_list" that no longer works as a result.

In order to fix, can you quickly do the following:
- Add a default (0) to the application_list function (see inline comment)
- in postgresConnect.ts, line 694 (isFullyAssignedLevel1 method), change `FROM application_list` to `FROM application_list()`.

Otherwise, looks good, although it'd be good to have some data in there so we can see the fields being correctly populated. (Front end not currently working with this, so can't create new applications)

nmadruga:
> A bit out of scope, but it's quite an important quick fix, if you could squeeze it in here:
> Was just talking to Andrei -- I wasn't up to date with the "application_list" being a function rather than a View. I noticed that there is one Action that makes a call to "application_list" that no longer works as a result.
> 
> In order to fix, can you quickly do the following:
> 
> * Add a default (0) to the application_list function (see inline comment)
> * in postgresConnect.ts, line 694 (isFullyAssignedLevel1 method), change `FROM application_list` to `FROM application_list()`.
> 
> Otherwise, looks good, although it'd be good to have some data in there so we can see the fields being correctly populated. (Front end not currently working with this, so can't create new applications)

Done : )

database/insertData/04_users.js
```diff
@@ -34,12 +28,6 @@ exports.queries = [
           passwordHash: "$2a$10$3Z1cXVI.GzE9F2QYePzbMOg5CGtf6VnNKRiaiRGkzlBXJ0aiMN4JG"
           username: "carl"
           firstName: "Carl", lastName: "Smith"
-          permissionJoinsUsingId: { 
-            create: [
-              # Self assign
-              { permissionNameId: 10500 }
-            ]
-          }
```
CarlosNZ, line 42: Hey, just FYI -- I have gone an done a fairly big update to all these in #303, so maybe avoid too many changes to the templates that aren't directly related to the PR's functionality, otherwise it's going to be messy to fix.

nmadruga: Yes, that's a good idea. I just removed this because it was giving a mis-leading result to the PR on the front-end. But that's the only changes I've done to templates.

CarlosNZ (approved):
Looks good -- haven't been able to fully test the logic for the functions, but we will put it through its paces properly once we add more data.

CarlosNZ (approved):
Looks good -- haven't been able to fully test the logic for the functions, but we will put it through its paces properly once we add more data.

PR #305, opened 2021-04-14
#609F-fix-application-start-bug

andreievg:
Fixes front end: https://github.com/openmsupply/application-manager-web-app/issues/609

As per comment: https://github.com/openmsupply/application-manager-web-app/issues/609#issuecomment-819158126


CarlosNZ (approved):
LGTM! Thanks.
(I think we must of assumed that because there was only one Action for "onApplicationCreate" that a sequence wouldn't be necessary. But yeah, it is in this case.)

Issue #306, opened 2021-04-14
Prepare back end build


andreievg:
Prepare back end build for dockerisation

* plugins can be built individually
* core and plugins can be built in combined way
* can launch build with node `node ./build/src/server.js`
* 
From epic: https://github.com/openmsupply/application-manager-web-app/issues/543

Issue #282, opened 2021-03-24
Actions -- suggested improvements and tweaks


CarlosNZ:
1. plugin.json file should have "optionalParameters" as well as "requiredParameters", ~~and the triggersAndActions module should check that all required parameters are supplied~~.
2. Change shape of All Action input parameters to be more consistent and clear, as @andreievg suggested:  
  A single "input" parameter, shaped as follows:  
  ```
  {
    parameters: <required and optional parameters>
    applicationData: ..,
    outputCumulative:...
    DBConnect:...,
  }
```  
And the first line of every plugin can therefore be like:
```
async function ({parameters,  applicationData, outputCumulative, DBConnect}) 
```
So it's clear what's what (and you'd only need to destructure values you actually need to use, so maybe not outputCumulative a lot of the time)

3. After discussion with @nmadruga, we thought it would be best if the main triggersAndActions module re-runs and updates "applicationData" between each Action. That way every Action can rely on having up-to-date applicationData passed into it and won't need to make its own version of the same basic query.
4. Bearing in mind what info is available in "applicationData", simplify required parameters to only require values that won't be available in applicationData. I think (on a case by case basic), Actions could have "override" parameters (which would be "optional") -- i.e. normally it gets the value from applicationData, but if something is specified (reviewId, say), that would override the applicationData value). I like Andrei's suggestion about having parameters like `isReview`, but it gets the actual review value from applicationData
5. This one's a "maybe" -- it might not be so straightforward: currently all the conditions in a sequence are evaluated at the beginning before any Actions execute, so the condition evaluator doesn't get to see the "Output" object. Might see if it's not too big a change to evaluate each condition in sequence at the start of each Action. This would make some of the condition queries much simpler (e.g. Checking that an application has been "Approved" would just be one output field from "ChangeOutcome" rather than the gnarly thing that's currently there)
6. Update templates to reflect the above changes.
7. Actions that need review levels should get them from the template definition rather than deducing them like they do now.

CarlosNZ:
Have started doing this in branch `#282-action-improvements` in between other jobs. See UPDATES on original issue for progress. Will publish a PR soon-ish.

PR #307, opened 2021-04-14
#306 prepare back end build

andreievg:
Fixes: #306 

There is quite a bit of diff, but the bulk of it is in plugin changes (which are quite minor and repetitive).
I've tried to minimise the diff with `git mv`, but still some file were not recognised as moves and as deletes : /

### Changes

* Relocated plugins and images to root
* Remove dependencies from plugins, remove jest and tsconfig from plugins
* Package json in each plugin to use scripts in `utils/pluginScripts` (using root jest and tsc at root level)
* Relocate plugin.json to plugin root
* Remove reliance on function name for plugins (each plugin should have 'action' exported via index.js)
* Added post install script to auto yarn install all plugins when global yarn install is called
* Updated docs
* Global yarn build will build everything and move built plugins to overall build folder
* Updates to use new folder structure, including using `getAppEntryPointDir()` vs `getAppRootDir()`, maybe I could have used existing entry point function, but had some issues with it in built environment, happy to change back if there is a work around. (not super happy with the way I've changed it tbh, but it works)
 
##### utils/pluginScripts
* `buildPlugin.js` -> installs and build a plugin, saving output to `build/{pluginName}` (within plugin folder)
  * adds temporary tsconfig.json to build just the plugin
  * builds in temporary folder and relocates build content to `build/{pluginName}` (within plugin folder)
  * cleans up
* `buildPluginFromPluginFolder.js` -> checks working directory to identify plugin name and uses `buildPlugins.js` to build plugin
* `testPluginFromPluginFolder.js` -> checks working directory to identify plugin name and runs `yarn test {pluginname}`
* `yarnInstallPlugins.js` -> runs `yarn install` with working directory being plugin folder (for all plugins asynchronously), used in `postinstall` of global package.json

### Testing

I did basic test with both build version (after `yarn build` -> `node build/src/server.js`) and normal dev version.

Was watching console output for problems and went through making a few applications (making sure to check at least one file upload, making sure thumbnail shows correctly)

Also made sure that correct console log is shown for trim responses when application is submitted (to make sure not errors are thrown for deep-equal package, as it's dependency specific to that plugin)


CarlosNZ:
One quick comment: most of the plugin folders have all their dependencies in their folder root. A couple of them (e.g. update_review_assignments) have a `node_modules` folder, which seems much tidier.

But shouldn't `node_modules` be in .gitignore anyway? (Maybe each plugin needs its own .gitignore file?)

documentation/Action-plugin-specification.md
```diff
@@ -8,26 +8,28 @@ See [here](./List-of-Action-plugins.md) for a list of currently implemented Acti
...
-- `/src/<plugin-name>.ts` -- Typescript source code
-- `/src/<plugin-name>.js` -- Javascript code compiled from Typescript. This is what is actually loaded and run in the built server app.
+- `plugin.json` -- contains metadata for the plugin. The system reads this file (at startup) to load all the relevant information about the plugin into the database, so make sure it is accurate and up-to-date. Further details [below](#plugin).
+- `/src/index.ts` -- A file that export the plugin entrypoint method as `action`
+- `/src/*.ts` -- All other plugin files
 
 ## Source code file
 
-The main source code for the plugin is simply an exported function that the server app dynamically loads via `require` at run-time. Calling the function exported by each action plugin, sequentially, is the job of the **Action module**.
+Plugin can have any number of source files located in the source folder, there must be an exported entrypoint method in `index.ts`, this method will be dynamically loaded via `require` at run-time
```
CarlosNZ, line 23:
```suggestion
Plugin can have any number of source files located in the source folder, but there must be an exported entrypoint method in `index.ts`. This method will be dynamically loaded via `require` at run-time.
```

documentation/Action-plugin-specification.md
```diff
@@ -8,26 +8,28 @@ See [here](./List-of-Action-plugins.md) for a list of currently implemented Acti
...
+Calling the function exported by each action plugin, sequentially, is the job of the **Action module**.
 
-Plugin code should follow this basic structure:
+Plugin entrypoint method should follow this basic structure:
+
+TO_DO describe `parameters`
```
CarlosNZ, line 28: Yeah, I'll do this as part of #282

documentation/Action-plugin-specification.md
```diff
@@ -42,6 +44,8 @@ module.exports['consoleLog'] = function (parameters: any) {
     }
   }
 }
+// src/index.ts
+export {default as action} from './consoleLog'
```
CarlosNZ, line 48: We should probably re-write this more like a "template", with just some generic placeholder inside the `try`method, etc. I'll do this with #282.

documentation/Action-plugin-specification.md
```diff
@@ -110,28 +110,42 @@ The `action_plugin` table is the primary record of what actions are available to
...
-- `function_name`
+- `path` (determined by the folder the plugin was found in)
 - `required_parameters`
 
 <a name="development"></a>
 
 ## Development and build process
 
-Plugins are intended to be stand-alone units, in that a user should be able to simply copy a plugin folder to another system and have it work straight away. (Eventually, we envisage that plugins could be imported directly via the front-end UI.) To that end, each plugin is developed as its own package, with its own `package.json` and `node_modules` folder, as well as Typescript and testing configurations.
+Plugins are intended to be stand-alone units, in that a user should be able to simply `yarn build` and copy content inside `build` folder to another system and have it work straight away. (Eventually, we envisage that plugins could be imported directly via the front-end UI.) To that end, each plugin is developed as its own package, with its own `package.json` and `node_modules` folder. Root typescript and testing configurations are used by default but can be overwritten
```
CarlosNZ, line 120: It doesn't seem to work like this. I just tried running `yarn build` inside a plugin folder and it failed. Also -- when I just run `yarn install` within plugin folder, it says "already up-to-date", but there's no node_modules folder in there.

package.json
```diff
@@ -33,17 +33,18 @@
...
-    "build": "rimraf ./build && tsc && rsync -a --exclude \"*/node_modules\" ./src/plugins ./build",
+    "build": "./utils/build_all.sh",
     "test": "jest --runInBand",
     "pg": "exec ./utils/postgraphile_launch.sh",
     "pg_permissions": "exec ./utils/postgraphile_launch_permissions.sh",
     "dev_pg": "concurrently --kill-others-on-fail \"yarn dev\"  \"yarn pg\"",
     "dev_pg_permissions": "concurrently --kill-others-on-fail \"yarn dev\"  \"yarn pg_permissions\"",
-    "database_init": "./database/initialise_database.sh && ./database/insert_data.sh && ./database/post_data_insert.sh",
-    "build_plugins": "exec ./utils/build_plugins.sh",
+    "database_init": "exec ./database/initialise_database.sh",
```
CarlosNZ, line 42: Was this an intentional re-version of the database_init script? Don't you need that for the snapshots?

package.json
```diff
@@ -33,17 +33,18 @@
...
+    "build_plugins": "exec ./utils/pluginScripts/build_plugins.sh",
```
CarlosNZ, line 43: This file doesn't exist. Did you perhaps mean `build_all.sh`?

documentation/Action-plugin-specification.md
CarlosNZ: Okay, it's working now (had to yarn install at top level). But build all still not working (see comment below -- wrong filename I think)

CarlosNZ (changes requested):
Looking good, documentation is helpful, thanks. Just a couple of little comments in the code

Tried `yarn build` on the whole app and it seems to work well.

One thing not working properly for me is the tests. Tried to run `yarn test` from within a plugin folder and I get:
```
 FAIL  plugins/action_trim_responses/build/action_trim_responses/src/trimResponses.test.js
  ● Test suite failed to run
```

One suggestion: you've got a "build_plugins" script for building just the plugins. Wonder if we need an additional flag on the main "build" script to tell it not to re-build plugins if they're already built -- would save quite a bit of time in the build. Maybe `yarn build -noplugins` or something? I dunno, probably not a high priority for now.

package.json
andreievg: That's my clumsy conflict resolution during merge from master:
48190f2a4a08ffcf254b7b08afec66c6f4297c4a, thanks for noticing

package.json
andreievg: oops, thanks for noticing:

396a243682d04d1cef5d6209c85397ad70ed808e

documentation/Action-plugin-specification.md
andreievg: > It doesn't seem to work like this. I just tried running yarn build inside a plugin folder and it failed. Also -- when I just run yarn install within plugin folder, it says "already up-to-date", but there's no node_modules folder in there.

Hmm, if root level node_modules are inline with root level package.json, should be able to `yarn build` from within plugin folder, I've just checked and it seems to work.

CarlosNZ: Yeah, it works now (see prev comment).

andreievg:
> One thing not working properly for me is the tests. Tried to run yarn test from within a plugin folder and I get:

I think i stuffed up the tests completely. Will fix shortly

andreievg:
> One thing not working properly for me is the tests. Tried to run yarn test from within a plugin folder and I get:

I did some fixes:

* make sure serial number update is only done for snapshots
* change tests to use correct import
* updated ID on one of the tests (createUser)
* jest config to ignore build folder (we would need another jest config for build testing)

It looks like there are quite a lot of id mismatches now (trim action test cant' even insert data for the test), I think we need to make test more resilient to data changes, I've suggested something here: https://github.com/openmsupply/application-manager-server/blob/master/documentation/Snapshots.md#todo, I'll make and issue over the weekend, but the idea is to load snapshot (which is stored for each test or globally) before running test

andreievg:
54dd302210c522438cfc56a4755908b2ed5dda05

CarlosNZ:
Cool, tests are working now (by which I mean they run -- not necessarily pass lol).

CarlosNZ (commented):
Cool, looking good now. My only concern is reconciling this with the few small changes to the action plugins that we've done over the last few days that won't be in here. How should we check?

andreievg:
>Cool, looking good now. My only concern is reconciling this with the few small changes to the action plugins that we've done over the last few days that won't be in here. How should we check?

In github desktop, I selected this branch, then went to 'commits', found the first commit of this branch. used commit just before this branch started and compared master with it: 

https://github.com/openmsupply/application-manager-server/compare/bc3d41215a80bff953fa2e027bdc99f299b7721f...master. Couldn't see any src/plugins in the file changed list

CarlosNZ:
> [bc3d412...master](https://github.com/openmsupply/application-manager-server/compare/bc3d41215a80bff953fa2e027bdc99f299b7721f...master). Couldn't see any src/plugins in the file changed list

There must be though, cos at the very least you've changed the "wrapper" for each and the export setting.

andreievg:
> There must be though, cos at the very least you've changed the "wrapper" for each and the export setting.

I was comparing master with 'master' but before this branch was started, to see if we've done any changes to the plugins, doesn't look like it

CarlosNZ:
Ah, okay cool

CarlosNZ (approved):

Issue #308, opened 2021-04-14
Rename permission and template names to use org instead of company


nmadruga:
As per [comment](https://github.com/openmsupply/application-manager-server/pull/301#discussion_r612861098) on PR #293:

- Rename file `03_template_B_companyRegistration.js` to `03_template_B_orgRegistration.js`
- Rename file `03_template_F_companyJoin.js` to `03_template_F_joinOrganisation.js` (maybe the code too - and where it's used) 
- Rename permissions in file `02_permission_names` (and where it's used):
  - `applyCompanyRego` to `applyOrgRego`
  - `reviewCompanyRego` to `reviewOrgRego`

CarlosNZ:
Done just before I merged #303.

CarlosNZ:
Sorry, I didn't actually rename the permissions, just the files. Re-opening...

PR #311, opened 2021-04-19
#308 Replace "company" with "org"/"organisation"

CarlosNZ:
Replaced all permission names from "company" to "org" or "organisation", and some of the text in templates too.


nmadruga (changes requested):
Bravo!

I have done a global search for `company` and there are a few documentation point to it, which I think it's not too bad. Although some tests are point to company too, which I think is the cause of this:

Screen Shot 2021-04-22 at 12 32 44 PM

CarlosNZ:
> I have done a global search for `company` and there are a few documentation point to it, which I think it's not too bad. Although some tests are point to company too, which I think is the cause of this:

Okay, have updated the docs where appropriate. 2501e4d7

The tests are a bit of a shambles due to constantly shifting IDs -- I've made some fixes in #282 that address some of the problems (including replacing "Company" with "Organisation"), but to do it properly is a big job that requires some careful planning and discussion. Have made an issue for it: #316

nmadruga (approved):

PR #315, opened 2021-04-21
#282 Action improvements

CarlosNZ:
Implements the suggestions (mostly) outlined in #282. Involved quite a bit a re-factoring but I think it makes things more readable and robust overall, as well as simplifying the template queries.

Have updated templates and tested as best I can, except for Drug Registration -- haven't touched that one.

Sorry for the massive PR, but a lot of it is just repetitive adjustments to account for the new changes. The bulk of the "logic" changes are all in `triggersAndActions.ts` and related methods.

### Summary of changes:

(Including additional things/fixes that were needed in order to make the main elements (#282) work)

- added "outcome" to `application_stage_status` View (so it gets into "applicationData" -- surprised this wasn't already there)
- plugins specify "required_parameters" and "optional_parameters" in plugin.json. These are not actually checked at all at runtime, but they'll be important later for the Template Builder
- Condition is now evaluated at the start of each Action in a sequence (rather than all at once on a trigger). This meant that every Action goes into the Action Queue (previously only ones that met the condition did) and they're not checked until just before they run, so there is an additional "condition_expression" field in the action queue table. Also, there is a new action queue status value -- "Condition not met" -- that is applied to actions in the queue after checking
- applicationData is regenerated for every action, just before execution, so it'll always have the most up-to-date state, including updated values from previous actions. This largely removes the need for the cumulative "output" object, except for a few cases where a more obscure value not in applicationData is required. Because of this, we no longer need "application_data" in the action_queue table, so that field is gone.
- renamed "Test -- General Registration" (i.e. Feature Showcase) template to "Demo -- Feature Showcase" with templateCode = "Demo"
- all plugins take a standard input shape (see #282 description), and, for the most part, are able to deduce `applicationId` or `reviewId` from applicationData instead of needing to be provided as parameters. (I've updated the docs to explain exactly what is required for each and when they are optional, etc.)
- Removed a lot of the parameters from templates (as a result of previous change)
- Other template updates that simplify queries (e.g. condition can just check "Outcome" in applicationData rather than doing the more complex "reviewData" lookup every time)
- renamed `triggerFetchData` to `getApplicationData`, and a few small adjustments to shape the applicationData in a slightly more structured way
- Fix: getTemplateIdFromTrigger wasn't handling triggers from "review_assignment" table
- `getNumReviewLevels` method now looks directly at the new `template_stage_review_levels` table rather than the clunky calculation based on "template_permissions"
[...]

PR #321, opened 2021-04-26
#317 Graphql evaluator to accept external api

### LINKED ISSUE #317: ExpressionEvaluator to accept other endPoints to GraphQL queries so we can use it to make calls to external API

nmadruga:
Fixes #317 

### Changes:
- Fixed tests and simplified use of values direct as the type to be used instead of `{ "value": ...} `
- Added new tests to check usability of extension of `graphQL` operation to use external API,
- Updated the docs to explain how to use the operation for internal and external calls, for now to use our internal API it will expects the second parameters (url) as a string `graphQLEndpoint`. 
- Updated `graphQL` operation to also accept an external url to be used, and call the same way as how it was used before
- Attempt (not tested) to add a question to enter a code and a `testInfo` to display the country name as a result.

### Existing problem:
There is still one problem to use the GraphQL to call external API with the UC API, but this will be verified by the UC team.


CarlosNZ:
You'll want to re-base this to my branch I think, since most of the changes showing up here are from my PR (#315).

database/insertData/03_template_A_featureShowcase.js
```diff
@@ -689,12 +688,55 @@ exports.queries = [
...
+                              children: [
+                                "responses.Q2GraphQL.text"
+                                ""
+                              ]
+                            }
+                            "country"
+                          ]
+                        }
+                        visibilityCondition: {
+                          operator: "!="
```
CarlosNZ, line 733: As we discussed, this won't work with the currently installed package, so maybe save this addition to another PR once the package has been updated.

documentation/Query-Syntax.md
```diff
@@ -231,8 +228,9 @@ Performs queries on connected GraphQL interface.
 
 - Input:
   - 1st child node returns a **string** representing the GraphQL query
-  - 2nd child node returns an **array** of field names for the query's associated variables object. If no variables are required for the query, pass an empty array (i.e. `{ value: [] }`).
-  - 3rd...N-1 child nodes return the values of the fields for the variables object -- one node for each field in the previous node's array.
+  - 2nd child node returns a **string** containing the url to GrapqhQL (using "graphQLEndpoint") will the app internal GraphQL server.
```
CarlosNZ, line 231:
```suggestion
  - 2nd child node returns a **string** containing the url of the GrapqhQL endpoint. Using the value "graphQLEndpoint" (or empty string `""`) will the use the graphQL endpoint specified in the input parameter "GraphQLConnection" object.
```

src/modules/expression-evaluator/src/evaluateExpression.test.ts
```diff
@@ -451,20 +453,7 @@ test('Test Postgres get array of template names', () => {
...
-      expect(result).toEqual(4)
-    }
-  )
-})
-
-test('Test Postgres get template names -- no type', () => {
-  return evaluateExpression(testData.getListOfTemplates_noType, { pgConnection: pgConnect }).then(
-    (result: any) => {
-      expect(result).toEqual([
-        { name: 'Test -- General Registration' },
```
CarlosNZ, line 463: Sorry, this is not really for your PR, but can you update the tests that use this to the new name for the Demo template? (Currently 2 are failing due to this).

src/modules/expression-evaluator/src/evaluateExpressionTestData.ts
```diff
@@ -769,94 +744,121 @@ testData.getListOfApplications_withId = {
...
+          name
+          id
         }
-      }`,
-    },
-    { value: [] },
-    { value: 'applications.nodes' },
+      }
+    }`,
+    'graphQLEndpoint',
```
CarlosNZ, line 786: As we discussed, it'd be good if the empty string "" also defaulted to the internal endpoint, so could you make sure (at least) one of the test cases uses this? (And update the docs)

src/modules/expression-evaluator/src/evaluateExpression.ts
```diff
@@ -105,15 +105,17 @@ export default async function evaluateExpression(
...
+        let urlWithQuery, returnedProperty
         try {
-          ;[url, queryFields, queryValues, returnProperty] = assignChildNodesToQuery(
-            childrenResolved
-          )
+          const { url, fieldNames, values, returnProperty } = assignChildNodesToQuery([
+            '', // Extra unused field for API (query)
+            ...childrenResolved,
+          ])
+          returnedProperty = returnProperty
```
CarlosNZ, line 114: Just wondering why the need to rename the variable? Why not just keep using `returnProperty` all the way?

src/modules/expression-evaluator/src/evaluateExpression.ts
```diff
@@ -202,11 +215,16 @@ const extractAndSimplify = (
...
-  const fieldNames = childNodes[1]
-  const values = childNodes.slice(2, fieldNames.length + 2)
-  const returnProperty = childNodes[fieldNames.length + 2]
-  return [query, fieldNames, values, returnProperty]
+  const skipFields = 3 // skip query, url and fieldNames
+  const query: string = childNodes[0]
+  const url: string = childNodes[1]
+  const fieldNames: string[] = childNodes[2]
+
+  const lastField = fieldNames.length + skipFields
```
CarlosNZ, line 223: Perhaps `lastFieldIndex` for this variable?

src/modules/expression-evaluator/src/evaluateExpression.ts
```diff
@@ -243,8 +261,14 @@ const simplifyObject = (item: number | string | boolean | BasicObject) => {
...
 // Abstraction for GraphQL database query using Fetch
-const graphQLquery = async (query: string, variables: object, connection: IGraphQLConnection) => {
-  const queryResult = await connection.fetch(connection.endpoint, {
+const graphQLquery = async (
+  url: string,
+  query: string,
+  variables: object,
+  connection: IGraphQLConnection
+) => {
+  const endpoint = url !== 'graphQLEndpoint' && url !== '' ? url : connection.endpoint
```
CarlosNZ, line 270: Not a biggie, but maybe make `graphQLEndpoint` a case-insensitive requirement. So `graphqlendPOINT` should also be acceptable. But not super important if you can't be bothered for now.

Nice to have the empty string in there too. :) I wonder if we should also accept `null` for a default? (No need to worry about `undefined` as you can't pass `undefined` in a JSON string)

src/modules/expression-evaluator/src/evaluateExpression.ts
```diff
@@ -105,15 +105,17 @@ export default async function evaluateExpression(
...
 
       case 'API':
-        let url, urlWithQuery, queryFields, queryValues: string[], returnProperty
+        let urlWithQuery, returnedProperty
         try {
-          ;[url, queryFields, queryValues, returnProperty] = assignChildNodesToQuery(
-            childrenResolved
-          )
+          const { url, fieldNames, values, returnProperty } = assignChildNodesToQuery([
+            '', // Extra unused field for API (query)
```
CarlosNZ, line 111: Nice -- thanks :)

src/modules/expression-evaluator/src/evaluateExpressionTestData.ts
```diff
@@ -888,28 +890,22 @@ testData.emailValidation = {
...
-        {
-          value: 'http://localhost:8080/check-unique',
-        },
-        {
-          value: ['type', 'value'],
-        },
-        { value: 'email' },
+        'http://localhost:8080/check-unique',
+        ['type', 'value'],
+        'email',
```
CarlosNZ, line 903: It's a good idea to rewrite these using the simpler syntax, but we should make sure we leave some of the tests using the more verbose version so that gets tested too.

src/modules/expression-evaluator/src/evaluateExpression.test.ts
CarlosNZ: Also, why delete this test? Is there a problem with it?

src/modules/expression-evaluator/src/evaluateExpression.test.ts
```diff
@@ -585,7 +558,29 @@ test('Test GraphQL -- count Sections on current Application', () => {
...
+    expect(result).toEqual(testData.continentsResult)
+  })
+})
+
+test('Test GraphQL -- get country details by code from External Graphql API ', () => {
+  return evaluateExpression(testData.GraphQL_GetCountryByCode_ExternalAPI, {
+    graphQLConnection: {
+      fetch: fetch,
+      endpoint: graphQLendpoint,
+    },
```
CarlosNZ, line 581: The `graphQLConnection` object shouldn't be needed here, so it might be a good idea to remove it from _one_ of these tests to test for that case. (Like, if the GraphQL evaluator looks for a GraphQLConnection object and doesn't see one, it should still work, so we need to test for that)

CarlosNZ: Actually, scratch that -- it's still needed for the `fetch` object. So maybe instead omit the `endpoint` property from the object on one of these and make sure it still works. (You'll probably need to make "endpoint" an optional property in the types).

CarlosNZ (changes requested):
Great, thanks for this. Working well and seems robust. A couple of tests still failing, but I've marked that in my inline comments.

src/modules/expression-evaluator/src/evaluateExpression.test.ts
nmadruga: It seemed to be exactly the same as `getListOfTemplates`

src/modules/expression-evaluator/src/evaluateExpressionTestData.ts
nmadruga: Updated the docs - by committing your suggestion ✅ 
Adding new test to suite 📝

src/modules/expression-evaluator/src/evaluateExpression.test.ts
CarlosNZ: Okay, it's actually testing a slightly different thing:
```
testData.getListOfTemplates = {
  type: 'array',
  operator: 'pgSQL',
  children: [
    {
      value: 'SELECT name FROM template',
    },
  ],
}
```

vs

```
testData.getListOfTemplates_noType = {
  operator: 'pgSQL',
  children: [
    {
      value: 'SELECT name FROM template',
    },
  ],
}
```

The "type" parameter (which is not used much, but can be important) causes the query to return a different shaped array, so I think we should keep it in. In fact, I think I made these 2 different tests in order to demonstrate the difference.

From the docs:
> The type property is optional in most cases, but some operators will return their results in a different format depending on the type (e.g. pgSQL -- see below).

database/insertData/03_template_A_featureShowcase.js
nmadruga: Done: 30a1c6c

src/modules/expression-evaluator/src/evaluateExpression.test.ts
nmadruga: Oh, I see. Sorry. Will revert that removal now : )

src/modules/expression-evaluator/src/evaluateExpressionTestData.ts
nmadruga: Done: 2496d24

src/modules/expression-evaluator/src/evaluateExpression.ts
nmadruga: Not accessible when Ii changed this to be a destructed object (received inside the `try` block), not the prettiest like that I agree.

src/modules/expression-evaluator/src/evaluateExpression.ts
nmadruga: cd95ce1

src/modules/expression-evaluator/src/evaluateExpression.ts
nmadruga: Sure, I'll add these 2 options, but not specific tests, since this will be just a prevention for missusage

nmadruga: Done: 152e1e5

Checked that the case insensitive works fine using the "Evaluate expression gui", but couldn't test using `null`:
Screen Shot 2021-04-28 at 10 07 49 AM

src/modules/expression-evaluator/src/evaluateExpressionTestData.ts
nmadruga: Oh, sorry. I haven't though about that! I can add a new one specific to test that now

src/modules/expression-evaluator/src/evaluateExpression.test.ts
nmadruga: No, that's where it gets the `default` one from. As discussed in person, we could have changed that so front-end doesn't specify the default endpoint, but for now that's what is used for it. We can change that later, I don't think it's a priority for now. Maybe it can be a note to take as part of the Build to production Epic.

src/modules/expression-evaluator/src/evaluateExpressionTestData.ts
nmadruga: Done: 59783cd

src/modules/expression-evaluator/src/evaluateExpression.test.ts
nmadruga: Done: be26805

src/modules/expression-evaluator/src/evaluateExpression.test.ts
CarlosNZ: Yeah, but if you're passing an external URL in the query expression, you should be able to omit the `endpoint` field in the `graphQLConnection` object that you pass in, and the evaluator should still work, hence we should make sure there is a test that handles that case.

(Use case: if using the evaluator for something else (i.e. not IRIMS), you might not even have a local GraphQL connection, so why require an endpoint field for external queries?)

src/modules/expression-evaluator/src/evaluateExpression.ts
CarlosNZ: Oh -- I can see the problem: 
`const endpoint = url.toLowerCase() !== 'graphqlendpoint' && url !== '' && url !== null`

When `url` is `null`, that first evaluation `url.toLowerCase()` is throwing an error. It works as expected if you put the null check first:
`const endpoint = url !== null && url.toLowerCase() !== 'graphqlendpoint' && url !== ''`

Screen Shot 2021-04-28 at 11 05 42 AM

:)

src/modules/expression-evaluator/src/evaluateExpressionTestData.ts
CarlosNZ: Great, thanks. And it looks like there's still plenty of older tests using the old format for other operators, so that's all good.

src/modules/expression-evaluator/src/evaluateExpression.ts
nmadruga: Thanks I'l update that

src/modules/expression-evaluator/src/evaluateExpression.test.ts
nmadruga: I'll set this as a separated issue, I feel this is extra and really would like to continue with the progress of getting the changes ready to interact with UC codes server. And this will be an improvement to be done - when needed.

src/modules/expression-evaluator/src/evaluateExpression.ts
CarlosNZ: Ah yes, I see. I've come across that particular problem in a few places, actually. Wish there was a more elegant solution for it.

nmadruga:
Can this one be merged now? I guess after doing this I need to follow this [docs](https://github.com/openmsupply/application-manager-server/wiki/Query-Syntax#installation) to publish changes to evaluator? I'm very excited to get this one included so I can finally work on changes to issue [Front-end#582](https://github.com/openmsupply/application-manager-web-app/issues/582)

CarlosNZ (approved):
Great, thanks. :)

If you want to publish this package once you've merged it, there's instructions at: https://github.com/openmsupply/application-manager-server/wiki/Query-Syntax#development

Although last time I did it, something had changed slightly (sorry can't remember what), so let me know if you have any probs with it.

Cheers.

PR #323, opened 2021-04-28
Fix nasty bug in postinstall

andreievg:
Getting freezes with postinstall (when plugins are installed), where plugin folder doesn't have package json (i.e. git ignored  node_modules folder exists, but no code). It also seems like cpu usage goes insane (without any log of what's happening, can hear my working overtime)

### DIFF
--- a/utils/pluginScripts/yarnInstallPlugins.js
+++ b/utils/pluginScripts/yarnInstallPlugins.js
@@ -1,4 +1,5 @@
 const fs = require('fs')
+const path = require('path')
 const { exec } = require('child_process')
 // should run from root folder yarn scripts
 const pluginsFolder = './plugins/'
@@ -13,6 +14,10 @@ console.log('running yarn install for plugins')
 const command = 'yarn install'
 plugins.forEach((plugin) => {
   const { fullPluginFolder, pluginName } = plugin
+  if (!fs.existsSync(path.join(fullPluginFolder, 'package.json'))) {
+    console.log('no package.json: ' + pluginName)
+    return
+  }
   console.log('running yarn install for ' + pluginName + ' ... ')
 
   const options = { cwd: fullPluginFolder }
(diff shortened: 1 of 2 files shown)


utils/pluginScripts/yarnInstallPlugins.js
```diff
@@ -13,6 +14,10 @@ console.log('running yarn install for plugins')
 const command = 'yarn install'
 plugins.forEach((plugin) => {
   const { fullPluginFolder, pluginName } = plugin
+  if (!fs.existsSync(path.join(fullPluginFolder, 'package.json'))) {
+    console.log('no package.json: ' + pluginName)
+    return
+  }
```
CarlosNZ, line 20: When this bug happened for me, it was when there was empty plugin folders (from previously renamed plugins on another branch), so it wasn't just the lack of a `plugin.json`. What I mean is that, will this fix also handle folders within "plugins" that have _nothing_ in them, too?

CarlosNZ (approved):
Thanks, looks good. I just made one quick comment about potential cases that might cause a problem, so maybe just double-check it'll handle empty plugin folders too.

utils/pluginScripts/yarnInstallPlugins.js
andreievg: ohh it's looking for `package.json`, that should be enough

CarlosNZ: Sweet. So it never looks for anything else? What about `index.ts`?

andreievg: nope, it just stops it from running 'yarn install' which causes an error to be thrown and something funky happens in `postinstall`, i think every plugin should have `package.json` as it has build instructions for it.

and you've done good checking on application startup for other things (in case of empty folders like this)

PR #325, opened 2021-04-28
#317 Part 2 - Bump to version 1.6.2

nmadruga:
Just published package @expression-evaluator with new version 1.6.2

Please also check front-end [PR](https://github.com/openmsupply/application-manager-web-app/pull/643) for bumping @expression-evaluator.


database/insertData/03_template_A_featureShowcase.js
```diff
@@ -512,6 +512,7 @@ exports.queries = [
                           operator: "graphQL",
                           children: [
                             "query getOrgs {organisations {nodes {name}}}",
+                            "graphQLEndpoint",
```
andreievg, line 515: should this be the end point ?

CarlosNZ: Nah, it's the default, so it'll just use the existing endpoint from the passed-in connection object. Could also be "" or `null`. (Personally I think `""` is better to use in this context -- precisely to avoid the exact confusion you just had  😄)

CarlosNZ (changes requested):
Line 54 of "package.json":
```
    "@openmsupply/expression-evaluator": "^1.5.2",
```
You haven't actually bumped the version it's using.

nmadruga:
Hmmm not sure what happened. I'm certain I had run the `yarn upgrade @openmsupply/expression-evaluator` and it resulted with bumping the @expression-evaluator version, with this message:
Screen Shot 2021-04-29 at 10 03 21 AM
Maybe it was missing the `yarn` after that idk, I ended up adding the correct to `package.json` manually then running `yarn`. Hope it is enough.

CarlosNZ (approved):
👍

CarlosNZ:
> I ended up adding the correct to `package.json` manually then running `yarn`

That's what I've always done. :)

PR #327, opened 2021-04-28
#326 Fix to allow result of GraphQL operation to return Null instead of fallback

### LINKED ISSUE #326: Allow returning null from result of GraphQL query in Expression evaluator

nmadruga:
Fixes #326 

Another quick fix, needed to get proper result for when the query results on `null` as "no results" for the given input.

### DIFF
--- a/src/modules/expression-evaluator/src/evaluateExpression.test.ts
+++ b/src/modules/expression-evaluator/src/evaluateExpression.test.ts
@@ -622,6 +622,17 @@ test('Test GraphQL -- get country details by code from External Graphql API ', (
   })
 })
 
+test('Test GraphQL -- Check result of field can be null', () => {
+  return evaluateExpression(testData.GraphQL_CheckDefaultResultIsNull, {
+    graphQLConnection: {
+      fetch: fetch,
+      endpoint: graphQLendpoint,
+    },
+  }).then((result: any) => {
+    expect(result).toBe(null)
+  })
+})
+
 // TO-DO: Test with multiple variables and dynamic values
 
 // More complex combinations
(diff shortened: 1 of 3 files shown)


CarlosNZ:
Cool, but needs test cases for it. Just add one more test using the query you're planning to use in the Drug reg template.

nmadruga:
Ah! I just realised that now haha. Will do that in a minute

nmadruga:
Hi @CarlosNZ it is still failing on one test: "Test unresolved object with Null fallback" which I'm not sure what to do about it
```
testData.objectPropertyUnresolvedWithNullFallback = {
  operator: 'objectProperties',
  children: ['application.querstions.q2', null],
}
```

CarlosNZ (changes requested):
🙁
Screen Shot 2021-04-29 at 1 03 01 PM

nmadruga:
> 🙁
> Screen Shot 2021-04-29 at 1 03 01 PM

Please see comment above ^

CarlosNZ (approved):
Great -- thanks!

Once you merge this, publish it and then bump the version number directly on master to save doing another PR.

nmadruga:
> Great -- thanks!
> 
> Once you merge this, publish it and then bump the version number directly on master to save doing another PR.

Ok. Thanks! I'll probably need to create one for Front-end, or should just do straight on master as well?

CarlosNZ:
> Ok. Thanks! I'll probably need to create one for Front-end, or should just do straight on master as well?

Yeah, you're literally just changing the minimum version number in `package.json`, (and re-run `yarn` so `yarn.lock` updates too).

PR #332, opened 2021-05-02
#571F User Edit template using "default" values

CarlosNZ:
This is the back-end branch for [front-end PR #652](https://github.com/openmsupply/application-manager-web-app/pull/652).

Have created a rudimentary "Edit User" template which demonstrates pre-filled "default" values for text inputs. There is a lot of functionality not implemented yet, as that's not the primary focus for this issue.

Just so we have a record of it, here are the things that became apparent we need to do to make editing user details possible:

- a back-end Action to update user entities (prob a generic one that can work for Orgs and other entities too)
- a way to confirm (existing) passwords from within a form -- will need a POST operator for the evaluator so we can check the `/login` endpoint.
- some way to handle password input in a form without storing the response. We could perhaps extend the existing "password" input plugin to handle password checking as well as creation?


andreievg (commented):
should we just use the core actions ?

andreievg:
i.e. do we need to trim responses etc ?

CarlosNZ:
> should we just use the core actions ?

I'm thinking not, since this is more like the "UserRegistration" template, which doesn't go through any review, etc. In fact, it's pretty much exactly the same as User Registration, except we need an action to *update* a user rather than create one.

andreievg (approved):

Issue #241, opened 2021-03-09
Convert Postgres Enums to use ALL_CAPS for values


CarlosNZ:
This is so we don't get problems with GraphQL and SQL returning incompatible values, e.g. ("SUBMITTED" vs "Submitted")

Other considerations:
- SPACES_BECOME_UNDERSCORES
- No apostrophes or other punctuation.

PR #336, opened 2021-05-06
re-formatting using pgFormatter

andreievg:
Reformat for changes needed in fix for #241 (can reformat again, this was needed for changes in postgres insert data, I had pgFormatter turned on and it was a lot of hassle turning format on save to on and off, and wanted to reduce diff in upcoming PR)

using `pgFormatter` vs code extension


CarlosNZ (approved):
Sweet. Like the consistent use of CAPS statements now :)

PR #337, opened 2021-05-06
#241 Changing enums

### LINKED ISSUE #241: Convert Postgres Enums to use ALL_CAPS for values

andreievg:
Fixes #241 

Branched from re-formatting-postgres-using-pg-formatter

Needs front end [front end pr](https://github.com/openmsupply/application-manager-web-app/pull/676) (to make sure permissions checking is using aligned permission policy types in application list)

Wanted to do this now, since i needed to check use roles on front end for Dashboard and the types returned from pg was miss matched. (i did a hacky way in demo branch).

### Changes

* Global replaced all enum values to capitalised snake case
* I could see all places that were are using string literals, so replaced them all with types from graphQL
* Wanted to make sure that all plugins conform to API, so change index.js to check type correctness
* Changed most plugins to use ActionPluginType type
* Went through git diff and cleaned up some changes that were incorrectly global replaced 

### Testing

* Made sure it built correctly
* Did a basis test of UI, actions seem to work
* Ran yarn test on both master and this branch, and compared test results (almost 100% match apart from the test that failed to run due to typescript compilation, which is fixed in this branch)


database/buildSchema/06_trigger_queue.sql
```diff
@@ -40,8 +40,8 @@ CREATE OR REPLACE FUNCTION public.add_event_to_trigger_queue ()
     AS $trigger_queue$
 BEGIN
     INSERT INTO trigger_queue (trigger_type, "table", record_id, timestamp, status)
-        VALUES (NEW.trigger::public.trigger, TG_TABLE_NAME, NEW.id, CURRENT_TIMESTAMP, 'Triggered');
-    EXECUTE format('UPDATE %s SET trigger = ''Processing'' WHERE id = %s', TG_TABLE_NAME, NEW.id);
+        VALUES (NEW.trigger::public.trigger, TG_TABLE_NAME, NEW.id, CURRENT_TIMESTAMP, 'TRIGGERED');
+    EXECUTE format('UPDATE %s SET trigger = %L WHERE id = %s', TG_TABLE_NAME, 'PROCESSING', NEW.id);
```
CarlosNZ, line 44: What is `%L`?

documentation/Glossary.md
```diff
@@ -28,7 +28,7 @@ In Old IRIMS terms: PROCEDURE.
 
 **Outcome** — The final result of an application. Will most likely be either “Approved” or “Rejected” once it reaches the “Complete” status, and before that would be “Pending” (or perhaps Null).
 
-**Trigger** — A type of Event listener that responds to changes in the database or environment to launch “Actions”. Examples might be “OnSubmit”, “OnReviewSubmit”, “OnApproval”, etc. Can be in response to user-initiated events (like those examples) or time-based, such as schedulers (e.g. if an application must be responded to within a certain time limit, an Action would be Triggered when that time is reached)
+**Trigger** — A type of Event listener that responds to changes in the database or environment to launch “Actions”. Examples might be “OnSubmit”, “ON_REVIEW_SUBMIT”, “OnApproval”, etc. Can be in response to user-initiated events (like those examples) or time-based, such as schedulers (e.g. if an application must be responded to within a certain time limit, an Action would be TRIGGERED when that time is reached)
```
CarlosNZ, line 31: Don't need this TRIGGERED, it's just regular prose here.

Global find and replace, huh? ;)

documentation/Triggers-and-Actions.md
```diff
@@ -33,18 +33,18 @@ Actions also return an **output** object (e.g. `createUser` Action returns user
...
+| `ON_APPLICATION_CREATE`    | Postgres trigger on `Application` table                          |
+| `ON_APPLICATION_SUBMIT`    | Postgres trigger on `Application` table                          |
+| `ON_APPLICATION_SAVE`      | Postgres trigger on `Application` table                          |
+| `ON_APPLICATION_WITHDRAWn` | Postgres trigger on `Application` table                          |
+| `ON_REVIEW_CREATE`         | Postgres trigger on `Review` table                               |
+| `ON_REVIEW_START`          | Postgres trigger on `Review` table                               |
+| `ON_REVIEW_SUBMIT`         | Postgres trigger on `Review` table                               |
+| `ON_REVIEW_ASSIGN`         | To be decided                                                    |
+| `ON_APPROVAL_SUBMIT`       | To be decided                                                    |
+| `onScheduledTime`          | Server scheduled service (see [here](link to Scheduled actions)) |
```
CarlosNZ, line 47: We haven't implemented ON_SCHEDULED_TIME -- shall we remove this line for now, or what?

plugins/types.ts
```diff
@@ -14,11 +13,4 @@ export interface ActionPluginOutput {
...
 
-export type Status =
-  | 'Draft'
-  | 'Withdrawn'
-  | 'Submitted'
-  | 'Changes Required'
-  | 'Re-submitted'
-  | 'Completed'
-  | 'Expired'
+export type ActionPluginType = (props: ActionPluginInput) => Promise<ActionPluginOutput>
```
CarlosNZ, line 16: What does this do?

documentation/Glossary.md
CarlosNZ: Also, “OnSubmit”, “ON_REVIEW_SUBMIT”, “OnApproval” -- 2 didn't get changed -- maybe change them to actual examples, not these obsolete ones.

CarlosNZ (changes requested):
Awesome, that looks like a lot of fiddly work. 

How thoroughly did you test the actions and stuff? Would be good to make sure you can still go through a whole Register User -> Create Org -> Join Org process and everything still works properly.

database/buildSchema/06_trigger_queue.sql
andreievg: string literal, I could have kept it as is, just capitalise, but the first time i changed it i remove the double `'` and broke it, so thought to just change to `%L`, %s doesn't work cause it needs to be quoted (unlike table name and id)

documentation/Glossary.md
andreievg: oops missed that capitalise trigger, yeah there were quite a lot that i global replaced and then change after reviewing my own PR, also change to actual trigger examples in this doc
9abb617d4bd0f5005adbbff7f4d7f3db0471ecbe

documentation/Triggers-and-Actions.md
andreievg: it would be used eventually ? cc0c57e2bf943d6bfaddf9d3973afc47e24a5494

plugins/types.ts
andreievg: single type definition for the method (including props and return type), used it in index files to check that plugin conforms to API (since it's dynamically imported it's not checked by TS for overall build, and individual plugin)
Also ended up using it in a few actions (although i didn't need to use it in some)
```
async function action(props: ActionPluginInput): Promise<ActionPluginOutput> { ..
```
vs
```
const action: ActionPluginType = async(props)  => { ...
```

andreievg:
> How thoroughly did you test the actions and stuff? Would be good to make sure you can still go through a whole Register User -> Create Org -> Join Org process and everything still works properly.

Will check now

andreievg:
I am glad i checked, 85ae17375cd891667075de5399606d67ab306b9a
i did a full test after, files are correctly trimmed, status change, organisation is created, review assignment created (checked for self assign and for assignment), users created. should be pretty good  now

CarlosNZ (approved):
Awesome, thanks for checking (good to catch those template values) :)

Issue #313, opened 2021-04-20
registerPlugins shouldn't put the full system path in the "path" field


CarlosNZ:
Relative path should be sufficient

PR #340, opened 2021-05-10
#313 Action plugin file path fix

CarlosNZ:
Change registerPlugins.ts to store the action path relative to the project root, not the absolute path. Full path is reconstructed (using `getAppEntryPointDir()`) as actions are loaded in triggersAndActions.ts

### DIFF
--- a/src/components/registerPlugins.ts
+++ b/src/components/registerPlugins.ts
@@ -42,7 +42,6 @@ export default async function registerPlugins() {
         console.log('Failed to parse plugin.json in: ' + pluginPath)
         throw e
       }
-      // console.log('pluginDirName', pluginDirName)
       const pluginObject = {
         ...pluginJson,
         path: getPluginIndexPath(pluginPath),
--- a/src/components/registerPlugins.ts
+++ b/src/components/registerPlugins.ts
@@ -13,21 +13,20 @@ import * as config from '../config.json'
 import DBConnect from './databaseConnect'
 import { ActionPlugin } from '../types'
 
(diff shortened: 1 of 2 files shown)


src/components/registerPlugins.ts
```diff
@@ -13,30 +13,33 @@ import * as config from '../config.json'
...
   console.log('Scanning plugins folder...')
   const plugins = fs
-    .readdirSync(pluginsFolder)
-    .map((pluginFolder) => {
-      return path.join(pluginsFolder, pluginFolder)
-    })
-    .filter((pluginPath) => fs.statSync(pluginPath).isDirectory())
-    .filter((pluginPath) => fs.existsSync(path.join(pluginPath, pluginJsonFilename)))
+    .readdirSync(pluginsFolderFull)
+    .filter((pluginPath) => fs.statSync(path.join(pluginsFolderFull, pluginPath)).isDirectory())
```
andreievg, line 29: would be good to map into {pluginName, pluginFullPath, pluginJsonFilenamePath} before filtering for isDirectoy and then existence of pluginJsonFilenamePath. But that's probably stylistic, can still follow code pretty easy.

andreievg (approved):

src/components/registerPlugins.ts
CarlosNZ: Not entirely clear what you mean. I tried different ways of combining different path chunks into different variables but couldn't really find anything else that was simpler in all places. Can you be more specific?

Issue #348, opened 2021-05-10
Modify Entity Action


CarlosNZ:
Currently we have a "Create User" and a "Create Org" action. We also need an Action to "Update User" and "Update Org", as well as being able to (in future) create and/or update entities in any arbitrary table (e.g. Product).

Proposed solution is to combine all of these into one Action -- **Modify Entity**. Input parameters would be exactly as per current Create User/Org Actions, except would require an additional "table" parameter to specify what to update (e.g. user, organisation). It would also accept an "id" parameter, in which case the intention is to update an *existing* record. If no ID is supplied a new record will be created using the input parameters.

(It would be good to be able to not require an ID and do an UPDATE on any conflict, but this doesn't seem that easy since the ON CONFLICT command requires a specified constraint, and we can't ensure that our action would know in advance the unique fields of any particular table)

andreievg:
I would call it 'record' rather than entity. To me entity sounds too ambiguous

CarlosNZ:
> I would call it 'record' rather than entity. To me entity sounds too ambiguous

Sure.

PR #349, opened 2021-05-11
#348 Modify Record Action

CarlosNZ:
Implements #348 

This will replace (and extend) the "createUser" and "createOrg" actions, ~~but I'll wait until this general approach is approved before I delete those ones and update their templates. (I'll also update the docs then too)~~ (Done)

For now, it's added to the UserEdit template, accessible at: http://localhost:3000/application/new?type=UserEdit

You should be able to edit your existing user details and have them updated in the database on Submit.

There is also a test suite that can be run after a fresh `database_init`

### Additional comments/considerations

- I had hoped to not need the "tableName" parameter, and just pass in a whole "user" or "org" object, and the Action would determine the table from the object key. However, this wasn't possible, since the evaluator for parameters only looks at the first level fields, not nested fields, and the template action needed to populate these values from responses.
- There was a problem when using a template to define entity fields in that it wasn't clear how to get the "update" method to ignore a field if it wasn't entered. (E.g. "password" field in UserEdit, where if the user doesn't change the password, the "default" original hash is not available). So for now, I've configured it to, if an input parameter has a Null value, it'll be removed by the Action before updating the database entity (i.e. it'll be unchanged). The only downside to this is that you won't be able to actually set a field to NULL if it previously had a value in it, but I can't think of a use case where that would be needed. Any thoughts about this?
- It just checks for an "id" input parameter to decide whether to update a record or create a new one. That means this would only work for tables that have "id" as their primary key, which is okay for now. I had hoped to use the database ON CONFLICT feature to let the database handle whether to update or create, but the problem is ON CONFLICT requires you to name the fields to check for conflicts, which is not useful for updating arbitrary tables.


andreievg (commented):
Looks good to me:

do we want to change existing templates to use this ?

> I had hoped to not need the "tableName" parameter, and just pass in a whole "user" or "org" object, and the Action would determine the table from the object key. However, this wasn't possible, since the evaluator for parameters only looks at the first level fields, not nested fields, and the template action needed to populate these values from responses

I think passing through tableName is the only way to do this

> There was a problem when using a template to define entity fields in that it wasn't clear how to get the "update" method to ignore a field if it wasn't entered. (E.g. "password" field in UserEdit, where if the user doesn't change the password, the "default" original hash is not available). So for now, I've configured it to, if an input parameter has a Null value, it'll be removed by the Action before updating the database entity (i.e. it'll be unchanged). The only downside to this is that you won't be able to actually set a field to NULL if it previously had a value in it, but I can't think of a use case where that would be needed. Any thoughts about this?

I am just wondering, what's wrong with updating all of the fields ?

> It just checks for an "id" input parameter to decide whether to update a record or create a new one. That means this would only work for tables that have "id" as their primary key, which is okay for now. I had hoped to use the database ON CONFLICT feature to let the database handle whether to update or create, but the problem is ON CONFLICT requires you to name the fields to check for conflicts, which is not useful for updating arbitrary tables.

tbh, I found ON CONFLICT logic in other actions quite messy (where sql statement is constructed based on value being null or not) and sql execute of statements that are computed from inputs that user can enter is risky (sql injections). I think it would have been better to:
* query for conflicts manually 
* use graphQL mutations for dynamic inserts

anyways, in reply to your comment about match by ID, i do see a use case where we would want to insert by something other then ID (say for company registration number, if we don't want to query and store it in the form). We can just add a query for that field (likewise in this implementation can do a manual query for id or 'fieldToMatch'/'fieldValueToMatch' and decide to either insert or update). Happy to approve as is, see what you think

CarlosNZ:
> do we want to change existing templates to use this ?

Yes, I plan to update user registration and Org registration once the general approach here is approved.

> I am just wondering, what's wrong with updating all of the fields ?

That would be fine, except for the case of the Password field, we don't actually have access to the previous hash, so unless the user has **changed** their password, we have no way of accessing the previous hash value to (re-) update the password_hash field.

Actually, I just thought of another way to handle it -- what do you think: we put this Action in the UserEdit template TWICE -- the first one in the sequence updates all fields EXCEPT password, and the second one updates the password field, but it only runs on the condition that the "Change Password" checkbox has been selected. _Voila!_ 🤷‍♂️


> * query for conflicts manually
> 
> * use graphQL mutations for dynamic inserts

But how do we "query for conflicts" manually? What field do we look up existing data by (other than ID?) Because this can be used on any table, the Action has no way to know what fields should be unique and therefore can be considered a "conflict" to be updated.

andreievg:
> But how do we "query for conflicts" manually? What field do we look up existing data by (other than ID?) Because this can be used on any table, the Action has no way to know what fields should be unique and therefore can be considered a "conflict" to be updated.

I mean two extra parameters  'fieldToMatch'/'fieldValueToMatch' (maybe more descriptive names to indicate that would be the field and value to match for update)

> Actually, I just thought of another way to handle it -- what do you think: we put this Action in the UserEdit template TWICE -- the first one in the sequence updates all fields EXCEPT password, and the second one updates the password field, but it only runs on the condition that the "Change Password" checkbox has been selected. Voila

hmm maybe we can do this without two actions ?

evaluation expression, for password field ` is checkbox ticked ? value from password field : null `, and we make sure this action doesn't try updating null params

andreievg:
Man i didn't PR this properly, you are already doing "evaluation expression, for password field is checkbox ticked ? value from password field : null, and we make sure this action doesn't try updating null params"

CarlosNZ:
> evaluation expression, for password field `is checkbox ticked ? value from password field : null`, and we make sure this action doesn't try updating null params

That's exactly what it's doing now 😆 (My only concern was if there was ever a case when we *wanted* to replace a non-null value with Null, but I can't think of such a use case)

CarlosNZ:
> > I mean two extra parameters 'fieldToMatch'/'fieldValueToMatch' (maybe more descriptive names to indicate that would be the field and value to match for update)
> 
> > Just to be clear, we do two queries ->
> > query to see if there is a match
> > if yes update
> > otherwise insert
> 
> Got it. Okay, will update first thing tomorrow morn. Will include the "matchField" parameter, but it should default to "id" if not provided. :)

@andrei, have updated this now, including the docs and the tests (to include various cases where `matchField` and `matchValue` are included and aren't)

Going to update existing templates now.

andreievg (approved):
lgtm

Issue #342, opened 2021-05-10
Create 3rd field for restriction on the Application table as JSON to be used on RowLevel restrictions - new action for JoinOrg template to add org_id


nmadruga:

CarlosNZ:
Can I just clarify the new field should get:

- a new "user registration" application should get value `{ uniqueId: <random-string> }`
- a new "join org" application should get value `{ org_id: <organisation ID from application response> }`

PR #350, opened 2021-05-12
#342 extra restriction (sessionId) field for application table

CarlosNZ:
Implements #342, as discussed. 

Requires [front-end branch from PR #693](https://github.com/openmsupply/application-manager-web-app/pull/693) 

### Summary of changes

- new field "session_id" on "application" table
- random sessionId (using nanoid) generated by server as part of `/user-info` method.
- sessionId returned as part of JWT and response body by `/login`, `/login-org` and `/user-info` endpoints
- add action to OrgJoin template to insert `org_id` (the org id of the Org applicant is applying to join) into application table on Submit (using new modifyRecord Action)


andreievg (changes requested):
Looks good, but I've commented in front end PR to:
```
Back end should add logic to make sure sessionID does not change if JWT is validated (but recalculate other things, templatePermission etc..)
```

i think we pass sessionID to getUserInfo, as optional and only generate it if it's not passed through

CarlosNZ:
> i think we pass sessionID to getUserInfo, as optional and only generate it if it's not passed through

Cool, done: f16f1c08

andreievg (approved):

Issue #344, opened 2021-05-10
Create new 3 policies with and only add the view permissions specific enforced (on application table)


nmadruga:
## Policies

- **nonRegisteredApply**:  view: match by sessionID and userId = nonregistered (ideally, set `nonRegistered` user to id = 1 in database)
- **applyByUser**: view: match by userId??
- **applyByOrg**: view: match by orgId

Do we need one that matches by userId AND orgId, or are these two enough?

CarlosNZ:
I think this is done, right?

PR #353, opened 2021-05-12
#344 Apply (View) policies

CarlosNZ:
Implements #344. 

I think this is all that's required for this issue, please let me know if there's more to it.

- Edited/Updated the following policies:
   - **applyNonRegistered**:  view: match by sessionID and userId = nonregistered (ideally, set `nonRegistered` user to id = 1 in database)
   - **applyUserRestricted**: view: match by userId??
   - **applyOrgRestricted**: view: match by orgId 
- updated permissionName links to these, and minor amendments to template permissions
- hard-coded INSERT and UPDATE policies (in rowLevelPolicyHelper.ts) with no restrictions (until we add them to policies)

You'll need to turn on permissions in PG (`yarn dev_pg_permissions`) to see it working. (e.g. only nonRegistered user with the same sessionId they started with can see a UserRegistration form)


andreievg (approved):
Look good `policyName` in the permission_policy view could prove handy

Issue #331, opened 2021-05-02
Expression Evaluator: create a POST request operator


CarlosNZ:
Seems like we will need the ability to make POST requests with the evaluator (for checking login, for example).

### Suggestions

- change the name of the API operator to "GET", and make another one called "POST"
- make the node structure similar to the GET operator, but will need another one for "body"
- do we need to make query parameters possible for POST, or should we assume all user data should be in body fields?

PR #354, opened 2021-05-12
#331 POST operator for evaluator

CarlosNZ:
Changed name of "API" operator to "GET", and added additional "POST" operator.

GET and POST share the same method, they only difference is that POST also sends a body object with the request, so this is quite a simple improvement.

### Also tidied up:
- simplified input handling for evaluator -- no longer allows stringified JSON (it will be returned just as a string), but we never use this, and it's trivial to parse JSON before evaluating if we ever need it.
- updated tests (include POST test, remove stringified JSON tests)
- Improved docs:
  - removed all references to the `value` property (we should probably deprecate this soon)
  - add table of contents

**NOTE**: Will update existing templates etc. in a separate PR once this is merged and I've pushed a new package version.


src/modules/expression-evaluator/src/evaluateExpression.ts
```diff
@@ -290,9 +287,21 @@ const graphQLquery = async (
...
+}
+
+// GET/POST request using fetch (node or browser variety)
+const fetchAPIrequest = async ({ url, APIfetch, method = 'GET', body }: APIrequestProps) => {
+  const result = await APIfetch(url, {
+    method,
+    headers: {
+      'Content-Type': 'application/json',
+    },
+    body: JSON.stringify(body),
```
andreievg, line 304: Although this work, would prefer a condition:

Screen Shot 2021-05-13 at 11 23 16 AM

```
body: body instanceof Object ?  JSON.stringify(body) : undefined
```

src/modules/expression-evaluator/src/evaluateExpression.ts
```diff
@@ -1,149 +1,146 @@
...
-        } catch {
-          throw new Error('Problem with API call')
-        }
-        try {
-          return extractAndSimplify(data, returnedProperty, "API - Can't resolve property")
-        } catch {
-          throw new Error('Problem parsing requested node from API result')
-        }
+    case 'POST':
+    case 'GET':
```
andreievg, line 95: can leave this as three options, 'POST', 'API' 'GET' (API === GET), might be safer and would avoid need to refactor configs

src/modules/expression-evaluator/src/types.ts
```diff
@@ -26,25 +26,26 @@ export interface IParameters {
...
+  value?: any
   type?: NodeType
-  operator?: Operator
-  children?: Array<IQueryNode | string | boolean | number | null | undefined>
+  operator: Operator
+  children: Array<IQueryNode | string | boolean | number | null | object | undefined>
 }
 
-type NodeType = 'string' | 'number' | 'boolean' | 'array'
+export type NodeType = 'string' | 'number' | 'boolean' | 'array'
```
andreievg, line 35: can NodeType be an object ?

src/modules/expression-evaluator/src/evaluateExpression.ts
```diff
@@ -1,149 +1,146 @@
...
+  IParameters,
+  IGraphQLConnection,
+  NodeType,
+  BasicObject,
+} from './types'
 
 const defaultParameters: IParameters = {}
 
 export default async function evaluateExpression(
   inputQuery: IQueryNode | string | number | boolean | any[],
```
andreievg, line 13: or object ?

andreievg (changes requested):
Looks pretty good, a few comments

src/modules/expression-evaluator/src/evaluateExpression.ts
CarlosNZ: Not sure what the difference is, since `JSON.stringify(body)` returns undefined anyway, if body is undefined. And any other primitive type is valid JSON.

src/modules/expression-evaluator/src/evaluateExpression.ts
CarlosNZ: Good idea. 6f981707

src/modules/expression-evaluator/src/evaluateExpression.ts
CarlosNZ: Okay, in light of the changes here (and gradual evolution since this was born), I've cleaned up and simplified the types overall. 7a953d6d

Every node is now either a `OperatorNode` or a `ValueNode` type, defined as:

```
export interface OperatorNode {
  operator: Operator
  type?: OutputType
  children: Array<OperatorNode | ValueNode>
  value?: ValueNode // deprecated
}

export type ValueNode = string | boolean | number | BasicObject | null | undefined | any[]
```

The evaluator takes `OperatorNode | ValueNode` as input, and returns `Promise<ValueNode>`

src/modules/expression-evaluator/src/evaluateExpression.ts
andreievg: I was worried about error being thrown (if body is not valid), but it's all wrapped with error handlers, so should be all good

andreievg (approved):
Looks good! cheers

Issue #352, opened 2021-05-12
Configure insertData script to work when row-level permissions are enabled


CarlosNZ:
- Create a policy that allows reading/writing to whole database
- Create a user with this permission (or just use existing Admin user)
- Login as this user, store the JWT
- Make insertData script include this JWT in its mutation headers

PR #355, opened 2021-05-13
#352 Graphile user can insert data

CarlosNZ:
Came up with a simpler way for insertData to work while PostGraphile is logged in as `graphile_user` (i.e. row-level security enforced)

Basically, I just moved the commands to turn on ROW LEVEL SECURITY out of the main "buildSchema" folder, and now they run separately _after_  insertData has finished.


andreievg (approved):
sweet.

Would envision us turning row level policies on/off based on policies generated via server (someday)

PR #356, opened 2021-05-13
698F and 322 Dashboard

andreievg:
Fixes: #322 

Require for front end branch `dashboard`

Changes move from demo branch

### Changes

* Add filters and categories to schema
* Create default filters for most templates
* Categories templates
* Fixed an issue where mutliple permissions were returned for user
* Also added changes to initialise scripts to make sure it stops on errors


database/buildSchema/07B_template.sql
```diff
@@ -8,11 +8,12 @@ CREATE TYPE public.template_status AS ENUM (
 CREATE TABLE public.template (
     id serial PRIMARY KEY,
     name varchar,
-    code varchar NOT NULL,
+    code varchar NOT NULL UNIQUE,
```
CarlosNZ, line 11: I don't think `code` can be UNIQUE since we can have multiple versions of the same template -- the `AVAILABLE` one is the "real" one.

database/buildSchema/07B_template.sql
```diff
@@ -8,11 +8,12 @@ CREATE TYPE public.template_status AS ENUM (
...
     is_linear boolean DEFAULT TRUE,
     start_message jsonb,
     status public.template_status,
     submission_message jsonb DEFAULT '{"value": "Thank you! Your application has been submitted."}' ::jsonb,
```
CarlosNZ, line 15:
```suggestion
    submission_message jsonb DEFAULT '"Thank you! Your application has been submitted."' ::jsonb,
```
No need to write expressions in this old {value:...} style (and string literals are valid JSON).

database/buildSchema/10B_template_filter.sql
```diff
@@ -0,0 +1,17 @@
+-- template filter tables
+CREATE TABLE public.filter (
+    id serial PRIMARY KEY,
+    code varchar NOT NULL UNIQUE,
+    icon_color varchar DEFAULT 'blue',
+    icon varchar,
+    title varchar,
+    query jsonb,
+    user_role public.permission_policy_type
+);
```
CarlosNZ, line 10: So this table could also be used to store any filter, right? So if we allowed users to "bookmark" certain filtering configs for their list, this is where we'd store them?

database/buildSchema/10B_template_filter.sql
```diff
@@ -0,0 +1,17 @@
...
+
+CREATE TABLE public.template_filter_join (
+    id serial PRIMARY KEY,
+    template_id integer REFERENCES public.template (id),
+    template_filter_id integer REFERENCES public.filter (id)
```
CarlosNZ, line 15: This column should probably just be called `filter_id`

database/insertData/08_template_filters.js
```diff
@@ -0,0 +1,343 @@
...
+                }
+              }
+              {
+                templateToTemplateId: {
+                  connectByCode: { code: "UserRegistration" }
+                }
+              }
+              { templateToTemplateId: { connectByCode: { code: "ReviewTest" } } }
+              { templateToTemplateId: { connectByCode: { code: "DrugRegoGen" } } }
+              { templateToTemplateId: { connectByCode: { code: "OrgJoin" } } }
```
CarlosNZ, line 30: So this is why you made `code` unique (above)?

Perhaps the "template_id" field in template_filter_join needs to be a "soft" reference to template_code, rather than a foreign key. That's how we've done it elsewhere (for template elements, for example).

And also, if the AVAILABLE (Active) version of a template changes, we'd have to update all the template_id references. So linking by "code" would be better, it just wouldn't be baked into the database.

But that's okay -- sometimes we want a hard-link to an exact template versions (e.g. for existing applications), whereas sometimes we want a "soft" link to a template code, which will always stay linked to the "AVAILABLE" version of the template.

So, conclusion? Perhaps change "template_id" field in "template_filter_join" to "template_code" and not make it a foreign key.

CarlosNZ (changes requested):

database/buildSchema/10B_template_filter.sql
andreievg: yep, can add some more join tables to user etc..

database/buildSchema/07B_template.sql
andreievg: * remove constraint
* added categories and filters before templates
* joins categories and filters to templates in template mutation
* changes core_actions to core_mutations and added joinFilters to it

069c0c86f7f6df89fc2741ac7537c022bcfa10e5

database/buildSchema/07B_template.sql
andreievg: 447877af67391a8a7fa0f6d883766a8b8e999bcb

database/buildSchema/10B_template_filter.sql
andreievg: 447877af67391a8a7fa0f6d883766a8b8e999bcb

database/insertData/08_template_filters.js
andreievg: https://github.com/openmsupply/application-manager-server/pull/356#discussion_r632177974

CarlosNZ (approved):
Looks good -- thanks :)

database/insertData/03_template_E_drugRegistration.js
```diff
@@ -932,6 +932,8 @@ exports.queries = [
               }
```
nmadruga, line 932: Please remove lines 20-22 not needed anymore : )

nmadruga (approved):
Nice, just some quick things to add, please

database/buildSchema/10B_template_filter.sql
```diff
@@ -0,0 +1,17 @@
...
+    title varchar,
+    query jsonb,
+    user_role public.permission_policy_type
+);
+
+CREATE TABLE public.template_filter_join (
+    id serial PRIMARY KEY,
+    template_id integer REFERENCES public.template (id),
+    filter_id integer REFERENCES public.filter (id)
+);
```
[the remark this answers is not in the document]
CarlosNZ: Nooooo....

database/insertData/03_template_E_drugRegistration.js
andreievg: 7dfb3ea28ca36d95152f8e8aa518d6aaa17c035b goneski

database/buildSchema/10B_template_filter.sql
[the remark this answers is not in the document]
andreievg: not sure if there is an agreement here. so would leave it for now

PR #357, opened 2021-05-13
#351 Evaluator - Combine CONCAT/ + operators, also merges objects

### LINKED ISSUE #351: Expression Evaluator: deprecate "CONCAT" operator (just use "+"), and extend "+" operator to also merge objects.
"+" should act on Objects by combining properties.

So if one node returns `{ one: 1, two: 2 }` and the other node returns `{ three: 3 }`, the operation result will be:
```
{
  one: 1,
  two: 2,
  three: 3
}
```

CarlosNZ:
Fixes #351.

Removed "CONCAT" operator (it's actually still active, but it falls back to "+") since everything can be done using "+":
- Arrays and strings get joined into either an array or a string, depending on the "type" parameter, or the type of the first node (e.g. `[1, 2, 3] + "three" => [1,2,3,"three"]`, but `"three" + [1,2,3] => "three1,2,3"`
- objects get merged together (if all children are objects), e.g. `{ one: 1, two: 2 } + { three: 3 } => { one: 1, two: 2, three: 3 }`
- any other types, or mixtures of, get combined using whatever the javascript "+" would return (e.g. ` 3 + 3 =>6`, `true + true => 2`, etc.)

Have updated tests:
- existing "CONCAT" tests have been moved into the "+" section, and their operators changed, but the test expressions and results are the same
- added a couple of Object merging tests


src/modules/expression-evaluator/src/evaluateExpression.ts
```diff
@@ -60,10 +50,27 @@ export default async function evaluateExpression(
...
+      if (childrenResolved.every((child) => typeof child === 'string' || Array.isArray(child)))
+        return childrenResolved.reduce((acc, child) => acc.concat(child))
+
+      // Merge objects
+      if (childrenResolved.every((child) => child instanceof Object && !Array.isArray(child))) {
+        return childrenResolved.reduce((acc, child) => ({ ...acc, ...child }), {})
+      }
+
+      // Or just try to add any other types
+      return childrenResolved.reduce((acc: number, child: number) => acc + child)
```
andreievg, line 73: should probably check if it's 'CONCAT' and do default string concat , otherwise  + ?

andreievg (changes requested):
One minor use case, as per blow comment.

src/modules/expression-evaluator/src/evaluateExpression.ts
CarlosNZ: I was trying to deprecate "CONCAT" completely, I only left it in as a fallback for backwards compatibility, so I'd rather they didn't behave differently.

The idea is, if you want to force concatenation, you set `type: "string"`  e.g.
```
{
operator: "+",
type: "string",
children: [1, 2, 3]
}
```
will return `"123"` (as opposed to `6`)

andreievg: ok, i was kind of referring to backwards compatibility, I am not sure if it's the case in our config, but concat logic is different now and it would add number and booleans instead of concating them. I guess it's not a biggy, i'll approve but would be good to check if we are concating numbers of some reason

andreievg (approved):

src/modules/expression-evaluator/src/evaluateExpression.ts
CarlosNZ: Yeah, I tried to cover a pretty wide range of possibilities in the test cases, and none of them have changed their output result, so we should be good. (We're mostly using stringSubstitution for string manipulation in the wild currently, and I just checked and all CONCAT usage in the templates are for simple string joins, no numbers)

PR #359, opened 2021-05-14
#345 #346 review/assign org policy

### LINKED ISSUE #345: Create new policy reviewJoinCompany and add only view permissions specific enforced (on application, review and reviewAssignment)

CarlosNZ:
Fixes #345 #346

Minor extensions to existing policies to also consider `review` and `review_assignment` table (view) restrictions.

Also, added more global "INSERT/UPDATE all" policies (hard-coded) until we add these restrictions to policies

And added action to UserRegistration to grant permissions to OrgRego on submit.


database/insertData/01_permission_policies.js
```diff
@@ -95,16 +138,26 @@ exports.queries = [
...
         permissionPolicy: {
-          name: "basicAssign"
+          name: "assignBasic"
           rules: {
             application: {
               view: { template_id: "jwtPermission_bigint_templateId" }
             }
+            review: {
+              view: {
+                reviewer_id: "jwtUserDetails_bigint_userId"
```
nmadruga, line 153: Shouldn't be `assigner_id` instead? Also wouldn't all reviews be visible to Assigner that have permission to assign? So not sure how to restrict visible review/review_assignment to them. Because it would be either if: 
- assigned by this user
- Not fully assigned and user has permission to assign template.

src/components/permissions/rowLevelPolicyHelpers.ts
```diff
@@ -86,11 +86,18 @@ const updateRowPolicies = async () => {
...
 
-  // Temporarily hard-code create/update permissions on application table
+  // Temporarily hard-code create/update permissions RLS-enabled tables
   await databaseConnect.query({
     text: `
-    CREATE POLICY "create_all" ON application FOR INSERT WITH CHECK (true);
-    CREATE POLICY "update_all" ON application FOR UPDATE USING(true) WITH CHECK (true);`,
+    CREATE POLICY "create_all_application" ON application FOR INSERT WITH CHECK (true);
+    CREATE POLICY "update_all_application" ON application FOR UPDATE USING(true) WITH CHECK (true);
+    CREATE POLICY "update_all_application_response" ON application_response FOR INSERT WITH CHECK (true);
```
nmadruga, line 94:
```suggestion
    CREATE POLICY "create_all_application_response" ON application_response FOR INSERT WITH CHECK (true);
```

src/components/permissions/rowLevelPolicyHelpers.ts
```diff
@@ -86,11 +86,18 @@ const updateRowPolicies = async () => {
...
+    CREATE POLICY "update_all_application_response" ON application_response FOR UPDATE USING(true) WITH CHECK (true);
+    CREATE POLICY "update_all_review" ON review FOR INSERT WITH CHECK (true);
```
nmadruga, line 96:
```suggestion
    CREATE POLICY "create_all_review" ON review FOR INSERT WITH CHECK (true);
```

src/components/permissions/rowLevelPolicyHelpers.ts
```diff
@@ -86,11 +86,18 @@ const updateRowPolicies = async () => {
...
+    CREATE POLICY "update_all_review" ON review FOR UPDATE USING(true) WITH CHECK (true);
+    CREATE POLICY "update_all_review_assignment" ON review_assignment FOR INSERT WITH CHECK (true);
```
nmadruga, line 98:
```suggestion
    CREATE POLICY "create_all_review_assignment" ON review_assignment FOR INSERT WITH CHECK (true);
```

nmadruga (changes requested):
Just a few changes requests please

nmadruga:
Hmmm I'm not really sure how to test this. I tried logging with different Reveiwer/Assigner and clicking to see the list of application of ReviewTest, but all users can see all applications. Do you have some user cases on how to test it?

CarlosNZ:
> Hmmm I'm not really sure how to test this. I tried logging with different Reveiwer/Assigner and clicking to see the list of application of ReviewTest, but all users can see all applications. Do you have some user cases on how to test it?

You need to launch the server with `yarn dev_pg_permissions`. Then only those who are allowed to see certain applications will have access to view them.

database/insertData/01_permission_policies.js
CarlosNZ: Yeah, I'm not sure either. I think for assigner we just want to restrict access to review_assignment by template only (so they can see all the assignment possibilities). But because `template_id` is not a field on review_assignment, it'll probably have to be one of those nested queries Andrei was talking about. Will figure it out. Thanks.

src/components/permissions/rowLevelPolicyHelpers.ts
CarlosNZ: Oops, yes. d6b5056c

database/insertData/01_permission_policies.js
andreievg: create check file, and run it with ts-node
```
import { getSqlConditionFromJSON } from './helpersUtilities'
console.log(
  getSqlConditionFromJSON({
    reviewer_id: {
      $in: {
        $select: {
          id: true,
          $from: 'application',
          $where: { template_id: 'jwtPermission_bigint_templateId' },
        },
      },
    },
  })
)
```

CarlosNZ: Okay, _finally_ managed to get this insert to work. Put the rules into a variable as per @andreievg 's suggestion, and updated insertData.js to look for queries containing variables (by checking for Object or string)

ee29828a

src/components/permissions/rowLevelPolicyHelpers.ts
```diff
@@ -86,11 +86,18 @@ const updateRowPolicies = async () => {
...
 
-  // Temporarily hard-code create/update permissions on application table
+  // Temporarily hard-code create/update permissions RLS-enabled tables
   await databaseConnect.query({
     text: `
-    CREATE POLICY "create_all" ON application FOR INSERT WITH CHECK (true);
-    CREATE POLICY "update_all" ON application FOR UPDATE USING(true) WITH CHECK (true);`,
+    CREATE POLICY "create_all_application" ON application FOR INSERT WITH CHECK (true);
+    CREATE POLICY "update_all_application" ON application FOR UPDATE USING(true) WITH CHECK (true);
+    CREATE POLICY "create_all_application_response" ON application_response FOR INSERT WITH CHECK (true);
```
andreievg, line 94: this is not turned on yet, i guess it doesn't hurt having it here.

andreievg (approved):
LGTM.

fyi, in demo server i had something like this (permission name wise, just showing the structure, can't remember what the names were)

noOrgApply
* granted on user registration
* APPLY -> org join
* APPLY -> org registration

reviewOrgJoin
* granted on org rego approval
* REVIEW -> org registration

applyWithOrg
* granted on org rego approval or org join approval
* APPLY -> all others

orgRegoReview
* granted (not configured yet, but had a couple of self assign reviewers for it)
* REVIEW -> org rego

database/insertData/01_permission_policies.js
nmadruga: 🎉 🎉 🎉 🎉 🎉

nmadruga:
> > Hmmm I'm not really sure how to test this. I tried logging with different Reveiwer/Assigner and clicking to see the list of application of ReviewTest, but all users can see all applications. Do you have some user cases on how to test it?
> 
> You need to launch the server with `yarn dev_pg_permissions`. Then only those who are allowed to see certain applications will have access to view them.

Thanks! I was not aware. I'll try this now : )

nmadruga:
Hmmm noticed that no Dashboards are showing for any user -- Do you know what could be blocking that? Do we need to upgrade all queries to use the new permissions, so we make sure these are always linking the JWT token or is this already in place?

nmadruga:
> Hmmm noticed that no `Dashboard` are showing for any user -- on latest front-end `master` branch. Do you know what could be blocking that? Do we need to upgrade all queries to use the new permissions, so we make sure these are always linking the JWT token or is this already in place?

Cool - This seems to be fixed after getting latest changes on this PR and running the `yarn`.

nmadruga:
I'm still seeing one thing which seems like this should be happening:
- Logged in as **Reviewer1**, I can see the list of `ReviewTest` with all 4 applications, even though this user only has been assigned to 2 of them (in the current stage). If I try to open the review page it correctly say "Page not found", but the list should also reflect the permissions for this user.

nmadruga:
Another problem - Not sure if related (I have already create one for refreshing the list based on current user in a separated [Front-end Issue#719](https://github.com/openmsupply/application-manager-web-app/issues/719)):
- When I go to this page: http://localhost:3000/application/ABC123/review it shows the result "Not found" to user **Reviewer2** correctly, but when changing to user **Reviewer1** which has permissions it show this error:
Screen Shot 2021-05-19 at 11 26 28 AM

PS: This is running back-end as `yarn pg_dev_permissions`

CarlosNZ:
> I'm still seeing one thing which seems like this should be happening:
> 
>     * Logged in as **Reviewer2**, I can see the list of `ReviewTest` with all 4 applications, even though this user only has been assigned to 2 of them (in the current stage).
> 
>     * If I try to open the review page it correctly say "Page not found"
> 
> 
> But the list should also reflect the permissions for this user.

This is behaving correctly in terms of what is currently defined in the policies. The policy "reviewBasic" is only restricted by template_id for viewing _applications_ (so all reviewers who have this permissions will see them in the list), whereas the _reviews_ are restricted by reviewer as well, hence they won't be able to see the review page for ones they're not a reviewer on. 

We could add an additional restriction to restrict application viewing by "assigned", but not sure we want to quite yet. Do we?

CarlosNZ:
>     * When I go to this page: http://localhost:3000/application/ABC123/review it shows the result "Not found" to user **Reviewer2** correctly, but when changing to user **Reviewer1** which has permissions it show this error:

Refreshing the page works, so it must be a front-end error in our "user switcher" tool. Looks like your [new issue ](https://github.com/openmsupply/application-manager-web-app/issues/719) is exactly what's going on.

nmadruga:
> We could add an additional restriction to restrict application viewing by "assigned", but not sure we want to quite yet. Do we?

I think that's the idea for regular reviewers. So it's basically 2 different permissions. The users who can see any applications (which would be more common for Chief in the department) but usually Consolidators would also get this permission. And the Reviewer level one who only can view application assigned to them.

nmadruga:
> > ```
> > * When I go to this page: http://localhost:3000/application/ABC123/review it shows the result "Not found" to user **Reviewer2** correctly, but when changing to user **Reviewer1** which has permissions it show this error:
> > ```
> 
> Refreshing the page works, so it must be a front-end error in our "user switcher" tool. Looks like your [new issue ](https://github.com/openmsupply/application-manager-web-app/issues/719) is exactly what's going on.

Yes, indeed. To be investigated. 🕵️

nmadruga (approved):

andreievg:
@CarlosNZ as per our conversation.

It's quite possible to create circular policies (in any system tbh, but since the current design is configurable, it's quite susceptible to this effect).

We kind of need to make internal rules for order of dependencies in policy configurations (they don't need to be enforced, just documented and presented to administrators). This potentially would require creating extra generated columns.

For existing queries the dependency tree would need to be as follows: [diagram](https://app.diagrams.net/#G1Y151loQbPkMlgpoftUODdWI_PfR4Yw78)

Permission Policy Dependency Graph

This means that review cannot use review_response or application or application response in it's policy, but can use review_assignment.

Application cannot use application-response in it's policy but can use review and review assignment

etc..

For existing use cases, we need generated column of template_id at least on the review_assignment

There is another way to solve this, but it's more complex and less performant, so decided not to explore it any further at this stage.

CarlosNZ:
Okay, @andreievg -- I have added `template_id` generated col to review_assignment, and tidied up the policies, if you wanted to check it?

database/insertData/01_permission_policies.js
```diff
@@ -95,24 +108,99 @@ exports.queries = [
...
+                  id: true,
+                  $from: 'application',
+                  $where: {
+                    template_id: 'jwtPermission_bigint_templateId',
+                  },
+                },
+              },
+            },
+          },
+        },
```
andreievg, line 197:
```suggestion
        review: {
          view: {
            review_assignment_id: {
              $in: {
                $select: {
                  id: true,
                  $from: 'review_assignment',
                  $where: {
                    template_id: 'jwtPermission_bigint_templateId',
                  },
                },
              },
            },
          },
        },
```

andreievg:
> Okay, @andreievg -- I have added template_id generated col to review_assignment, and tidied up the policies. Seems to be behaving as expected with the different reviewers. If you wanted to have a quick look-over?

Just one more little change, and ready to go

andreievg (approved):

PR #360, opened 2021-05-14
Quick fix -- error thrown on bad login

CarlosNZ:
Just noticed while doing something else that when an invalid username was provided for `/login` the response was an error message rather than `success: false`.

2 quick changes:
- wrapped whole login method in `try/catch`
- additional check for array length of `userOrgInfo` after querying database -- return false if it's `0`

Also, there was some formatting weirdness in `postgresConnect` that Prettier has tidied up -- no actual changes.


andreievg (changes requested):
lgtm

PR #361, opened 2021-05-16
#Update evaluator, and UserEdit template

CarlosNZ:
Bumped evaluator version with recent additions (POST operator). (required updating a couple of types that use Evaluator types)

Then updated UserEdit template to use new POST operator for password check. It requires entry of correctly verified current password before new password can be entered (see front-end PR for changes made to Password element)

Requires [front-end branch #701](https://github.com/openmsupply/application-manager-web-app/pull/701)

**To consider**: after user details have been updated, the "currentUser" still has info from the previous login, so if you've changed your name (for example), it won't show up in the UI. What's the best way to handle this? For now, I think we should probably force a logout, so the user must login with their new details. Any other suggestions?


nmadruga (approved):
Seems legit. Will be testing on Front-end review.

nmadruga (commented):
👍 Thanks for changes

PR #310, opened 2021-04-15
#309 dockerise local build of combined back and front end

andreievg:
Fixes: #309

Branched from #306 and from `snapshot` (maybe best to merge that one before reviewing)

### Changes

* Added documentation
* Added `#!/bin/bash` to all scripts
* Added build files in `Docker` folder

### Testing

Make sure you have the latest version of docker (and that it's running), I use Docker app for Mac OS

`cd Docker`

`./dockerise.sh` -> this would take awhile (make sure to add githubtoken.txt, as per instructions in script and in docs)

then (make sure to stop anything on port 3000)

`./run.sh` go to http://localhost:3000 after '--- Ready to go -> http://localhost:3000'

PR #365, opened 2021-05-18
#310 new filters by reviewer and assigner

nmadruga:
Fixes #310 

## Front-end branch to use
[PR#310B](https://github.com/openmsupply/application-manager-web-app/pull/715)

### Changes
- Add another function `assignment_list` to create 2 arrays of usernames for `reviewer_usernames` and `assigner_usernames` that only returns associated reviewers and assigners to each application (on the current stage).
- Problem: assigners are always set to NULL, and this returns an array like: [NULL]. Couldn't figure out how to change this to an empty array. Any ideas are welcome. The commented WHERE statement was what I was trying but not sure where to add that to make it work only for `assigner_usernames`


andreievg (approved):
Looks great !

PR #367, opened 2021-05-19
#674F replace actions filter from number to enum type

nmadruga:
### To be run with Front-end PR:
[#674](https://github.com/openmsupply/application-manager-web-app/pull/722)

### Changes
- Add new ENUMs `ReviewerAction` and `AssignerAction` to be returned for each application on the `application_list_view`
- Create 2 new function to generated the `reviewer_actin` and `assigner_action` - Updated 20/05: **FIXED**
- UPDATED 20/05: Revert to have functions `assigned_questions_count` and `template_questions_count ` in same file
- UPDATED 20/05: Call above function on `assigner_action_list` and return `is_fully_assigned_level_1` which is used only for Level 1 (for now at least) along with new ENUM for assigner_action - seemed more straight forward than passing through the `is_fully_assigned_level_1` in.
- Updated filters to work on Dashboard using ENUM instead of counts


database/buildSchema/28B_review_is_fully_assigned_list.sql
```diff
@@ -0,0 +1,64 @@
...
+    RETURNS TABLE (
+        application_id int,
+        is_fully_assigned boolean,
+        assigned_questions bigint,
+        total_questions bigint
+    )
+    AS $$
+    SELECT
+        app.id AS application_id,
+        CASE
```
andreievg, line 53: i haven't checked, but can this be done without a case ? assigned_questions_count(app, $1, $2) = template_questions_count(app) as is_fully_assigned

database/buildSchema/38_application_list_view.sql
```diff
@@ -72,8 +64,9 @@ CREATE FUNCTION application_list (userid int DEFAULT 0)
     LEFT JOIN application_stage_status_latest AS stage_status ON app.id = stage_status.application_id
     LEFT JOIN organisation org ON app.org_id = org.id
     LEFT JOIN assignment_list (stage_status.stage_id) ON app.id = assignment_list.application_id
-    LEFT JOIN review_list ($1) ON app.id = review_list.application_id
-    LEFT JOIN assigner_list ($1) ON app.id = assigner_list.application_id
+    LEFT JOIN review_list (stage_status.stage_id, $1) ON app.id = review_list.application_id
+    LEFT JOIN review_is_fully_assigned_list (stage_status.stage_id, 1) AS assignment_status ON app.id = assignment_status.application_id
+    LEFT JOIN assigner_list (stage_status.stage_id, assignment_status.is_fully_assigned, $1) ON app.id = assigner_list.application_id
```
andreievg, line 69: how would this work ? I think I missed this in the previous PR of filter by reviewer and assigner. i.e. what happens when you pass a table field (which is different for every row), to a method that returns a table that relies on it ?

andreievg (commented):
Still reviewing, but had a question, wonder if you know the answer

database/buildSchema/28B_review_is_fully_assigned_list.sql
nmadruga: Good call - I was doing something else before and it seemed better to do that, but now definitely I can change to return as suggested. I'll have a look on this tomorrow though - have been working on it for so long, I need a break : D

database/buildSchema/38_application_list_view.sql
nmadruga: Not sure if I understood. I'll check with you in person in a sec

nmadruga: Hope this is better to read now. But basically the idea is that `application_list_view` would generate a table with all applications (visible to the current user) and fields as `reviewer_action`, `assigner_action` if this user has any actions associated and in another fields `reviewer_usernames` and `assigner_usernames` we return associated usernames which are reviewers and assigners to this application. And now that I have said that, It seems better to use some rowLevel permissions on all these fields - to avoid retrieving sensitive information for Applicant users.

nmadruga:
This is ready for a re-review now. I have finished all the work-in-progress

database/buildSchema/38_application_list_view.sql
andreievg: I like the solution I just wasn't sure how postrgres handles a method call with a column key of selected records (since there is a 'where' statement within the condition.

andreievg (approved):
looks really good! I'll test it when reviewing front end branch

Issue #364, opened 2021-05-17
When LOQ or non conformity decision is made lock other reviews


nmadruga:

andreievg:
https://app.diagrams.net/#G1VIyFc7NjAEShXOQxgJoOvKgKrxbtsY0q

As per above diagram, should lock reviews when consolidation is submitted for anything other then changes requested

nmadruga:
@andreievg I think it would be the case for a consolidation submitted as Changes requested as well. Imagine this case:
1. Reviewer 1 has a draft partial review of Application A
2. Reviewer 2 SUBMIT a partial review of Application A
3. Consolidator SUBMIT back to Reviewer2 with CHANGES REQUESTED
4. Reviewer1 should have their review LOCKED to avoid SUBMITTING and changing Consolidator back to PENDING

If we allow Reviewer1 to SUBMIT while Reviewer2 has CHANGES_REQUESTED status, how would this show to each user?

nmadruga:
Let's try to describe the possible cases here (this will also be available on updated docs I'm planning to create a PR for today BTW - issue #726), when a review is LOCKED/UNLOCKED: 
1. Locking reviews after CHANGES_REQUESTED to Reviewer:
   - Update status to LOCKED for reviews is in DRAFT (lower level to consolidation submitted)
   - Update status to DRAFT for reviews in same level on a review re-submission after CHANGES_REQUEST
2. Locking reviews after LOQ or NO_CONFORM decision to Applicant:
   - Update status to LOCKED for reviews is in DRAFT (lower level to consolidation submitted or same-stage if we consider more than one reviewer with `is-last-level` on that stage)
   - Update status to DRAFT for reviews in first level on a application re-submission after CHANGES_REQUIRED
Q: Locking start of review with assignment as ASSIGNED when CHANGES_REQUESTED to reviewers or when consolidation submits a LOQ or NO_CONFORM to applicant.

andreievg:
As far as I can see its, it's fine to have concurrent change requests to reviewers.

andreievg:
So in my view the logic is this (I think the diagram tried to convey it, but maybe it's too complex):

When consolidator submits LOQ or non conform. Lock all previous level reviews that are draft.

When applicant submits an application we unlock locked ones (which is already implemented)

andreievg:
What to do for reviews that are not started, but LOQ or non conform was submitted ?

If consolidation was only one level we can add extra check to see if application is 'submitted'.

[...]

PR #368, opened 2021-05-20
363 update other reviews on changes request or submitting updated review

### LINKED ISSUE #363: Change status of reviews to "changes_requested" & consolidation to "pending" after a related review is submitted

nmadruga:
Fixes #363 

### Changes
- Update docs in `List-od-Action-plugin`, `plugin.json` and in `core_mutations` with latest changes to be extending `updateReviewStatuses` to also be used for **Review submitted**. 
- Updated insert file `07_review_assignments`:
  - Application 4000 has `consolidator1` already existing in DRAFT
  - Application 4001 has Changes required to the applicant SUBMITTED (by level 1 Reviewer in Stage 1), 
- Updated tests on file `updateReviewStatuses.test.ts` out-dated and added 2 new ones to check new cases for Review submission: 1) Check it is changing to CHANGES REQUESTED when upper level review submits a review with disagreements & 2) Check it's changing upper level reviews to PENDING after the re-submission is done 
- Current implementation will change status of `lower-level` when a consolidation review is submitted to CHANGES_REQUESTED (only if any review_question_assignment matches disagreements)
- Current implementation will change the current status of `same-level` reviews and `higher-level` reviews to:
- For `same-level` reviews, when LOCKED they will change to DRAFT ~~(or kept as LOCKED if other CHANGES_REQUESTED reviews in the same level)~~ REMOVED
- For `higher-level` reviews, when SUBMITTED they will change to PENDING


nmadruga:
To test, run this suit test only - probably we still have the problem to run others - since ids keep changing (and I added another review...): `yarn test plugins/action_update_review_status/src/updateReviewsStatuses.test.ts`

CarlosNZ (commented):
Great, thanks heaps for updating this and making the tests work, it's a pretty complicated one to make sense of.

Just a couple of very minor comments.

database/buildSchema/31_review_decision.sql
```diff
@@ -1,16 +1,16 @@
...
-    'List of questions',
-    'Conform',
-    'Non-conform',
-    'CHANGES_REQUESTED',
-    'No Decision'
+    'List of questions', -- TODO: 'LIST_OF_QUESTIONS'
+    'Conform', -- TODO: 'CONFORM'
+    'Non-conform', -- TODO: 'NON_CONFORM'
+    'CHANGES_REQUESTED', 
+    'No Decision' -- TODO: 'NO_DECISION'
```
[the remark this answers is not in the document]
nmadruga: He is checking this now - will be probably in another PR

documentation/List-of-Action-plugins.md
```diff
@@ -246,12 +246,17 @@ When an applicant re-submits an application after making changes, this Action up
...
-| `changedApplicationResponses` [Array of applicationResponseIds] |                   |
+| Input parameters<br />(\*required) <br/>               | Output properties |
+| ------------------------------------------------------ | ----------------- |
+| `applicationId`                                        | `updatedReviews`  |
+| `reviewId`                                             |                   |
+| `changedResponses`\* [Array of applicationResponseIds] |                   |
 
-**Note:** If `applicationId` is not provided, the plugin will attempt to fetch it from `applicationData`
+**Note:** - If `applicationId` or `reviewId` is not provided, the plugin will attempt to fetch it from `applicationData`
+**Note 2:** - If `reviewId` is received, the Action will be updating status of related reviews of same stage in the current and next level reviews.
```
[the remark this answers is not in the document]
nmadruga: Hmm good point. The way this is configured, it will run as a Review submission in this case, so maybe we need to prevent that somehow when an application is submitted (and the `reviewId` is also passed through `applicationData`)

I'll update to docs

plugins/action_update_review_status/src/updateReviewsStatuses.ts
```diff
@@ -18,38 +18,96 @@ const updateReviewsStatuses: ActionPluginType = async ({
...
-  const { changedApplicationResponses = [] } = parameters
+  const isReview = !!reviewId
+  const changedResponses = parameters.changedResponses || []
   const stageId = parameters?.stageId || applicationData?.stageId
-  const currentReviewLevel = applicationData?.reviewData?.levelNumber || 0
-  const level = parameters?.level || currentReviewLevel + 1
+  const currentReviewLevel = parameters.level || applicationData?.reviewData?.levelNumber || 0
+
+  if (isReview) console.log('Review submitted id', reviewId, 'level ', currentReviewLevel)
+  else console.log('Application submitted id', applicationId)
```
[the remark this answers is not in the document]
nmadruga: Cool. Will update this one

documentation/List-of-Action-plugins.md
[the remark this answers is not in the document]
CarlosNZ: > so maybe we need to prevent that somehow when an application is submitted (and the `reviewId` is also passed through `applicationData`

In which case, should we implement the explicit `isReview` parameter as per some of the other Actions?

plugins/action_update_review_status/src/updateReviewsStatuses.ts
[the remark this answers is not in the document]
nmadruga: Done: 1f3d3cf

plugins/action_update_review_status/src/updateReviewsStatuses.ts
```diff
@@ -18,38 +18,96 @@ const updateReviewsStatuses: ActionPluginType = async ({
...
 
-  console.log('Updating reviews status...')
-
+  // Get application/reviewId from applicationData if not provided in parameters
   const applicationId = parameters?.applicationId ?? applicationData?.applicationId
+  const reviewId = parameters?.reviewId ?? applicationData?.reviewData?.reviewId
+  const decision = applicationData?.reviewData?.latestDecision?.decision || Decision.NoDecision
 
-  const { changedApplicationResponses = [] } = parameters
+  const isReview = !!reviewId
```
andreievg, line 26: would prefer something a little bit more descriptive, `isTriggeredByReview` or `triggeredBy: 'REVIEW' | 'APPLICATION'`

plugins/action_update_review_status/src/updateReviewsStatuses.ts
```diff
@@ -18,38 +18,96 @@ const updateReviewsStatuses: ActionPluginType = async ({
...
+      if (decision === Decision.ChangesRequested) {
+        const previousLevelReview = currentReviewLevel - 1
+        // Update lower level reviews with assigned responses on the array changed responses ...
+        const reviewsSubmitted = await getReviewsByLevelAndStatus(previousLevelReview, [
+          ReviewStatus.Submitted,
+        ])
+        // Deduce which ones should be updated to Pending
+        for (const review of reviewsSubmitted) {
+          const { reviewAssignmentId } = review
+          if (await haveAssignedResponsesChanged(reviewAssignmentId))
```
andreievg, line 65: There is a [diagram](https://app.diagrams.net/#G1VIyFc7NjAEShXOQxgJoOvKgKrxbtsY0q]), but it doesn't quite describe this in enough details. We should also check that change review response is of status 'disagree'

Screen Shot 2021-05-21 at 12 16 00 PM

plugins/action_update_review_status/src/updateReviewsStatuses.ts
```diff
@@ -18,38 +18,96 @@ const updateReviewsStatuses: ActionPluginType = async ({
...
+          ReviewStatus.Submitted,
+        ])
+        submittedReviews.forEach((review) =>
+          reviewsToUpdate.push({ ...review, reviewStatus: ReviewStatus.Pending })
+        )
+      }
+    } else {
+      // For Application submission
+      // Get reviews/review assignments (with status) matching application_id & current stage
+      const reviews = await getReviewsByLevelAndStatus(currentReviewLevel || 1, [
```
andreievg, line 99: maybe we should always just use level 1 when application submitted

plugins/action_update_review_status/src/updateReviewsStatuses.ts
```diff
@@ -18,38 +18,96 @@ const updateReviewsStatuses: ActionPluginType = async ({
...
+        ReviewStatus.Submitted,
+        ReviewStatus.Locked,
+        ReviewStatus.Draft,
+      ])
+      // Deduce which ones should be updated to Pending
+      for (const review of reviews) {
+        const { reviewAssignmentId, levelNumber, reviewStatus } = review
+        if (levelNumber > 1) reviewsToUpdate.push({ ...review, reviewStatus: ReviewStatus.Pending })
```
andreievg, line 107: I think this was previously not differentiating between submission from review or application, now that it does we don't need this condition

plugins/action_update_review_status/src/updateReviewsStatuses.ts
```diff
@@ -18,38 +18,96 @@ const updateReviewsStatuses: ActionPluginType = async ({
...
+          reviewsToUpdate.push({ ...review, reviewStatus: ReviewStatus.Pending })
+        )
+      }
+    } else {
+      // For Application submission
+      // Get reviews/review assignments (with status) matching application_id & current stage
+      const reviews = await getReviewsByLevelAndStatus(currentReviewLevel || 1, [
+        ReviewStatus.Submitted,
+        ReviewStatus.Locked,
+        ReviewStatus.Draft,
```
andreievg, line 102: Hmm i think we should change `statusToUpdate` in `getReviewByLevelAndStatus` to `filterByStatus` and it can be optional parameter (in which case don't filter), and dont't need a filter here I think

plugins/action_update_review_status/src/updateReviewsStatuses.ts
```diff
@@ -18,38 +18,96 @@ const updateReviewsStatuses: ActionPluginType = async ({
...
+        // Deduce which ones should be updated to Pending
+        for (const review of reviewsSubmitted) {
+          const { reviewAssignmentId } = review
+          if (await haveAssignedResponsesChanged(reviewAssignmentId))
+            reviewsToUpdate.push({ ...review, reviewStatus: ReviewStatus.ChangesRequested })
+        }
+      } else {
+        // Review submitted from lower level to upper level
+        // Update same level reviews as submitted if no other still have changes requested status...
+        const othersChangeRequested = await getReviewsByLevelAndStatus(currentReviewLevel, [
```
andreievg, line 71: hmm, I am not sure about the logic in this bit I think you describe it here:

> For same-level reviews, when LOCKED they will change to DRAFT (or kept as LOCKED if other CHANGES_REQUESTED reviews in the same level)

I think it's not needed. locked reviews will be changed to pending when applicant re-submits application, and the only time they will be locked is when consolidator (lvl + 1) submits a decision other then `change request`

plugins/action_update_review_status/src/updateReviewsStatuses.ts
```diff
@@ -18,38 +18,96 @@ const updateReviewsStatuses: ActionPluginType = async ({
...
+          // Locked -> to avoid other reviews submitted while awaiting changes requests
+          reviewsLocked.forEach((review) =>
+            reviewsToUpdate.push({ ...review, reviewStatus: ReviewStatus.Draft })
+          )
+        }
+
+        // Update upper level reviews submitted
+        const nextReviewLevel = currentReviewLevel + 1
+        const submittedReviews = await getReviewsByLevelAndStatus(nextReviewLevel, [
+          ReviewStatus.Submitted,
```
andreievg, line 90: consolidation could be in draft, i don't think we need to filter by status in this case

andreievg (changes requested):
I mainly just reviewed updateReviewStatus.ts, did't check docs etc..

A few changes.

There might be a need for another bit of logic when we do three levels of consolidation, but can leave that for now

plugins/action_update_review_status/src/updateReviewsStatuses.ts
nmadruga: Done on next PR: #369

plugins/action_update_review_status/src/updateReviewsStatuses.ts
nmadruga: Done: da60a4c

plugins/action_update_review_status/src/updateReviewsStatuses.ts
nmadruga: Yes, I was very confused by this diagram. Couldn't really follow for what was required to update reviews:
- From lower level to upper level
- From upper level to lower level

I think the idea is to abstract and the flow is not very intuitive - at least not when I was reading on my own

plugins/action_update_review_status/src/updateReviewsStatuses.ts
nmadruga: Yes, I think that would be ok

plugins/action_update_review_status/src/updateReviewsStatuses.ts
nmadruga: Hmmm I see, I think I agree. Because from this point we would expected that Reviews would be either:
- `Submitted` back to Applicant
- `Locked` for previous `Draft`

But might be a very edge case (when we start changing this current structure of 1 review level in stage 1 and more in other stages, and there could be cases when the level 1 reviewers are in `Changes_requested` or else. So for this we would be changing all of them to `pending`.

nmadruga: Done: bddc898

plugins/action_update_review_status/src/updateReviewsStatuses.ts
nmadruga: Done: bddc898

plugins/action_update_review_status/src/updateReviewsStatuses.ts
nmadruga: Done: bddc898

plugins/action_update_review_status/src/updateReviewsStatuses.ts
nmadruga: Hmmm I think this will require a bigger change, the `changedResponses` need to pass each decision to this action. And this is one array passed as `outputCumulative` from a previous method. I'll ask @CarlosNZ how is the best way to update that.

plugins/action_update_review_status/src/updateReviewsStatuses.ts
nmadruga: Ok, I can remove this

plugins/action_update_review_status/src/updateReviewsStatuses.ts
andreievg: I guess the logic is not very simple, In this case the idea is that when there is a change request submitted, we change status of change requested for submitted lvl -1 review where
* at least one `new` consolidation review response is disagree for template element that's assigned to lvl -1 reviewer

andreievg: new = updatedResponse

plugins/action_update_review_status/src/updateReviewsStatuses.ts
andreievg: thanks for change, as per edge case:

there would be an issue if any review statuses are not `submitted` or `locked` when application is being re-submitted by applicant. I think by trying to avoid this use case we avoid complicated logic further (i.e. having multiple streams of LOQ or decisions)

plugins/action_update_review_status/src/updateReviewsStatuses.ts
andreievg: yeah can do a better job on the diagram, i can make one from the updateReviewsStatuses logic, once it's finalised

nmadruga: Done: 6d7bd4b

plugins/action_update_review_status/src/updateReviewsStatuses.ts
andreievg: Thanks, that be great

andreievg:
Thanks for the changes so far, looking good, only a couple of minor tidy ups now:

https://github.com/openmsupply/application-manager-server/pull/368#discussion_r636622493 -> would be good to remove that bit of logic

Also good use case of `some`:

```JS
    return changedResponses.reduce(
      (isInAssigned: boolean, { templateElementId, reviewDecision }: any) => {
        const assignedResponse = questionAssignments.includes(templateElementId)
        if (triggeredBy === 'REVIEW')
          return (
            isInAssigned || (reviewDecision === ReviewResponseDecision.Disagree && assignedResponse)
          )
        // triggeredBy === 'APPLICATION'
        return isInAssigned || assignedResponse
      },
      false
    )
```
vs
```JS
    return changedResponses.some(
      ({ templateElementId, reviewDecision }: any) => {
        const assignedResponse = questionAssignments.includes(templateElementId)
        if (triggeredBy === 'REVIEW')
          return (
            reviewResponseDecision === ReviewResponseDecision.Disagree && assignedResponse
          )
        // triggeredBy === 'APPLICATION'
        return assignedResponse
      }
    )
```

also reviewResponseDecision seems like a more descriptive name

documentation/List-of-Action-plugins.md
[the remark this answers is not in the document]
nmadruga: Have done something else as per [this suggestion](https://github.com/openmsupply/application-manager-server/pull/368#discussion_r636553093) by Andrei: 
> would prefer something a little bit more descriptive, `isTriggeredByReview` or `triggeredBy: 'REVIEW' | 'APPLICATION'`

nmadruga:
> [#368 (comment)](https://github.com/openmsupply/application-manager-server/pull/368#discussion_r636622493) -> would be good to remove that bit of logic

Done

nmadruga:
> Also good use case of `some`:
> 
> ```js
>     return changedResponses.reduce(
>       (isInAssigned: boolean, { templateElementId, reviewDecision }: any) => {
>         const assignedResponse = questionAssignments.includes(templateElementId)
>         if (triggeredBy === 'REVIEW')
>           return (
>             isInAssigned || (reviewDecision === ReviewResponseDecision.Disagree && assignedResponse)
>           )
>         // triggeredBy === 'APPLICATION'
>         return isInAssigned || assignedResponse
>       },
>       false
>     )
> ```
> 
> vs
> 
> ```js
>     return changedResponses.some(
>       ({ templateElementId, reviewDecision }: any) => {
>         const assignedResponse = questionAssignments.includes(templateElementId)
>         if (triggeredBy === 'REVIEW')
>           return (
>             reviewResponseDecision === ReviewResponseDecision.Disagree && assignedResponse
>           )
>         // triggeredBy === 'APPLICATION'
>         return assignedResponse
>       }
>     )
> ```
> 
> also reviewResponseDecision seems like a more descriptive name

Yes. I agree. Changing to suggestion now : )

nmadruga:
Related [suggestion](https://github.com/openmsupply/application-manager-server/pull/368#issuecomment-845658386)
> Yes. I agree. Changing to suggestion now : )

Done

plugins/action_update_review_status/src/updateReviewsStatuses.ts
```diff
@@ -18,38 +24,91 @@ const updateReviewsStatuses: ActionPluginType = async ({
...
+        // Get all Locked reviews matching application_id, current stage, current level and in Locked status
+        const reviewsLocked = await getReviewsByLevelAndStatus(currentReviewLevel, [
+          ReviewStatus.Locked,
+        ])
+
+        // Now previous locked  to be allowed to continue
+        // Locked -> to avoid other reviews submitted while awaiting changes requests
+        reviewsLocked.forEach((review) =>
+          reviewsToUpdate.push({ ...review, reviewStatus: ReviewStatus.Draft })
+        )
```
andreievg, line 86:
```suggestion

```

https://github.com/openmsupply/application-manager-server/pull/368#discussion_r636638083

Good to go after this one, thanks for changes

plugins/action_update_review_status/src/updateReviewsStatuses.ts
nmadruga: Hi, moving back to this. I though your request was just to remove the logic to check if other CHANGES_REQUEST have been done before changing from LOCKED to DRAFT. 

I agree in removing this for with test and then we can work on a more general solution on issue #364 : )

plugins/action_update_review_status/src/updateReviewsStatuses.ts
nmadruga: Hi, sorry it wasn't clear on your previous suggestion. To me we would LOCK/UNLOCK reviews in the same level when another review gets CHANGE_REQUESTED by consolidator... I'm commenting the code now. And hopefully the confusion can be sorted on the next issue that I have in progress.

andreievg (approved):

PR #369, opened 2021-05-21
#675F - Part 1: Update upper level status on new lower level submission with consolidation in progress

nmadruga:
## To be run with Front-end PR:
[675](https://github.com/openmsupply/application-manager-web-app/pull/730)

### Based oh branch of PR
#368 

### Changes
- New test case to check review submitted for on-going consolidation -> will update to PENDING status
- Quick fix to return `reviewerId` instead of `review_id` in databaseMethods
- Ideally wouldn't be using generate `ids` for reviewResponses... Since these will pretty much all change as soon as we add another review/review_responses on our insert data. But setting ids for all `reviews`, `review_responses` and `decisions` proved to be a too-long-and-difficult task for now. I've it going on in a separated PR, will probably be ready to PR next week.


andreievg (approved):
lgtm

pending small change in 363

PR #373, opened 2021-05-26
Update review_decision and tests using old enum types

nmadruga:
Just a tiny change that was on my TODO list

I haven't checked with all tests... But I think many still need to be updated after latest changed to the test data.

### DIFF
--- a/database/insertData/06_applications_D_reviewTest.js
+++ b/database/insertData/06_applications_D_reviewTest.js
@@ -561,4 +561,118 @@ exports.queries = [
       }
     }
   }`,
+  // Application 5 for Review Testing -- on Stage 2 with consolidation assigned
+  `mutation ReviewTestApplication4 {
+    createApplication(
+      input: {
+        application: {
+          id: 4004
+          serial: "45678"
+          applicationSectionsUsingId: {
+            create: [{ templateSectionId: 1005 }, { templateSectionId: 1006 }]
+          }
+          name: "Test Review -- Amoxicilin"
+          outcome: PENDING
+          isActive: true
+          templateId: 4
+          userId: 2
+          applicationStageHistoriesUsingId: {
+            create: [
+              {
+                isCurrent: false
+                templateStageToStageId: { connectById: { id: 4 } }
+                applicationStatusHistoriesUsingId: {
+                  create: [
+                    { isCurrent: false, status: SUBMITTED }
+                    { isCurrent: false, status: COMPLETED }
+                  ]
+                }
+              }
+              {
+                isCurrent: true
+                templateStageToStageId: { connectById: { id: 5 } }
+                applicationStatusHistoriesUsingId: {
+                  create: { isCurrent: true, status: SUBMITTED }
+                }
+              }
+            ]
+          }
+          applicationResponsesUsingId: {
+            create: [
+              {
+                id: 4150
+                isValid: true
+                templateElementId: 4001
+                status: SUBMITTED 
+                value: { text: "Another" }
+              }
+              {
+                id: 4151
+                isValid: true
+                templateElementId: 4002
+                status: SUBMITTED 
+                value: { text: "Test" }
+              }
+              {
+                id: 4152
+                isValid: true
+                templateElementId: 4003
+                status: SUBMITTED 
+                value: { text: "this@test.com" }
+              }
+              {
+                id: 4153
+                isValid: true
+                templateElementId: 4005
+                status: SUBMITTED 
+                value: { text: "18" }
+              }
+              {
+                id: 4154
+                isValid: true
+                templateElementId: 4006
+                status: SUBMITTED 
+                value: { text: "Brazil" }
+              }
+              {
+                id: 4155
+                isValid: true
+                templateElementId: 4008
+                status: SUBMITTED 
+                value: { text: "Amoxicilin" }
+              }
+              {
+                id: 4156
+                isValid: true
+                templateElementId: 4009
+                status: SUBMITTED 
+                value: { text: "Medicine", optionIndex: 2 }
+              }
+              {
+                id: 4157
+                isValid: true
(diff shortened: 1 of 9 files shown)


database/insertData/06_applications_D_reviewTest.js
```diff
@@ -572,4 +572,118 @@ exports.queries = [
...
+                templateElementId: 4008
+                status: SUBMITTED 
+                value: { text: "Amoxicilin" }
+              }
+              {
+                id: 4156
+                isValid: true
+                templateElementId: 4009
+                status: SUBMITTED 
+                value: { text: "Medicine", optionIndex: 2 }
```
CarlosNZ, line 660: Dropdown responses need to be in the form
```
{"text": "Medicine", "selection": "Medicine", "optionIndex": 0}
```
Also -- in the template "Medicine" is Option Index 0 (first item)

plugins/action_update_review_status/src/databaseMethods.ts
```diff
@@ -9,13 +9,39 @@ const databaseMethods = (DBConnect: any) => ({
...
-      ON review.review_assignment_id = review_assignment.id
-      JOIN review_status_history ON review.id = review_status_history.review_id
-      WHERE review_status_history.is_current = true
-      AND review_assignment.application_id = $1
-      AND stage_id = $2
-      AND review_assignment.level_number = $3
+      FROM review_assignment 
+        LEFT JOIN review ON review_assignment.id = review.review_assignment_id 
+        LEFT JOIN review_status_history ON review.id = review_status_history.review_id 
+          AND review_status_history.is_current = true
```
CarlosNZ, line 15: Just wondering -- we currently have created Views for application_stage and status history (`application_stage_status_all` and `application_stage_status_latest`) which has been quite useful. Do you think we should have one for review histories too?

plugins/action_update_review_status/src/updateReviewsStatuses.ts
```diff
@@ -99,12 +121,16 @@ const updateReviewsStatuses: ActionPluginType = async ({
...
+
+      // Get reviews matching application_id, current stage, in level 1
       const level = 1
-      // Get reviews/review assignments (with status) matching application_id & current stage
       const reviews = await getReviewsByLevelAndStatus(level)
       // Deduce which ones should be updated to Pending
       for (const review of reviews) {
         const { reviewAssignmentId, reviewStatus } = review
+        if (reviewStatus === ReviewStatus.Locked) {
+          // TODO: Check with team should also lock the reviewAssignment ?
```
CarlosNZ, line 132: Did you get an answer on this, and do you want to figure it out before merging this PR, or wait till later?

src/components/postgresConnect.ts
```diff
@@ -551,6 +551,18 @@ class PostgresDB {
...
+      'INSERT into review_assignment is_locked VALUES ($2) WHERE id = $1 RETURNING id, is_locked'
+    try {
+      const result = await this.query({ text, values: [reviewAssignmnetId, isLocked] })
+      return result.rows[0]
+    } catch (err) {
+      console.log(err.message)
+      throw err
+    }
+  }
+
```
CarlosNZ, line 565: Should probably keep this in the databaseMethods file of the plugin for consistency (unless this is used/intended to be used elsewhere?)

CarlosNZ (changes requested):
Looks great, and tests all pass. Just a couple of fairly minor comments.

src/components/postgresConnect.ts
CarlosNZ: Never mind, I see you already did this in your next PR :)

database/insertData/06_applications_D_reviewTest.js
nmadruga: Oops! Good point, I did that and end up not testing.

nmadruga: Done: 3c3ef77

plugins/action_update_review_status/src/databaseMethods.ts
nmadruga: Hmm perhaps. Do you think I could include as part of this PR?

plugins/action_update_review_status/src/updateReviewsStatuses.ts
nmadruga: No... I'll leave this one there until we decide. Will add a discussion issue to check that.

plugins/action_update_review_status/src/databaseMethods.ts
CarlosNZ: If it's not too much extra work. But can make a new issue if you'd prefer.

plugins/action_update_review_status/src/updateReviewsStatuses.ts
nmadruga: Issue #757

plugins/action_update_review_status/src/databaseMethods.ts
nmadruga: Hmm I was checking out file `32_review_status_history` and saw that the latest `status` is already exposed in the `review` table:
https://github.com/openmsupply/application-manager-server/blob/9da7a3cb43932aac717af2267430b56387d19039/database/buildSchema/32_review_status_history.sql#L42-L52
So actually we don't need this view... I can just change this code to use review.status instead.

CarlosNZ: Looks like "status" is visible on `review` table in GraphQL, but not in SQL, as it hasn't been added as a generated column.

nmadruga: Yes, tried changing and realised that too. So will just leave this one as is.

CarlosNZ (approved):

PR #380, opened 2021-05-28
#120F Search plugin (API lookup)

CarlosNZ:
Adds "search" question elements to FeatureShowcase template, to demonstrate the Search element from [front-end PR 747](https://github.com/openmsupply/application-manager-web-app/pull/747)

### DIFF
--- a/database/insertData/03_template_A_featureShowcase.js
+++ b/database/insertData/03_template_A_featureShowcase.js
@@ -1138,6 +1138,10 @@ exports.queries = [
                             "users.nodes"
                           ]
                         }
+                        resultFormat: {
+                          title: "\${firstName} \${lastName}"
+                          description: "\${username}"
+                        }
                       }
                     }
                   ]
--- a/database/insertData/03_template_A_featureShowcase.js
+++ b/database/insertData/03_template_A_featureShowcase.js
@@ -1144,6 +1144,42 @@ exports.queries = [
                         }
                       }
                     }
+                    {
+                      code: "searchCountries"
+                      index: 8
+                      title: "Search Demo (countries)"
+                      elementTypePluginCode: "search"
+                      category: QUESTION
+                      isRequired: false
+                      parameters: {
+                        label: "Search for countries"
+                        description: "Type the two-character country code (must be CAPS sorry)"
+                        placeholder: "Search countries"
+                        icon: "world"
+                        minCharacters: 2
+                        source: {
+                          operator: "graphQL",
+                          children: [
+                            "query countries($code: ID!) {country(code: $code) {name, capital, emoji, code}}",
+                            "https://countries.trevorblades.com",
+                            [
+                              "code"
+                            ],
+                            {
+                              operator: "objectProperties",
+                              children: [
+                                "search.text"
+                              ]
+                            },
+                            "country"
+                          ]
+                        }
+                        resultFormat: {
+                          title: "\${emoji} \${name}"
+                          description: "Capital: \${capital}"
+                        }
+                      }
+                    }
                   ]
                 }
               }
--- a/database/insertData/03_template_A_featureShowcase.js
+++ b/database/insertData/03_template_A_featureShowcase.js
@@ -1116,7 +1116,7 @@ exports.queries = [
                       elementTypePluginCode: "search"
                       category: QUESTION
                       isRequired: false
-                      helpText: "The next two pages contain examples of the **Search** element. It can make API queries in response to user input and display the results, which can be selected."
+                      helpText: "This page contain examples of the **Search** element. It can make API queries in response to user input and display the results, which can be selected."
                       parameters: {
                         label: "Search for users"
                         description: "Start typing to search database for usernames"
@@ -1230,6 +1230,7 @@ exports.queries = [
                       elementTypePluginCode: "search"
                       category: QUESTION
                       isRequired: false
+                      helpText: "This one has no \`displayFormat\` prop, so it falls back to a generic \\"default\\" display"
                       parameters: {
                         label: "Search for an organisation"
                         description: "Please enter the organisation's registration code (min 6 chars)"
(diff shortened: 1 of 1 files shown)


database/insertData/03_template_A_featureShowcase.js
```diff
@@ -1102,6 +1102,161 @@ exports.queries = [
...
+                      parameters: {
+                        label: "Search for users"
+                        description: "Start typing to search database for usernames"
+                        placeholder: "Search usernames"
+                        icon: "user"
+                        source: {
+                          operator: "graphQL",
+                          children: [
+                            "query GetUsers($user: String!) { users(filter: { username: {includesInsensitive: $user } }) {nodes { username, firstName, lastName }}}",
+                            "",
```
nmadruga, line 1129: I see, this is using the local server, should be better to send that `graphqlEndpoint` string instead? Just to it's easier to remember what this field is...

database/insertData/03_template_A_featureShowcase.js
```diff
@@ -1102,6 +1102,161 @@ exports.queries = [
...
+                              children: [
+                                "search.text"
+                              ]
+                            },
+                            "entities.data"
+                          ]
+                        }
+                        displayFormat: {
+                          title: "\${description}",
+                          description: "ATC code: \${code}"
```
nmadruga, line 1218: Just to be clearer about this one, what is returned from UC isn't the ATC code, that used to be one of the parameters in the properties array of a drug, but has been removed temporary from UC server, so now what this is showing is actually the mSupply universal code for this drug.

database/insertData/03_template_A_featureShowcase.js
```diff
@@ -1102,6 +1102,161 @@ exports.queries = [
...
+                        label: "Search for an organisation"
+                        description: "Please enter the organisation's registration code (min 6 chars)"
+                        placeholder: "Search organisations"
+                        icon: "building"
+                        minCharacters: 6
+                        source: {
+                          operator: "graphQL",
+                          children: [
+                            "query GetOrgs($registration: String!) { organisations(filter: {registration: {startsWithInsensitive: $registration}}) { nodes { name registration } } }",
+                            "",
```
nmadruga, line 1244: Also maybe better to use `graphqlEndpoint` here so it's more descriptive

nmadruga (changes requested):
Looks great. Thanks for doing that, it will be very simple to reuse this on Drug Registration, Create organisation and Import Permit (biggest focus for this week). I would recommend adding as part of this PR just a few lines on the docs about this, so we have a easy place to check how to use this new plugin. 

And I'll be checking the functionality on front-end PR now.

database/insertData/03_template_A_featureShowcase.js
```diff
@@ -1102,6 +1102,161 @@ exports.queries = [
...
+                      elementTypePluginCode: "search"
+                      category: QUESTION
+                      isRequired: false
+                      parameters: {
+                        label: "Lookup ATC codes"
+                        description: "Start typing a drug name"
+                        placeholder: "Search medicines"
+                        icon: "pills"
+                        minCharacters: 3
+                        multiSelect: true
```
nmadruga, line 1198: While testing...

Something seemed wrong with the multi-selection or UC API calls... Not sure which. Can you investigate this please?

My test: typed "pam" and there were 4-5 results. And selected one:
Screen Shot 2021-06-01 at 10 04 05 AM

Then tried to search for another one start with "pam", it stays loading for ever... even if I erase what I type. It only goes back to normal if I refresh the page:
Screen Shot 2021-06-01 at 10 03 48 AM

database/insertData/03_template_A_featureShowcase.js
```diff
@@ -1102,6 +1102,161 @@ exports.queries = [
...
+                        placeholder: "Search organisations"
+                        icon: "building"
+                        minCharacters: 6
+                        source: {
+                          operator: "graphQL",
+                          children: [
+                            "query GetOrgs($registration: String!) { organisations(filter: {registration: {startsWithInsensitive: $registration}}) { nodes { name registration } } }",
+                            "",
+                            [
+                              "registration"
```
nmadruga, line 1246: Can we add a search for both "registration" or "company name"? I think it would be useful. Not sure if it's feasible, but I imagine just passing the same `search.text` to 2 different search options on the GraphQL query?

nmadruga:
> I would recommend adding as part of this PR just a few lines on the docs about this, so we have a easy place to check how to use this new plugin.

Sorry, now I checked and there is docs on the Front-end PR: `documentation/internal/Element-Type-Specs.md`

database/insertData/03_template_A_featureShowcase.js
CarlosNZ: Right, you meant the more specific default value (I was confused at first). Anyway, done dcf2e3f

database/insertData/03_template_A_featureShowcase.js
CarlosNZ: Ah okay, got it. dcf2e3f

database/insertData/03_template_A_featureShowcase.js
CarlosNZ: Yup, done with [this one](https://github.com/openmsupply/application-manager-server/pull/380#discussion_r642700688)

database/insertData/03_template_A_featureShowcase.js
CarlosNZ: Sure, this will be a front-end issue, I'll look into it for that PR.

database/insertData/03_template_A_featureShowcase.js
CarlosNZ: Yes, the query would be quite straightforward (just nest it within an OR filter), but this was just trying to demonstrate being able to look up a company without being able to just search freely for names, like @andreievg first described.

Not sure which implementation we will do in the actual template, but let's save figuring out what the specs need to be when we update the actual "joinOrg" template, shall we?

database/insertData/03_template_A_featureShowcase.js
CarlosNZ: Done :) [Explanation](https://github.com/openmsupply/application-manager-web-app/pull/747#discussion_r642718661)

Thanks for catching this one :)

nmadruga (approved):

PR #390, opened 2021-06-02
#383 Subfolder arguments for insertData script

### LINKED ISSUE #383: Create new separate insertData file for adding templates and minimal user config for Laos
The configuration should have a quick way to define which `insert data` file to use. So we can quickly change between by selecting the one to use when initialising the db.

CarlosNZ:
Fixes #383. 

Add an argument to `yarn database_init` to specify a subfolder to load (in addition to the base-level `insertData` files)

E.g. `yarn database_init laos` or `yarn database_init dev`

Most of the large diff here is due to moving/renaming the files in insertData folder, the only files with "real" code changes in them are:

- package.json
- insert_data.sh
- insertData.js

### Summary of changes

- `insert_data.sh` takes an optional parameter which is passed down to `insertData.js`
- `insertData.js` uses this parameter to read the files from the specified subfolder and insert them after the base-level ones
- moved files into `dev` and `laos` subfolders. `dev` contains all our existing templates and data, `laos` contains just a placeholder template for now.
- moved "nonRegisteredUser" into its own file, as it's required (with ID=1) for all implementations
- Note: core_mutations and dev_actions are COPIED to the relevant subfolder each time the script runs, so we only have to keep the base-level one up to date.


nmadruga:
@CarlosNZ please resolve the conflict with this branch. I'm having a look on it now : )

Just a initial comment of one thing I noticed:
- When running `yarn database_init dev` it doesn't create the previous application (for DemoTest template)

nmadruga (commented):
I see that you kept insert_data files `template_filters` and `template_categories` for both, which is fine, but I noticed a tiny thing on the `categories`, couldn't comment on the file so just adding a overall comment here: Needs to update the `userCategory` title, for now it's showing "Drug Registration", this example is showing Dashboard on "laos" config right after I registered a user:
Screen Shot 2021-06-03 at 10 39 57 AM

CarlosNZ:
> @CarlosNZ please resolve the conflict with this branch. I'm having a look on it now : )

Done!

>     * When running `yarn database_init dev` it doesn't create the previous application (for DemoTest template)

Hmmm, that's odd -- the applications are in the database, just not showing up in front-end. Maybe some permissions got messed up? Will take a look...

package.json
```diff
@@ -40,7 +40,8 @@
     "pg_permissions": "exec ./utils/postgraphile_launch_permissions.sh",
     "dev_pg": "concurrently --kill-others-on-fail \"yarn dev\"  \"yarn pg\"",
     "dev_pg_permissions": "concurrently --kill-others-on-fail \"yarn dev\"  \"yarn pg_permissions\"",
-    "database_init": "./database/initialise_database.sh && ./database/insert_data.sh && ./database/post_data_insert.sh",
+    "database_init": "./database/initialise_database.sh && ./database/insert_data.sh",
+    "postdatabase_init": "./database/post_data_insert.sh",
```
nmadruga, line 44: When is this one being used?

CarlosNZ: The `post` prefix means it automatically runs after `database_init` (see [here](https://docs.npmjs.com/cli/v7/using-npm/scripts#pre--post-scripts). It used to be bundled in the same command (see previous diff), but the arguments were only getting passed to the last command, so needed to make them separate.

CarlosNZ:
> > @CarlosNZ please resolve the conflict with this branch. I'm having a look on it now : )
> 
> Done!
> 
> > ```
> > * When running `yarn database_init dev` it doesn't create the previous application (for DemoTest template)
> > ```
> 
> Hmmm, that's odd -- the applications are in the database, just not showing up in front-end. Maybe some permissions got messed up? Will take a look...

I see, the application mutations have hard-coded templateSectionIds in them, so they're getting linked to the wrong template now. Guess I'll need to update that. Thanks for spotting.

CarlosNZ:
Okay, I think I've fixed them all now: 46469375

database/insertData.js
```diff
@@ -1,12 +1,35 @@
...
-    .readdirSync('./database/insertData')
+    .readdirSync(dataFolder, { withFileTypes: true })
+    .filter((dirent) => dirent.isFile())
+    .map((dirent) => dirent.name)
     .filter((file) => !file.match(/^\./)) // Ignore hidden files
-    .filter((file) => !['core_mutations.js', 'dev_actions.js'].includes(file))
+    .filter((file) => !sharedDataFiles.includes(file))
+
+  // Add locale-specific files
+  const localeFolder = process.argv[3]
```
andreievg, line 18: can this default to `dev` by default, if not supplied ?

database/insertData.js
```diff
@@ -1,12 +1,35 @@
...
+    console.log(`Locale: ${localeFolder}`)
+    const subfolderFilesToProcess = fs
+      .readdirSync(path.join(dataFolder, localeFolder))
+      .filter((file) => !file.match(/^\./)) // Ignore hidden files
+      .filter((file) => !sharedDataFiles.includes(file))
+    filesToProcess.push(
+      ...subfolderFilesToProcess.map((filename) => path.join(localeFolder, filename))
+    )
+    // Copy current version of core mutations/actions to subfolder
+    sharedDataFiles.forEach((file) =>
```
andreievg, line 29: why do we need to do this (can we not just import them from parent folder ? )

nmadruga (commented):
Cool, now the applications show up on the list! But there still something not quite right... Probably about ids again. I checked the `review_question_assignment` but that looks ok, there are specific ids for each, but in the Review home page, no review or assignment are listed, not either the actions for Reviewer on each application in the list:
Screen Shot 2021-06-03 at 11 23 14 AM

The review page (e.g. Review id: 1) are displaying the reviews completed correctly though:
Screen Shot 2021-06-03 at 11 27 31 AM

andreievg (commented):
A couple of change request (as comments).

Also was wondering if we can structure things so that we don't need to filter out sharedDataFiles.

i.e.
```
insertData
-> common
   ... (01-03)
-> specific
   -> dev
   -> laos
-> helpers 
    ... dev_actions, core_mutations
```

nmadruga:
Hmmm something that may be what is causing that, when I load the Review home page (on the same example above - for a Review that already has been submitted in **Screening** stage, and should be in the **Assessment** stage now), it comes without the current stage selected:

Screen Shot 2021-06-03 at 11 29 04 AM

Then when I go to select the stage, there is only **Screening** available - when should be Screening or Assessment, and selecting **Screening** display reviews in the **Assessment** stage:
Screen Shot 2021-06-03 at 11 29 13 AM

database/insertData.js
CarlosNZ: I just wanted it to be possible easily move the insert mutations between the base level and subfolders freely, without having to remember to change the import path every time. I thought we might be doing it quite a bit...

database/insertData.js
CarlosNZ: Yes! 2204d26b

CarlosNZ:
> Also was wondering if we can structure things so that we don't need to filter out sharedDataFiles.

Yeah, will do. How about:

```
insertData
-> _common
   ... (01-03)
-> _helpers 
    ... dev_actions, core_mutations
-> dev
-> laos
...
```
 
And that way the "common" ones and the specific ones will have the same relative path to helpers, so won't need to worry about updating when moving from common to specific (or vice versa), as per [this comment](https://github.com/openmsupply/application-manager-server/pull/390#discussion_r644395120)

CarlosNZ:
> > Also was wondering if we can structure things so that we don't need to filter out sharedDataFiles.
> 
> Yeah, will do. How about:
> 
> ```
> insertData
> -> _common
>    ... (01-03)
> -> _helpers 
>     ... dev_actions, core_mutations
> -> dev
> -> laos
> ...
> ```
> 
> And that way the "common" ones and the specific ones will have the same relative path to helpers, so won't need to worry about updating when moving from common to specific (or vice versa), as per [this comment](https://github.com/openmsupply/application-manager-server/pull/390#discussion_r644395120)

Okay, let me know if this works for you: 9ef0cfef

database/insertData.js
CarlosNZ: This is taken care of [here](https://github.com/openmsupply/application-manager-server/pull/390#issuecomment-853469430)

CarlosNZ:
> Cool, now the applications show up on the list! But there still something not quite right... Probably about ids again. I checked the `review_question_assignment` but that looks ok, there are specific ids for each, but in the Review home page, no review or assignment are listed, not either the actions for Reviewer on each application in the list:

Okay, had to bump the stageIds in review_assignment too: 58d2cd3b 🙄

database/insertData.js
```diff
@@ -1,12 +1,30 @@
...
+    .readdirSync(path.join(dataFolder, '_common'), { withFileTypes: true })
+    .filter((dirent) => dirent.isFile())
+    .map((dirent) => path.join('_common', dirent.name))
     .filter((file) => !file.match(/^\./)) // Ignore hidden files
-    .filter((file) => !['core_mutations.js', 'dev_actions.js'].includes(file))
+
+  console.log('filesToProcess', filesToProcess)
+  // Add locale-specific files
+  const localeFolder = process.argv[3] || 'dev'
+  if (localeFolder) {
```
andreievg, line 19: don't need the if anymore

andreievg (approved):
looks good, thanks for the changes. pending the if statement removal.

database/insertData.js
CarlosNZ: 🙄

b4f2b3b0

nmadruga (approved):
LGTM

Issue #376, opened 2021-05-27
Extensions to User and Org tables -- and how to handle custom fields


CarlosNZ:
In order to create the templates required by Valerio for creating a company and applying for licences, the User and Org tables need more information (e.g. Phone number, place of birth, Professional ID Number, etc.)

Need to discuss what new columns to add, but also how to handle "Custom" fields, as every installation will be different.

To-do:

- [ ] Agree on a schema and means of extending with Custom data (e.g. JSON field for additional info?)
- [ ] Implement schema
- [ ] Check that all user/org info is retrieved and saved in local state on login
- [ ] Update User and Org registration templates to input additional information

andreievg:
It seems this is a big part of the system that's been postponed.

What I was thinking is something like this:

* modifyRecords action (maybe better named as createOutcome, based on this suggestion), does a bit more
  * creates table if doesn't exist
  * create fields if they dont' exist
  * able to save full response
  * able to link to application
* outcomes route that shows outcomes, and .. yep, you guessed it, it's configurable, check out the diagram.
   * show cards of all of the outcomes
   * show when card is pressed show a table
   * when table element is pressed show details for that record (this shows summary views)
 
This diagram is not very good, but should describe the basic concept: https://app.diagrams.net/#G1y76nr7TnYXpLs2u32Aj4xr-92mhKzA8y

I did a proof of concept implementation `outcomes-proof-of-concept` branch of back end and front end. To see it working, have to create a new user, and then go to `/outcomes` route, click through to the newly created user. (should be able to navigate to application)

Changes

 * modifyRecord does as per above, creates table, and fields and join to an application (the database info view is from the new snapshot functionality, which I am slow to PR)
 * added an upload identification form element
 * outcomes are configured for one table (user), with table view and details view
 * have to use dynamic query building (similar to lookup table functionality).

I can explain this tomorrow.

Btw we don't need to add any extra information local state on login (and don't think we needed extra fields that were added for user edit, we can do graphQL query for defaults)

CarlosNZ:
Sounds good. Look forward to discussing. Couple of things to consider:

- Are we sure having a dynamic schema is a good idea with GraphQL/Postgraphile? How quickly after creating a field on a table does it "appear" in Postgraphile for querying, and is it reliable? (Maybe you've tried?)
- "dynamic query building" -- strongly discouraged by GraphQL people. Can we achieve what we want with static queries and variable objects?\

[...]

PR #392, opened 2021-06-02
#376 custom fields and tables outcomes

andreievg:
Fixes: #376 

Required for front end PR: https://github.com/openmsupply/application-manager-web-app/pull/758
Testable with front end branch: #376-outcome-displays

### Changes
* Transferred a view and database method to get database info (from work in progress snapshot branch)
* Added database outcome displays to the schema
* Added data to outcome displays (organisation and user)
* Saving logo as json value when organisation is approved
* modifyRecord now adds a table fields and join table, also adds entry to join table

### Testing
* As per front end PR testing

TODO
* link to epic: https://github.com/openmsupply/application-manager-web-app/issues/534
* create documentation issue


plugins/action_modify_record/src/databaseMethods.ts
```diff
@@ -1,56 +1,121 @@
...
+    const result = await DBConnect.query({ text, values: Object.values(record) })
+    const firstRow = result.rows[0]
+    return { success: true, [tableName]: firstRow, recordId: firstRow.id }
+  } catch (err) {
+    console.log(err.message)
+    throw err
+  }
+}
+
+export type DatabaseMethodsType = ReturnType<typeof databaseMethods>
```
CarlosNZ, line 119: Since all Actions have (or will have) a databaseMethods import, should this type be defined in `/plugins/types` and shared to all Actions? (No need to adjust existing actions, but it'd be good to have it available)

CarlosNZ (changes requested):
Looks really good, and fairly easy to follow what's going on.

Just wondering -- why do we need a new join table for every outcome type? Can't we just put an "application_id" field on the outcome record itself? (There'll only be one won't there?)

And just a suggestion -- if we really _do_ need the join tables, perhaps change the naming so that each join table will be next to its "outcome entity" table -- e.g. instead of `application_user_join`, we have `user_application_join`, so we can see it alongside the "user" table.

And was also wondering --what's the purpose of all the extra info in the schema Views. Don't we only need a list of tables and their columns (which you could get quite easily without making a View)?

database/insertData/03_template_C_userRegistration.js
```diff
@@ -181,6 +181,14 @@ exports.queries = [
                     operator: "objectProperties"
                     children: ["applicationData.responses.Q5.hash"]
                   }
+                  username_value: {
+                    operator: "objectProperties"
+                    children: ["applicationData.responses.Q3"]
```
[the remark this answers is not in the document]
CarlosNZ: Oh I see, this is so it can display the SummaryView on the table.

I wonder if this is not a good case for a jsonb field to collect all of these together. Seems like we're going to have a lot of extraneous columns just to hold response data, so perhaps they should all go into one column called "application_responses", and have a field for each response you want to keep.

andreievg:
> Just wondering -- why do we need a new join table for every outcome type? Can't we just put an "application_id" field on the outcome record itself? (There'll only be one won't there?)

I think they will be many to one:

user -> `create`, `edit`

licence -> `create`, `review` 

etc..

> and just a suggestion -- if we really do need the join tables, perhaps change the naming so that each join table will be next to its "outcome entity" table -- e.g. instead of application_user_join, we have user_application_join, so we can see it alongside the "user" table.

good point:

957dd36006c7b3ae928fda87fb8611021e98f78c

also makes front end slightly simpler

https://github.com/openmsupply/application-manager-web-app/pull/758/commits/27424241f6fcaf7e694a86aff3464e777ae3bede

> And was also wondering --what's the purpose of all the extra info in the schema Views. Don't we only need a list of tables and their columns (which you could get quite easily without making a View)?

It was a transfer from new snapshot/import-export work, I just copied and pasted it in

plugins/action_modify_record/src/databaseMethods.ts
```diff
@@ -1,56 +1,121 @@
...
-    matchField: string,
-    matchValue: any,
-    record: { [key: string]: any }
-  ) => {
+  updateRecord: async (tableName: string, id: string, record: { [key: string]: any }) => {
     const placeholders = DBConnect.getValuesPlaceholders(record)
     const matchValuePlaceholder = `$${placeholders.length + 1}`
     const text = `
       UPDATE "${tableName}" SET (${Object.keys(record)})
       = (${DBConnect.getValuesPlaceholders(record)})
```
nmadruga, line 22:
```suggestion
    const text = `
      UPDATE "${tableName}" SET (${Object.keys(record)})
      = (${placeholders})
```

Can probably use this variable here.

Although I'm trying to understand what this new WHERE statement is checking, what is in `matchValuePlaceholder`, and how an `id` can be compared with `$${placeholders.length + 1}` ?

nmadruga: Hmmm, I think I understood now. this will be getting the last value from `values: [...Object.values(record), id]` which is the id, but not sure if this is in the right format, is it? Just trying to understand what is this one doing.

database/insertData/03_template_C_userRegistration.js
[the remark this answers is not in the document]
andreievg: This one in particular was an artefact of testing.

> I wonder if this is not a good case for a jsonb field to collect all of these together. Seems like we're going to have a lot of extraneous columns just to hold response data, so perhaps they should all go into one column called "application_responses", and have a field for each response you want to keep.

I am not sure if column quantity is an issue ? (still same amount of data right ? it's just hidden in a jsonb field)

I think it's much easier to see what 'things' are in the outcome by looking at the table vs looking at possible values inside jsonb field of one of the records.

Anyways I've removes saving username_value:

d2de7a94364f1d5b9509c9c2437320bb337d8e7d

database/insertData/03_template_C_userRegistration.js
```diff
@@ -181,6 +181,14 @@ exports.queries = [
...
+                  }
+                  identification: {
+                    operator: "objectProperties"
+                    children: ["applicationData.responses.identification"]
```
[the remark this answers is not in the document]
andreievg: Oops: 

d2de7a94364f1d5b9509c9c2437320bb337d8e7d

nmadruga (commented):
Partial review - will continue from file `plugins/action_modify_record/src/databaseMethods.ts` when I get back to this later on today.

plugins/action_modify_record/src/databaseMethods.ts
```diff
@@ -1,56 +1,121 @@
...
-      const result = await DBConnect.query({ text, values: Object.values(record) })
-      return { success: true, [tableName]: result.rows[0] }
+      await DBConnect.query({
+        text,
+        value: [tableName],
+      })
     } catch (err) {
       console.log(err.message)
-      throw err
+      throw new Error(`Filed to create table: ${tableName}`)
```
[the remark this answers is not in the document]
andreievg: 51c31a586883cc929196905523147e79c0902190

plugins/action_modify_record/src/databaseMethods.ts
```diff
@@ -1,56 +1,121 @@
...
+                  id serial PRIMARY KEY, 
+                  application_id integer references application(id),
+                  ${tableName}_id integer references "${tableName}"(id));`
+
+    console.log('creating join table with statement: ', text)
+    try {
+      await DBConnect.query({ text })
+    } catch (err) {
+      console.log(err.message)
+      throw new Error(`Filed to create join table: ${joinTableName}`)
```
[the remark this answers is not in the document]
andreievg: 51c31a586883cc929196905523147e79c0902190

plugins/action_modify_record/src/databaseMethods.ts
andreievg: This one is specific to modifyRecord action.

Since i've added a method `createOrUpdateTable` to the action, which takes action specific instance of databaseMethods, needed to know what type it is.

Also I've noticed that `DBConnect?: any` is the parameter of ActionPluginInput, we should probably use the `DBConnectType` now. I didn't want to bloat this PR though, incase there are problems with types after that

plugins/action_modify_record/src/databaseMethods.ts
andreievg: f667700abb4426796a8d4664bb03cc6cdfa2e50d

> Hmmm, I think I understood now. this will be getting the last value from values: [...Object.values(record), id] which is the id, but not sure if this is in the right format, is it? Just trying to understand what is this one doing.

it's adding a new placeholder, see below `values: [...Object.values(record), id]`

plugins/action_modify_record/src/databaseMethods.ts
```diff
@@ -1,56 +1,121 @@
...
     }
   },
-  createRecord: async (tableName: string, record: { [key: string]: any }) => {
-    const text = `
-      INSERT INTO "${tableName}" (${Object.keys(record)}) 
-      VALUES (${DBConnect.getValuesPlaceholders(record)})
-      RETURNING *
-      `
+  createRecord: async (tableName: string, record: { [key: string]: any }) =>
+    await createRecord(tableName, record, DBConnect),
```
nmadruga, line 38: Simpler to just have the local createRecord exposed here? Is there a reason why not? It would be easier to read.

nmadruga (changes requested):
Looks great. And the functionality is also pretty good one to have now : )

I just noticed something in the tests: there is one failing...


And the output is showing this problem:
  ● Test: Update existing organisation

```
  console.log
    insert or update on table "organisation_application_join" violates foreign key constraint "organisation_application_join_application_id_fkey"

      at plugins/action_modify_record/src/databaseMethods.ts:114:13

  console.log
    insert or update on table "organisation_application_join" violates foreign key constraint "organisation_application_join_application_id_fkey"
```

plugins/action_modify_record/src/databaseMethods.ts
andreievg: how do you mean ? Not sure how to pass through DBConnect to it (without refactoring quite a bit)

plugins/action_modify_record/src/databaseMethods.ts
nmadruga: Yep. Got it : )

andreievg:
@nmadruga 

> Looks great. And the functionality is also pretty good one to have now : )

thanks

> I just noticed something in the tests: there is one failing...

Oops, thanks for checking, i should have done this before PR: d5ce52874a88af726d9bcbc8f843f09c212a777a

CarlosNZ:
@andreievg  FYI -- I see you've merged master into here and dealt with the file structure changes. However, the file `08_outcome_displays.js` is still in the top-level folder, not `_common`, so isn't getting imported.

CarlosNZ:
Cool -- are you able to modify the orgRego template and organisation outcome display to use the imgDisplay element and the "logo_url" field like we discussed (and ditch the other "logo" field)? Then this is pretty much good to go, I think.

andreievg:
> @andreievg FYI -- I see you've merged master into here and dealt with the file structure changes. However, the file 08_outcome_displays.js is still in the top-level folder, not _common, so isn't getting imported.

thanks, good catch

6f934593fc42299186773598d00a5b4cd9673d7a

andreievg:
> Cool -- are you able to modify the orgRego template and organisation outcome display to use the imgDisplay element and the "logo_url" field like we discussed (and ditch the other "logo" field)? Then this is pretty much good to go, I think.

Made the changes but not happy with them (way more verbose):

d58562203f892779c6c234626e5ad31495281cc9

and requires a hacky front end commit:

https://github.com/openmsupply/application-manager-web-app/pull/758/commits/3c1cf07d74a26efc7e8bcf7e77ffb6b0583b0cb8

plugins/action_modify_record/src/databaseMethods.ts
nmadruga: It should be accessible since this is received by `const databaseMethods = (DBConnect: any) =>`, couldn't be in use the same way that it's done by `getRecordId` ?

andreievg: createRecord is now used in two methods, wasn't sure how to invoke it from another method (technically `this` would work but had typescript errors), so I've made it a local method that require DBConnect.  

Actually this would work: b2d0440fda951a740a1991a08a96a59683322979, retested, seems to still work

Is this what you had in mind ?

andreievg:
@nmadruga @CarlosNZ all requested changes done/commented

database/insertData/_common/04_outcome_displays.js
```diff
@@ -81,11 +81,35 @@ exports.queries = [
                 title:  "Address"
               }
               {
-                columnName: "logo"
+                columnName: "logoUrl"
+                elementTypePluginCode: "imageDisplay"
+                isTextColumn: true
+                parameters: {
+                  url: {
+                    operator: "CONCAT"
```
CarlosNZ, line 89: This works, but strictly speaking we were trying to deprecate "CONCAT" and just use "+". (You can just swap it out and it will do exactly the same thing)

CarlosNZ (approved):
Looks good, thanks. (One tiny nitpick)

database/insertData/_common/04_outcome_displays.js
andreievg: 095dfa0bdd7973e6b49c981ba80c412462415d2a

nmadruga (approved):

PR #395, opened 2021-06-07
#761F generic default value handling to work for all plugins

andreievg:
Required for front end PR https://github.com/openmsupply/application-manager-web-app/pull/764: 

note: when this PR is merged we need to push evaluator

### Changes

* Default value field on element (note this is the full response value, plugin specific)
* Use new defaultValue in userEdit (now requires buildObject evaluation)
* Use EvaluatorNode type for evaluator

#### Evaluator Changes
* Fix deeply nested query bug, new test added (it won't pass in master)
* New buildObject evaluation (see below notes about evaluator)
* Would need to use new evaluator for this PR to compile, need to repeat the same evaluator related steps as per Testing in front end PR

### Testing
* See front end testing in PR 'X'
* Use back end branch #761F-testing (`https://github.com/openmsupply/application-manager-server/compare/%23761F-generic-default-value-handling-to-work-for-all-plugins...%23761F-testing`)
* Make sure to run the same steps for evaluator import from local file as per front end branch

### Notes about Evaluator

I've added a new resolve `buildObject`, see documentation change for the new format, which doesn't use the children syntax, which i think is very cumbersome and hard to use, I think it's too much work to change existing resolvers to conform with the new format but I think it might be worth doing at some stage.

I think it's easier to have resolver specific configuration shape then the generic children shape

```
{
    operator: "graphQL"
    query: "query getUser($id: Int!){user(id: $id) {listDemo}}"
    variables: [
        {
            key: "id"
            value: {
                "operator": "objectProperties", 
                "children": ["currentUser.id"]
            }
        }
    ]
    extractor: "user.listDemo"
}
```
VS
```
{
    operator: "graphQL",
    children: [
        "query getUser($id: Int!){user(id: $id) {listDemo}}",
        "",
        ["id"],
        {
            "operator": "objectProperties", 
            "children": ["currentUser.id"]
        },
        "user.listDemo"
    ]
}
```

### DIFF
--- a/documentation/Query-Syntax.md
+++ b/documentation/Query-Syntax.md
@@ -50,6 +50,7 @@ For more complex lookups, we would hide the complexity from the user in the Temp
   - [graphQL](#graphql)
   - [GET](#get)
   - [POST](#post)
+  - [buildObject](#buildObject)
 - [Usage](#usage)
 - [Examples](#examples)
 - [To Do](#to-do)
--- a/documentation/Query-Syntax.md
+++ b/documentation/Query-Syntax.md
@@ -355,7 +355,7 @@ properties: [{
 ]
 ```
 
-- Output evaluated object, empty object by default. If value or key evaluates to undefined it wont' be present in the output
+- Output evaluated object, empty object by default. If value or key evaluates to `undefined`, the property will be omitted from the output.
 
 **See test in {evaluator module}/src/resolvers/buildObject.test.ts for further examples**
 
--- a/documentation/Query-Syntax.md
+++ b/documentation/Query-Syntax.md
@@ -386,19 +386,21 @@ Input
 ```
 {
   operator: "buildObject",
-  properties: [{
-    key: "someKey",
-    value: "someValue"
-  },{
-    key: {
-      operator: "+",
-      children: ["evaluted", "key"]
+  properties: [
+    {
+      key: "someKey",
+      value: "someValue"
     },
-    value: {
-      operator: "AND",
-      children: [true, false]
-     },
-   }
+    {
+      key: {
+        operator: "+",
+        children: ["evaluated", "key"]
+      },
+      value: {
+        operator: "AND",
+        children: [true, false]
+       },
+     }
   ]
 }
 ```
--- a/documentation/Query-Syntax.md
+++ b/documentation/Query-Syntax.md
@@ -350,7 +350,9 @@ Using custom object shape for configuration vs children array
 properties: [{
   key: {evaluation or value}
   value: {evaluation or value}
-}, {more key values}
+},
+  ... { more key-values }
+]
 ```
 
 - Output evaluated object, empty object by default. If value or key evaluates to undefined it wont' be present in the output
--- a/documentation/Query-Syntax.md
+++ b/documentation/Query-Syntax.md
@@ -337,6 +337,78 @@ This expression queries our `/login` endpoint to check a user's credentials, and
 }
 ```
 
+## buildObject
+
+Build a key/value object with evaluatable keys and values
+
+Using custom object shape for configuration vs children array
+
+- Input format
+
+```
+properites: [{
+  key: {evaluation or value}
+  value: {evaluation or value}
+}, {more key values}
+```
+
+- Output evaluated object, empty object by default. If value or key evaluates to undefined it wont' be present in the output
+
+**See test in {evaluator module}/src/resolvers/buildObject.test.ts for further examples**
+
+**Example**:
+
+Input
+
+```
+{
+  operator: "buildObject",
+  properties: [{
+    key: "someKey",
+    value: true
+  }]
+}
+```
+
+Output
+
+```
+{
+  "someKey": true
+}
+```
+
+Input
+
+```
+{
+  operator: "buildObject",
+  properties: [{
+    key: "someKey",
+    value: "someValue"
+  },{
+    key: {
+      operator: "+",
+      children: ["evaluted", "key"]
+    },
+    value: {
+      operator: "AND",
+      children: [true, false]
+     },
+   }
+  ]
+}
+```
+
+Output
+
+```
+{
+  "someKey": "someValue",
+  "evaluatedKey": false
+}
+```
+
 # Usage
 
 The query evaluator is implemented in the `evaluateExpression` function:
--- a/documentation/Query-Syntax.md
+++ b/documentation/Query-Syntax.md
@@ -347,7 +347,7 @@ Using custom object shape for configuration vs children array
 - Input format
 
 ```
-properites: [{
+properties: [{
   key: {evaluation or value}
   value: {evaluation or value}
 }, {more key values}
--- a/src/modules/expression-evaluator/src/resolvers/buildObject.test.ts
+++ b/src/modules/expression-evaluator/src/resolvers/buildObject.test.ts
@@ -1,0 +1,123 @@
+import evaluateExpression from '../evaluateExpression'
+
+test('Testing basic buildObject', () => {
+  const testIn = {
+    operator: 'buildObject',
+    properties: [
+      {
+        key: 'key',
+        value: 'value',
+      },
+    ],
+  }
+  const testOut = {
+    key: 'value',
+  }
+  return evaluateExpression(testIn).then((result: any) => {
+    expect(result).toEqual(testOut)
+  })
+})
+
+test('Testing buildObject for handling erronouse input', () => {
+  const testIn = {
+    operator: 'buildObject',
+    properties: [
+      {
+        value: 'missing key',
+      },
+      {
+        key: 'key',
+        value: 'value',
+      },
+      {},
+      {
+        key: 'missing value',
+      },
+    ],
+  }
+  const testOut = {
+    key: 'value',
+  }
+  return evaluateExpression(testIn).then((result: any) => {
+    expect(result).toEqual(testOut)
+  })
+})
+
+test('Testing buildObject with evaluated key and value', () => {
+  const testIn = {
+    operator: 'buildObject',
+    properties: [
+      {
+        key: 'key',
+        value: 'value',
+      },
+      {
+        key: {
+          operator: 'objectProperties',
+          children: ['key'],
+        },
+        value: {
+          operator: 'objectProperties',
+          children: ['value'],
+        },
+      },
+    ],
+  }
+
+  const testOut = {
+    key: 'value',
+    keyFromObjects: 'valueFromObjects',
+  }
+
+  const params = {
+    objects: { key: 'keyFromObjects', value: 'valueFromObjects' },
+  }
+
+  return evaluateExpression(testIn, params).then((result: any) => {
+    expect(result).toEqual(testOut)
+  })
+})
+
+test('Testing buildObject with evaluations and nesting', () => {
+  const testIn = {
+    operator: 'buildObject',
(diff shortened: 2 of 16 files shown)


CarlosNZ:
> I think it's too much work to change existing resolvers to conform with the new format but I think it might be worth doing at some stage.

How would you structure it?

CarlosNZ:
> I think it's easier to have resolver specific configuration shape then the generic children shape

Yeah that makes sense for the more complex operators that require a very specific input set. Just need to make sure we don't make it more complicated for the simpler operators that just take an array of similar children (the logical operators, +, etc.)

documentation/Query-Syntax.md
```diff
@@ -337,6 +337,78 @@ This expression queries our `/login` endpoint to check a user's credentials, and
...
+  },{
+    key: {
+      operator: "+",
+      children: ["evaluted", "key"]
+    },
+    value: {
+      operator: "AND",
+      children: [true, false]
+     },
+   }
```
nmadruga, line 398: This input seems to have 2 objects and the output 1... Is the first object left there by mistake?

src/modules/expression-evaluator/src/resolvers/buildObject.test.ts
```diff
@@ -0,0 +1,123 @@
...
+  }
+  const testOut = {
+    key: 'value',
+  }
+  return evaluateExpression(testIn).then((result: any) => {
+    expect(result).toEqual(testOut)
+  })
+})
+
+test('Testing buildObject for handling erronouse input', () => {
```
nmadruga, line 21: `erronouse` ? Also I don't really like the tests using:
`{        key: 'key',        value: 'value',      }` because it's so abstract that makes a not in the head. Could you change the actual key and value to something else please?

nmadruga (approved):
Just a few comments. Code looks good, still trying to understand this, but the changes are helpful for understanding the whole evaluationExpression easier. Liked the types, not sure where isEvaluationExpression is used yet, but will have another look when reviewing front-end...

documentation/Query-Syntax.md
```diff
@@ -337,6 +337,78 @@ This expression queries our `/login` endpoint to check a user's credentials, and
...
+  ]
+}
```
CarlosNZ, line 400:
```suggestion
{
  operator: "buildObject",
  properties: [
    {
      key: "someKey",
      value: "someValue"
    },
    {
      key: {
        operator: "+",
        children: ["evaluated", "key"]
      },
      value: {
        operator: "AND",
        children: [true, false]
       },
     }
  ]
}
```

src/modules/expression-evaluator/src/resolvers/buildObject.ts
```diff
@@ -0,0 +1,35 @@
...
+) => Promise<ValueNode>
+
+const buildObject: BuildObject = async (buildObjectQuery, evaluateExpressionInstance) => {
+  const result: { [key: string]: ValueNode } = {}
+
+  const evaluateKeyValue = async (key: EvaluatorNode, value: EvaluatorNode) => {
+    const [resultKey, resultValue] = await Promise.all<ValueNode>([
+      evaluateExpressionInstance(key),
+      evaluateExpressionInstance(value),
+    ])
```
CarlosNZ, line 23: I think I can see why you did it like this, but I think it's cleaner if each operator/resolver receives pre-evaluated "children"/"properties". If we extend later this as per your suggestions (i.e. named parameters rather than "children" array), we don't want to have to write this same "evaluate all my children" code inside each operator. Surely it's simpler to evaluate all children/properties at the top level and pass results to each operator (like it does with the current "children" array).

I think it would be fairly straightforward to extend the existing "evaluate the children" code to also iterate over additional object keys and evaluate them too, don't you think?

CarlosNZ (changes requested):
Cool, I agree that migrating to named properties for each evaluator type rather than "children" is a good idea in the long-term. I think initially, the operators all acted on multiple children of the same type (AND, =, +, etc), so a simple array made sense,but the later ones have become more strictly defined, hence the complication.

I do think we should handle iterating and evaluating these properties/children at the top level though, rather than adding recursive calls inside each operator like you've done with buildObject. (See my comment inline) 


Also -- the previous test suite is failing, I think because the names of the templates have changed. No worries if it's too much hassle to fix now, I'll re-address it sometime soon.

src/modules/expression-evaluator/src/resolvers/buildObject.ts
CarlosNZ: I think this would also prevent the need to pass in a "evaluateExpressionInstance" -- it would all be done already.

documentation/Query-Syntax.md
nmadruga: Hmm. Yea, that makes sense! I found to be very confusing with the examples.

src/modules/expression-evaluator/src/resolvers/buildObject.test.ts
andreievg: 745d643c3dee8ea136dd7377ee0c021562db37aa

documentation/Query-Syntax.md
```diff
@@ -337,6 +337,78 @@ This expression queries our `/login` endpoint to check a user's credentials, and
 }
 ```
 
+## buildObject
```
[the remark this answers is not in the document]
andreievg: 60d49438ca293fd85ccf809983f1d7e8ce9bf889

src/modules/expression-evaluator/src/resolvers/buildObject.ts
andreievg: Sure, if there is obvious similarity in some shapes of configurations we can either add helper methods or find out a generic way of passing through configurations of evaluations that can be done by core evaluator logic. (would prefer adding helper methods, in evaluation instance that is passed through, i.e. evaluateArray(array), evaluateArrayObjects(array, ['key', 'value'])

Although I dont' think it would be that simple for some:

I guess a good way to address this point is -> if we see repetitions of evaluation in future resolvers we can consolidate them with some common logic (as long we don't sacrifice readability for the sake of dry)

andreievg:
> Also -- the previous test suite is failing, I think because the names of the templates have changed. No worries if it's too much hassle to fix now, I'll re-address it sometime soon.

I just did the test on buildObject, and it seems to pass (had issues with other test in master also)

src/modules/expression-evaluator/src/resolvers/buildObject.ts
CarlosNZ: We currently iterate over each element in "children" and evaluate them before passing the children array to the relevant operator. Surely we can do the same here -- just iterate over the keys of the object (excluding "operator" and "children"), and evaluate them before passing to "buildObject". I don't think any operators need to be doing recursive calls to their nested elements when we already have functionality to handle this at the top level. It just makes it very confusing to understand to have two different ways of handling child elements.

andreievg:
> Cool, I agree that migrating to named properties for each evaluator type rather than "children" is a good idea in the long-term. I think initially, the operators all acted on multiple children of the same type (AND, =, +, etc), so a simple array made sense,but the later ones have become more strictly defined, hence the complication.

Sure, I see why common evaluation of children was used for simpler types, and yeah that makes sense (although could simplify config slightly with custom shapes for some like `=` which checks the first childrenResolved against others, which would be `{ operator: '=', compareValue: {evaluation or value}, againstValues: [{evaluations or values}]}`

andreievg:
@nmadruga 
>  Liked the types, not sure where isEvaluationExpression is used yet, but will have another look when reviewing front-end...

It's used in front end, which previously did it's own check to see if value or evaluation is an evaluation (I previously commented that it should be exported method from evaluator, thus extracted it in this PR)

src/modules/expression-evaluator/src/resolvers/buildObject.ts
CarlosNZ: Just to demonstrate, I made a quick branch off this branch called `pre-evaluate-other-fields`. It's pretty bare bones and doesn't handle errors or anything, but it demonstrates how much simpler the "buildQuery" operator can (should?) be. I really think we should stick to the pattern of keeping the operators just doing their basic operation, and all the recursive calling in the main "evaluateExpression" function and not mix and match them. 

My changes:
- drastically simplify `buildObject` operator to only do its "operator" duties and nothing else
- In the same place that all "children" are resolved, wrote extra code to check and resolve expressions on any other properties of the passed-in expression (including arrays). This means we can re-use the same logic when we refactor/create other operators based on your named parameters suggestion. It's a bit ugly currently, but just trying to make the point. 

Only tested in GUI builder thus far, but working fine for basic cases.

Let me know what you think.

andreievg: Did you test in GUI after updating to use new `buildObject` ? Good way to check if it works is to run `yarn test buildObject` from evaluator module folder. 

To work for buildObject, the logic would also need to iterate over keys of objects inside the array, and try to evaluate them, and a small change in buildObject to make sure undefined values or keys are not included. I am not yet convinced that generic evaluation is worth the added complexity, unless it can be written in a readable way and all use cases are considered.

>  It just makes it very confusing to understand to have two different ways of handling child elements.

I would have thought that outsourcing logic of evaluation to resolvers simplifies things.

> It just makes it very confusing to understand to have two different ways of handling child elements.
> I don't think any operators need to be doing recursive calls to their nested elements when we already have functionality to handle this at the top.

It's quite a bit different, since the shape of configuration object is different (and may vary for future resolvers), in this case we just want to evaluate two keys (key and value) inside objects inside an array, vs evaluating children in an array. 

Anyways, I dont' think this is a priority for now, but if you come up with a readable solution, be keen to check it out. (just be wary of trying to do too much in one place to accomodate all use cases)

CarlosNZ: Yeah, the evaluation of the `{key..., value:...}` objects is the only bit that can't easily be handled by a generic process, so it makes sense the way you've done it. Although I'm not sure I see the need for a unique format, when we could have just done:
```
{
 operator: 'buildObject',
 children: [ <key>, <value>, <key>, <value>... ]
}
```
which would have made it a lot simpler.

I get that named parameters makes sense for the operators that have a very specific, ordered set of nodes (i.e. GraphQL, API etc.), but when the nodes are all a homogeneous collection (or close to it), what's wrong with `children` array? (You could always name it "properties" instead of "children", but make the shape the same)

Anyway, all good for now, but I think we should plan for further extensions so we don't end up with every operator having its own logic and style. I think named parameters/fields is a good idea, but I think it should be possible to do it all in a generic way so as to keep the operators simple.

CarlosNZ (approved):

andreievg:
Thanks for the reviews !

@CarlosNZ 

Happy to keep conversation going re structure of configurations. To answer your latest question:

> I get that named parameters makes sense for the operators that have a very specific, ordered set of nodes (i.e. GraphQL, API etc.), but when the nodes are all a homogeneous collection (or close to it), what's wrong with children array?

I am trying to break down the reason for my preference for more structured configuration shape, I think it's the same reason map or objects are used in modern dev, can still use arrays for all of the things objects are used in any programming language, but it seems more readable. I don't really want to delve on aspects of human psychology to explain why I feel like object shape is easier to digest than array shape (I think config should be pretty self explanatory, without deducing data, especially if data is also an evaluation expression)

PR #397, opened 2021-06-08
change for snapshot 2.0

andreievg:
Didn't mean to do a massive PR like this (sorry not sure how to review it), but had to bring snapshot functionality up to this standard, since we are going to be using snapshots to share configurations to be imported in demo server. I'll try to list changes later on, but for now a summary of `what it does`

(need to `yarn install` since a few new packages are added)

* `yarn snapshot take`, `yarn snapshot use` (can provide 'profile' via --options {optionsName from snapshotOptions folder}), just did a quick test and snapshot name is broken, will fix in the morning **fixed**
* now creates zip file in snapshot folder
* options can include a filter, there is an example option for template export with filter 
* options are saved in snapshot, use snapshot will use that option unless other option is specified (technically can take full snapshot and import only subset)
* schema changes are included in snapshot (i.e. look up tables and custom outcomes)
* uses postgres view vs graphql introspection (don't need to specify generated columns etc...)
* if non re-initilisation import, will create new ids (and link them on the go), can be used for export and import of templates only (also filter when exporting will trickle down to related tables, i.e. if exporting templates and template sections and templates elements, if template is filtered then only template sections and template elements for that template will be exported)
* stronger typed database scripts

@nmadruga, i just tested the scenario you were talking about, I've initialised data, started a new feature showcase application, entered some things in there, then did `yarn snapshot take --options applicationData new`, then changed something in feature showcase template, then did `yarn snapshot use new`, and after that, reloaded application page (saw my application changes and template changes), `applicationData.json` in `database/snapshotOptions` can be adjusted/duplicated to work for laos locale (i think changing locale alone might work), would need to include outcome tables and perhaps exclude some _common scripts if you want to test default value insertion

TODO documentation


nmadruga:
Is this one ready for a review?

CarlosNZ:
Just a quick comment -- I think we should keep snapshots out of the repo (i.e. put the snapshots folder in .gitignore). I think it would be easier to manage and share snapshots with Google Drive, don't you think? (We don't want to have to make PRs for every little snapshot change, and it makes the diff messy if you include them with other PRs) Is there any advantage to including them?

andreievg:
> Just a quick comment -- I think we should keep snapshots out of the repo (i.e. put the snapshots folder in .gitignore). I think it would be easier to manage and share snapshots with Google Drive, don't you think? (We don't want to have to make PRs for every little snapshot change, and it makes the diff messy if you include them with other PRs) Is there any advantage to including them?

Ok, I liked the idea of showing diff between snapshots, but that was implemented in a different way now. 8526705edc64c6b3282d52c22e9b60079e356c50

andreievg:
@nmadruga

> Is this one ready for a review?

 yep it should be, although the diff is massive

database/insert_data.sh
```diff
@@ -2,17 +2,13 @@
...
 # Makes script wait until async node script has completed
 PID=$!
 wait $PID
 
-# Turn on Row-level Security
-echo -e "\nEnabling Row-level Security..."
-psql -U postgres -q -b -d tmf_app_manager -f "./database/turn_on_row_level_security.sql"
-
-node ./database/updateRowPolicies.js --from_insert_data.sh &
+yarn ts-node ./database/updateRowPolicies.ts &
```
CarlosNZ, line 11: Should us updateRowPoliciesCLI.ts

package.json
```diff
@@ -40,21 +40,25 @@
...
-    "use_snapshot": "node ./database/useSnapshot.js"
+    "snapshot": "ts-node ./database/snapshotCLI.ts"
   },
   "dependencies": {
     "@graphile-contrib/pg-simplify-inflector": "^6.0.0",
     "@openmsupply/deep-comparison": "^1.0.1",
+    "@types/adm-zip": "^0.4.34",
     "@openmsupply/expression-evaluator": "^1.8.0",
     "@types/jsonwebtoken": "^8.5.0",
+    "@types/lodash": "^4.14.169",
```
CarlosNZ, line 58: Add lodash to imports?

database/snapshotOptions/default.json
```diff
@@ -0,0 +1,16 @@
+{
+  "filters": {},
+  "includeTables": [],
+  "excludeTables": [
+    "actionQueue",
+    "triggerQueue",
+    "applicationListShape",
+    "elementTypePlugin",
+    "notification",
+    "trigger_queue"
```
CarlosNZ, line 10: Don't need underscore case

database/snapshotOptions/exampleSingleTemplateExport.json
```diff
@@ -0,0 +1,9 @@
+{
+  "filters": { "template": { "id": { "in": [2, 8] } } },
+  "shouldReInitilise": false,
+  "includeTables": ["templateElement", "template", "templateSection"],
```
CarlosNZ, line 4: Stage, reviewLevel, etc.

src/components/snapshots/takeSnapshot.ts
```diff
@@ -0,0 +1,141 @@
...
+      path.join(newSnapshotFolder, `${SNAPSHOT_FILE_NAME}.json`),
+      JSON.stringify(snapshotObject, null, ' ')
+    )
+
+    await fs.writeFile(
+      path.join(newSnapshotFolder, `${OPTIONS_FILE_NAME}.json`),
+      JSON.stringify(options, null, ' ')
+    )
+
+    await getScheamDiff(newSnapshotFolder)
```
CarlosNZ, line 57: Spelling

CarlosNZ (approved):
Sweet, just fix those little things we noticed and merge!

package.json
andreievg: cb1eec9d61fefc6686e4f9e6adcefb234f14c9ac

database/snapshotOptions/default.json
andreievg: 44f0b219897cae57d585bc9587010f0f905bf522

database/snapshotOptions/exampleSingleTemplateExport.json
andreievg: 59b58014f0aa34f0023b863511888491e8e1525a

src/components/snapshots/takeSnapshot.ts
andreievg: 5bc70af608b06e5fbe3d589139a19430fc8e47b7

database/insert_data.sh
andreievg: 323d4f9e47bba89b67e9c8cebad868c9f51f88c7

PR #401, opened 2021-06-09
Build updates

andreievg:
Build changes from demo branch

### Changes

* Change to using ts file vs json file for config, to make sure graphql url adjust correctly in production
* Make sure there are no references to localhost:8080 in configs
* change permissions of postgres inside docker to allow access to root user (without gosu to postgres)
* When building docker image, tag it in appropriate way
* Copy database folder to build

To test, change dockerise.sh to have 'build-updates' for both back and and front end.

run that script file from docker folder

after build (would take awhile) change `testbuild` in run.sh to `irims-demo: front-build-updates_back-build-updates_pg-12_node-14` (or image id from `docker images` output), and ./run.sh (make sure to shut down dev web app)


database/insertData/_common/03_template_C_userRegistration.js
```diff
@@ -73,7 +73,16 @@ exports.queries = [
                       validation: {
                         operator: "API"
                         children: [
-                          "http://localhost:8080/check-unique"
+                          {
+                            operator: "CONCAT"
```
CarlosNZ, line 77:
```suggestion
                            operator: "+"
``` 
;)

database/insertData/_common/03_template_G_userEdit.js
```diff
@@ -110,7 +110,16 @@ exports.queries = [
                           {
                             operator: "API"
                             children: [
-                              "http://localhost:8080/check-unique"
+                              {
+                                operator: "CONCAT"
```
CarlosNZ, line 114:
```suggestion
                                operator: "+"
```

database/insertData/dev/03_template_A_featureShowcase.js
```diff
@@ -155,7 +155,16 @@ exports.queries = [
                       validation: {
                         operator: "API"
                         children: [
-                          { value: "http://localhost:8080/check-unique" }
+                          {
+                            operator: "CONCAT"
```
CarlosNZ, line 159:
```suggestion
                            operator: "+"
```

CarlosNZ (commented):
> Change to using ts file vs json file for config, to make sure graphql url adjust correctly in production

That's cool, but if we want to make a UI to change configurations, it'll have to save it to a JSON file (or .ini file), won't it? (That would be simpler at least)

andreievg:
Thanks, I keep on forgetting about 'concat' -> '+'.

> That's cool, but if we want to make a UI to change configurations, it'll have to save it to a JSON file (or .ini file), won't it? (That would be simpler at least)

good point. I guess it should still be json, but the config file is replaced with production one. I'll aim to do that further upstream

nmadruga (commented):
Just had a look in the code and tried to run both back-end and front-end on `dev`, now I'm trying to run the docker - after having to install a more recent version than my current v17.

Will get back to you after have finished testing this local dockerise version of the App : )

nmadruga (approved):
Build seems to be working well. Thanks for the updates and easy to follow steps for building. I thought it was interesting to have a counting on the `step` during the building process.

I noticed only that this is using the `dev` config, so just a warning maybe to change to use the laos config as default? Since we are using this to upgrade the server after new changes are done - I would assume that now this update makes the production work correctly with paths to server we would be using this in a very straight forward way when we want to upload latest versions of the App for others. 

Just another question - how to update the server without changing the current data in use - just use the snapshot before and then after the upgrade?

database/buildSchema/41_databse_info.sql
```diff
@@ -1,38 +0,0 @@
...
-    JOIN information_schema.columns AS columns_info ON tables_info.table_name = columns_info.table_name
-    LEFT JOIN constraints_info ON columns_info.table_name = constraints_info.from_table_name
-        AND columns_info.column_name = constraints_info.from_column_name
-WHERE
-    tables_info.table_schema != 'pg_catalog'
-    AND tables_info.table_schema != 'information_schema'
-ORDER BY
-    columns_info.table_name,
-    columns_info.column_name;
-
```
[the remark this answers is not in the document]
andreievg: yeah, somehow it's a duplicate

andreievg:
> I noticed only that this is using the dev config, so just a warning maybe to change to use the laos config as default? Since we are using this to upgrade the server after new changes are done - I would assume that now this update makes the production work correctly with paths to server we would be using this in a very straight forward way when we want to upload latest versions of the App for others.

I've made the change in b0-0-4 to use laos, should have changed it here also.
Changed now to be specified by input args for docker build: d468a01b5a7b3c496e31daa581a3b78aa3f76f82

> Just another question - how to update the server without changing the current data in use - just use the snapshot before and then after the upgrade?

Yeah, unless there are schema changes or breaking changes to schema logic, this should work. We haven't thought about migration strategy yet, so it's possible that data won't comply with above two, and we dont' have a way to handle this yet. Ideally we would have proper migration, which would be -> placeholder issue: (https://github.com/openmsupply/application-manager-server/issues/414)

PR #402, opened 2021-06-10
bind snapshot routes

andreievg:
Similar to what was done for demo server, add routes for:

`/snapshot/take`
`/snapshot/use`
`/snapshot/list`
`/snapshot/upload`
`/snapshot/files`

Need for front end PR: https://github.com/openmsupply/application-manager-web-app/pull/781

Will try to link to epic and create issue for this after demo server is updated


src/components/snapshots/routes/index.ts
```diff
@@ -0,0 +1,22 @@
...
+import routeUploadSnapshot from './routeUploadSnapshot'
+
+const snapshotRoutes: FastifyPluginCallback<{ prefix: string }> = (server, _, done) => {
+  server.register(fastifyStatic, {
+    root: SNAPSHOT_FOLDER,
+    prefix: '/files/',
+  })
+  server.get('/list', routeListSnapshots)
+  server.post('/take', routeTakeSnapshot)
+  server.post('/use', routeUseSnapshot)
```
CarlosNZ, line 16: Since these two just use a single query parameter, why not GET methods? Seems like "/upload" is the only one that needs to be POST?

src/components/snapshots/takeSnapshot.ts
```diff
@@ -72,7 +67,7 @@ const getOptions = async (optionsName?: string, options?: ExportAndImportOptions
     return options
   }
 
-  const optionsFile = path.join(optionsFolder, `${optionsName}.json`)
+  const optionsFile = path.join(SNAPSHOT_OPTOINS_FOLDER, `${optionsName}.json`)
```
CarlosNZ, line 70: Spelling "SNAPSHOT_OPT**IO**NS_FOLDER"

src/server.ts
```diff
@@ -71,6 +72,9 @@ const startServer = async () => {
 
   server.register(lookupTableRoutes, { prefix: '/lookup-table' })
 
+  // Snapshout routes
```
CarlosNZ, line 75:
```suggestion
  // Snapshot routes
```

CarlosNZ (changes requested):
Looks great, works well :) 

Just a couple of little things

src/components/snapshots/takeSnapshot.ts
andreievg: 051ec5e8a637d984913cbd645c0f5a78e336c990

src/server.ts
andreievg: Aggressive routes
051ec5e8a637d984913cbd645c0f5a78e336c990

src/components/snapshots/routes/index.ts
CarlosNZ: Although I guess it fits the convention of a POST request causing a "change" or side-effect on server, so fine as is 👍

andreievg: I was thinking that config (options) could be passed through from front end (like the filter for template etc..). Can change to get for now, would TO-DO be ok for now ?

8d8a13b03351e358d3ff5b6a1f3bd4574624b181

CarlosNZ: Yeah, sounds good -- you might have missed my prev comment before yours, so fine to leave as-is.

CarlosNZ (approved):

PR #404, opened 2021-06-10
#719F permission policy getters can deal with empty string claims

andreievg:
When restricting by row level policies, I was getting error messages `"message": "invalid input syntax for type boolean: \"\"",` (or for boolean) after changing user, the problem would go away after some delay, but it made policy restricted demo unusable. 

I think if claim is already set, postgraphile would make it an empty string '' when jwt changed. And existing jwt helpers could handle absence of claims but not empty string when casting to bigint and boolean. I think the problem went away after a delay because session (pool?) was reset and no claims were set.

Replication:
`yarn pg_permissions` instead of just `yarn pg`

go to application list with some user, then change to another user that can also see this application list (but through different permission name), you should see an error (instead of empty list)

### DIFF
--- a/database/buildSchema/02_jwt_config.sql
+++ b/database/buildSchema/02_jwt_config.sql
@@ -18,10 +18,15 @@ STABLE;
 CREATE FUNCTION jwt_get_boolean (jwt_key text)
   RETURNS boolean
   AS $$
-  SELECT
-    COALESCE(current_setting('jwt.claims.' || $1, TRUE)::bool, FALSE)
+BEGIN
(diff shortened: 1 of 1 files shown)


nmadruga:
Oh no! PR #404 🚫

nmadruga (approved):
Wonderful!! This fix problem described in #719! Yey 🎉

andreievg:
> Oh no! PR #404 🚫

Yeah i've noticed, quick merge merge

> Wonderful!! This fix problem described in Front-end#719! Yey 🎉

Nice !

PR #407, opened 2021-06-11
Server to use admin jwt for graphQL queries

andreievg:
There was a problem running graphQL queries with server when permissions are enabled. 

### Changes

* Starting postgraphile with row level permission with postgres user but with default role 
* Add role as 'postgres' to admin JWT token (loaded into singleton graphQL provider)

### Problem Replication
* start PG with row policies enabled, try to start/create a review, graphQL query fails to get reviewData and review is locked (due to changeStatus action for locked returning 'can't resolve object')

### DIFF
--- a/docker/database.sh
+++ b/docker/database.sh
@@ -15,4 +15,4 @@ yarn postgraphile -c "postgres://postgres@localhost/tmf_app_manager" --watch --d
 sleep 3
 
 echo '--- ADDING DATA'
-./database/insert_data.sh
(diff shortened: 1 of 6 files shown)


nmadruga:
Not really sure what is this fix for, do we need to prevent something else?
Do you know why it would only break when trying to create a new review?

nmadruga:
I would like to merge this one as well @andreievg, if that's all right with you. Cheers

nmadruga (approved):

PR #408, opened 2021-06-11
#406 grant permissions to new templates on approval of company license

### LINKED ISSUE #406: Grant permission to new blank-templates when company license is approved

nmadruga:
Fixes #406

### Changes
- Add `grantPermission` to apply to either `importPermit`, `drugRegoMMD` and `drugRegoMMC` depending on what was selected in the `companyLicense` application.
- Update action `grantPermission` to also take `orgId` (instead of `orgName`) when create a new permission_join between user, organisation and template.

### DIFF
--- a/src/components/postgresConnect.ts
+++ b/src/components/postgresConnect.ts
@@ -651,26 +651,17 @@ class PostgresDB {
     org: string | number,
     permissionName: string
   ) => {
-    const text =
-      typeof org === 'number'
-        ? `
-      INSERT INTO permission_join (user_id, organisation_id, permission_name_id)
-      VALUES (
-        (select id from "user" where username = $1),
-        $2,
-        (select id from permission_name where name = $3)
-      ) RETURNING id`
-        : `
-    insert into permission_join (user_id, organisation_id, permission_name_id) 
+    const text = `
+    INSERT INTO permission_join (user_id, organisation_id, permission_name_id) 
     values (
         (select id from "user" where username = $1),
-        (select id from organisation where name = $2),
+        ${typeof org === 'number' ? '$2' : '(select id from organisation where name = $2)'},
         (select id from permission_name where name = $3))
     ON CONFLICT (user_id, organisation_id, permission_name_id)
       WHERE organisation_id IS NOT NULL
     DO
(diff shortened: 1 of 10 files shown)


CarlosNZ (approved):

nmadruga:
I went investigating this one, and realised what is the problem... In the `grantPermissions` we try to get the organisation_name by this:
```
orgName: {
                      operator: "objectProperties"
                      children: ["outputCumulative.organisation.name"]
                    }
```

but there is none on the `outputCumulative`, I think when I was testing I was using create company template and that's why this was working there. I need a hand with this @CarlosNZ I had a check today but I'm not sure where to get this `orgName` from, we have acess to the `orgId` without changes though. Should I create something similar to `joinPermissionNameToUserOrg` in the Action plugin to `grantPermissions` that uses the orgId instead?

src/components/postgresConnect.ts
```diff
@@ -648,10 +648,19 @@ class PostgresDB {
 
   public joinPermissionNameToUserOrg = async (
     username: string,
-    orgName: string,
+    org: string | number,
     permissionName: string
   ) => {
-    const text = `
+    const text =
```
CarlosNZ, line 654: How about just:
```
const text =
    INSERT INTO permission_join (user_id, organisation_id, permission_name_id) 
    values (
        (select id from "user" where username = $1),
        ${typeof org === 'number' ? "$2" : "(select id from organisation where name = $2)"},
        (select id from permission_name where name = $3))
    ON CONFLICT (user_id, organisation_id, permission_name_id)
      WHERE organisation_id IS NOT NULL
    DO
    		UPDATE SET user_id = (select id from "user" where username = $1)
    RETURNING id
    `
```
I think the `ON CONFLICT...  DO....` stuff is important too, and you didn't have it in your query, was that intentional?

CarlosNZ (changes requested):
Haven't tested, but looks good. Just a suggestion about simplifying the query, since most of it is duplicated.

src/components/postgresConnect.ts
nmadruga: Good point, updating to this now

CarlosNZ (approved):
Sweet. Thanks :)

PR #412, opened 2021-06-18
#477F - part 1 - Update review assignment with stage date

nmadruga:
### Linked PRs
[Front-end](https://github.com/openmsupply/application-manager-web-app/pull/798)

Separated PR to update `review_assignment` to also have `stage_date` when created. This is useful for when querying for all assignment and related reviews, the stage date is present and can be set in the Review structure on the front-end.

This might be not the best way to do this (since is another property being passed from one place to another) but that's how we have been doing with stage properties to get added to Review entity - they are set on the reviewAssignment and are automatically generated on the Review record as well. If you think in a clearer way more straight forward, let me know.


plugins/action_generate_review_assignment_records/src/generateReviewAssignments.ts
```diff
@@ -150,8 +153,9 @@ const generateNextReviewAssignments = async ({
       reviewAssignments[userOrgKey] = {
         reviewerId: userId,
         orgId,
-        stageId: nextStageId,
-        stageNumber: nextStageNumber,
+        stageId,
+        stageNumber,
+        stageDate,
```
andreievg, line 158: i think you'll need to add this to `addReviewAssignments` db method

nmadruga: Yes, thanks!

CarlosNZ (commented):
Cool -- as discussed, I think the name `stageDate` is a bit unclear. What about either keeping the original field name `stageHistoryTimeCreated`, or maybe just `stageTimestamp` if you want something more succint?

nmadruga:
> `stageHistoryTimeCreated`

Doing this change now. Thanks

PR #415, opened 2021-06-20
#604F filter UI

andreievg:
Required for front end branch in PR: https://github.com/openmsupply/application-manager-web-app/pull/799

### Changes

* application list returns users with first and last name concatenation vs username
* set all user data to have last name
* make sure boolean field either false or true (no nulls as that breaks graphQL equal or notEqual filter) 
* add filter

### Testing

* Best way to test is to play around with filters in application list in the related front end branch


database/insertData/dev/04_users.js
```diff
@@ -93,6 +94,7 @@ exports.queries = [
           passwordHash: "$2a$10$ne2WcPISMw/Do3JzlwThYeO2GcodrumjI3FwGu1ZUoKgRQyAgNS3e"
           username: "valerio"
           firstName: "Valerio"
+          lastName: "R"
```
CarlosNZ, line 97: This would still work if lastName wasn't specified, wouldn't it?

(I think it should)

Perhaps omit lastName from a couple of the users so we can see what happens/how it appears?

database/buildSchema/38b_application_list_filters.sql
```diff
@@ -0,0 +1,81 @@
+CREATE FUNCTION application_list_filter_applicant (applicant varchar, template_code varchar)
```
CarlosNZ, line 1: I'm not entirely clear why these need to be functions, as opposed to just queries on existing tables. Can you briefly explain it?

andreievg: there is not distinct aggregate on graphQL queries. (I couldn't find any plugins that do it either)

To get a list of distinct values in graphQL we need to query all applications and compute distinct values on front end : /

nmadruga:
I think this one is pending changes-requests on [Front-end-PR](https://github.com/openmsupply/application-manager-web-app/pull/799)

database/insertData/dev/04_users.js
andreievg: Oops i never replied, I think the issue was with extra space at the end of applicant/assigners/reviewers column if last name is missing, which is not a biggy i guess

Issue #413, opened 2021-06-18
Bug: Reviewer and Consolidator can't see each other reviews


andreievg:
Permission Policy issue, as per title.

Replication -> try consolidation with row policy disabled and then enable row level policies and check how consolidation looks.

andreievg:
We can't use applications to determine visibility of reviews as per: https://github.com/openmsupply/application-manager-server/pull/359#issuecomment-847620635

Thus we are left with review_assignment, I think we can just use application_id from review_assignment to filter out reviews (rather then review_assignment_id)

PR #421, opened 2021-06-21
#413-adjust-review-visibility-policy

andreievg:
@nmadruga does this work now as per the use case you've mentioned ? (I've tested, seems to work with my example, although in my example consolidator's response were visible, but not their name, after this change their name appeared, which suggests review became visible)

### Changes

* restrict review by review assignment application_id (vs reviewer_id)

### DIFF
--- a/database/insertData/_common/01_permission_policies.js
+++ b/database/insertData/_common/01_permission_policies.js
@@ -145,7 +145,17 @@ exports.queries = [
         },
         review: {
           view: {
-            reviewer_id: 'jwtUserDetails_bigint_userId',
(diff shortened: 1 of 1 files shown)


nmadruga (approved):
Thanks, that seems to solve the problem well.

nmadruga:
@andreievg can this one be merged?

andreievg:
Sure can be merged, although the permission system is quite slow in dockerised postgres (fast on our local postgres though), I have a change that I've done in b0-0-9 which will be transferred to a PR to

PR #422, opened 2021-06-22
#57 Send notification/email Action

### LINKED ISSUE #57: Create action to send the email
The email action plugin should already consider parameter to add PDF when the Emails are sent for some Application update. For the user registration:
- Email would be sent with link to finish registration of user.
- Email is send after registration is complete with welcome message (template_message).

CarlosNZ:
Fixes #57 (Initial implementation)

This Action creates a record in the notification table (we don't do anything with this yet, it's basically a log for now) and sends an email to specified user(s)

## Config

- Run `yarn install` to add packages
- Add a `.env` file to the project root with the line `SMTP_PASSWORD=<password>` where "password" can be found in Bitwarden named "IRIMS test email address"
- The email server configuration is in the plugin's own "config.json" file, but it should already be set up for the test email account above

## Testing

- I've only added it to the dev "orgJoin" template currently -- a notification/email will be sent once the application is approved by a reviewer (with an attachment, just for testing purposes)
- You can play round with the actions in `devActions` to see the effects of various parameters -- trigger them by manually applying "DEV_TEST" trigger to any application (but please change the email recipients otherwise you'll be spamming me!)

## Issues / Discussion

- I'm not super happy with how attachments are handled. Currently, we take an array of file "uniqueIDs" (from file table), OR pre-packaged `{path, filename}` objects (the type required by the sendMail method) for attaching files that are not in the file table (or external files). I think we will have to extend/re-factor this somewhat once we have a clearer idea of where input files might be coming from (e.g. PDFs created on the fly)
- The email sending using this server is pretty slow, and sometimes the trigger even times out in the front-end before it's finished. We can't add this action as an "Async" one, since it needs the result of previous actions. I was thinking I should make the actual email sending async within the action (and have it update the notification table using `.then()`), but then we don't get a "email_sent" confirmation in the output of the action itself. This is probably okay, since the notification table would still record if the email was successfully sent or not (as it does currently -- that's what the "email_sent" field is for), but I thought I'd get your input to see if you have any other suggestions.
- There seems to be a problem the imported @types file for nodemailer -- the `SentMessageInfo` type (the response from sent email) doesn't have some fields defined (e.g. "response"), so when I attempt to use them it gives an error. Currently just using "any" in the meantime, but if you have any ideas? (Suspect nodemailer has been updated but the types file is a few versions behind, not sure best way to deal with)

## To-Do:
- documentation


database/insertData/_helpers/dev_actions.js
```diff
@@ -3,18 +3,29 @@ GraphQL Fragment - DEVELOPMENT Actions
...
       trigger: DEV_TEST
       sequence: 1
       parameterQueries: {
-        tableName: "user"
-        matchField: "username"
-        first_name: "Boba"
-        last_name: "Fett"
-        username: "js"
-        email: "nowhere@outerrim.space"
+        email: ["carl@sussol.net", "ceejaysmith@gmail.com"]
```
nmadruga, line 12: probably better to change this to some sussol test email?

plugins/action_send_notification/src/sendNotification.ts
```diff
@@ -0,0 +1,127 @@
...
+    let notificationResult
+    notificationResult = await db.createNotification({
+      userId,
+      applicationId: applicationData?.applicationId,
+      emailAddressString,
+      subject,
+      message,
+      attachments,
+    })
+
```
nmadruga, line 54: Should check or the try catch would stop here if notification not correctly created?

plugins/action_send_notification/src/sendNotification.ts
```diff
@@ -0,0 +1,127 @@
...
+    // Send email
+    if (sendEmail) {
+      console.log('Sending email...')
+      // Note: Using "any" type as imported @types defintions is incorrect, doesn't recognise some fields on "SentMessageInfo" type
+      const emailResult: any = await transporter.sendMail({
+        from: `${fromName} <${fromEmail}>`,
+        to: emailAddressString,
+        subject,
+        text: message,
+        html: marked(message),
```
nmadruga, line 64: Curious what does `marked` package does?

plugins/action_send_notification/src/sendNotification.ts
```diff
@@ -0,0 +1,127 @@
...
+  appRootFolder: string,
+  filesFolder: string
+): Promise<Attachment[]> => {
+  const attachmentObjects = []
+  for (const file of attachments) {
+    if (typeof file === 'object') {
+      if (!file?.path || !file?.filename) throw new Error('Invalid attachment')
+      attachmentObjects.push(file)
+    } else {
+      const { original_filename, file_path } = await getFilePath(file)
```
nmadruga, line 119: Is this `original_filename` that you said would be good to get from uploaded files that aren't getting the proper original name from file but a random number instead?

nmadruga (changes requested):
That's great Carl. I make a few comments, mostly just questions to understand how to continue this work and have all remaining tasks for notifications setup.
I really liked how this is already set up to be used for other notifications as well. And we can easily add the feature to display the list of notifications to the user when the time comes.

database/insertData/dev/03_template_F_organisationJoin.js
```diff
@@ -273,6 +273,58 @@ exports.queries = [
...
+                      "### Congratulations, %1!\\n\\nYour application to join organisation %2 has been APPROVED.\\n\\n[Application Dashboard](%3)\\n\\nPlease find attached the ID file you uploaded (testing attachment handling)"
+                      {
+                        operator: "objectProperties"
+                        children: ["applicationData.firstName", ""]
+                      }
+                      {
+                        operator: "objectProperties"
+                        children: ["applicationData.responses.S1Q1.text", ""]
+                      }
+                      "http://localhost:3000" #TO-DO: add website URL to back-end config
```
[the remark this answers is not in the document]
nmadruga: Also I think would be better to link the summary page for this application, right?

Result I got on email after applying to Join organisation (as Valerio user - but using my email)
Screen Shot 2021-07-07 at 11 04 06 AM

CarlosNZ:
> I found this thread about some other problem with types from nodemailer where they say the types are unoficial
> [nodemailer/nodemailer#1094](https://github.com/nodemailer/nodemailer/issues/1094)

Cool, thanks, as I suspected. Prob easiest to just leave this particular case as `any` type for now.

database/buildSchema/34_notification.sql
```diff
@@ -3,10 +3,11 @@ CREATE TABLE public.notification (
     id serial PRIMARY KEY,
     user_id integer REFERENCES public.user (id),
     application_id integer REFERENCES public.application (id),
-    review_id integer REFERENCES public.review (id),
```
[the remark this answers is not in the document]
CarlosNZ: Yeah, good call. I've added it to the Action as well so this value gets written if available in applicationData: ca1da710

database/insertData/dev/03_template_F_organisationJoin.js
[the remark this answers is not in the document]
CarlosNZ: > Does this required a separated issue?

Yeah, done: #432

> Also I think would be better to link the summary page for this application, right?

I dunno, I though after you've been approved for a company the first thing you'd want to do would be to go to the Dashboard and see all the new things you can now do. I don't think there's much need to re-visit old applications once they're approved. 

But we can configure this more properly when we do the "real" one, anyway. It's really just a demo/proof-of-concept at this stage.

plugins/action_send_notification/plugin.json
```diff
@@ -0,0 +1,24 @@
+{
+  "type": "action_plugin",
+  "code": "sendNotification",
+  "name": "Send Email",
+  "description": "Creates a notification and send an email",
+  "required_parameters": ["subject", "message"],
+  "optional_parameters": ["userId", "fromName", "fromEmail", "attachments", "sendEmail"],
+  "output_properties": ["[record]"],
```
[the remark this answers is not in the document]
CarlosNZ: Not sure what was going on there, I think this was leftover from another action (the modifyEntity one I think). Anyway, corrected it now: 11521072

database/insertData/_helpers/dev_actions.js
CarlosNZ: Yeah, okay. The dev_actions are really meant to be just for playing round/testing and are over-written in subsequent PRs, but I've just put dummy addresses in there for now with a comment to change them: 62e7d7a4

plugins/action_send_notification/src/sendNotification.ts
CarlosNZ: If the database method throws an error, the Action will escape from the "Try" block and run "catch" -- is that what you meant?

plugins/action_send_notification/src/sendNotification.ts
CarlosNZ: It's the Markdown parser. Same core as the one we use in front-end, but the front-end one has a React wrapper.

plugins/action_send_notification/src/sendNotification.ts
CarlosNZ: It's okay in this context, as the emailer can set the filename of the outgoing file, which is what's happening here. The problem I was referring to is specifically to do with the "sendFile" method used by the `/file` download endpoint.

nmadruga:
Just something that I noticed after approving the review of application for template "Join organisation". So using user **Nicole** that has been assigned to review this Join organisation application and...
1) Clicking to **conform** and submit:
Screen Shot 2021-07-07 at 12 12 49 PM

2) It loads while waiting for actions to run: 
Screen Shot 2021-07-07 at 12 12 56 PM

3) Failed to load after 10 attempts:
Screen Shot 2021-07-07 at 12 13 04 PM

Note: refreshing will fix, it's just taking too long to finish the trigger set on this review

CarlosNZ:
> Note: refreshing will fix, it's just taking too long to finish the trigger set on this review

Yeah, the sendMail is taking too long (it's our Sussol mail server!). 

A couple of options:
- Set the sendNotification action in this template to Async -- but it'll have to work independently of other sequential actions, including its condition checking. I think it'll work okay, but there might be unforeseen issues
- I could make the `sendMail` part of the Action itself async. That would just mean that there would be no confirmation that the email actually sent as part of the Outcome object (it would just show the notification record info), but there would be a record in the "email_sent" field of the notification table.

Any particular preference, or any other suggestion?

plugins/action_send_notification/src/sendNotification.ts
nmadruga: So when we upload files?

plugins/action_send_notification/src/sendNotification.ts
nmadruga: Yes, I was unsure if the notification could not get created and not errors thrown, but I think it would throw because of the return is accessing the row[0]

database/insertData/dev/03_template_F_organisationJoin.js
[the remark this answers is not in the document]
nmadruga: Yes, it's good for now : )

nmadruga:
> * I could make the `sendMail` part of the Action itself async. That would just mean that there would be no confirmation that the email actually sent as part of the Outcome object (it would just show the notification record info), but there would be a record in the "email_sent" field of the notification table.

I vote for this solution

plugins/action_send_notification/src/sendNotification.ts
CarlosNZ: Sorry, I'm not following. Hit me up on Telegram if you want further clarification.

plugins/action_send_notification/src/sendNotification.ts
CarlosNZ: > > * I could make the `sendMail` part of the Action itself async. That would just mean that there would be no confirmation that the email actually sent as part of the Outcome object (it would just show the notification record info), but there would be a record in the "email_sent" field of the notification table.
> 
> I vote for this solution

Yeah, I think that's probably best -- will do it now.

CarlosNZ: Righto -- that's done now, you shouldn't get the timeout on Join Org now: d35dd23b

nmadruga (approved):
Thanks all good now!

Issue #417, opened 2021-06-21
Snapshot + Template builder: Update User Rego templates to use confirmation of email action


nmadruga:

nmadruga:
Already done in `core_templates` and in latest released demos...

PR #424, opened 2021-06-23
#416 Verification -- endpoint and action

### LINKED ISSUE #416: Verification endpoint - link to an action
### Back-end
- new table in DB -- "verification", with fields "unique_id", "is_verified", "timestamp", "expiry" and "trigger"
- new Trigger "ON_VERIFICATION"
- new endpoint `/verify` which takes uniqueId as query parameter and sets "is_verified" to `true` and sets trigger to "ON_VERIFICATION", returns `{verified: true}` to caller
- if current time > "expiry", reject verification 
- new Action "createVerification", which generates a uniqueId, stores it in the verification table, set to `is_verified = false` and outputs the uid to Output object (for use by subsequent Actions)

### Front-end
- new route `/verify`, where our verification links direct to. Passes the query parameter (?id=xxxx) to the `/verify` endpoint above
  - UI if not verified yet -- shows and confirms verification, link back to Dashboard (similar to Submission page)
  - If already verified, show "Already used this link" 

### Templates
- update templates (start with UserRegistration) to use this implementation:
  - ON_SUBMIT -- create verification action, and send notification with verification link in it
  - ON_VERIFICATION: create user

### Questions
- should "expiry" be a timestamp, or a duration (e.g. 6 hours). Probably best for "createVerification" action to take a duration as input, but set "expiry" timestamp at creation time, right? And what unit to accept? (I'm thinking "hours")

CarlosNZ:
Fixes #416 

Implements solution as suggested in [the issue](https://github.com/openmsupply/application-manager-server/issues/416), with some amendments basic on subsequent [discussion](https://github.com/openmsupply/application-manager-server/issues/416#issuecomment-866442834)

Requires [front-end PR #810](https://github.com/openmsupply/application-manager-web-app/pull/810)

**Note**: the additional fields on the "verification" table (code, data, etc.) exist, and can be passed as parameters, but I'm not doing anything with them yet.

### Testing

- make sure you've set up for the Send Email action to work as described in my previous PR #422 
- fresh `yarn database init dev`
- from Login page, click "create new account" and submit a new user registration form (I've put the more complex user reg into "laos" folder, and used the shortened one here in "dev")
- On submission, you should receive a verification email. Note that the user has not yet been created in the system (or any other "approval" actions)
- Click the verification link. You should see a "Successful verification" page, and the user has been created, so you can now log in with your new user account


database/buildSchema/34_b_verification.sql
```diff
@@ -0,0 +1,21 @@
+-- verification
+CREATE TABLE public.verification (
+    id serial PRIMARY KEY,
+    unique_id varchar UNIQUE NOT NULL,
+    application_id integer REFERENCES public.application (id) NOT NULL,
+    code varchar,
+    message varchar,
+    data jsonb,
+    created_time timestamptz DEFAULT NOW(),
+    expiry_time timestamptz,
```
nmadruga, line 10: I would suggest `time_created` so it has the same pattern as other places that are using similar timestamps. And I'm still reading this but `expiry_time` maybe better as `expiry_date`?

database/buildSchema/34_b_verification.sql
```diff
@@ -0,0 +1,21 @@
...
+    TRIGGER public.trigger
+);
+
+-- TRIGGER (Listener) on verification table
+CREATE TRIGGER verification_trigger
+    AFTER INSERT OR UPDATE OF trigger ON public.verification
+    FOR EACH ROW
+    WHEN (NEW.trigger IS NOT NULL AND NEW.trigger <> 'PROCESSING' AND NEW.trigger <> 'ERROR')
+    EXECUTE FUNCTION public.add_event_to_trigger_queue ();
+
```
nmadruga, line 21: Wondering if we need to list the tables that have the TRIGGER column. As ones that need to be accounted on row_level_permission. There was something about allowing these to be changed by front-end that I can't remember what it was.

database/buildSchema/34_b_verification.sql
nmadruga: Now that I found that this is a duration and not a date, I suggest `duration_until_expired` and perhaps another one saying its expired?

src/components/permissions/routes.ts
```diff
@@ -91,4 +91,35 @@ const routeUpdateRowPolicies = async (request: any, reply: any) => {
   return reply.send(await updateRowPolicies())
 }
 
-export { routeUserInfo, routeLogin, routeLoginOrg, routeUpdateRowPolicies, routeCreateHash }
+const routeVerification = async (request: any, reply: any) => {
+  const { uid } = request.query
```
nmadruga, line 95: Hmmm was wondering how we would do this server-side localisation. Have this been considered by @sworup? I have seen some strings un-localised as part of plugins, which should be tricker to solve now, but this ones could possibly be listed in a similar file as in the Front-end.

nmadruga (changes requested):
Overall comment:
- Noticed the user_registration is now only for `dev` should we have a separated issue to update laos user_registration to also use this verification?
- Email is not being check as unique. During the user_registration would be good to check that, at least that's how most of websites do. Can be anoter issue as well.

CarlosNZ:
> * Noticed the user_registration is now only for `dev` should we have a separated issue to update laos user_registration to also use this verification?

Yes, I planned to update this PR to do that for Laos once we were happy with the other aspects of it.

> * Email is not being check as unique. During the user_registration would be good to check that, at least that's how most of websites do. Can be anoter issue as well.

I thought we were not enforcing uniqueness on email, for now at least? Since Valerio mentioned that their might be a common email address for multiple users. (It can be added easily to a template though, if we need to)

src/components/permissions/routes.ts
CarlosNZ: Good point, but I'd say localising server-generated messages would be low priority for now.

database/buildSchema/34_b_verification.sql
CarlosNZ: Okay, how about `expiry_hours`, so it's clear what the units are as well?

I considered making a field for showing whether it's expired or not, but I figured that the logic required to ensure it's always up to date and accurate would be more complicated than just deducing it every time you need to check it (ie. `if time_created + expiry_hours < current time...`). Can you think of a use case that would justify stating it explicitly?

database/buildSchema/34_b_verification.sql
CarlosNZ: Yeah, I should update the "Triggers & Actions" documentation to include this. Will do.

nmadruga:
> * Noticed the user_registration is now only for `dev` should we have a separated issue to update laos user_registration to also use this verification?

Just noticed we already have this issue that can be dealing with adding this to Laos user_registration template: Update template to use confirmation of email action #417

database/buildSchema/34_b_verification.sql
CarlosNZ: Okay, have updated the list of trigger tables in docs: 1c31ee5c

database/buildSchema/34_b_verification.sql
CarlosNZ: > Now that I found that this is a duration and not a date, I suggest `duration_until_expired` and perhaps another one saying its expired?

Hang on, I was mistaken --  the `expiry_time` field _is_ a date/timestamp -- the "expiry" field in the Action input is a duration. The expiry time is calculated when the record is created:
```js
 const expiryTime = expiry ? DateTime.now().plus({ hours: expiry }).toISO() 
```

So I'll call the fields `time_created` and `time_expiring`, okay? (the expiring one is a bit ambiguous because it could actually be "expiring" or "expired" depending on if the time has passed or not.)

CarlosNZ: > (the expiring one is a bit ambiguous because it could actually be "expiring" or "expired" depending on if the time has passed or not.)

Which is why `expiry_time` is clearer, although I take your point about consistency with other cases. I think I'll go with `time_expired` -- at least the tense is consistent with others, even if it's strictly incorrect if that time hasn't passed yet.

CarlosNZ: Righto -- done! 5b5eeb76

nmadruga: Ok. Sorry I didn't read that before. So now `time_expired` is the actual time when it will expire. That's all good by me

database/buildSchema/34_b_verification.sql
nmadruga: Thanks. That's great

nmadruga (approved):

PR #425, opened 2021-07-05
#420 Generate PDF Action and endpoint

### LINKED ISSUE #420: Action to generate a PDF based on a Carbone template - temporary file that gets send via email
Carbone is the one that turn xml with a templating language with data and the next package npm (can't remember the name - in demo1 branch) to transform into PDF

CarlosNZ:
Fixes #420.

Created a general function ("generatePDF", in "documentGenerate.tsx"), and created a POST endpoint and Action that use this function to output PDFs.
The two required input parameters are "fileId" , which is the uniqueId of the carbone template, and "data", which is an object containing the data to be inserted into the carbone template replacement fields. The function also takes other input pertaining to the application, such as userId, applicationSerial, etc -- when used as an Action, this info will be filled by default with info from applicationData. 

You'll need [LibreOffice](https://www.libreoffice.org/download/download/) installed for the PDF conversion to work. And a fresh `yarn install` to get the appropriate packages.

### generatePDF function
- interacts with the "file" table in PostGres to register created files (i.e. it inserts the appropriate fields and returns a unique ID for each generated file)
- As well as the input parameters described above, I've also added an optional "subFolder" parameter. Normally files are stored in a subfolder named for the application Serial, but if "subFolder" is explicitly specified, this will take precedence. For examples, I've been putting carbone docs in "carbone_templates" subfolder (Note: this parameter is also available to the "/upload" endpoint

### `/generate-pdf` endpoint
- same parameters as above
- returns the generated document, so we can use this endpoint to create preview PDFs for example
- parameters can be provided either as query parameters or body JSON (data must be in bodyJSON, since it's a nested object)

### "Generate document" Action
- does the same as the endpoint, but all parameters other than "fileID" (of the carbone template) can be inferred from applicationData (but can be specified explicitly if required)

### Additional
- made a few additional "helper" functions (in utilityFunctions.ts):
   - `objectKeysToCamelCase` / `objectKeysToSnakeCase` to avoid problems with inconsistency of input/output parameters and make conversion between them simple. In general, we should accept both types of case for input (and convert to appropriate at start of Action), and all output fields should be camelCase. (These use lodash's "mapKeys" method)
   - `combineRequestParams` -- merges http request query params and body JSON into a single object, similar to our "query" property in useRouter in the front-end -- can be used in conjuction with the above by optionally specifying either "snake" or "camel" as 2nd parameter to apply to output object
   - `makeFolder` - simple method to create a folder if it doesn't exist, but carry on if it does (since node's `mkdir` method throws an error if you try to create a folder that already exists)

## Testing

Here's a Postman collection with the "generate-pdf" endpoint included
[Application Manager.postman_collection.json.zip](https://github.com/openmsupply/application-manager-server/files/6760786/Application.Manager.postman_collection.json.zip)

And here is a simple carbone template for testing
[doc_action_demo.odt.zip](https://github.com/openmsupply/application-manager-server/files/6760788/doc_action_demo.odt.zip)

You'll need to use the `/upload` endpoint to insert the carbone template in the database, then use its uniqueID as the "fileId" parameter in the "generate-pdf" endpoint. When successful, Postman will show the generated PDF as its response.

I've only tested the Action using "dev_actions" so far, will add to "real" templates once we've finalised the details

So for now, easiest way to test the Action is:

1. Upload the carbone template using endpoint.
2. In the database, manually change the uniqueId of that uploaded file to `9eRMD5mgLGuMfntQ1I-_9` (as that's what's specified as parameter in "dev_actions.js")
3. Run a "DEV_TEST" trigger on one of the existing applications (preferably one that is associated with your user, so you'll get the email)
4. A simple report will be generated and emailed to the applicant

## To-do

- Documentation


nmadruga:
> I've only tested the Action using "dev_actions" so far, will add to "real" templates once we've finalised the details

I think it would be nice to already plan what templates we can already add to have this PDF generation as part of some workflow. If you have some ideas can you add placeholder cards so we can discuss on our next Planning please?

CarlosNZ:
> I think it would be nice to already plan what templates we can already add to have this PDF generation as part of some workflow. If you have some ideas can you add placeholder cards so we can discuss on our next Planning please?

Okay, I've made an initial issue where we can start making a list before breaking them out into separate issues: #440

nmadruga (approved):
Everything looks good to me. Congrats on the good work Carl 😀 
Maybe @andreievg wants to have a look too.

PR #431, opened 2021-07-06
#477F part2 fix order with new timestamps fields for responses

nmadruga:
Backend PR for front-end issue [#477](https://github.com/openmsupply/application-manager-web-app/pull/803)

### Linked PR
- Based on PR #412 

### Changes
- Add new fields `timeCreated` and `timeSubmitted` for `review_responses` and `application_responses`
- on the `TrimResponses` action change to update the `timeSubmitted` instead of `timeUpdated` (that only changes when the response mutation happens) - to fix the order of responses displayed in the Show History (check in the front-end PR)
- Add some docs related to new field + few missing fields from previous documentation of Review schema.
- Updated all responses on dev insertData to have timestamps in order to show in the examples

Tests:
- Needs to check if `trimResponses` and `updateReviewStatus` are still running correctly after changes


database/buildSchema/22_application_response.sql
```diff
@@ -11,6 +11,8 @@ CREATE TABLE public.application_response (
     status public.application_response_status DEFAULT 'DRAFT',
     value jsonb,
     is_valid boolean,
-    time_updated timestamptz DEFAULT CURRENT_TIMESTAMP
+    time_created timestamptz DEFAULT CURRENT_TIMESTAMP,
+    time_updated timestamptz DEFAULT CURRENT_TIMESTAMP,
+    time_submitted timestamptz DEFAULT NULL
```
CarlosNZ, line 16: Prob don't need DEFAULT NULL -- NULL is always the default.

database/buildSchema/30_review_response.sql
```diff
@@ -25,7 +25,9 @@ CREATE TABLE public.review_response (
     review_response_link_id integer REFERENCES public.review_response (id),
     original_review_response_id integer REFERENCES public.review_response (id),
     review_id integer REFERENCES public.review (id),
+    time_created timestamptz DEFAULT CURRENT_TIMESTAMP,
     time_updated timestamptz DEFAULT CURRENT_TIMESTAMP,
+    time_submitted timestamptz DEFAULT NULL,
```
CarlosNZ, line 30: Same

plugins/action_trim_responses/src/trimResponses.test.ts
```diff
@@ -9,23 +9,23 @@ beforeAll(async (done) => {
...
+  INSERT INTO "public".application_response (id, template_element_id, application_id, "value", is_valid, time_submitted) VALUES (DEFAULT, 1009, 1000, '{"text": "Manufacturer B", "optionIndex": 1}', NULL, 'NOW()');
+  INSERT INTO "public".application_response (id, template_element_id, application_id, "value", is_valid, time_submitted) VALUES (DEFAULT, 1007, 1000, '{"text": "Manufacturer", "optionIndex": 1}', NULL, 'NOW()');
+  INSERT INTO "public".application_response (id, template_element_id, application_id, "value", is_valid, time_submitted) VALUES (DEFAULT, 1006, 1000, '{"text": "12345678"}', NULL, 'NOW()');
+  INSERT INTO "public".application_response (id, template_element_id, application_id, "value", is_valid, time_submitted) VALUES (DEFAULT, 1005, 1000, '{"text": "craig@sussol.net"}', NULL, 'NOW()');
+  INSERT INTO "public".application_response (id, template_element_id, application_id, "value", is_valid, time_submitted) VALUES (DEFAULT, 1004, 1000, '{"text": "craig_drown"}', NULL, 'NOW()');
+  INSERT INTO "public".application_response (id, template_element_id, application_id, "value", is_valid, time_submitted) VALUES (DEFAULT, 1002, 1000, '{"text": "Drown"}', NULL, 'NOW()');
```
CarlosNZ, line 17: The IDs are causing a failure -- they're referring to IDs that no longer exist. 🤷‍♂️

CarlosNZ (changes requested):
Cool, just fix those template_element_ids in the tests if you can. Thanks.

CarlosNZ:
Oh, problem with the Update_review_statuses test as well:
```
error: column "time_submitted" of relation "review_decision" does not exist
```

plugins/action_trim_responses/src/trimResponses.test.ts
nmadruga: Yes, I checked this now and there isn't an application 1000 for template "03_tempalte_A_featuresShowcase" anymore. This needs to be updated. I'll do this in a separated PR

CarlosNZ: Okay, I think we'll probably want to start using Snapshots for our test suites. Maybe discuss this at our next meeting?

PR #434, opened 2021-07-07
#433 fix trim response and update review statuses tests

### LINKED ISSUE #433: trimResponses and updateReviewStatuses tests are all failing - needs a quick fix

nmadruga:
Fixes #433 

While merging PR #431 noticed some tests were broken. In particular the ones that I had made changes (not the reason the tests were broken) and now this has a quick fix for those. The stages `ids` keep changing and I try to fix them, but then tests were failing again and I didn't want to spend more time on this. So, ideally on next fixing of tests we could fix the stages `id` by creating them with fixed ids and using the specific id instead of counting on the order.


CarlosNZ (approved):
Great! Thanks for sorting this 👍

PR #436, opened 2021-07-08
#427 Generate Name action

### LINKED ISSUE #427: New action-plugin to generate the application name on Application on Start and Submission (to update with specific fields)

CarlosNZ:
Fixes #427 

**EDIT**: This PR will be superceded by the more generic `generateTextString` Action I will be PR-ing today, so don't review the generateName action for now. The other "Additional related changes" still apply though.

It occurred to me that we could actually already do this using existing actions (ModifyRecord with an evaluator expression to create the name string using objectProperties and buildObject), so this Action is really just a simplified version of that. Due to that, this Action actually calls the modifyRecord function itself rather than duplicating existing functionality. 

The action has no compulsory parameters, and will simply update the application name with a generic string (TemplateName - Serial) if none provided. However, optional parameters can be provided to specify a formatting string (either evaluator expression OR `${ }` substitutions), and even specify the table and field to be updated, so it can be used to generate values for other things if we need it to. See Documentation for full parameter list.

### ApplicationData

I extended the `getApplicationData` method to be a little more flexible and return more data, namely:
- additionally returns templateName, templateCode, and applicationName
- updated the View `application_stage_status_all` (and indirectly "_latest") to include the above (as that is where getApplicationData gets its application info from)
- `getApplicationData` can take a more flexible input -- either an ActionPayload (as previously) OR a plain applicationId and/or reviewId, in case we wish to call it from other places (such as front-end/REST endpoint)

### ModifyRecord Action

Couple of small tweaks to modifyRecord so it plays nicely with the new action:
- added parameter `shouldCreateJoinTable` (default `true`), so the Action can be used without creating a join table for the entity
- fixed a small bug where the update database method was failing if only one field was being updated. Turns out that the SQL syntax `SET (fields) = (values)` doesn't work if only one field -- need to use `SET field = (value)` instead, so I've tweaked the `getKeys` method to handle this case

## Testing

There's a demo action in "dev_actions", so trigger a "DEV_TEST" trigger on any application to update its name

Other than that, I've added an Action to "Basic Review Testing" to update the application name to "TemplateName -- Product Name" ON_APPLICATION_SUBMIT

Regarding whether this should be part of Core Actions -- if we wanted to generate a name from a specific question response, then it would have to be different for each template. Currently the front-end creates a generic name on application creation (Template - Serial), so we figured it made sense to leave this as-is, and then you only need to add this Action if the template wants to have something different on Submit.


plugins/action_generate_name/src/generateName.ts
```diff
@@ -0,0 +1,102 @@
...
+    additionalData = {},
+    fallbackString = 'UNRESOLVED',
+    formatExpression = '${applicationData.templateName} — ${applicationData.applicationSerial}',
+  } = parameters
+
+  try {
+    const generatedString = await generateString(
+      formatExpression,
+      {
+        applicationData,
```
nmadruga, line 28: Why not `...applicationData` so in the formatted string you don't have to specify `applicationData` on each field?

src/components/getApplicationData.ts
```diff
@@ -1,27 +1,30 @@
...
+export const getApplicationData = async (input: {
+  payload?: ActionPayload
+  applicationId?: number
+  reviewId?: number
+}): Promise<ActionApplicationData> => {
+  // Requires either application OR trigger_payload, so throw error if neither provided
+  if (!input?.payload?.trigger_payload && !input?.applicationId)
+    throw new Error('trigger_payload or applicationId required')
+
+  const { trigger_payload } = input?.payload as ActionPayload
```
nmadruga, line 17: If you would receive only an applicationId would it be a problem here?

nmadruga: Also I don't. see the case where only `reviewId` is received... Would in both cases `applicationId` and `reviewId` one payload gets passed?

src/components/getApplicationData.ts
```diff
@@ -34,10 +37,12 @@ export const getApplicationData = async (payload: ActionPayload) => {
...
+    input?.reviewId || trigger_payload?.table === 'review'
       ? {
-          reviewId: trigger_payload.record_id,
-          ...(await DBConnect.getReviewData(trigger_payload.record_id)),
+          reviewId: input?.reviewId ?? trigger_payload?.record_id,
+          ...(await DBConnect.getReviewData(
+            input?.reviewId ?? (trigger_payload?.record_id as number)
+          )),
         }
       : {}
```
nmadruga, line 47: This is a hard to read as it is, I would suggest replace with:
```
const isReview = input?.reviewId ||  trigger_payload?.table === 'review'
const review = isReview ? await DBConnect.getReviewData(
            input?.reviewId ?? (trigger_payload?.record_id as number)
          ) : undefined

const reviewData = isReview ? {
      reviewId: input?.reviewId ?? trigger_payload?.record_id,
          ...{review},
     } : {}
```

src/components/postgresConnect.ts
```diff
@@ -240,18 +240,19 @@ class PostgresDB {
...
+      template_code as "templateCode",
       stage_id as "stageId", stage_number as "stageNumber", stage,
       stage_history_id as "stageHistoryId",
       stage_history_time_created as "stageHistoryTimeCreated",
       status_history_id as "statusHistoryId", status, status_history_time_created as "statusHistoryTimeCreated",
       user_id as "userId",
       org_id as "orgId",
       outcome
-      FROM application_stage_status_all
+      FROM application_stage_status_latest
```
nmadruga, line 254: Hmmm

nmadruga (changes requested):
Looks good. I'll start testing now. I wasn't sure why the change from application_stage_status_all to application_stage_status_latest, but I think you explained here:
> updated the View application_stage_status_all (and indirectly "_latest") to include the above (as that is where getApplicationData gets its application info from)

nmadruga:
Quick tests on `dev` seems good. Yey! Great job! I noticed that the page reloads a few times probably waiting on the trigger to run, I wonder how we can deal with that (in the future) that might take a little longer to generate the application names.

Screen Shot 2021-07-15 at 4 42 14 PM

CarlosNZ:
> I wasn't sure why the change from application_stage_status_all to application_stage_status_latest

With "_all", we have to add "AND stage_is_current = true AND status_is_current = true" -- the whole point of the "_latest" was to avoid this. This was just one case we'd missed moving it over to use latest -- it does exactly the same thing: `application_stage_status_latest` is just `application_stage_status_all ` with the is_current=true clauses applied.

plugins/action_generate_name/src/generateName.ts
CarlosNZ: This is a good point, and I tried it both ways (and it's the same issue for the generateTextString, hence it's still relevant here). Since we can add any amount of additional data to this object, I thought it would be cleaner (and maybe safer) to keep "applicationData" encapsulated in its own object rather than mixing it up with all the other properties -- at the expense of more verbose expressions. (and this is the way we do it for evaluator expressions) But I'm happy to be convinced otherwise -- let's have a quick chat about it after you've looked at #454.

src/components/getApplicationData.ts
CarlosNZ: > If you would receive only an applicationId would it be a problem here?

No, it's fine. That's what this change does -- it means we can generate `applicationData` from just an application Id rather than needing a whole Action payload.

> Also I don't. see the case where only `reviewId` is received... Would in both cases `applicationId` and `reviewId` one payload gets passed?

You can't pass only `reviewId` -- it's "extra" on top of basic application data -- it would cause problems to generate only the review part of this without the relevant application info.

So it throws an error if only `reviewId` is supplied.

src/components/getApplicationData.ts
CarlosNZ: Okay, I did something slightly different which I think is much clearer. b11f088f

src/components/postgresConnect.ts
CarlosNZ: See [comment](https://github.com/openmsupply/application-manager-server/pull/436#issuecomment-880534206)

CarlosNZ:
Merging this -- even though the "generateName" action is deleted in the next PR, the small changes to modifyRecord are useful.

PR #437, opened 2021-07-08
#824F Dashboard layout, pluralisation and remove icons in filters

nmadruga:
Back-end for Front-end https://github.com/openmsupply/application-manager-web-app/pull/833

### Changes
- Added `plural` field to template table and updated all existing templates
- Removed `icon` and `iconColour` from `template_filter` table and js file


database/insertData/_common/03_template_C_userRegistration.js
```diff
@@ -13,6 +13,7 @@ exports.queries = [
         template: {
           code: "UserRegistration"
           name: "User Registration"
+          plural: "User Registations"
```
CarlosNZ, line 16: Why require a plural field, seems a lot of extra hassle to add it for every template? If we really need to pluralise template names (and I'd argue we probably don't), we can use the localisation library to handle this for each language. (In English, the rules are very simple). Personally I don't think we need to refer to template names in the plural at all.

nmadruga: I'll add the fallback for now. And we can discuss the idea for extending this to also use the localisation for pluralisation considering different languages on Monday (because as discussed in person the fallback would only work for English now: `${template_name} applications`)

CarlosNZ: Sure. Would you mind also naming the field "name_plural" rather than just "plural", it makes it clear what its referring to? Seeing "plural" on its own in the front-end with no reference to "name" at all is a bit unclear.

nmadruga: Done in front-end

nmadruga: > Sure. Would you mind also naming the field "name_plural" rather than just "plural", it makes it clear what its referring to? Seeing "plural" on its own in the front-end with no reference to "name" at all is a bit unclear.

No problem. Doing this change now

CarlosNZ: Looks like you'll need to update the field names in `insertData` templates as well.

CarlosNZ: Done: 2239d9e3

CarlosNZ (approved):

PR #441, opened 2021-07-09
#430(2) Trigger/Function on review_response

CarlosNZ:
Should have really included this as part of #438 PR -- this does exactly the same thing, but for the review_response table. 

(Plus, fixed a couple of small naming issues in the last one 😳)


nmadruga (approved):
LGTM

PR #449, opened 2021-07-12
#443 File upload updates

### LINKED ISSUE #443: Updates to file upload
1. Allow /upload to specify a uid
2. add "template_id" field to file table (and make sure it gets written)

CarlosNZ:
Fixes #443.

- the `/upload` endpoint can now take 2 more optional query parameters:
  - `uid` -- allows a uniqueId to be explicitly specified (default behaviour it is randomly generated)
  - `template_id` -- set the template_id in the file table
- add `template_id` field to "file" table
- generateDoc action (and generatePDF function) insert the `template_id`

You should be able to test it out with Postman.


nmadruga (changes requested):
All good here. It seems ok to passing one `template_id` to the upload endpoint, although while testing with postman to pass this new entry in the query params didn't quite work. The file is updated but the field `template_id` isn't added as I specified in Postman:
Screen Shot 2021-07-15 at 4 58 01 PM

CarlosNZ:
> All good here. It seems ok to passing one `template_id` to the upload endpoint, although while testing with postman to pass this new entry in the query params didn't quite work. The file is updated but the field `template_id` isn't added as I specified in Postman:

Ah -- in this case the `template_id` needs to be passed in through query params, since the "body" is not JSON -- it's "form-data". I don't know how to include body JSON as well as binary file data -- do you?

But this works:
`http://localhost:8080/upload?subFolder=carbone_templates&template_id=1`

andreievg (approved):

Issue #385, opened 2021-06-02
Create new template to apply for "Renew company license: RMMC  RMMM  RMDC  RMDM  RWSL  RRIT"


nmadruga:
https://docs.google.com/presentation/d/1ZznkU_eWpXqmWMv91-5OqZYI74Vxv-Zc/edit#slide=id.p44

- Select which license to renew
- The form can be exactly the same as company license for the same type

We need to check if we can be using the new "Outcome table" and dynamic listing done by Andrei to add the different licenses connected to a company. And how to select this license when applying for Drug Registration, for example if different licenses can allow applying for the same template. 

The simplest way to create "outcomes" for licenses for this is to add 4 new tables with just at least 2 fields "license number" and document attached related to this.

PR #451, opened 2021-07-12
#385 - New template company license renewal mmd mmc (and some help please)

nmadruga:
Attempt to add template for "Renew Company License", the search for license isn't working though. Creating this PR now just so others can help with that, I've already spent to much time trying to understand what is wrong but no lucky.

I check that the objectProperty is building the "orgId" correctly as string - when using the EvaluationExpression UI, I don't know how I can check if `currentUser.organisation.orgId` it's coming correct though.

The problem that I'm facing is on the search element in the template "03_template_license_renewal_MMC_MMD` source field where the operator "graphQL" should be getting 2 inputs, but one of then is always failing:
 

Would be easier to test once Snapshot PR is in since to test this will require:
1) Create a company
2) Get approval to have access to template "Create License"
3) Create a company license
4) Get approval on the company license
5) Restart rowLevelPolicies with: http://localhost:8080/updateRowPolicies
6) Applying to a company license renewal:
Screen Shot 2021-07-12 at 3 37 34 PM
Screen Shot 2021-07-12 at 3 30 36 PM

### DIFF
--- a/database/insertData/_common/02A_permission_names.js
+++ b/database/insertData/_common/02A_permission_names.js
@@ -53,6 +53,23 @@ exports.queries = [
       }
     }
   }`,
+  // applyOrgLicense
+  `mutation createPermissionName {
+    createPermissionName(
+      input: {
+        permissionName: {
+          name: "applyRenewLicense"
+          permissionPolicyToPermissionPolicyId: {
+            connectByName: { name: "applyUserRestricted" }
+          }
+        }
+      }
+    ) {
+      permissionName {
+        name
+      }
+    }
+  }`,
   // applyImportPermit
   `mutation createPermissionName {
     createPermissionName(
--- a/database/insertData/_common/04_outcome_displays.js
+++ b/database/insertData/_common/04_outcome_displays.js
@@ -213,6 +213,13 @@ exports.queries = [
           outcomeDisplayDetailsUsingId: {
             create: [
               {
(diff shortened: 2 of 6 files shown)


CarlosNZ:
> The problem that I'm facing is on the search element in the template "03_template_license_renewal_MMC_MMD` source field where the operator "graphQL" should be getting 2 inputs, but one of then is always failing:

The problem is that the Search plugin is not passing the "currentUser" object to its internal evaluator!
Screen Shot 2021-07-13 at 10 03 30 AM
Sorry -- I'll add this now.

CarlosNZ:
> The problem is that the Search plugin is not passing the "currentUser" object to its internal evaluator!

Okay, I've fixed this now: https://github.com/openmsupply/application-manager-web-app/pull/839

Please let me know if it works -- I haven't tested it with your template, I've only made sure that the data is there.

nmadruga:
> > The problem is that the Search plugin is not passing the "currentUser" object to its internal evaluator!
> 
> Okay, I've fixed this now: [openmsupply/application-manager-web-app#839](https://github.com/openmsupply/application-manager-web-app/pull/839)
> 
> Please let me know if it works -- I haven't tested it with your template, I've only made sure that the data is there.

Thanks! It works now : )
Screen Shot 2021-07-15 at 3 56 39 PM

Issue #447, opened 2021-07-12
Update action changeOutcome to deal with final decision (and reviews without a final decision required)


nmadruga:
Change core_action changeOutcome to run in case its isLastStage and isLastLevel trigged by ON_REVIEW_SUBMITTED with an specific changeOutcome for DrugRegistration to also changeOutcome on first stage + isLastLevel in case the review decision is NON_CONFORM

----
Non conform at first stage - means Rejection
At stage previous to final decision Reject/Approve - it means jump to the next stage
At final decision stage decision Reject/Approve then change Outcome

nmadruga:
Deal with problem raised in this PR [comment](https://github.com/openmsupply/application-manager-server/pull/452#discussion_r670008747) about when to do an `incrementStage` if the NON_CONFORM decision was selected by last_level reviewer - it should only go to next stage if its a Final Decision stage.

PR #452, opened 2021-07-12
#445 add canMakeFinalDecision to user and template

### LINKED ISSUE #445: Configure templates that have Final decision or no-final-decision needed

nmadruga:
Fixes #445 

### Briefing
Initial changes as discussed for [EPIC#34](https://github.com/openmsupply/application-manager-web-app/issues/334). So this PR doesn't actually has the full functionality yet, is just adding a new field and allowing the stage to go to next stages when either decision is CONFORM or NON_CONFORM and `is_last_level` reviewer.

❓ Can we guarantee that when the review is submitted on the last level of one given stage the `is_last_level` is set to true? Since this isn't an automatic generated column in `review_assignment` I wasn't sure.

### Changes
- Added field to template_permission as discussed in Planning
- 2 new users **finalDecision1** and **finalDecision2** already created with new permission_name to make final approval (existing so I didn't change the name, let me know if best to): `reviewReviewTestApproval`.
- Add new field `canMakeFinalDecision` on creation of `template_permission` in Review Test template. 
- Updated conditions sent to action `incrementStage` in **core_actions** to also accept decision of NON_CONFORM to move to the next stage.

### Tests
- I was only testing to CONFORM/NON_CONFORM one application of Review Test template to go from stage Assessment to Final Decision. To do this just log in with either Consolidator1 or Consolidator2 and submit one application on the correct stage with one of the decisions.

### DIFF
--- a/database/insertData/_helpers/core_mutations.js
+++ b/database/insertData/_helpers/core_mutations.js
@@ -149,15 +149,32 @@ exports.coreActions = `
         operator: "AND"
         children: [
           {
-            operator: "="
+            operator: "OR",
             children: [
               {
-                operator: "objectProperties"
+                operator: "=",
                 children: [
-                  "applicationData.reviewData.latestDecision.decision"
+                  {
+                    operator: "objectProperties",
+                    children: [
+                      "applicationData.reviewData.latestDecision.decision"
+                    ]
+                  },
+                  "CONFORM"
+                ]
+              },
+              {
+                operator: "=",
+                children: [
+                  {
+                    operator: "objectProperties",
+                    children: [
+                      "applicationData.reviewData.latestDecision.decision"
+                    ]
+                  },
+                  "NON_CONFORM"
                 ]
               }
-              "CONFORM"
             ]
           }
           {
--- a/database/insertData/dev/04_users.js
+++ b/database/insertData/dev/04_users.js
@@ -266,11 +266,6 @@ exports.queries = [
                   connectByName: { name: "reviewReviewTestAssessmentLvl2" }
                 }
               }
-              {
-                permissionNameToPermissionNameId: {
-                  connectByName: { name: "reviewReviewTestApproval" }
-                }
-              }
             ]
           }
         }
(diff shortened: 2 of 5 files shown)


database/insertData/_helpers/core_mutations.js
```diff
@@ -149,15 +149,32 @@ exports.coreActions = `
...
+              {
+                operator: "=",
+                children: [
+                  {
+                    operator: "objectProperties",
+                    children: [
+                      "applicationData.reviewData.latestDecision.decision"
+                    ]
+                  },
+                  "NON_CONFORM"
```
CarlosNZ, line 175: As we discussed, the stage needs to NOT increment if the decision is "NON_CONFORM" and the next stage is not Final Decision. This logic should probably be inside the incrementStage action, so maybe make a new issue for that, or perhaps we need to discuss it to agree on a definitive solution to handling this logic.

database/insertData/dev/04_users.js
```diff
@@ -488,6 +483,57 @@ exports.queries = [
...
   }`,
+  // Final decision users
+  `mutation {
+    createUser(
+      input: {
+        user: {
+          username: "finalDecision1"
+          firstName: "Final"
+          lastName: "Decision Maker 1"
+          passwordHash: "$2a$10$5R5ruFOLgrjOox5oH0I67.Rez7qGCEwf2a60Pe2TpfmIN99Dr0uW."
```
CarlosNZ, line 494: Create new hashes for these guys, if you can be bothered: https://www.browserling.com/tools/bcrypt 

Not really important for "dev" users, I know, so feel free to ignore.

CarlosNZ (approved):
Cool, looks good. Just need to figure out the logic for when to increment the stage, and where it happens, so please make a new issue if you merge this one as-is.

database/insertData/_helpers/core_mutations.js
nmadruga: I think we can deal with this on issue #447

database/insertData/dev/04_users.js
nmadruga: Thanks Done

PR #453, opened 2021-07-12
#446 generate review assignments for final decision

### LINKED ISSUE #446: Update action generateReviewAssignment to create reviewAssignments already assigned to final decision users
if user has `canMakeDecision` in template_permission table and `canSelfAssign` = `false` to allow straight assignment (so reviewAssignment starts as ASSIGNED) when a final decision user clicks to start this reviewAssignment.
In the front-end should create the review (and this can be created for any users with reviewAssignment not locked)

nmadruga:
Fixes #446

### Linked PR
- Based on #452 - Thanks @CarlosNZ 

### Briefing
- The **Final Decision** stage has only the difference of having a new field `canMakeFinalDecision` set to **true** on `template_permission` on the template configuration. Any user associated with this permission will be allowed to start the final decision (WIP on issue [Front-end#835](https://github.com/openmsupply/application-manager-web-app/issues/835)).
- The `template_permission` for final decision doesn't set `canSelfAssign` field (default **false**) because the `reviewAssignment` will be created with status ASSIGNED when this stage is a final decision. 
  - Not sure how more than one level would work in this scenario though (do we need to consider that `canMakeFinalDecision` and `isLastLevel` when making the change on outcome)
- What should happen next: 
  - Reviewer of Final decision **starts** a review: 
    - No reviewQuestionAssignment is created
    - Once the reviewer starts a **Review** is created and display the Final decision page (should use same as existing)
    - All sections show collapsed, but can be opened to see the original_review_decision and last_level_review_decision
    - The last level decision + original_review decision overall should be displayed at the bottom of sections
    - The Final decision (editable) box with options to: Approve or Reject
  - Reviewer of Final decision **submits** a review:
    - Other reviewAssignment/Review (for other Reviewer of Final decision) should get locked
    - The application outcome is changed
    - Email is sent to applicant with final decision

### Changes
- Added field `can_make_final_decision` on `template_permission` table
  - Used existing `permission_name`: "reviewReviewTestApproval" to link new users with final decision making
  - Added stage for **Final Decision** on[ Test Review template](https://github.com/openmsupply/application-manager-server/pull/453/files?file-filters%5B%5D=.js&file-filters%5B%5D=.ts#diff-a1585c72430f4666e589182523a07abdab95c62a36796e864bf9d3fddf7e6b97) with new field `canMakeFinalDecision`
- Update action to `generateReviewAssignment` to consider that when next level reviewer `canMakeFinalDecision` a `reviewAssignment` is to be create with status as **Assigned** so this reviewer doesn't need to self-assign.
- Updated `core_mutation` action `incrementStage` to also consider when the last_level reviewer (of previous stage) sends a decision of NON_CONFORM. The Final Decision (last stage) needs to consider either previous decision for CONFORM or NON_CONFORM.  

### Tests
- Fixed both tests `generateReviewAssignment` and `updateReviewStatus` by adding ids to Reviews and ReviewAssignments and edited one application (ABC123) to be on **Final Decision** stage (and new test to check if reviewAssignment generation works correctly). I have also added a new application (ABC345) to have no reviews started as original ABC123 was.


CarlosNZ:
>     * Not sure how more than one level would work in this scenario though (do we need to consider that `canMakeFinalDecision` and `isLastLevel` when making the change on outcome)

I was thinking that "Final Decision" would imply that it's the last level. But I guess there could be a situation where there is an initial manager who makes a "recommended" final decision, and a chief who approves it.

CarlosNZ:
> Reviewer of Final decision **submits** a review:
> 
>     * Other reviewAssignment/Review (for other Reviewer of Final decision) should get locked
> 
>     * The application outcome is changed
> 
>     * Email is sent to applicant with final decision

So in regards to the last comment, those last two items would only happen if where at last level, right? There could be a more senior decision maker.

nmadruga:
> > Reviewer of Final decision **submits** a review:
> > ```
> > * Other reviewAssignment/Review (for other Reviewer of Final decision) should get locked
> > 
> > * The application outcome is changed
> > 
> > * Email is sent to applicant with final decision
> > ```
> 
> So in regards to the last comment, those last two items would only happen if where at last level, right? There could be a more senior decision maker.

I think all 3 of them. Right? It's only when the final decision is made that you would stop other reviewer (in the same level) to submit other decisions for the same application

plugins/action_generate_review_assignment_records/src/generateReviewAssignments.ts
```diff
@@ -138,12 +138,13 @@ const generateNextReviewAssignments = async ({
...
   // and merge their section code restrictions
   nextLevelReviewers.forEach((reviewer: Reviewer) => {
-    const { userId, orgId, allowedSections, canSelfAssign } = reviewer
+    const { userId, orgId, allowedSections, canSelfAssign, canMakeFinalDecision } = reviewer
 
-    const status =
-      canSelfAssign || nextReviewLevel > 1
-        ? ReviewAssignmentStatus.AvailableForSelfAssignment
-        : ReviewAssignmentStatus.Available
+    const status = canMakeFinalDecision
```
andreievg, line 143: I find nested inline condition statements really hard to read. not a required change, since it's a stylistic preference, but would prefer something like:
```JS
getAssignmentStatus = () => {
  if(canMakeFinalDecision) return ReviewAssignmentStatus.Assigned;
  if(canSelfAssign || nextReviewLevel > 1) return ReviewAssignmentStatus.AvailableForSelfAssignment
  return ReviewAssignmentStatus.Available
}
```
above would also help make it extendable if more conditions needs to be added

plugins/action_generate_review_assignment_records/src/generateReviewAssignments.ts
```diff
@@ -192,11 +193,19 @@ const generateNextReviewAssignments = async ({
 
   console.log('ReviewAssignmentAssignerJoinIds', reviewAssignmentAssignerJoinIds)
 
+  // Remove timeStageCreated from log - to help with tests
+  const reviewAssignmentsLog: ReviewAssignment[] = []
+
+  Object.values(reviewAssignments).forEach((reviewAssignment) => {
+    delete reviewAssignment.timeStageCreated
```
andreievg, line 200: hmm, maybe we can do this in tests ?

before expect.toEqual(), can clean up reviewAssignments (with generic method), i.e.:

btw we have lodash now in repo, might be good to use it (https://lodash.com/docs/4.17.15#mapValues, https://lodash.com/docs/4.17.15#omit)

```
removeTimeStageCreated(reviewAssignments: any) => mapValues(reviewAssignments, value => omit(value, ['timeStageCreated']))
...
clearResult(result: any) => ({...result, output: {...output, reviewAssignments: removeTimeStageCreated(result.output.reviewAssignments) }})
...
expect(cleanResult(result)).toEqual({
```

andreievg (approved):
looks good, thanks Nicole, it's a lot of work !

one stylistic comment, and one suggestion, not stopping approval though

plugins/action_generate_review_assignment_records/src/generateReviewAssignments.ts
nmadruga: Yes, good point. Done: e135430

plugins/action_generate_review_assignment_records/src/generateReviewAssignments.ts
nmadruga: Nice. Thanks for the advise. Should be a better use than changing the result on the action.

nmadruga: Done: 9e9c78c

PR #454, opened 2021-07-14
#426 Generate Serials

### LINKED ISSUE #426: New action-plugin to run on Application Start to generate a number for application serial (Default)

CarlosNZ:
Fixes #426 

Edit: can now be tested with front-end. Needs [branch from PR #844](https://github.com/openmsupply/application-manager-web-app/pull/844)

This uses my package "[custom_string_patterns](https://www.npmjs.com/package/custom_string_patterns)", which allows us to create simple regex-based patterns for generating strings/serials that can also include persistent counters and substitutions from applicationData (and anywhere if we need it to).

Note that this replaces the need for a specific "generateName" Action, so I've deleted that and incorporated all its functionality into here.

Easiest way to see it in use is to use the `/run-action` endpoint.
Here is a Postman collection with a query already configured: [Application Manager.postman_collection.json.zip](https://github.com/openmsupply/application-manager-server/files/6813235/Application.Manager.postman_collection.json.zip)

This JSON query:
```js
{
    "actionCode": "generateTextString",
    "applicationId": 4001,
    "parameters": {
        "pattern": "<?tempCode>-[A-Z]{3}-<+dddd>",
        "counterName": "demo",
        "counterInit": 100,
        "customFields": {
            "tempCode": "applicationData.templateCode"
        }
    }
}
```
will generate strings such as:
ReviewTest-VZE-0100
ReviewTest-BGC-0101
ReviewTest-UOM-0102

From the `pattern` field, that's a value from applicationData (templateCode) followed by 3 random A-Z chars, and a four-digit incrementing counter (seperated by "-").

**Edit**: I've updated it so you can just put the long property names directly in the pattern rather than use customFields -- it's probably a bit easier to follow at the expense of more verbose patterns, so we can use either format.

i.e. this will give the same results as the above query:

```js
{
    "actionCode": "generateTextString",
    "applicationId": 4001,
    "parameters": {
        "pattern": "<applicationData.templateCode>-[A-Z]{3}-<+dddd>",
        "counterName": "demo",
        "counterInit": 100,
    }
}
```

We can also use it to generate an application name and update the application table record:

```js
{
    "actionCode": "generateTextString",
    "applicationId": 4001,
    "parameters": {
        "pattern": "<?templateName>—<?productName>",
        "customFields": {
            "templateName": "applicationData.templateName",
            "productName": "applicationData.responses.Q20.text"
        },
        "updateRecord": true
    }
}
```
The "updateRecord" parameter defaults to using "application" table, "name" field, but any record can be specified using additional parameters.

### Summary of Additional changes (other than the Action)

1. Added a "counter" table to schema, and a function in PostGresConnect to fetch the current value and increment the counter. There can be any number of different counters, just need to specify a "name" for each.
2. Added this Action to "coreActions" for "ON_APPLICATION_CREATE" -- it sets a serial and a generic name for the application. (We still need to update front-end to work with this -- see [this issue](https://github.com/openmsupply/application-manager-web-app/issues/840))
3. While testing the above core actions, I ran into a problem -- the actions failed, as they weren't receiving "applicationData". Turns out this was because the initial database query in "getApplicationData" looks up the "application_stage_status_latest" table (view), and there were no entries for the application if no stage or status had been set yet. So I've refactored the View queries a little so that the application table info has a LEFT JOIN with the stage/status stuff, so that we can still see applications that have NULL values for stage and status. Also had to make a small tweak to the "incrementStage" action to handle this change.
4. Updated the "generateName" action in Basic Review Testing to use "generateTextString"

### TO DO

- [x]  Documentation (Done: 846466cb)
- [x]  Decide best way to store each template's serial pattern. I think the initial serial generation _needs_ to be part of core_actions, since it's a fundamental requirement for front-end functioning. However, if we want each template type to have its own distinct serial number pattern, we have to store it somewhere specific to each template. I vote for a new "serial_pattern" field on "template", which would be part of applicationData so it could be inserted by this core action. #459 
- [ ] ~~Implement re-naming of application ON_APPLICATION_SUBMIT for all templates~~ To do in issue #457


database/insertData/_helpers/core_mutations.js
```diff
@@ -5,10 +5,43 @@ GraphQL Fragment - CORE ACTIONS
...
     # change status to DRAFT
+    # generate serial
+    # generate initial name
     {
-        actionCode: "incrementStage"
+        actionCode: "generateTextString"
         sequence: 1
         trigger: ON_APPLICATION_CREATE
+        parameterQueries: {
+          pattern: "S-[A-Z]{3}-<+dddd>"
```
nmadruga, line 15: Maybe good to have a comment here on the top explaining what this code means? Just briefly

plugins/action_generate_text_string/package.json
```diff
@@ -1,7 +1,7 @@
 {
-  "name": "generate_name",
+  "name": "generate_text_string",
   "version": "0.1",
-  "description": "Creates a text string and (optionally) applies it to a record, e.g. an application name",
+  "description": "Generates serial numbers or arbitrary text strings using a specified pattern",
```
nmadruga, line 4: Should it keep the `e.g. an application name` as this is the default location? Or it isn't ?

yarn.lock
```diff
@@ -1125,6 +1125,26 @@
...
+"@tsconfig/node14@^1.0.0":
+  version "1.0.1"
+  resolved "https://registry.yarnpkg.com/@tsconfig/node14/-/node14-1.0.1.tgz#95f2d167ffb9b8d2068b0b235302fafd4df711f2"
+  integrity sha512-509r2+yARFfHHE7T6Puu2jjkoycftovhXRqW328PDXTVGKihlb1P8Z9mMZH04ebyajfRY7dedfGynlrFHJUQCg==
+
+"@tsconfig/node16@^1.0.1":
+  version "1.0.1"
+  resolved "https://registry.yarnpkg.com/@tsconfig/node16/-/node16-1.0.1.tgz#a6ca6a9a0ff366af433f42f5f0e124794ff6b8f1"
+  integrity sha512-FTgBI767POY/lKNDNbIzgAX6miIDBs6NTCbdlDb8TrWovHsSvaVIZDlTqym29C6UqhzwcJx4CYr+AlrMywA0cA==
+
```
nmadruga, line 1147: Do you know why those are getting added?

plugins/action_increment_stage/src/incrementStage.ts
```diff
@@ -24,7 +24,7 @@ async function incrementStage({
         : await DBConnect.getCurrentStageStatusHistory(applicationId)
```
nmadruga, line 24: Should this also cover the case when is comming from `applicationData`?

nmadruga (changes requested):
Thanks for doing that, Carl. Looks very generic and really helpful to have the counter already included, that's great. I just added a few initial comments and plan to have more tests on this tomorrow.

database/insertData/_helpers/core_mutations.js
CarlosNZ: Sure, but I'll wait until we know what the actual pattern should be. I need to do overall documentation though, so will do now.

plugins/action_generate_text_string/package.json
CarlosNZ: Good call: b149be66

yarn.lock
CarlosNZ: Hmm, it looks like `ts-node` has been updated (in package.json). Not sure why exactly, it seems like it automatically bumps the version if one of the dependencies specifies a higher version (as long as there's no maximum version specified in the main package.json)

plugins/action_increment_stage/src/incrementStage.ts
CarlosNZ: Not sure what you mean. The logic for getting the applicationId is handled at the start with: 
`const applicationId = parameters?.applicationId ?? applicationData?.applicationId`

By the time we get to here it shouldn't make any difference where it came from -- parameters or applicationData.

Is that what you meant?

database/insertData/_helpers/core_mutations.js
nmadruga: Thanks. And regarding this TODO:
> Decide best way to store each template's serial pattern. I think the initial serial generation needs to be part of core_actions, since it's a fundamental requirement for front-end functioning. However, if we want each template type to have its own distinct serial number pattern, we have to store it somewhere specific to each template. I vote for a new "serial_pattern" field on "template", which would be part of applicationData so it could be inserted by this core action.

I think that's a good way to store the pattern in the template, that's nice because when we duplicate it also will have the pattern persisted in the previous template version, which can be updated if the user would like to. Are you willing to do this change as part of this PR?

nmadruga:
> * Implement re-naming of application ON_APPLICATION_SUBMIT for all templates

I think this can be done in a separated PR

database/insertData/_helpers/core_mutations.js
CarlosNZ: > I think that's a good way to store the pattern in the template, that's nice because when we duplicate it also will have the pattern persisted in the previous template version, which can be updated if the user would like to. Are you willing to do this change as part of this PR?

Yes, for sure. I want to, but Andrei had some reservations about it, so we should probably have a quick discussion about it as a group first.

yarn.lock
nmadruga: I think we use a minimum node-version on our project, so in this case it seems we would need to update that.
Check out this config in file:
https://github.com/openmsupply/application-manager-server/blob/c406cdc95346ea43c7965dd9aecd26c31340ca98/package.json#L10-L12

documentation/List-of-Action-plugins.md
```diff
@@ -142,31 +143,86 @@ This will look for a user record with `username = "js"` and update it with the _
...
+| `matchValue` (default `applicationId`)   |                   |
+| `additionalData`                         |                   |
+
+This action uses the [custom_string_patterns](https://www.npmjs.com/package/custom_string_patterns), which allows the generation of random strings using Regular Expressions, but with the added functionality of also including:
+
+- incrementing, persistent counters
+- custom replacement functions
+- object data replacement.
+
+Please see the custom_string_patterns documentation for full detail, but basically the `pattern` field specified the string to be generated using:
```
nmadruga, line 173: Where is this `custom_string_patterns` documentation? Maybe a link here?

documentation/List-of-Action-plugins.md
```diff
@@ -142,31 +143,86 @@ This will look for a user record with `username = "js"` and update it with the _
...
+  customFields: {
+    tempCode: "applicationData.templateCode"
+}
+```
+
+The pattern `<?tempCode>-[A-Z]{3}-<+dddd>` consists of:
+
+- `<?tempCode>` - this will be replaced with whatever data field is specified in `customFields` -- in this case `applicationData.templateCode`. Note that this is equivalent to just putting `<applicationData.templateCode>` directly in the pattern.
+- `[A-Z]{3}` a regular expression specifying 3 random upper-case characters in the A to Z range
+- `<+dddd>` will be replaced with a four-digit number, which will be taken from the current value of the counter specifed in `counterName`, in this case "demo".
```
nmadruga, line 196: It's so nice that you have explained each par of the pattern used in the example, very good for future reference. I just am a little lost with why do you have that top part and then after the example you add explanation about `<?tempCode>` (one example of <?field>) and `<+ddd>` again. Isn't the case to combine both these blocks?

CarlosNZ:
> > * Implement re-naming of application ON_APPLICATION_SUBMIT for all templates
> 
> I think this can be done in a separated PR

Sure, made an issue: #457

documentation/List-of-Action-plugins.md
```diff
@@ -142,31 +143,86 @@ This will look for a user record with `username = "js"` and update it with the _
...
+      templateName: "applicationData.templateName",
+      productName: "applicationData.responses.Q20.text"
+  }
+```
+
+which is the same as just:
+
+```
+  pattern: "<applicationData.templateName>—<applicationData.responses.Q20.text>"
+```
```
nmadruga, line 225: I think its confusing that this example is showing how to use this action with its default option, when this topic is about using this to be "Updating a record" that isn't the applicationName. Can you move this example to the top and add another one here on how to use this to update a different record?

documentation/Triggers-and-Actions.md
```diff
@@ -150,3 +151,24 @@ Note that `applicationData` is regenerated between each Action in a sequence, so
...
+Endpoint: `/run-action`
+Type: `POST`
+Parameters (via body JSON):
+
+```js
+{
+    "actionCode": "generateTextString", // Code of the action
+    "applicationId": 4001, // for generating applicationData, so not required for all actions
+    "parameters": {
+        <...input parameters for the action itself>
```
nmadruga, line 168: probably a good idea to add the parameters here as well. So the example is clearer

plugins/action_generate_text_string/package.json
```diff
@@ -1,7 +1,7 @@
 {
-  "name": "generate_name",
+  "name": "generate_text_string",
   "version": "0.1",
-  "description": "Creates a text string and (optionally) applies it to a record, e.g. an application name",
+  "description": "enerates serial numbers or arbitrary text strings (e.g. for application name) using a specified pattern, and updates a database record if requested",
```
nmadruga, line 4:
```suggestion
  "description": "Generates serial numbers or arbitrary text strings (e.g. for application name) using a specified pattern, and updates a database record if requested",
```

nmadruga (commented):
A few comments on the documentation - testing now. .

I'll just see the application EditUser name generation for now. But I'm missing more details (examples on how to use this to update other fields) it seems that this PR is touching different areas, and I think we need to add at list some ways to manually test that. What would happen if trying to update a table without permission for example?

plugins/action_generate_text_string/package.json
CarlosNZ: Oops 😳

nmadruga:
When trying to run (using `pg_permissions` with this back-end branch and `master` front-end) when I click to register user I'm getting a "Page not found":
Screen Shot 2021-07-16 at 10 48 29 AM

nmadruga:
Hmmm ok, seems like master was missing the `yarn generate` to get latest types for counters : ) I'll try again now

nmadruga:
Nah, same problem : (

Screen Shot 2021-07-16 at 10 51 43 AM

Screen Shot 2021-07-16 at 10 51 51 AM

CarlosNZ:
> When trying to run (using `pg_permissions` with this back-end branch and `master` front-end) when I click to register user I'm getting a "Page not found":
> Screen Shot 2021-07-16 at 10 48 29 AM

That's right -- it needs the front-end PR I made yesterday: https://github.com/openmsupply/application-manager-web-app/pull/844

(Sorry, when I wrote this, I didn't have any front-end integration yet -- I'll update the original post)

documentation/List-of-Action-plugins.md
CarlosNZ: The link is just a few lines further up.

I might add it in here again, but I'll re-visit it after addressing your other comments below, since I'll probably restructure the documentation somewhat.

CarlosNZ:
> What would happen if trying to update a table without permission for example?

If it's run as a standard triggered Action, it would be fine, since back-end has access to everything. 

If you run it from the `/run-action` endpoint, it'd also currently allow anything, but we need to address this at some point. Have talked about it briefly with Andrei, but I've just made a placeholder issue for it so we don't forget about it: #458

nmadruga:
> > When trying to run (using `pg_permissions` with this back-end branch and `master` front-end) when I click to register user I'm getting a "Page not found":
> 
> That's right -- it needs the front-end PR I made yesterday: [openmsupply/application-manager-web-app#844](https://github.com/openmsupply/application-manager-web-app/pull/844)
> 
> (Sorry, when I wrote this, I didn't have any front-end integration yet -- I'll update the original post)

Yes. That makes sense now. Have been testing it now and seems all good. Besides the time to load every application now - as we discussed in person (which might need a solution in the early future), but nothing to do now. We need to keep our focus on merge what we can and finish this Sprint issues.

nmadruga:
> If you run it from the `/run-action` endpoint, it'd also currently allow anything, but we need to address this at some point. Have talked about it briefly with Andrei, but I've just made a placeholder issue for it so we don't forget about it: #458

Yes. That's very important indeed. If we just expose this end-point as it is, then we allow anyone changing things in the system which isn't a good idea.

documentation/List-of-Action-plugins.md
```diff
@@ -164,66 +164,119 @@ Generates serial numbers or arbitrary text strings (e.g. for application name) u
...
+**For example:**
 
-The pattern `<?tempCode>-[A-Z]{3}-<+dddd>` consists of:
+The pattern: `<?templateCode>-[A-Z]{3}-<+dddd>` consists of:
 
-- `<?tempCode>` - this will be replaced with whatever data field is specified in `customFields` -- in this case `applicationData.templateCode`. Note that this is equivalent to just putting `<applicationData.templateCode>` directly in the pattern.
+- `<?templateCode>` - will be replaced by the template code
 - `[A-Z]{3}` a regular expression specifying 3 random upper-case characters in the A to Z range
-- `<+dddd>` will be replaced with a four-digit number, which will be taken from the current value of the counter specifed in `counterName`, in this case "demo".
+- `<+dddd>` will be replaced with a four-digit number from a counter.
```
nmadruga, line 185: > (will be padded with leading 0's to fill the required length)

Nice one to have I though that was very informative

nmadruga:
Thanks for final updates to the documentation. I think it's looking way better and easier to understand. Probably we need to add more exampled on dealing with other tables in the future if that's get used more broadly. But so far I'm happy with this, as long as we also get some prevention on the end-point using Authentication. I think that can be dealt separated in the previous branch anyway.

nmadruga (approved):

CarlosNZ:
Hi @nmadruga -- I've re-structured the documentation substantially -- hopefully it addresses most of your issues. Thanks for the feedback, it was really hard to write this doc, as every new explanation referred to several other things that also required explanation, so it quickly became quite messy.

I'm fairly happy with how it's structured now, hopefully it makes sense to. 2d20f740

Note: I just tried using Github Desktop's new "merge commits" feature to squash the last two changes together into this commit. It gave me a warning about other users needing to update their local branch, so my apologies if this is messy for you.

database/insertData/_helpers/core_mutations.js
CarlosNZ: I've made a PR based off this one that implements this in a fairly straightforward way. See what you think: #459

PR #459, opened 2021-07-16
426(part 2) "serial_pattern" field for template

CarlosNZ:
Proposed way of defining different serial formats (and counters) for each template, as briefly discussed [here](https://github.com/openmsupply/application-manager-server/pull/454#discussion_r670839268) and with @andreievg IRL earlier in the week.

### Summary

- add `serial_pattern` field to "template" table, in jsonb format, with a default specified (GEN-XXX-0001)
- get "templateSerialPattern" as part of "applicationData" (via `application_stage_status_all`)
- actions in core_mutations for ON_APPLICATION_CREATE updated to pass these values through to `generateTextString` for creating the serial

Let me know what you think, or if you have alternative suggestions. Bear in mind that if a particular template requires a more complex serial than can be defined in "serial_pattern", we can add another action to that template that immediately overwrites the one generated in core actions.

### DIFF
--- a/database/buildSchema/07B_template.sql
+++ b/database/buildSchema/07B_template.sql
@@ -9,6 +9,10 @@ CREATE TABLE public.template (
     id serial PRIMARY KEY,
     name varchar,
     code varchar NOT NULL,
+    serial_pattern jsonb DEFAULT '{
+        "pattern": "GEN-[A-Z]{3}-<+dddd>",
+        "counterName": "general"
+    }' ::jsonb,
     is_linear boolean DEFAULT TRUE,
     start_message jsonb,
     status public.template_status,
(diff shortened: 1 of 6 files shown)


database/buildSchema/07B_template.sql
```diff
@@ -9,6 +9,10 @@ CREATE TABLE public.template (
     id serial PRIMARY KEY,
     name varchar,
     code varchar NOT NULL,
+    serial_pattern jsonb DEFAULT '{
```
andreievg, line 12: sorry I missed the discussion mentioned in PR comments, actions are also duplicated when template is duplicated, which seemed to have been the main reason for this ? Any other cases where this can be useful ? Can chat tomorrow

CarlosNZ:
Summary of what I plan to do after discussion with @andreievg:

- don't need the "serial_pattern" field
- a core action which generates a very basic serial, but uses the templateCode as the name of its counter
- when new templates are created/duplicated, core actions are copied, so it's easy for template creators to change the generate serial action if they want to make a more specific serial generating pattern

CarlosNZ:
Closing -- replaced with #468

Issue #439, opened 2021-07-08
Generate a Carbone document template for an application with all responses by applicant and review_responses


CarlosNZ:

CarlosNZ:
Hey, just a thought about applicationData. I'm working on the Carbone template for Feature Showcase and we need some way to make all the "template" data available to the Action (as in all the question elements and their parameters, we currently only have application responses, not the associated questions). Possible approaches, with their various pros/cons that I can think of:

1. Just add more stuff to `getApplicationData`. This would be the simplest, but I'm wary of just shoving more stuff in here considering we run this (at least) once per action, so adding more database queries that give the same result every time seems quite clunky.
2. Add a "getTemplateData" function similar to `getApplicationData` which effectively builds the structure into an object for back-end consumption. And we run it once per _trigger_, not over and over for each action, since the response won't change for a given template (and would presumably return cached data by postgres?). This is passed as input to the actions along with "applicationData" and "outputCumulative", etc. 
3. Just add an additional query for getting template info to the `generateDoc` action. This seems like a sensible solution, but it's implying that this info won't be useful to other actions, which may not be the case. It would get silly if multiple actions all had their own "getTemplateData" script. Perhaps we just make a general "getTemplateData" function, and multiple actions could call it when needed?
4. Write database evaluator queries that fetch specific data as required as part of the generateDoc action. This is obviously a very tedious approach and would be quite inefficient since there'd be a separate query for each question. Not really a practical approach, just listing here for comparative purposes.
5. Any other options?

Personally, I'd be inclined to Option 2. Actually, we can avoid making changes to the input fields on actions by just bundling it into `applicationData` but without re-querying. When the trigger runs, it can run a `getTemplateData` function, which it passes into `getApplicationData`, which bundles it into the return object. We can also have a check in `applicationData` that re-queries `getTemplateData` only if it's not passed in.

PR #462, opened 2021-07-21
#439 (Part 1) getTemplateData

CarlosNZ:
This is just the preliminary stuff that needs to happen before I can really do #439 properly.

In Summary, this adds "templateData" to applicationData, so we can have access to all questions, etc in Actions. Just wanted to PR this first to check that we were all happy with this approach.

See [this comment](https://github.com/openmsupply/application-manager-server/issues/439#issuecomment-883848014) for context.

### In Summary:

- a "getTemplateData" query runs at the top of each Trigger (in `processTrigger`)
- this is passed to `getApplicationData`, and embedded into the resulting `applicationData` object (so it doesn't need to be re-queried between multiple actions in a sequence)
- if `templateData` is NOT passed to `getApplicationData` (e.g. if running from endpoint, say), it will be re-queried.

The query is done with GraphQL (using GraphQLConnect) -- are we happy with the shape of the resulting object?

### DIFF
--- a/src/components/graphQLConnect.ts
+++ b/src/components/graphQLConnect.ts
@@ -1,6 +1,7 @@
 import fetch from 'node-fetch'
 import config from '../config'
 import { getAdminJWT } from './permissions/loginHelpers'
+import { Template } from '../generated/graphql'
 
 const endpoint = config.graphQLendpoint
 
@@ -71,7 +72,7 @@ class GraphQLdb {
     // Not implemented yet -- needs more data in DB
   }
 
-  public getTemplateData = async (templateId: number) => {
+  public getTemplateData = async (templateId: number): Promise<Template> => {
     const data = await this.gqlQuery(
       `
         query getTemplateData($templateId: Int!) {
--- a/src/components/graphQLConnect.ts
+++ b/src/components/graphQLConnect.ts
@@ -70,6 +70,92 @@ class GraphQLdb {
     }
     // Not implemented yet -- needs more data in DB
   }
+
+  public getTemplateData = async (templateId: number) => {
+    const data = await this.gqlQuery(
+      `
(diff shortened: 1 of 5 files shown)


src/components/graphQLConnect.ts
```diff
@@ -70,6 +71,92 @@ class GraphQLdb {
     }
     // Not implemented yet -- needs more data in DB
   }
+
+  public getTemplateData = async (templateId: number): Promise<Template> => {
```
andreievg, line 75: Hmm, i am actually not sure about this, I think we've decided that dynamically displaying carbone template for the whole application is too difficult, and wanted to just provide a template that shows how to render each individual type of element (and their variations), and for carbone templates to be build for each type of application ?

I want to have a chat about this as I have an idea of how extra data can be provided to the generate carbone action

CarlosNZ: > I think we've decided that dynamically displaying carbone template for the whole application is too difficult, and wanted to just provide a template that shows how to render each individual type of element (and their variations)

Yes, but will still need access to the question data from the template one way or another. So we can display something like (to user in doc):

> **Q1:** Company Name
> _Pharmac Importers_

Without the question data (labels, etc.), the responses alone won't be much use.

> I want to have a chat about this as I have an idea of how extra data can be provided to the generate carbone action

I'm happy for getting this extra data to be specific to the "generateDoc" action, I just thought it might be useful elsewhere too. We can have a chat in the morn.

CarlosNZ:
See [this issue](https://github.com/openmsupply/application-manager-web-app/issues/852) for a potential solution that will remove the need for this PR. 

Will close this once I submit a new PR for the Carbone template.

CarlosNZ:
Closing -- #469 is adequate for now.

PR #468, opened 2021-07-23
#426(2) core action for serial generation

CarlosNZ:
Replaces PR #459, as discussed [here](https://github.com/openmsupply/application-manager-server/pull/459#issuecomment-885342227)

- a core action which generates a very basic serial, but uses the templateCode as the name of its counter
   - when new templates are created/duplicated, core actions are copied, so it's easy for template creators to change the generate serial action (in Template Builder) if they want to make a more specific serial generating pattern
- remove default "table" and "field" values from "generateTextStrings", as per suggestion (and add them to templates accordingly)

### DIFF
--- a/database/insertData/_common/03_template_G_userEdit.js
+++ b/database/insertData/_common/03_template_G_userEdit.js
@@ -310,6 +310,7 @@ exports.queries = [
                       # TBD
                     }
                     updateRecord: true
+                    table: "application"
                     fieldName: "serial"
                   }
               }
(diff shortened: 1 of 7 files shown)


nmadruga (approved):
Just removing the default table from the action - what would happen if it isn't passed?

Seems good to me - working on merging to #498 now

CarlosNZ:
> Just removing the default table from the action - what would happen if it isn't passed?

It will fail. Andrei preferred it this way. 🤷‍♂️

Issue #474, opened 2021-07-28
Summarise changes to start using snapshots + template builder to keep our dev insertData and laos insertData as snapshots considering tests which would require assignment and reviews already created


nmadruga:
### Tests and initialising database
- Create 2 snapshots: one for `dev` another for `laos`
- Perhaps one without any templates - just schema loaded. What are the benefits?
- When loading the `dev` snapshot - reviews and reviewAssignments for ReviewTest would be created - so changes to this template would be a problem since fields/table might not match correctly anymore

### Storing: version control vs sharing folder
- Where do we store our snapshots?
- If keeping on a repository, we can use diff to compare changes. Would that be helpful?

### Exporting snapshots
- Change/add options on the script for snapshot to only export latest or just the **active** version of each template - so we remove all different versions created previously for the same template

### Create new templates
- Carl and I were trying to figure out the best strategy to duplicate same actions that are likely to be used in every template. We have core actions - and one possibility would be to just copy all core_actions to a new template (not duplication of template). But actions for notifications - which would be very repetitive between different templates -  shouldn't be core_actions (since the definition of this is required for the correct flow of the App). The idea was to have a simple **base template** that would be copied in case the Admin user would like to start a new template - from scratch. The base template would have the most basic setup, but considering all notifications and other specific actions to easy the creation of a new template

### Create new core_actions
- Ideally adding new core actions can be done from any template, so having a flag to say this is a core action and when saving the template a copy of this action would be sent to where other core_actions are kept (for now `core_mutations.js` file)
- Any template could import all new **core_actions** that aren't already in this template - to get all latests additions
- Not a good idea to overwrite/or manage diffs between existing core actions and current changes on specific template.  
- Having a **code** per action/template would be also helpful here so we can check what core_actions are missing by looking if all actions by codes from core_actions are included in the current template

nmadruga:
My initial plan of issues to create and get the setup started:
1. Create clean snapshot without templates - should have only base template?
2. Work out how to change script to generate snapshot only with latest version of each template
3. Create & implement button to create new template (copying base template)
4. Start second snapshot only with Laos config using base template and copy other fields
5. Start third snapshot only with dev config. It should also have already all 6 existing applications/reviews created
6. Make sure tests are running well on dev snapshot
7. Publish 3 snapshots with only one version of each template
8. Write documentation on how to use new environment

CarlosNZ:
Sounds good. 

[...]

PR #479, opened 2021-07-29
#690F fix review assignments duplicated for level 1 after LOQ is submitted

### LINKED ISSUE #690: New company registration - Email templates

nmadruga:
Fixes  front-end https://github.com/openmsupply/application-manager-web-app/issues/690

Finally! It took me so long to fix the tests... `generateReviewAssignments` but now all the **existing** reviewAssignments aren't returned anymore, which is what was causing the problem.

### Changes
- Update Application 4 (serial: 45678) to not be already in Stage 2 Lvl 2 - so this case can be tested in `generateReviewAssignments` to generate reviewAssignments for next level in same stage
- Fix tests in `generateReviewAssignments` and `updateReviewStatuses` to be aligned with current application/assignments
- Add test specific to show no new reviewAssignments are created if there are already some for the same: `reviewer`, `stage`, `level` and `application`.

### Testing
- Use application ABC456 logged in with user **Reviewer1**
- Start review and DECLINE one response and submit the LOQ to applicant
- Log in with user **Valerio** and fix the response then re-submit
- Log in again as **Reviewer1** to Re-review and APPROVE all then CONFORM and submit
- Now the application is on the next stage. No other reviewAssignments have been created for other reviewer in stage 1

Check in the master to see the bug happening - you can check the logs to see new reviewAssignments created after re-submitting the application with LOQ back to the reviewer.

### DIFF
--- a/plugins/action_generate_review_assignment_records/src/generateReviewAssignments.test.ts
+++ b/plugins/action_generate_review_assignment_records/src/generateReviewAssignments.test.ts
@@ -31,12 +31,13 @@ test('Test: Submit Application ID#4001 - Stage 1 (Last level) - reviewAssignment
     expect(clearResult(result)).toEqual({
       status: ActionQueueStatus.Success,
       error_log: '',
-      output: {},
+      output: {
+        reviewAssignments: [],
+      },
     })
   })
 })
 
-// Simulate application submissions:
 describe('Submit new application on first stage to generate reviewAssignments on Stage 1', () => {
   // Setup database
   beforeAll(async (done) => {
@@ -252,8 +253,6 @@ describe('Re-submit new application on second stage to generate new reviewAssign
   })
 })
 
-// Recreated test case for First Review submission (Stage 1 - generate assignment for Stage 2)
-
 // Simulate review submission:
 
 test('Test: Submit Review ID#7002 for Application ID#4003 - Stage 2 Lvl 1 to generate level 2 assignments', () => {
@@ -308,14 +307,14 @@ describe('Create new Review to simulate application moving to last stage (Final 
   beforeAll(async (done) => {
     await DBConnect.query({
       text: `
-      INSERT INTO public.review (id, application_id, review_assignment_id)
-        VALUES (8000, 4004, 10)
+      INSERT INTO public.review (id, review_assignment_id)
+        VALUES (8000, 10);
       INSERT INTO public.review_decision (id, decision, review_id)
         VALUES (DEFAULT, 'NON_CONFORM', 8000);
       INSERT INTO public.review_status_history (id, review_id, status)
-        VALUES (DEFAULT, 7004, 'SUBMITTED');
+        VALUES (DEFAULT, 8000, 'SUBMITTED');
       INSERT INTO public.application_stage_history (id, application_id, stage_id, is_current)
-        VALUES (2000, 4004, 7, 'True');
+        VALUES (2000, 4003, 7, 'True');
       INSERT INTO public.application_status_history (id, application_stage_history_id, status, is_current)
         VALUES (DEFAULT, 2000, 'SUBMITTED', 'True');
       `,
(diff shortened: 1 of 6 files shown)


CarlosNZ:
I'll have a look at this this morning, can you just fix the conflicts first? Thx

nmadruga:
Coo. Updated - also I made sure 2 tests were working.

I think this is one of the PR I will need to merge to branch #496 to get setup for dev tests on `dev` snapshots.

plugins/action_generate_review_assignment_records/src/generateReviewAssignments.test.ts
```diff
@@ -61,14 +61,14 @@ describe('Submit new application on first stage to generate reviewAssignments on
...
-      expect(result).toEqual({
+      expect(clearResult(result)).toEqual({
         status: ActionQueueStatus.Success,
         error_log: '',
         output: {
           reviewAssignments: [
             {
               reviewerId: 2,
-              orgId: null,
+              organisationId: null,
```
CarlosNZ, line 71: Hmmm, when did this change from `orgId` to `organisationId`? Pretty much all the action plugins use `orgId` for their input and output parameter names (we had a discussion about it ages ago) -- wondering why this one is different?

nmadruga: This is only something store locally on the `generateReviewAssignment` function, using the type defined in that plugin
Screen Shot 2021-08-06 at 10 31 04 AM

nmadruga:
When you are ready please approve/changes request @CarlosNZ 

I have already merged this PR into #474 (along with #480) and I'm planning to also merge #426 today. I think that's all schema/template changes we have so far? So planning to have a good snapshot for basic/dev/laos on our new repo today - and from there we can continue with configs next week.

CarlosNZ:
Okay, as discussed in IRL:

- the functionality to update existing records rather than creating new ones already exists in the database methods, so no need to "check for existing" on each generated review_assignment
- we have an action to change review_assignment status to "NOT_AVAILABLE" when one reviewer is assigned, but the problem in this case is that it was creating NEW review_assignments for reviewers who would normally have already had them (since this data was from a manual entry). This could still happen in a real-world situation though, if new reviewers are given permissions after the initial review_assignments have been created.
- so... the bug we actually need to fix is making sure NEW review_assignments check if there is someone already assigned as part of deducing their new status.

Here is the function that is at fault, I'd say:
```js
const getAssignmentStatus = () => {
      if (canMakeFinalDecision) return ReviewAssignmentStatus.Assigned
      if (canSelfAssign || nextReviewLevel > 1)
        return ReviewAssignmentStatus.AvailableForSelfAssignment
      return ReviewAssignmentStatus.Available
    }
```

It's not considering the status of OTHER review_assignments that already exist. I guess it probably needs an additional database query just to get all the current ones and return if ANY has "ASSIGNED" already.

Good news is that it will be quite a small additional from what's currently in master, I think. Maybe just make a new PR?

plugins/action_generate_review_assignment_records/src/generateReviewAssignments.test.ts
```diff
@@ -31,184 +32,232 @@ test('Test: Submit Application ID#4001 - Stage 1 (Last level)', () => {
...
-            levelNumber: 1,
-            isLastLevel: true,
-            isLastStage: false,
-            isFinalDecision: false,
-          },
-        ],
-        reviewAssignmentIds: [1, 2, 3, 4, 1005, 6, 7],
-        reviewAssignmentAssignerJoins: [],
-        reviewAssignmentAssignerJoinIds: [],
-        nextStageNumber: 1,
```
nmadruga, line 130: As pointed out by @CarlosNZ the reason this assignments are getting created/updated is because when we first manually generated reviewAssignments for applicationId 4001, only 1 was created, but there are 7 users with proper permission. So, 6 new are being generated and 1 is updated (reviewAssignment 1005, with new status: `AvailableForSelfAssignment`. 
Screen Shot 2021-08-10 at 12 35 34 PM

Screen Shot 2021-08-10 at 12 35 44 PM

The proper fix for this issue will be to prevent new reviewAssignments generated to have this status if one is already **Assigned**. A new PR will be created to proper fix this. And this one can be closed - without merging.

Issue #480, opened 2021-07-29
Scheduled Actions: Main infrastructure


CarlosNZ:
- Schema updates
- Schedule Action action
- Process Triggers updates
- Node-scheduler

PR #486, opened 2021-07-31
#485 objectFunctions for evaluator

### LINKED ISSUE #485: evaluator: "Date" operator
Needed so we can generate expiry date values for outcome tables

Suggested implementation:

operator: "dateTime"
children: 
1. Luxon DateTime method name (e.g. "fromISO", "fromObject", "now", "local", etc)
2. argument for (1) (e.g { year: 2021, month: 12, day: 18 } )
3. (optional): further DateTime chained methods (e.g. "plus")
4. (optional): arg for (3).
5. ... chained as required

e.g. if you want to return a Date that is one year from the current time, expression would be:
```
{
  operator: "dateTime",
  children: [
    "now",
    null,
    "plus",
    { year: 1 }
  ]
}

Output will be Javascript Date object

CarlosNZ:
Fixes #485 (indirectly) -- see [this comment](https://github.com/openmsupply/application-manager-server/issues/485#issuecomment-890276358)

A general solution for adding arbitrary functions to evaluator, so we can add any additional functionality we require without adding loads of operators.

This PR just has the evaluation and a couple of tests to demonstrate it, but I've used this to generate expiry dates for products/outcomes in the Scheduled Actions (PR coming soon), and it works nicely for that.

- Also updated existing test due to changing database data

### DIFF
--- a/src/modules/expression-evaluator/package.json
+++ b/src/modules/expression-evaluator/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@openmsupply/expression-evaluator",
-  "version": "1.8.0",
+  "version": "1.9.0",
   "description": "Module to evaluate dynamic expressions for the openMsupply Application Manager project",
   "main": "lib/evaluateExpression.js",
   "types": "lib/evaluateExpression.d.ts",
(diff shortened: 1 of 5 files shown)


CarlosNZ:
Note, have already published this package so I can use it in #480, but can easily publish a patch update if changes required.

nmadruga (approved):
All good tested running the `src/modules/expression-evaluator/src/evaluateExpression.test.ts` and all seems ok

PR #492, opened 2021-08-04
#491 Category-specific layout settings

### LINKED ISSUE #491: Implement a way to specify which template categories should show up where
The main thing is we don't want "non-template" templates to show up in places like the Dashboard, as they just make it confusing to the user.

There are also several different places template links can go, so it would be good to be able to specify that, namely:
- Dashboard
- User menu
- Main menu bar
- Applications list (we probably want to never show certain templates, such as UserEdit, or ChangePassword)
- Other?

### Suggested solution 

- new enum "ui_location" with values corresponding to ui placement: DASHBOARD, MENU, USER, ?
- new field on template_category called "ui_location" which is an **array** of the above enum values eg `[DASHBOARD, MENU]`
- front end ui then just filters by ui_location accordingly

Default value for ui_location field will be `[DASHBOARD, MENU]`

CarlosNZ:
Fixes #491 as per agreed solution.

Test with [front-end PR #890](https://github.com/openmsupply/application-manager-web-app/pull/890)

Only extra thing I did was to add an 'icon' field to "template", so each template _can_ have its own icon. Only used in "User" menu for "Edit User details" currently. (If a template icon is called and its not defined, it'll fall back to the category icon)

### DIFF
--- a/database/buildSchema/07A_template_category.sql
+++ b/database/buildSchema/07A_template_category.sql
@@ -1,8 +1,17 @@
 -- template category table
+--
+-- places in ui where template_categories can show up
+CREATE TYPE public.ui_location AS ENUM (
+    'DASHBOARD',
+    'MENU',
+    'USER'
+);
+
 CREATE TABLE public.template_category (
     id serial PRIMARY KEY,
     code varchar NOT NULL UNIQUE,
     title varchar,
-    icon varchar
+    icon varchar,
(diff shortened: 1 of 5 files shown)


andreievg (approved):

CarlosNZ:
@andreievg -- there's a front-end partner to this too, which will need to be merged at the same time, if you want to take a quick look?: https://github.com/openmsupply/application-manager-web-app/pull/890

PR #495, opened 2021-08-04
#481 #482 Grant/Revoke permissions updates

### LINKED ISSUE #481: Scheduled Actions: Update Grant Permissions to update existing, and add a "active/inactive" flag

CarlosNZ:
Fixes #481 #482

### Summary of Changes

- add `is_active` field to "permission_join" table
- "grantPermissions" action ensures "is_active" gets turned back on if the permission_joins already exist
- new "revokePermissions" action -- same input format as "grantPermissions", but it just sets "is_active" to `false` on existing permission_joins

### Testing -- use Postman `/run-action`

Grant a bunch of permissions to user without org:
```
{
    "actionCode": "grantPermissions",
    "applicationId": 4002,
    "parameters": {
        "username": "carl",
        "permissionNames": [
            "applyRenewLicense", "applyImportPermit", "assignGeneral", "applyDrugRegoMMC", "applyDrugRegoMMD"
        ]
    }
}
```

Revoke some of them:
```
{
    "actionCode": "revokePermissions",
    "applicationId": 4002,
    "parameters": {
        "username": "carl",
        "permissionNames": [
            "applyDrugRegoMMC", "applyDrugRegoMMD"
        ]
    }
}
```

Try re-granting them using the original query and note that `is_active` gets set back to `true`

Now grant permissions to user with an org:

```
{
    "actionCode": "grantPermissions",
    "applicationId": 4002,
    "parameters": {
        "username": "carl",
        "orgName": "Drugs-R-Us",
        "permissionNames": [
            "applyRenewLicense", "applyImportPermit", "assignGeneral", "applyDrugRegoMMC", "applyDrugRegoMMD"
        ]
    }
````

And try revoking some of them:

```
{
    "actionCode": "revokePermissions",
    "applicationId": 4002,
    "parameters": {
        "username": "carl",
        "orgName": "Drugs-R-Us",
        "permissionNames": [
            "applyDrugRegoMMC", "applyDrugRegoMMD"
        ]
    }
}
```

Please try playing with different org/user settings to ensure permissions are added and revoked as expected

### CODE REVIEW

File: plugins/action_revoke_permissions/plugin.json [hunk 1]
```diff
@@ -0,0 +1,24 @@
+{
+  "type": "action_revoke_permissions",
+  "code": "revokePermissions",
+  "name": "Revoke Permissions",
+  "description": "Revokes existing permission(s) from user/org -- only sets existing permissions to inactive, doesn't actually delete records",
+  "required_parameters": ["username", "permissionNames"],
```
nmadruga, line 6: Maybe username doesn't need to be required, since can be fetched from applicationData

nmadruga: Similarly for `grantPermission`. I'm only suggesting to update the docs and `plugin.json` description

plugins/action_revoke_permissions/plugin.json
```diff
@@ -0,0 +1,24 @@
...
+  "optional_parameters": ["orgName", "orgId"],
+  "output_properties": ["permissionJoinIds", "permissionNames"],
```
nmadruga, line 8: It would be nice to output `permissionJoinIds` and `permissionNames`, but currently it's outputs other fields. Look what I'm getting after running the action (Postman):
```
{
    "status": "SUCCESS",
    "error_log": "",
    "output": {
        "revokedPermissions": [
            {
                "permission_name_id": 7,
                "name": "applyDrugRegoMMC"
            },
            {
                "permission_name_id": 8,
                "name": "applyDrugRegoMMD"
            }
        ]
    }
}
```

nmadruga (approved):
Looking good! Approved - Was thinking if we could do something already to prevent permissions set to `isActive` = false to not get returned on JWT token for user permissions?

CarlosNZ:
> Was thinking if we could do something already to prevent permissions set to `isActive` = false to not get returned on JWT token for user permissions?

Yes, but we don't want to completely prevent access, which that would do. Users should always be able to see their previous applications, even if they've lost the right to create new ones. So the idea was that if the permission exists but "is_active = false" then they can read but not create for that type.

But this would have to be implemented at the policy level (as well as some front-end filtering probably)

plugins/action_revoke_permissions/plugin.json
CarlosNZ: The permissionName is there (just "name"), but yeah okay, I've updated to include the permissionJoinIds and renamed "name" to "permissionName"

091f12f9

CarlosNZ:
I still need to do docs for this, so will do so later this afternoon

CarlosNZ:
> I still need to do docs for this, so will do so later this afternoon

Done 1254e7a2

(Some jobs are easy enough to do while sitting on the couch with a toddler)

plugins/action_revoke_permissions/plugin.json
CarlosNZ: Yup. Done 4bd37270

PR #498, opened 2021-08-06
#474 move templates to submodule main

### LINKED ISSUE #474: Summarise changes to start using snapshots + template builder to keep our dev insertData and laos insertData as snapshots considering tests which would require assignment and reviews already created

nmadruga:
Fixes #474

## Summary of changes
Starting setup to be only using Snapshots import and template builder for config from now on.
Basically added new private repo: https://github.com/openmsupply/application-manager-templates
Change to use snapshots in database_init and move templates to new submodule **database/_snapshots**

**Note**: The `basic_snapshot` containing only 5 templates and 1 user was added to database/insertData folder, and after the submodule is initialised this folder gets copied to _snapshots so all snapshots can be imported/exported from the same place. The submodule will fail to initialise if it already contain some file/folder and that's why it doesn't start already there.

List of TODOs to finish up this setup:
- [x] Create basic snapshot to be added to main repo
- [x] Test user without access
- [x] Use **basic** snapshot when no params are passed (first copy datafile/insertData/snapshot_basic to datafile/_snapshots)
- [x] Make sure users without access can load basic snapshot and export new snapshots to folder database/_snapshots
- [ ] Change export templates to add to current folder (in _snapshots)
- [ ] Remove old config and js files **Last one until we need to merge things in js files will be kept**

## To be decided
- [ ] Improve script to pushing commits on _snapshots folder after exporting snapshot/template 
- [ ] Show **upload** button in UI to get latest templates from private repo during development


nmadruga:
Screen Shot 2021-08-06 at 5 19 46 PM

Problem happening when opening any template after loading a snapshot

nmadruga:
Updates:

- Renamed `snapshot_basic` to `basic_snapshot`
- Only one username available to log in: **admin**
- New template added code CoreActions, has description on each action. This template should be used when duplicating templates to start a new one - plan to start `demo_snapshot` based on `basic_snapshot`
- List of templates available in `basic_snapshot`:
- Other 2 snapshots `dev` has all ReviewTest applications and tests should be running using this config. `laos` has the previous setup for laos with templates to be used as example as we configure `demo_snapshot`. I suggest running it on a second instance on docker.

nmadruga:
Problems encountered:
- `OrgRego` template is not opening on Template builder. Also in the console this logs are showing up:
Screen Shot 2021-08-11 at 12 36 13 PM
- `CoreActions` has wrong `namePlural` set from when it was duplicated and show in with the wrong title in the Dashboard.

nmadruga:
And here is the list of templates in `laos` snapshot:

Screen Shot 2021-08-11 at 12 52 43 PM

Although testing to load this one now I see there are problems to open some of the templates - I'm getting the "Serial is missing" error for example when trying to open `applyForPermisisons` template:
Screen Shot 2021-08-11 at 12 54 31 PM

CarlosNZ:
One thing we need to sort for this is to update the build scripts to handle the change.
In particular:
- "build_all.sh" will need updating
- maybe need to check the submodule gets pulled properly as part of the process

Can you make sure the full Docker build process works okay with this new setup before we merge it master?

(P.S. I'm fiddling around with this on the demo branch now, so I'll report back any thing relevant).

CarlosNZ:
So this seems to be working pretty well, having used it in the demo branch for a few days.

One thing I'd prefer though, and I may be in the minority here, but I think it's better if the "working" folder that the snapshot module reads and writes from/to by default is _different_ from the repo (and in .gitignore). Cos I much prefer to be free to make a mess of my snapshots (with many versions etc) and only share it with the repo once its in a state I consider stable.

We can leave it like this if you'd prefer, but I'll just need to be careful not to commit anything and just use Dropbox or something for keeping my WIP snapshots synced (which I'm doing currently)

nmadruga:
Closing this PR - a new one was created pointing to new branch `develop`.

CarlosNZ:
> Closing this PR - a new one was created pointing to new branch `develop`.

Since this is a major restructure, with changes to git settings, we should probably merge this to "main" (or the other PR, specifically) then merge main back into develop, don't you think? (could get messy otherwise when switching back and forth)

nmadruga:
> > Closing this PR - a new one was created pointing to new branch `develop`.
> 
> Since this is a major restructure, with changes to git settings, we should probably merge this to "main" (or the other PR, specifically) then merge main back into develop, don't you think? (could get messy otherwise when switching back and forth)

Yes. Good point. I still needed to get latest `main` as base for this to keep it up-to-date (so, can change the other PR to base on `main` instead)

PR #499, opened 2021-08-06
Demo build changes

andreievg:
Sorry no issue:

* add libre office to docker build
* show how environmental variables can be passed to server when docker image is being run
* make sure that missing values are not imported as undefined when loading snapshot

### DIFF
--- a/docker/run.sh
+++ b/docker/run.sh
@@ -5,6 +5,7 @@
 docker run \
     -ti \
     -p 3000:3000 \
+    -e "SMTP_PASSWORD=smt_password_goes_here" \
     testbuild
 
 # -ti -> interactive (connect to shell on startup)
--- a/src/components/exportAndImport/importFromJson.ts
+++ b/src/components/exportAndImport/importFromJson.ts
@@ -61,6 +61,7 @@ const constructInsertAndGetter = (
   columns.forEach(({ isPrimary, isGenerated, columnName }) => {
     if (!preserveIds && isPrimary) return
     if (isGenerated) return
+    if (values[columnName] === undefined) return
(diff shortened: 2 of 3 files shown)


src/components/exportAndImport/importFromJson.ts
```diff
@@ -61,6 +61,7 @@ const constructInsertAndGetter = (
   columns.forEach(({ isPrimary, isGenerated, columnName }) => {
     if (!preserveIds && isPrimary) return
     if (isGenerated) return
+    if (values[columnName] === undefined) return
```
CarlosNZ, line 64: Can this be:
```
if (values?.[columnName]) return
```
That would cover "undefined", "Null", empty strings, and just missing.

docker/run.sh
```diff
@@ -5,6 +5,7 @@
 docker run \
     -ti \
     -p 3000:3000 \
+    -e "SMTP_PASSWORD=smt_password_goes_here" \
```
CarlosNZ, line 8: So how do you put the password in here without saving it to the repo?

src/components/exportAndImport/importFromJson.ts
andreievg: I thought it's good to differentiate between null or empty string to undefined (since those values would have been exported and perhaps should be imported), vs not exported fields:values which (for columns that were added since export)

docker/run.sh
andreievg: ohh this is just an example to show that env variable can be passed when running built image in demo server

CarlosNZ (approved):

PR #500, opened 2021-08-09
#768F Number plugin template updates

CarlosNZ:
Updates to FeatureShowcase template to demonstrate the "Number" input plugin (front-end PR ???)

Questions are in Section 3, page 5 - 6

(Note PR also has changes from DatePicker plugin which were never merged into back-end)

Will need to manually transfer useful elements over to Snapshot version of this template at some point


nmadruga:
> Will need to manually transfer useful elements over to Snapshot version of this template at some point

Oh! That's quite big... Could it be done before we merge this one? I think it would be as part of `dev` and probably also added to `basic_snapshot` as well? I think it would be a hard task to do it **only** using the Template builder though... Maybe if we merge this and update #474 to have this it can be merged with current js files in dev and a new snapshot generated - I guess that would be the easiest way to do this.

nmadruga (approved):
Looks good! And very nice to have this in.

Issue #476, opened 2021-07-28
Should always lock partial reviews undone or not submitted when moved to next stage


nmadruga:

CarlosNZ:
I think this could be done by a Postgres trigger/function, perhaps?

CarlosNZ:
Done

PR #501, opened 2021-08-10
#690F fix review assignments duplicated for level 1 after LOQ is submitted

nmadruga:
Fixes  front-end https://github.com/openmsupply/application-manager-web-app/issues/690

### Changes
- Created new method `getExistingReviewAssignments` in **action_generate_review_assignment** to get existing ReviewAssignments
- When generating (or updating existing) ReviewAssignment will set `isLocked` to true if other ReviewAssignment already existing is Assigned, and keep the same status and isLocked on existing ReviewAssignments.
- Fix tests in `generateReviewAssignments` to be receiving new `isLocked` as part of resulted reviewAssignments
- Fix problem with `updateReviewStatuses` to be aligned with current application/assignments

### Manual testing
- Use application ABC456 logged in with user **Reviewer1**
- Start review and DECLINE one response and submit the LOQ to applicant
- Log in with user **Valerio** and fix the response then re-submit
- Log in again as **Reviewer1** to Re-review and APPROVE all then CONFORM and submit
- Now the application is on the next stage. No other reviewAssignments have been created for other reviewer in stage 1

Check in the master to see the bug happening - you can check the logs to see new reviewAssignments created after re-submitting the application with LOQ back to the reviewer.


plugins/action_generate_review_assignment_records/src/generateReviewAssignments.ts
```diff
@@ -146,17 +146,47 @@ const generateNextReviewAssignments = async ({
...
+  // Check if other reviewAssignmet is already assigned to create new ones LOCKED
+  const existingReviewAssignments: ExistingReviewAssignment[] =
+    await db.getExistingReviewAssignments(applicationId, stageNumber, nextReviewLevel)
+
+  const getNewOrExistingAssignmentStatus = (
+    userId: number,
+    canMakeFinalDecision: boolean,
+    canSelfAssign: boolean
+  ) => {
+    // Temporarly final decision shouldn't be locked if there are other reviewAssignemt assigned
```
CarlosNZ, line 158: Is it "temporarily"? I thought it would be permanent (at least as far as this action is concerned)

plugins/action_generate_review_assignment_records/src/generateReviewAssignments.ts
```diff
@@ -146,17 +146,47 @@ const generateNextReviewAssignments = async ({
...
+    if (canMakeFinalDecision) return { status: ReviewAssignmentStatus.Assigned, isLocked: false }
+
+    // Check if existing review assignment
+    const existingAssignment = existingReviewAssignments.find(
+      (reviewAssignment) => reviewAssignment.userId === userId
+    )
+    // If user already has reviewAssignment return same status and isLocked flag
+    if (existingAssignment)
+      return { status: existingAssignment.status, isLocked: existingAssignment.isLocked }
```
CarlosNZ, line 167: What about if the review_assignment is NOT locked, but there is *another* reviewer who has assignment "ASSIGNED"? Then surely we need to set *this* review_assignment to "LOCKED", right?

(In theory, this shouldn't happen, since it would have already been set previously, but it's probably a good idea to cover that case anyway, like in the situation of our manual records that weren't consistent)

plugins/action_generate_review_assignment_records/src/generateReviewAssignments.ts
```diff
@@ -146,17 +146,47 @@ const generateNextReviewAssignments = async ({
...
+
+    // Non-existing review assignments
+    const isLocked = existingReviewAssignments.some(
+      ({ status }) => status == ReviewAssignmentStatus.Assigned
+    )
```
CarlosNZ, line 172: It seems inefficient to do the same search through the same array for every reviewer. Can this value just be deduced once for all?

Maybe just below `const existingReviewAssignments = ...`, but outside the loop:

```js
const anyAlreadyAssigned = existingReviewAssignments.some(
      ({ status }) => status == ReviewAssignmentStatus.Assigned
    )
```

And then just refer to `anyAlreadyAssigned` inside the loop?
```js
if (canSelfAssign || nextReviewLevel > 1)
  return { status: ReviewAssignmentStatus.AvailableForSelfAssignment, isLocked: anyAlreadyAssigned }
return {
  status: ReviewAssignmentStatus.Available,
  isLocked: anyAlreadyAssigned,
}
```

I think that would be easier to understand, too, personally.

plugins/action_generate_review_assignment_records/src/generateReviewAssignments.ts
CarlosNZ: Also, just wondering if the "existing and unchanged" case shouldn't be last in the list of checks (so it's the fallback). That way you won't have to add the `if (existingAssignment)...` check as you can just return the existing values since other cases will have returned earlier.

So it would go:
- If any other reviewers assigned -- return with amended status/isLocked values
- if (canSelfAssign || nextReviewLevel > 1) return with Available status
- else return unchanged (cos it doesn't matter if it already exists or not, there's no need to check that, it's already handled by the database update).

Is my understanding correct?

CarlosNZ (changes requested):
Cool, looks good. Took me a while to fully understand the logic of checking the existing assignments. I'm not 100% sure my suggestions are correct, I may have misunderstood somewhat, I just want to make sure the logic here is as straightforward as possible since it's getting pretty gnarly and we want to make sure we cover all possible cases with minimal confusion.

CarlosNZ:
Also -- is there any changes necessary for the "Generate review_assignment_assigner_joins"? 

I'm thinking not, but just wondered if you considered that at all in this issue.

plugins/action_generate_review_assignment_records/src/generateReviewAssignments.ts
nmadruga: hahahha yes, it would for now - I was thinking I would be working with issue #476 and check if this would be changed or just rephrase this comment in case not. Updating to "temporarily" anyways.

plugins/action_generate_review_assignment_records/src/generateReviewAssignments.ts
nmadruga: > What about if the review_assignment is NOT locked, but there is _another_ reviewer who has assignment "ASSIGNED"? Then surely we need to set _this_ review_assignment to "LOCKED", right?

Oh, I see if the reviewAssignment is existing... I think preventing this is a sword to fix a non-existing problem because this shouldn't have happened in the first place. When other reviewAssignment is ASSIGNED other reviewAssignments already get set to locked, so I don't see the need to do this (again) here to prevent some possible missed case.

nmadruga: > So it would go:
> 
> * If any other reviewers assigned -- return with amended status/isLocked values
> * if (canSelfAssign || nextReviewLevel > 1) return with Available or Available for self-assignment status
> * else return unchanged (cos it doesn't matter if it already exists or not, there's no need to check that, it's already handled by the database update).
> 
> Is my understanding correct?

I'm not sure I can see what do you mean... We would still have to check if non-existing, because it it is then the status is already set - it can be more than 2 possibilities.

nmadruga: If it's existing it can be either: AVAILABLE, SELF_ASSIGNED_BY_ANOTHER, ASSIGNED and AVAILABLE_FOR_SELF_ASSIGNMENT

CarlosNZ: >  think preventing this is a sword to fix a non-existing problem because this shouldn't have happened in the first place.

Sure, it shouldn't, but this is the exact situation that created this issue in the first place, since we had manually created review_assignments. I think it'd be good to cover this case -- otherwise we're assuming that all review_assignments will only ever be created by this action. Mostly that's true, but we shouldn't assume it -- the idea should be that whenever this action runs, the output is correct no matter how messed up or inconsistent the existing values might have been beforehand.

CarlosNZ: > > 
> 
> I'm not sure I can see what do you mean... We would still have to check if non-existing, because it it is then the status is already set - it can be more than 2 possibilities.

Yeah, but I thought you were returning the current status unchanged. My point was, if you're going to return the status unchanged, then there's no need to check if it exists -- if it exists, it won't change; if it doesn't exist, it'll get created.

nmadruga: Ok... I have changed to something, but while testing I see one case which this logic doesn't work - like I said previously the review assignment set to lock is done separately and more complex. So the case is, when you have reviews assigned to reviewers by an assigner. Other review assigments should be AVAILABLE (note it's not AVAILABLE_FOR_SELF_ASSIGNMENT). I'll have a thinking now to see how I can count for these cases as well.

nmadruga: I have changed the code to this (not less - it's actually more code so I don't like it):
```
// Check if existing review assignment
    const existingAssignment = existingReviewAssignments.find(
      (reviewAssignment) => reviewAssignment.userId === userId
    )

    const isExistingAndAssigned =
      existingAssignment && existingAssignment.status === ReviewAssignmentStatus.Assigned

    const someoneElseIsAssigned = existingReviewAssignments.some(
      ({ status }) => status == ReviewAssignmentStatus.Assigned
    )

    // Should lock other review assignments if one is already ASSIGNED
    const isLocked = isExistingAndAssigned
      ? false
      : someoneElseIsAssigned
      ? true
      : existingAssignment?.isLocked ?? false

    // Get current status - if already existing or set new status
    const status =
      existingAssignment?.status ||
      (canSelfAssign || nextReviewLevel > 1
        ? ReviewAssignmentStatus.AvailableForSelfAssignment
        : ReviewAssignmentStatus.Available)

    return {
      status,
      isLocked,
    }
```

The problem is - this is making another assumption which is to lock reviewAssignments when another one is ASSIGNED. But this is wrong in some cases:
- Many users than have permission to review S1, S2 or S3 they start with reviewAssignment with status AVAILABLE
- If only one section is assigned to one review Assignment, now the status is ASSIGNED
- If this code runs it would change other review Assigments to **locked**
- If the assigner needs to assign other sections no reviewers would show

nmadruga: if we want to also check for sections - this will start to be a major change on this behaviour. Remember this PR was only to prevent changing the reviewAssigment's status to something else after applicant re-submits to reviewers.

plugins/action_generate_review_assignment_records/src/generateReviewAssignments.ts
CarlosNZ: What I mean was why are you saying "temporarily" at all? Shouldn't it just say "Final decision shouldn't be locked if there are other reviewAssignemt assigned" -- what's temporary about it?

plugins/action_generate_review_assignment_records/src/generateReviewAssignments.ts
CarlosNZ: > I have changed the code to this (not less - it's actually more code so I don't like it):

Where is this? Not seeing it on the branch.

CarlosNZ: This conversation is getting a bit hard to follow. I think there is some confusion as to what exactly we're trying to achieve with this change, but I'm not sure if it's me or you. (I thought it was a fairly minor tweak)

I think I'll draw a little diagram (on paper) of what I think the logic needs to be, then I'll show it to you tomorrow and see if you agree. Or you can correct it if need be.

nmadruga:
> Also -- is there any changes necessary for the "Generate review_assignment_assigner_joins"?
> 
> I'm thinking not, but just wondered if you considered that at all in this issue.

That hasn't been touched, so it will stay the same

plugins/action_generate_review_assignment_records/src/generateReviewAssignments.ts
```diff
@@ -146,17 +146,48 @@ const generateNextReviewAssignments = async ({
   console.log('Next level reviewers', nextLevelReviewers)
   const reviewAssignments: ReviewAssignmentObject = {}
 
+  // Check if other reviewAssignment is already assigned to create new ones LOCKED
+  const existingReviewAssignments: ExistingReviewAssignment[] =
+    await db.getExistingReviewAssignments(applicationId, stageNumber, nextReviewLevel)
+
+  const isLocked = existingReviewAssignments.some(
```
CarlosNZ, line 153: Maybe just change the variable name now to something like "reviewAlreadyAssigned" ? It's a bit confusing because you also have "isLocked" below which means something similar but not the same.
 "shouldBeLocked" might be better, but why not just call it exactly what it is?

CarlosNZ (changes requested):
Okay, thanks for being patient with this. 

Last comments, based on IRL chat and small things:

- where it says "temporarily final decision should be locked....", please note that further check to be done in another issue (and maybe add issue no.)
- please add comment (somewhere around line 170) that this logic doesn't cover every possible case and that it needs to be re-addressed at some point (to also consider full assignment based on sections, etc.)
- see my inline comment about the `isLocked` variable (the first one, not the field name)

Thanks.

plugins/action_generate_review_assignment_records/src/generateReviewAssignments.ts
nmadruga: Changing to `reviewAlreadyAssigned`

CarlosNZ: Great, that's perfect. :)

CarlosNZ (approved):
Great, thanks for the changes

PR #516, opened 2021-08-17
#474(2) rebase to develop templates to snapshots

### LINKED ISSUE #474: Summarise changes to start using snapshots + template builder to keep our dev insertData and laos insertData as snapshots considering tests which would require assignment and reviews already created

nmadruga:
Fixes #474

## Summary of changes
Starting setup to be only using Snapshots import and template builder for config from now on.
Basically added new private repo: https://github.com/openmsupply/application-manager-templates
Change to use snapshots in database_init and move templates to new submodule **database/_snapshots**

**Note**: The `basic_snapshot` containing only 5 templates and 1 user was added to database/insertData folder, and after the submodule is initialised this folder gets copied to _snapshots so all snapshots can be imported/exported from the same place. The submodule will fail to initialise if it already contain some file/folder and that's why it doesn't start already there.

List of TODOs to finish up this setup:
- [x] Create basic snapshot to be added to main repo
- [x] Test user without access
- [x] Use **basic** snapshot when no params are passed (first copy datafile/insertData/snapshot_basic to datafile/_snapshots)
- [x] Make sure users without access can load basic snapshot and export new snapshots to folder database/_snapshots
- [ ] Change export templates to add to current folder (in _snapshots)
- [ ] Remove old config and js files **Last one until we need to merge things in js files will be kept**

## To be decided
- [ ] Improve script to pushing commits on _snapshots folder after exporting snapshot/template 
- [ ] Show **upload** button in UI to get latest templates from private repo during development


CarlosNZ:
One other suggestion: to avoid confusion, should we only save the "folder" version of snapshots intended for "initialising" with, and only save the "zip" versions for ones intended for the "Import" function of template builder. Just that, we may wish to make some manual edits to the "snapshot.json" file, and it's kind of a pain to have to re-zip it everytime, and could lead to inconsistency. It would also allow us to tell at a glance whether the saved snapshots are "whole system" snapshots or exported templates.

(BTW, I don't mean changing the export scripts at all, I just mean we delete one of the versions before pushing to the submodule)

nmadruga:
@CarlosNZ I think it will be good to merge this in to `main` so we get other PRs also aligned... They might need to do something about current snapshots when things have changed on the templates, but I guess it will be easier that way. Other than having this PR in between - already in demo but not in `main`.

CarlosNZ:
> @CarlosNZ I think it will be good to merge this in to `main` so we get other PRs also aligned... They might need to do something about current snapshots when things have changed on the templates, but I guess it will be easier that way. Other than having this PR in between - already in demo but not in `main`.

Sure, it seems to be working well, when you merge to main, can you then merge main back into develop?

CarlosNZ (approved):
Cool. I'd like to address the "working folder" situation at some point (as mentioned [here](https://github.com/openmsupply/application-manager-server/pull/498#issuecomment-899888922)), but we can address that later on. I have some other thoughts for how we might improve that, but I won't bother you with that now.

Merge away!

nmadruga:
> Cool. I'd like to address the "working folder" situation at some point (as mentioned [here](https://github.com/openmsupply/application-manager-server/pull/498#issuecomment-899888922)), but we can address that later on. I have some other thoughts for how we might improve that, but I won't bother you with that now.
> 
> Merge away!

Oh! It's not boring. I just haven't been able to play with this again. It's so time consuming to update the snapshots that I have only done that and didn't add extra requirements. It would be good to have the templates exported only as zips (and to the same folder as the project - if that's still a thing) and snapshots only exported as folders. I have removed the template's folders and snapshots' zips from the templates repo BTW.

PR #519, opened 2021-08-17
#432 Add webhost URL to applicationData

### LINKED ISSUE #432: Back-end config file: add website URL, or figure out how to access from front-end
And add it to the "environmentData" part of "getApplicationData" so we can create correct URLs in Actions, as per [this note](https://github.com/openmsupply/application-manager-server/pull/422#discussion_r664222761)

Update: As discussed with @andreievg  -- will be adding as an environment variable and loading it into "applicationData". In dev mode, will be stored in `.env` file, in Docker, will be passed in as a runtime argument (like SMTP password)

CarlosNZ:
Fixes #432

You'll need to also add this line to your `.env` file:
```
WEB_HOST = "http://localhost:3000"
```

I still need to update the templates to use this, but I'll do that on the demo server snapshot

### DIFF
--- a/README.md
+++ b/README.md
@@ -52,6 +52,13 @@ This repo has the following functionality implemented in basic form:
   - Create scheduled actions (using `node-schedule` which I'm not sure is the best way to go)
 - `evaluateExpression` module
 
+## Development
+
+* `main` branch - only has features which have been released (on a demo-tag)
+* `develop` - to be used for development (to create a feature-branch before making a new PR) and selected as base for PRs
+
+The organisation-team will transfer approved changes from `develop` into the `main` periodically once all new features are stable and tested.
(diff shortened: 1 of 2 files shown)


nmadruga (approved):
LGTM. 

Do you also have another one to merge this into `develop`? I think we should be using `develop` to merge PRs from now on, so things can get transfer to main just in one bulk, otherwise we will have to keep updating develop?

CarlosNZ:
> LGTM.
> 
> Do you also have another one to merge this into `develop`? I think we should be using `develop` to merge PRs from now on, so things can get transfer to main just in one bulk, otherwise we will have to keep updating develop?

Sure, I've changed the base branch here.

PR #520, opened 2021-08-17
Speed up application creation action sequence

andreievg:
Fixes: https://github.com/openmsupply/application-manager-web-app/issues/904

explicit search for table name. this is more evident in docker instance of pg vs local (when creating application schema lookup for modify record takes a bit longer then expected)


CarlosNZ (approved):
Cool, thanks. Can you just change the base branch to "develop" before merging thought? Nicole is keen to keep develop as the main working branch now.

Issue #522, opened 2021-08-18
Snapshot not importing


andreievg:
snapshot not importing with error on applicationStatusHistory duplicate PK when there is at least one completed application

PR #521, opened 2021-08-17
#505 and #506 user category template to always show and file snapshot management

andreievg:
fixes: https://github.com/openmsupply/application-manager-server/issues/505
partial fixes to

https://github.com/openmsupply/application-manager-server/issues/506

Was hoping to fix file export/import in snapshots such as files associated with template are copied and pasted, but realised that template_id on file is associated with the file for application files. I thought it would only be associated for files like carbone template, @CarlosNZ. Remove that association (so that template export functionality can still work), in front end PR: https://github.com/openmsupply/application-manager-web-app/pull/912

Btw are we still using fileCleanup action ?

PR #523, opened 2021-08-18
522-snapshot-to-work-when-importing-completed-application

andreievg:
Fixes: #522 
update for trigger vs insert, merged to demo already.

@CarlosNZ, I think we need to also do a condition of OLD.outcome = 'PENDING" ?

### DIFF
--- a/database/buildSchema/16_application.sql
+++ b/database/buildSchema/16_application.sql
@@ -50,7 +50,7 @@ LANGUAGE plpgsql;
 
 --TRIGGER to run above function when outcome is updated
 CREATE TRIGGER outcome_trigger
-    AFTER INSERT OR UPDATE OF outcome ON public.application
+    AFTER UPDATE OF outcome ON public.application
     FOR EACH ROW
     WHEN (NEW.outcome <> 'PENDING')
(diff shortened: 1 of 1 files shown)


CarlosNZ:
> @CarlosNZ, I think we need to also do a condition of OLD.outcome = 'PENDING" ?

Yeah, I guess, although once applications are "COMPLETED" (i.e. Outcome <> "Pending"), they should never change again. But feel free to add it.

andreievg:
I've added that check, I am surprised that there are no duplicates, since there is also an update after to change trigger to null

CarlosNZ:
> I am surprised that there are no duplicates, since there is also an update after to change trigger to null

I'm pretty sure I checked that when I updated it originally -- both functions have no overlap in their condition fields, so it should be all good.

CarlosNZ (approved):
Great, thanks.

nmadruga:
Can we merge this up @andreievg?

Was hoping we could do this today:

`main` would get PRs: #523, #521
`develop` should get changes from master + PRs #513, #524 & #528

andreievg:
rebased, no conflicts.

PR #528, opened 2021-08-20
Actions -- Better handling for condition check fail AND outputCumulative fix from scheduled trigger

CarlosNZ:
No issue sorry, just a quick update I made to demo, so adding it here as well.

Basically, was having a problem with the "condition" expression failing, which was causing the action in the queue to stay in a perpetual "PROCESSING" state, which affected other actions for the same template causing them to fail too.

This fix just specifically handles the case when the condition check fails, and aborts that action (sets it to FAIL rather than PROCESSING) and adds some error info to the error_log

EDIT: Also pushed a quick fix where "outputCumulative" was being returned from scheduled trigger event nested 1 level too deep.

### DIFF
--- a/src/components/actions/triggersAndActions.ts
+++ b/src/components/actions/triggersAndActions.ts
@@ -118,7 +118,7 @@ export async function processTrigger(payload: TriggerPayload) {
 
   // Collect output properties of actions in sequence
   // "data" is stored output from scheduled triggers or verifications
-  let outputCumulative = { ...data }
+  let outputCumulative = { ...data?.outputCumulative }
 
   // Execute sequential Actions one by one
   let actionFailed = ''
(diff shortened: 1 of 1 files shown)


nmadruga (approved):
Looks good. Thanks!

I have no idea how outputCumulative would be nested like that. Mistery...

CarlosNZ:
> I have no idea how outputCumulative would be nested like that.

I do. "outputCumulative" is stored as a field in the "data" object in the "trigger_schedule" table. And when restoring it I was doing `outputCumulative = { ...data }` when it needed to be `outputCumulative = { ...data?.outputCumulative }`.

What I'm really mystified about is how it managed to work okay during my original PR, and it only became a problem now.

PR #531, opened 2021-08-24
Unlink templates repo from submodule

CarlosNZ:
- Removes submodule from `database/_snapshots` folder

You may need to manually remove the "_snapshots" folder and recreate it for it to fully reset.

If you wish to keep using the "_snapshots" folder as your home of the templates repo (like it is currently), you should be able to just make "_snapshots" the home of your local templates repo, since it's in .gitignore. (Not 100% sure, haven't actually tried it)

Also added an "aliases" file for storing references to the paths of your various snapshots. 

So instead of using the command:
`yarn database_init ../../../application-manager-templates/demo/snapshots/demo_19_08_2021`
which points to the location of the snapshot, you could just use:
`yarn database_init demo`

The alias links are store in the file "snapshot_path_aliases.json" in the "database" folder, and contain info like this:
```
{
  "demo": "../../../application-manager-templates/demo/snapshots/demo_19_08_2021"
}
```

Note that snapshot_path_aliases.json is intended to be individualised to each person's system (depending on where you templates repo is stored), so I've added it to .gitignore.

Here is a base version you can use as a starting point:
[snapshot_path_aliases.json.zip](https://github.com/openmsupply/application-manager-server/files/7036541/snapshot_path_aliases.json.zip)

---

**EDIT**: Added another additional change here too, to allow easier switching between inserting a snapshot or inserting the previous .js insert scripts. 807604bd

Just add "js" as the first part of the argument after `database_init` to switch to insert js scripts (2nd argument for the name of the insert script folder.

Eg.
`yarn database_init js` -- runs insertData.js using default "dev" scripts
`yarn database_init js laos` -- runs insertData.js useing "laos" script folder

This seemed cleaner than turning lines on or off in the script.


CarlosNZ:
FYI -- I'll merge this into main and develop at the same time, so we've got a consistent base going forward.

nmadruga (approved):
Looks good - I haven't tested though. 

The idea of aliases is nice! And I liked the fix to use either js or snapshots.

Issue #530, opened 2021-08-24
Misc tweaks for next demo update


CarlosNZ:
Use this as an ongoing list of small things we should change before next demo build

(Most of these should be small enough that we can just push them straight to develop, no need for PR)

- [ ] Turn Scheduled Action check of "test" mode (so it checks once per hour, not every 30 seconds)
- [ ] Change application name on Login page
- [x] Typos in Company License (see below) - in latest snapshot demo-master-20
- [ ] Hide 'No Organisation' from first page of Company Registration application
- [ ] Unwrap Company License section header (or at least make the column a lot wider)

adamdewey:
Couple of minor typos: 

![image](https://user-images.githubusercontent.com/19318937/130549208-b9eb3f06-d7ea-409a-a567-eb220e73ad76.png)

adamdewey:
Login screen says IRIMS, next page says mFlow: 

![image](https://user-images.githubusercontent.com/19318937/130554515-18db39ca-4018-454c-99fa-8799f2e9599c.png)

![image](https://user-images.githubusercontent.com/19318937/130554524-29c6006d-643c-4f75-88b4-0dd92ae7fd37.png)

CarlosNZ:
> Login screen says IRIMS, next page says mFlow:

Yeah, that's already on the list:

> Change application name on Login page

;)

adamdewey:
D'oh - sorry! 🤦

adamdewey:
![image](https://user-images.githubusercontent.com/19318937/130559091-ed6668bc-cd6e-464f-92c1-18432bddbada.png)

adamdewey:
- [ ] Unwrap Company License section header (or at least make the column a lot wider)

![image](https://user-images.githubusercontent.com/19318937/130717868-8e69059e-9627-41bf-ae4d-c00c79bbcd2b.png)

nmadruga:
> Hide 'No Organisation' from first page of Company Registration application
> 
> ![image](https://user-images.githubusercontent.com/19318937/130559091-ed6668bc-cd6e-464f-92c1-18432bddbada.png)

That was the suggestion from the Designer - so it's clear when you are submitting as a company.
So when there is no company you think it would be better to show empty? That's a very easy fix, should worth a try.

adamdewey:
> That was the suggestion from the Designer - so it's clear when you are submitting as a company.
> So when there is no company you think it would be better to show empty? That's a very easy fix, should worth a try.

[...]

PR #535, opened 2021-08-25
#530 Misc tweaks

nmadruga:
Back-end for https://github.com/openmsupply/application-manager-web-app/pull/933

## Changes
- Just set the schedule actions to run once per day(?)


CarlosNZ (approved):
Cool thanks :)

> Just set the schedule actions to run once per day(?)

It's once per hour currently, configurable in config.ts. I think no less than once per hour, perhaps every half hour is good, but once per hour is fine for now. 

Cheers.

PR #543, opened 2021-08-30
#541 Snapshots - file import fixes

### LINKED ISSUE #541: Inconsistent/buggy file export / import with snapshots
Still getting weirdness with file import/export using snapshots:
- they seem to export okay (can check the output snapshot folder), but when re-importing they get nested a level too deep *inside* the files folder (so you get `/files/files/...`)
- sometimes not restored at all (actually being restored, but then deleted -- see Example 2 below).

I'll try and describe specific reproducible scenarios below. These are all using the latest `develop` branch for both ends.

### Example 1

1. Run a fresh `yarn database_init` so it loads the basic snapshot -- the four default icons are the only things /files
2. From the "templates" repo, copy `demo_2021_08_27` snapshot (from `/demo/snapshots/`) into your `_snapshots` folder, then load it using the Snapshot dev menu in the front-end
3. Note that the files been copied 1 level too deep, so they're in `/files/files` instead of root `/files`)

Same thing seems to happen when exporting/importing templates on their own.

### Example 2

1.  With that same snapshot folder still in `_snapshots`, install it from the command line using `yarn database_init demo_2021_08_27`
2. This time, when it's finished the "files" folder is empty! This is a particularly strange one, because the files are actually copied at the point they're supposed too (I can see them appearing in the folder during `copying files...`), but somehow they're being deleted right at the end of the script by some other process.

CarlosNZ:
Fixes #541

Two problems fixed:
- files were being imported a level too deep (i.e. `/files/files`), fixed so it now copies `/files/*` rather than `/files` from snapshot folder into (existing) files folder
- post insert script was clearing the files folder. Removed this (which means files folder will start to fill up so will need manual management)

### DIFF
--- a/package.json
+++ b/package.json
@@ -50,7 +50,7 @@
     "dev_pg": "concurrently --kill-others-on-fail \"yarn dev\"  \"yarn pg\"",
     "dev_pg_permissions": "concurrently --kill-others-on-fail \"yarn dev\"  \"yarn pg_permissions\"",
     "database_init": "./database/initialise_database.sh tmf_app_manager && ./database/insert_data.sh",
-    "postdatabase_init": "./database/turn_on_row_level_security.sh && ./database/post_data_insert.sh && rm -rf ./files && mkdir ./files",
+    "postdatabase_init": "./database/turn_on_row_level_security.sh && ./database/post_data_insert.sh",
     "build_plugins": "node ./utils/pluginScripts/buildPlugins.js",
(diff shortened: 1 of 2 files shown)


CarlosNZ:
> Good to get into `main` - would you like me to do the merge to main from develop after we merge this and any other pending (Approved) PRs?

Yeah, that'd be good thanks. Then we should be able to give the updated "main" a tag which we can make the latest build from, I think.

PR #544, opened 2021-08-31
Update demo server docs

CarlosNZ:
Removed the _README in ./demo_server and combined its info into existing "demo_server_guide.md"

A bit rushed, but wanted to get this in today. Strongly suggest whoever does a build/server update next make amendments to this guide if need be.


nmadruga (approved):
That's great! Thanks : ) I'll merge this one as well

PR #545, opened 2021-08-31
#926F fix not allow re-assign after completed

nmadruga:
## Linked PR
- Front-end PR https://github.com/openmsupply/application-manager-web-app/pull/943

### Changes
- Add new AssignerAction ASSIGN_LOCKED when all assignable question of an application in stage and level are submitted
- Create new SQL FUNCTION `submitted_assigned_questions_count` to get total questions assigned and submitted
- This is only checking if the current status is submitted TODO> Check ideally should be working for any review submitted at-some-time in review-status-history
- Rename function `template_questions_count ` to `assignable_questions_count `, simplify and also check that there is an application_response associated to be considered as "assignable".

### DIFF
--- a/database/buildSchema/41_application_list_view.sql
+++ b/database/buildSchema/41_application_list_view.sql
@@ -17,12 +17,11 @@ CREATE TABLE application_list_shape (
     reviewers varchar[],
     reviewer_action public.reviewer_action,
     assigner_action public.assigner_action,
-    is_fully_assigned_level_1 boolean,
-    assigned_questions_level_1 bigint,
+    -- is_fully_assigned_level_1 boolean,
+    -- assigned_questions_level_1 bigint,
     total_questions bigint,
     total_assigned bigint,
-    total_assign_locked bigint,
-    level_number int
+    total_assign_locked bigint
 );
 
 CREATE FUNCTION application_list (userid int DEFAULT 0)
@@ -45,16 +44,15 @@ CREATE FUNCTION application_list (userid int DEFAULT 0)
         reviewers,
         reviewer_action,
         assigner_action,
-        CASE WHEN is_fully_assigned_level_1 IS NULL THEN
-            FALSE
-        ELSE
-            is_fully_assigned_level_1
-        END,
-        assigned_questions_level_1,
+        -- CASE WHEN is_fully_assigned_level_1 IS NULL THEN
+        --     FALSE
+        -- ELSE
+        --     is_fully_assigned_level_1
+        -- END,
+        -- assigned_questions_level_1,
         total_questions,
         total_assigned,
-        total_assign_locked,
-        level_number
+        total_assign_locked
     FROM
         application app
     LEFT JOIN TEMPLATE ON app.template_id = template.id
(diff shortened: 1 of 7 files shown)


database/buildSchema/41_application_list_view.sql
```diff
@@ -42,13 +44,15 @@ CREATE FUNCTION application_list (userid int DEFAULT 0)
...
-            is_fully_assigned_level_1
-        END,
-        assigned_questions_level_1,
-        total_questions
+        -- CASE WHEN is_fully_assigned_level_1 IS NULL THEN
+        --     FALSE
+        -- ELSE
+        --     is_fully_assigned_level_1
+        -- END,
+        -- assigned_questions_level_1,
```
CarlosNZ, line 52: Is this meant to be commented? Why don't we need it? (And can you get rid of it if we don't)?

src/components/postgresConnect.ts
```diff
@@ -778,23 +778,23 @@ class PostgresDB {
...
+  //     const result = await this.query({
+  //       text,
+  //       values: [applicationId],
+  //     })
+  //     return result.rows[0].is_fully_assigned_level_1
+  //   } catch (err) {
+  //     console.log(err.message)
+  //     throw err
+  //   }
+  // }
```
CarlosNZ, line 797: Same comment, I guess.

CarlosNZ (commented):
Looks good, thanks. Don't suppose there is a snapshot or something that would demonstrate this in action?

nmadruga:
> Don't suppose there is a snapshot or something that would demonstrate this in action?

You can test with diifferent snapshots - just check if one assigner user can still have the dropdown available when assigner has submitted their review). One workflow that I just tested here, but has a few steps:
- Load `dev` snapshot
- Log in as `Valerio` user and update application serial **23456** and re-submit
- Log in as `Reviewer1` approve re-review and re-submit. Now it's in Assessment ready for assignment
- Log in as `Assigner1` and assign user `assessor1`, after assigning you still have the option to un-assign this user
- Log in as `Assessor1`, do the review and submit
- Log in back as `Assigner1` and you shound't get the action on the list to re-assign or assign and in the dropdown of the ReviewHome page it's showing as disabled:
Screen Shot 2021-09-16 at 10 47 05 AM

database/buildSchema/41_application_list_view.sql
nmadruga: The reason I didn't remove this was that I wasn't sure if we will use this at some point. So would rather keep it, in case we might need to use this, and if not we can remove later.

nmadruga:
@CarlosNZ can you have a look also in PR https://github.com/openmsupply/application-manager-web-app/pull/943 please? So we can merge both once this one is good to go? It might need some testing, but it's mainly using UI.

nmadruga:
Ok to merge this one along with front-end PR @CarlosNZ ?

CarlosNZ:
> Ok to merge this one along with front-end PR @CarlosNZ ?

Sure.

CarlosNZ (approved):

PR #547, opened 2021-09-01
Fix _snapshots folder missing from build

CarlosNZ:
Forgot about this in the last PR. The _snapshots folder was only being created as part of snapshot loading, which means that it wasn't yet present in the "build" folder when trying to upload a snapshot (before loading it).

So manually added "mkdir" to create this folder as part of build process.

Tested with full docker build and snapshot upload and import.

### DIFF
--- a/docker/run.sh
+++ b/docker/run.sh
@@ -5,7 +5,7 @@
 docker run \
     -ti \
     -p 3000:3000 \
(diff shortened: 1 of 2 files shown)


nmadruga (approved):
LGTM

Issue #549, opened 2021-09-02
Template import -- fails if filter and permissionName ids don't match


CarlosNZ:
It seems the filter IDs and the permissionName IDs have to be exactly the same when importing a template using the UI "Import" button.

## Test case

- Load snapshots "demo_2021_08_27" from templates repo
- Import "CompanyLicense-18.zip" from same repo -- it should work fine
- Now load "basic_snapshot" and try same import -- it fails, complaining of foreign-key constraints being broken

This is quite a big problem, as we often want to import a template into a snapshot that has had the permissionNames (and probably filters) changed quite a bit from other snapshots.

It *should* fail if it refers to a permissionName that doesn't exist, but there needs to be a way to be able to link to filters and permissions names by "name" rather than relying on ID.

PR #550, opened 2021-09-02
#549 Skip record with conflicting unique fields values in template import

andreievg:
Fixes: #549 

### Changes

* Changed tablesToUpdateOnInsertFail import/export option to skipTableOnInsertFail (made it backwards compatible with existing snapshots)
* Instead of trying to update on conflict, just query for that record (need id from that record to make sure we insert related record correctly

### Testing

I've tested a few scenarios, import doesn't show any errors, and seems logically correct, would be good to give it another test spin, to make sure correct filter joins and template permission are associated with corresponding filter and permission names, for those that conflict

### DIFF
--- a/database/snapshotOptions/templateExport.json
+++ b/database/snapshotOptions/templateExport.json
@@ -14,7 +14,9 @@
     "permissionName",
     "file"
   ],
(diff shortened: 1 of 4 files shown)


CarlosNZ (approved):
Thanks, Andrei -- tested and working well.

Issue #506, opened 2021-08-12
Misc Snapshot/Template Builder bugs


CarlosNZ:
Just summarising these all in one place as they're probably all fairly minor fixes (I hope):

### Done Sprint 48:
- [x]  ~~Permission Policy selector/display not showing the correct thing~~
![Screen Shot 2021-08-12 at 5 19 15 PM](https://user-images.githubusercontent.com/5456533/129166404-bbcaedb6-07f2-4536-b129-d5c605f71040.png)
- [x] Files not being imported from snapshots. It seems to just be the ones in subfolders -- the 4 default icons seem to come across okay. Here's a snapshot with some logo files to test import: 
[demo_master_02.zip](https://github.com/openmsupply/application-manager-server/files/6974025/demo_master_02.zip)
- [x] ~~False alarm on the Zipped files problem~~
- [x]  ~~Show fields in UI to allow changing: isLinear, namePlural, eventCode (IMPORTANT)~~
- [x] ~~Bug: Open element and change isRequired = **false** and save. Reopen and it gets changes automatically to isRequired = **true**~~
- [x] Bug: templateAction descriptions don't get imported when using import template
- [x] ~~Template builder evaluator UI should allow the "type" property to be set~~
- [x] Small fix - can't set "canMakeDecision" on review level for Final Decision in the Permissions tab on Template builder PR https://github.com/openmsupply/application-manager-web-app/pull/948
- [x] Small fix - fields `scheduledEventCode` getting set with `description` value on UI - doesn't seem to replace real eventCode... PR https://github.com/openmsupply/application-manager-web-app/pull/936
- [x] Small fix - When setting the "Start Message" to empty, it stores an empty string -- NULL would be better (empty string still works so not super critical, it'd just be more consistent with what the front-end expects) PR https://github.com/openmsupply/application-manager-web-app/pull/953
- [x] BUG: Snapshot imports with lookup tables in schema are failing (Add lookup table, export whole snapshot, then re-import => error!)
- [x] Template export is exporting ALL files, should only do the ones associated with this template
- [x] Nice to have - "Add action" at the begin or end of the list to avoid having to click to move all the way to the top PR https://github.com/openmsupply/application-manager-web-app/pull/957
- [x] Show selected page when moving page up/down PR https://github.com/openmsupply/application-manager-web-app/pull/963

CarlosNZ:
> * Need to add ability to add new permissionName
> 
[...]

PR #551, opened 2021-09-02
#506 files to export and import based on file table

andreievg:
Partial fix to #506

Make sure that only files that are linked to file records in import and export are imported and exported (and base files, in root files directory, so that base snapshot also cary over thumbnails)

### Changes
* Helper for getting base files name
* Pass file records from import and export to respective copy files methods
* Iterate over each file path (form file records and base files) and copy over to snapshot or files dir (try catch it all to make sure export and import still succeeds)

### Testing

tested with a few scenarios (new snapshots, snapshots with carbone templates), seems to work, and doesn't remove any existing files

### DIFF
--- a/src/components/exportAndImport/importFromJson.ts
+++ b/src/components/exportAndImport/importFromJson.ts
@@ -12,10 +12,16 @@ import {
   ObjectRecords,
 } from './types'
 
-const insertFromObject = async (
+type InsertFromObject = (
   records: ObjectRecords,
-  { includeTables, excludeTables, skipTableOnInsertFail = [] }: ExportAndImportOptions,
-  preserveIds: boolean = false
+  options: ExportAndImportOptions,
+  preserveIds?: boolean
+) => Promise<InsertedRecords>
+
+const insertFromObject: InsertFromObject = async (
+  records,
+  { includeTables, excludeTables, skipTableOnInsertFail = [] },
+  preserveIds = false
 ) => {
   const databaseTables = (await getDatabaseInfo()).filter(({ isView }) => !isView)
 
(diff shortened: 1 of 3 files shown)


CarlosNZ:
Cool. One question -- how does file import handle "merging" folders?

So say if we already have "/files/carbone_templates" with some files in it, and we import a template with additional files in its own "/files/carbone_templates" folder, will it successfully merge the contents of "carbone_templates", or will the original subfolder be over-written?

CarlosNZ:
Small additional request if it's not too much trouble:

I recently removed a  command in the post-insert script that deleted the contents of the files folder, as it was deleting files after snapshot import, which obviously wasn't useful. Problem is, now the files folder tends to fill up with garbage as it never gets cleaned out unless we do it manually.

Would it be possible to add an additional "options" field to snapshots called "resetFiles" or something, and if true it deletes everything in the files folder? Then we'd set this to "true" for full system snapshots, and "false" for template exports.

CarlosNZ:
> Small additional request if it's not too much trouble:

Actually, I can probably do this fairly easily, will have a crack later today.

CarlosNZ:
> (and base files, in root files directory, so that base snapshot also cary over thumbnails)

I don't know that this is necessary. I thought there was already code to copy the base thumbnails into "files" on server start (there definitely used to be), so no need to put these in with snapshots (particularly template export ones)

CarlosNZ (approved):
Tested, works great, thanks.

I might try and squeeze that additional file reset request into this, but feel free to merge anyway and I'll add it later.

andreievg:
> I don't know that this is necessary. I thought there was already code to copy the base thumbnails into "files" on server start (there definitely used to be), so no need to put these in with snapshots (particularly template export ones)

I think if you add the file reset, this would need to be in. (as server doesn't restart on snapshot loading anymore)

andreievg:
> Actually, I can probably do this fairly easily, will have a crack later today.

That would be good, also to get some familiarity with snapshot code

CarlosNZ:
> I think if you add the file reset, this would need to be in. (as server doesn't restart on snapshot loading anymore)

Yeah, good point.

PR #554, opened 2021-09-03
#553 Snapshots -- clear files option

### LINKED ISSUE #533: New external user can see all users / orgs
At the moment a new external user can see all users / orgs. 

image

Do we need to do something to restrict these?

### LINKED ISSUE #553: Snapshots: option to clear files folder on import
As per this comment: https://github.com/openmsupply/application-manager-server/pull/551#issuecomment-912081500

> I recently removed a  command in the post-insert script that deleted the contents of the files folder, as it was deleting files after snapshot import, which obviously wasn't useful. Problem is, now the files folder tends to fill up with garbage as it never gets cleaned out unless we do it manually.
> 
> Would it be possible to add an additional "options" field to snapshots called "resetFiles" or something, and if true it deletes everything in the files folder before importing the snapshot files? Then we'd set this to "true" for full system snapshots, and "false" for template exports

CarlosNZ:
Fixes #553  

Added the option "resetFiles", and added it to existing "snapshotOptions" files:
- for full system snapshots ("default"), files folder is cleared on import
- for templates, individual templates, application data, files folder is NOT cleared on import

For existing snapshots (that don't have this parameter in their options file), they WON'T clear files (i.e. default: false)

We should add this option to current demo snapshot.

(Fixed a spelling error in "initialiseDatabase" function name as well)

### DIFF
--- a/database/basic_snapshot/options.json
+++ b/database/basic_snapshot/options.json
@@ -1,17 +1,16 @@
 {
- "filters": {},
- "includeTables": [],
- "excludeTables": [
-  "actionQueue",
-  "triggerQueue",
-  "applicationListShape",
-  "elementTypePlugin",
-  "notification"
- ],
- "shouldReInitialise": true,
- "insertScriptsLocale": "dev",
- "includeInsertScripts": [
-  "none"
- ],
- "excludeInsertScripts": []
+  "filters": {},
+  "includeTables": [],
+  "excludeTables": [
+    "actionQueue",
+    "triggerQueue",
+    "applicationListShape",
+    "elementTypePlugin",
+    "notification"
+  ],
+  "shouldReInitialise": true,
+  "insertScriptsLocale": "dev",
+  "includeInsertScripts": ["none"],
+  "excludeInsertScripts": [],
+  "resetFiles": true
 }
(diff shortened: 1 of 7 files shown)


nmadruga:
> Fixes #533

Fixes #553 ?

CarlosNZ:
> > Fixes #533
> 
> Fixes #553 ?

Yes 😳

nmadruga:
Good to have this added to some domentation on Snapshots? If we don't have one yet, can we start now?

nmadruga (changes requested):
Testing:

- Have created a new snapshot and it good this option set by default: "resetFiles": **true** is that correct? Shouldn't it be **false**?

CarlosNZ:
> Have created a new snapshot and it good this option set by default: "resetFiles": **true** is that correct? Shouldn't it be **false**?

Yes, that is correct. Full snapshots (whole system) use the "default.json" options, and we want "resetFiles" to be true for those. It's "false" for all the others (templates etc).

CarlosNZ:
> Good to have this added to some documentation on Snapshots? If we don't have one yet, can we start now?

There is a "Snapshots.md" page, but it looks way out of date in terms of snapshot functionality so will need a substantial re-write. That'll be quite a big job, so I've made an issue for it: https://github.com/openmsupply/application-manager-server/issues/555

nmadruga (approved):

Issue #557, opened 2021-09-06
Snapshot - Errors in snapshot import with products


andreievg:
As per snapshot shared by Carl `outcomes_with_products`

* Can't import snapshot with ingredient list object in outcome table (it's a JSON array, snapshot can handle JSON but not json array)
* Product application join is not imported (schema patch in snapshot would fail silently and .sql file generated when taking snapshot is not ordered by reference, and since product application join table creation appears first in that file, creation of that table fail silently and subsequently is not imported)

PR #558, opened 2021-09-06
#557 fix snapshot import with product outcomes

andreievg:
Fixes: #557 

Hopefully commit title and content should explain the changes


CarlosNZ:
Cool -- will re-visit this tomorrow, it's not actually essential for demo server, since we'll no longer be uploading whole snapshots (just template updates).

nmadruga (approved):
LGTM - worked perfect to import the latest snapshot downloaded from demo 50002 in my local instance

PR #560, opened 2021-09-09
#865F Adding send notifications to applicant in basic_snapshot for APPROVE/REJECT and LOQ 

nmadruga:
Just changes to basic_snapshot with fix to changeOutcome core-action


CarlosNZ (approved):
LGTM :)

PR #563, opened 2021-09-09
#966F sessionId handling

CarlosNZ:
Back-end changes required for sessionId handling in the front-end (Front-end PR [#975]( https://github.com/openmsupply/application-manager-web-app/pull/975))

### Summary

- made sure `sessionId` field was passed through if present in "login" and "userInfo" endpoints, and the return object contains the same sessionId if present (or a random one if not, as before)
- added "sessionId" to "applicationData" (for back-end) so can be accessed in Actions


nmadruga (approved):
LGTM

PR #571, opened 2021-09-10
Portuguese strings v1

nmadruga:
PR for front-end https://github.com/openmsupply/application-manager-web-app/pull/979

### Changes:
Add local: pt-pt to generatePDF so dates show in the right format

Screen Shot 2021-09-10 at 6 18 16 PM

### DIFF
--- a/plugins/action_generate_document/src/generateDoc.ts
+++ b/plugins/action_generate_document/src/generateDoc.ts
@@ -8,7 +8,7 @@ const generateDoc: ActionPluginType = async ({
   // DBConnect,
   outputCumulative,
 }) => {
-  const { docTemplateId, additionalData } = parameters
+  const { options, docTemplateId, additionalData } = parameters
   const data = parameters?.data ?? { ...applicationData, ...outputCumulative, additionalData }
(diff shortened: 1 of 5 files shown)


CarlosNZ:
Looks cool :) 

Are you still having trouble with the Date formatting? (Is that why it's circled?)

nmadruga:
> Looks cool :)
> 
> Are you still having trouble with the Date formatting? (Is that why it's circled?)

No, I sorted that out - just circled to dislay what was this about. I think this is good to merge since it's defined per action/template

plugins/action_generate_document/src/generateDoc.ts
```diff
@@ -18,6 +18,9 @@ const generateDoc: ActionPluginType = async ({
     const result = await generatePDF({
       fileId: docTemplateId,
       data,
+      options: {
+        lang: 'pt-pt',
+      },
```
CarlosNZ, line 23: Doesn't this need to be a parameter, not hard-coded?

nmadruga: Yes - good point. Done in a hurry. I'll fix today

Issue #566, opened 2021-09-09
Implement outcomes/table/item endpoint


CarlosNZ:
According to spec: https://github.com/openmsupply/application-manager-server/issues/556

- based from #565

Issue #565, opened 2021-09-09
Implement /outcomes/table endpoint


CarlosNZ:
According to spec: https://github.com/openmsupply/application-manager-server/issues/556

- based from #564

Issue #556, opened 2021-09-06
Outcome improvements


CarlosNZ:
I originally started by commenting on a previous issue [here](https://github.com/openmsupply/application-manager-web-app/issues/903#issuecomment-899958966), but I thought it would be easier to make a new issue summarising the *requirements* of the Outcomes functionality in one place before we figure out an implementation.

Here are the things I think we need:

- ease of config: A new "Outcome" (i.e. a <outcome>_application_join table is created), should automatically have an entry in the outcomes list and be able to display basic data in table and detail view without any config. Config should just be for refining/improving/filtering the displays
- simpler way to define what columns show up in table and details view (include/exclude lists)
- robust against config errors (missing fields, etc.)
- default "text" display -- either the "text" field of the stored value, or the text itself. (and should have some way to respect line breaks, either Markdown or interpret `\n` correctly)
  - should be a default display for all common types, including Date, so can display in table without needing a SummaryView component (dates are the main use case for this, since ISO strings are no good in UI)
- permissions-based "layouts" -- groups of columns (as per above include/exclude lists) that are associated with permissionsNames. E.g for "User" outcomes, Admin might see all columns, whereas a regular user only sees name and username. Or can only see users associated with their own company.
- more flexibility over what can be displayed in a single column, i.e. we should be able to combine columns, for example, if we have 3 address fields (street, town, district), they should be able to be displayed in a single "address" fields combined. To do this we'd need access to more than just the current field value. My suggestion: when saving an outcome record, we also have an additional column which saves the whole "responsesByCode" object as JSON, and this is passed to display components for object lookups
- more flexibility in querying values for columns (not just fields in the outcome table). For example, in the Organisations outcomes, we should be able to display a list of users associated with the organisation in the outcome details, which would require a *query* in a column definition, not just a field. I like the idea of using the "SummaryView" components as the basis for displaying outcome details, but it'd be good to have more control over what we can send to a component.
- Should be able to export/import Outcome configs with templates (they really are an integral part of a template -- how to display the collected data), and we may want to have tighter associates between them (such as template_id in outcome_display table, or maybe just "code"?)
- Table list should be searchable, sortable and filterable, just like Applications list (in fact, we should be using a lot of the same core functionality if possible)
- Layout improvements:
  - table view should be basically the same as the Applications List (including filters, etc.)
  - detail view should be basically the same as application Summary page (vertical list, one component per line)
- (eventually) a UI to configure all the above, but let's get the design and implementation sorted first

[...]

PR #573, opened 2021-09-13
#564 #565 #566 Outcomes API

### LINKED ISSUE #564: Outcomes -- update schema and implement initial /oucomes endpoint
According to spec: https://github.com/openmsupply/application-manager-server/issues/556

CarlosNZ:
Fixes #564 #565 #566

Implements API for Outcomes as per spec details in #556 

Note: **Don't** use this with front-end yet, it won't work. This is just the API, so can only interact with endpoints.
(**Update:** there is now a [front-end PR](https://github.com/openmsupply/application-manager-web-app/pull/983), but suggest you still review this as stand-alone using the Postman collection)

Here is a Postman collection to demonstrate and test: 
[Outcomes.postman_collection.json.zip](https://github.com/openmsupply/application-manager-server/files/7150952/Outcomes.postman_collection.json.zip)

Please use the snapshot "outcomes_demo". I have put in the "dev/snapshots" folder in the templates repo on a new branch called `#outcomes-demo`

The changes are summarised in detail in the [original issue](#556). (I'll use that spec as the basis for documentation)

Please test:
- note the different results you get with different JWTs (Postman has Admin and nonRegistered JWT examples)
- please mess around with different tables etc, and try changing the query parameters (first/offset, etc.)
- have a mess around with the configs on "outcome_display" and "outcome_display_column_definitions" tables and see the results
- note that when you look at an Organisation details view, you also get a list of Users ("Members" field). (Unfortunately it's not really in a format we could display easily just yet, but that can be revisited when doing front-end)

TO-DO (once front-end is further along):
- more refinement with regards to the "formatting" options
- implement "sort" and "filter" queries. Left it for now, as not sure if front-end should build the filter objects (as per the Applications list), or if they should just be query parameters and we process the filter construction here
- it occurs to me it won't be possible to search or sort by any of the "custom" columns, since they only get generated for the fetched records, so you won't be sorting/searching through the whole possible set. (Suggested partial solution, but low priority -- specify a "sort field" in each custom column that corresponds to a database field. For example, in the "Full Name" column, the "sort field" could be "lastName", and so sorting by "full name", orders it by last name. I dunno, have a think).

### DIFF
--- a/tsconfig.json
+++ b/tsconfig.json
@@ -4,7 +4,7 @@
 
     /* Basic Options */
     // "incremental": true,                   /* Enable incremental compilation */
-    "target": "es5" /* Specify ECMAScript target version: 'ES3' (default), 'ES5', 'ES2015', 'ES2016', 'ES2017', 'ES2018', 'ES2019', 'ES2020', or 'ESNEXT'. */,
+    "target": "es6" /* Specify ECMAScript target version: 'ES3' (default), 'ES5', 'ES2015', 'ES2016', 'ES2017', 'ES2018', 'ES2019', 'ES2020', or 'ESNEXT'. */,
     "module": "commonjs" /* Specify module code generation: 'none', 'commonjs', 'amd', 'system', 'umd', 'es2015', 'es2020', or 'ESNext'. */,
     "lib": ["es2019"] /* Specify library files to be included in the compilation. */,
     "allowJs": true /* Allow javascript files to be compiled. */,
--- a/database/basic_snapshot/options.json
+++ b/database/basic_snapshot/options.json
@@ -1,16 +1,18 @@
 {
-  "filters": {},
-  "includeTables": [],
-  "excludeTables": [
-    "actionQueue",
-    "triggerQueue",
-    "applicationListShape",
-    "elementTypePlugin",
-    "notification"
-  ],
-  "shouldReInitialise": true,
-  "insertScriptsLocale": "dev",
-  "includeInsertScripts": ["none"],
-  "excludeInsertScripts": [],
-  "resetFiles": true
+ "filters": {},
+ "includeTables": [],
+ "excludeTables": [
+  "actionQueue",
+  "triggerQueue",
+  "applicationListShape",
+  "elementTypePlugin",
+  "notification"
+ ],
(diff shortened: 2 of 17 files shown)


tsconfig.json
```diff
@@ -4,7 +4,7 @@
 
     /* Basic Options */
     // "incremental": true,                   /* Enable incremental compilation */
-    "target": "es5" /* Specify ECMAScript target version: 'ES3' (default), 'ES5', 'ES2015', 'ES2016', 'ES2017', 'ES2018', 'ES2019', 'ES2020', or 'ESNEXT'. */,
+    "target": "es6" /* Specify ECMAScript target version: 'ES3' (default), 'ES5', 'ES2015', 'ES2016', 'ES2017', 'ES2018', 'ES2019', 'ES2020', or 'ESNEXT'. */,
```
nmadruga, line 7: Hmmm not sure if this could be meaningful on deployment. Any reasons for changing this now?

nmadruga (commented):
The code is indeed very extensive, but looks good. I didn't look with much details, just trying to get an overall picture. I think some documentation done now would be pretty important, so we have something to base on for future changes.

My feeling from this was:
- You already prepared the environment for `restrictions` using that on what is fetched back to the front-end when building the oucomesTable and outcomesDetails
- Now back-end is responsible for most of the logic on configuring what is displayed and do the evaluation and build different blocks depending on types.
- There are already some filtering, pagination and sorting included - to be continued in another PR or Issue?
- The tables aren't yet constructed automatically when a new entity is added (to be done in another issue ?)

I'll start doing some tests now and will approve this after if all seems good.

Thank you for putting this together, is looking great. and hopefully, we can add this to `main` branch soon when adding the remaining restriction that we are planning now.

Could you quicly just add a basic documentation on how it's currently working so we can have it on our wiki please?

tsconfig.json
CarlosNZ: One of the Set methods wasn't being allowed with ES5. 

I've already merged the front-end for this, so better merge this, but I'll make a note to check this again in develop to make sure it doesn't cause problems. 

Should be okay for back-end though, since we control the execution environment. Would be more risky in front-end since we don't know what browsers people will be using.

PR #574, opened 2021-09-13
#475 - Reassignment of reviews

### LINKED ISSUE #475: Add new logic for Re-assignment specs
If we don't consider Draft as personal (for reviewers) we can simply re-link everything (review, assignments and response) to new reviewer.
Using a helper graphql mutation exposed.

Diagrams:
Screen Shot 2021-09-09 at 9 23 02 PM

Screen Shot 2021-09-09 at 9 23 16 PM

nmadruga:
Fixes #475

### Linked PRs
- Front-end PR https://github.com/openmsupply/application-manager-web-app/pull/982
- Templates to use `dev/snashots/current` on branch `reassignment-changes` in application-manager-templates

### Changes
Update code to use ASSIGNED or AVAILABLE for reviewAssignment status 
Use `reviewAssignment.isSelfAssignable` to mark reviews that can be assigned by user
Update actions `generateReviewStatus` (quite a bit change also aiming to make the code looks simpler) and `updateReviewAssignmanetStatus` - work to be continued in issue #577 where we could use the same action to update the un-assigned reviewAssignment.


CarlosNZ (changes requested):
Problems noticed when testing:

- Unassign does nothing (I think you haven't got to that yet, right?)
- I thought once the review had been started, that it was no longer available for re-assignment (in this PR). However it still, is but it causes an error:
Screen Shot 2021-09-21 at 1 21 03 PM
- Also -- after that, I was able to start a review as "Assessor 2", but when I went back to the Review home page it showed as "Locked" and I couldn't continue.

UI Comments
- The "reassign" interface is weird. There shouldn't be a drop-down for "Unassign from", as that's not a choice, it's just information:
Screen Shot 2021-09-21 at 1 24 23 PM
- I don't this was you, but the Heading on the review page has become crazy big. Can you quickly set it back to h2 or h3, probably?
Screen Shot 2021-09-21 at 1 23 47 PM

nmadruga:
> * Unassign does nothing (I think you haven't got to that yet, right?)

Just replied in [here](https://github.com/openmsupply/application-manager-web-app/pull/982#issuecomment-923749852), the dropdown option no and the action to also un-assign review now in PR #577

nmadruga:
> * I thought once the review had been started, that it was no longer available for re-assignment (in this PR). However it still, is but it causes an error:
>   Screen Shot 2021-09-21 at 1 21 03 PM

No, as discussed in Epic#89 it's still allowed to transfer the review after it has started, just not allowed after submitted.

nmadruga:
> * Also -- after that, I was able to start a review as "Assessor 2", but when I went back to the Review home page it showed as "Locked" and I couldn't continue.

Ok - I think I used the wrong status to set `isLocked` to a review, have done some improvement on that in next PR #577. Please have a look and see if you see that happening in there.

nmadruga:
> * The "reassign" interface is weird. There shouldn't be a drop-down for "Unassign from", as that's not a choice, it's just information:
>   Screen Shot 2021-09-21 at 1 24 23 PM

I think the interface will be definitely improved afterward. Not sure how to deal with having the option to re-assign inside the dropdown either... Suggestions? After you click on re-assign, what happens?

nmadruga:
> * I don't this was you, but the Heading on the review page has become crazy big. Can you quickly set it back to h2 or h3, probably?
>   Screen Shot 2021-09-21 at 1 23 47 PM

Yes. Will do this now. Good idea

nmadruga:
Will also be combining this with next PR

PR #582, opened 2021-09-17
#575 Snapshot thumbnail export/import fix

### LINKED ISSUE #575: Bug fix -- snapshot export not copying thumbnail files.

CarlosNZ:
Fix #575 

Adds an extra line in both import and export to also add the list of thumbnail paths into the files to copy array

Note: Slightly sub-optimal, as it will copy the "generic" icons many times, but it's probably fine?


CarlosNZ:
FYI -- it took about 15 mins to do this PR, and over an hour to figure out a solution to batch re-create all the thumbnails for the Angola snapshot 😆

PR #584, opened 2021-09-20
#583 fix date of birth field

### LINKED ISSUE #583: Date of birth field in "User" table is varchar -- should be timestamptz

CarlosNZ:
Fix #583


nmadruga (approved):
It used to be date before - but because we were using the `modifyRecord` to update that it was only able to receive strings before. Goog to have this sorted now.

PR #585, opened 2021-09-20
#576 Increase file upload limit for server

### LINKED ISSUE #576: Add config to nginx to upload files bigger than 7MB as required for Portuguese demo when snapshot file can be a bit bigger
Need to update the scripts and make sure it works on the docker instance.

Andrei explanation on the fix done to the script already in the demo-server:
Andrei, [14.09.21 15:11]
yeah it’s the limit on nginx inside container
https://github.com/openmsupply/application-manager-server/blob/2310f8bd5f20d5f996c8bfa7e6d255c7162c95fc/docker/nginx_config/application_manager.conf#L5

Andrei, [14.09.21 15:11]
can solve by:
* increment in above config and rebuild/deploy
* or can expose that setting in a volume

that’s the location inside container:https://github.com/openmsupply/application-manager-server/blob/2310f8bd5f20d5f996c8bfa7e6d255c7162c95fc/docker/Dockerfile#L87

added a line in server nginx -> sudo nano /etc/nginx/nginx.conf, client_max_body_size 20M;, saved file and sudo service nginx restart <- this will persist
exec into container `sudo docker exec -ti mflow-on-8008_app_1 /bin/bash`, then edited nginx config inside container nano /etc/nginx/conf.d/application_manager.conf, changed client_max_body_size 7M; to client_max_body_size 20M;, saved filed, then service nginx restart <- this won’t persist (if container is restart)

nginx routes all traffic, and it has a settings to restrict maximum upload size, nginx sits inside docker container (to route all traffic through one port, i.e. both server, graphql and front end are all on the same port and are accessed with different url route), and another nginx is used on server to route between docker containers and do TSL (ssl)

CarlosNZ:
Fix #576 

Made it 500M, so it should be safe for a wee while.

### DIFF
--- a/docker/nginx_config/application_manager.conf
+++ b/docker/nginx_config/application_manager.conf
@@ -2,7 +2,7 @@ server {
      listen 3000;
      # gzip front end bundle
      gzip_types text/css application/javascript; 
-     client_max_body_size 7M;
+     client_max_body_size 500M;
      # this requires postgraphile to be started with there parameters: -q '/postgraphile/graphql' -i '/postgraphile/graphiql'
      location /postgraphile/ {
           proxy_pass http://127.0.0.1:5000/postgraphile/;


nmadruga (approved):
LGTM

PR #586, opened 2021-09-21
#577 Actions triggered on Re-assignment 

### LINKED ISSUE #577: Configure action to run on assignment or re-assignment to set previous Review status
re-assigned (from) should lock review (if existiing) in back-end and re-assigned (to) should create/unlock previous review

nmadruga:
Fixes #577 

## Linked PRs
- Front-end https://github.com/openmsupply/application-manager-web-app/pull/994
- Templates to use `dev.current` on branch `reassignment-changes` in application-manager-templates

### Changes
- Update code to use ASSIGNED or AVAILABLE for reviewAssignment status 
- Use `reviewAssignment.isSelfAssignable` to mark reviews that can be assigned by user
- Update actions `generateReviewStatus` (quite a bit change also aiming to make the code looks simpler) and `updateReviewAssignmanetStatus` - work to be continued in issue #577 where we could use the same action to update the un-assigned reviewAssignment.
- Added new method to retrieve review from reviewAssignmentId passed so ReviewData is present when the trigger is set in table `review_assignmen`
- Added for reference in `core-mutations` new core-action that have been added in `dev` (and should also be added to **basic_snapshot**) and other templates when this PR is merged. To allow the re-assignment to work as expected.
- New review.status of DISCONTINUED to be used when the review was un-assigned from this Reviewer.


nmadruga:
Just copying here what was suggested for changes required in previous #475 PR (close to be combined in one):

> Problems noticed when testing:
> 
> * Unassign does nothing (I think you haven't got to that yet, right?)

In a PR soon (already has front-end partially done) but should be ok to review after this is merged (not connected, just dependent on things changed in here)

> * I thought once the review had been started, that it was no longer available for re-assignment (in this PR). However it still, is but it causes an error:
>   Screen Shot 2021-09-21 at 1 21 03 PM

Probably fixes after the latest changes. Let me know if you still see this happening.

> * Also -- after that, I was able to start a review as "Assessor 2", but when I went back to the Review home page it showed as "Locked" and I couldn't continue.

Also I think this was due to unfinished changes to previous PR.

> UI Comments
> 
> * The "reassign" interface is weird. There shouldn't be a drop-down for "Unassign from", as that's not a choice, it's just information:
>   Screen Shot 2021-09-21 at 1 24 23 PM

Doing

> * I don't this was you, but the Heading on the review page has become crazy big. Can you quickly set it back to h2 or h3, probably?
>   Screen Shot 2021-09-21 at 1 23 47 PM

Done

database/insertData/dev/07_review_assignments.js
```diff
@@ -435,6 +435,7 @@ exports.queries = [
           stageNumber: 2
           levelNumber: 2
           isLastLevel: true
+          isSelfAssignable: true
```
CarlosNZ, line 438: Do you think we still need to bother maintaining these? 

There's a few template features in Feature Showcase I think that could do with being pulled into the main snapshot, but apart from that, we may as well deprecate these scripts, do you think?

plugins/action_update_review_assignments_status/src/updateReviewAssignmentsStatus.ts
```diff
@@ -21,29 +23,40 @@ async function updateReviewAssignmentsStatus({
       level_number: reviewLevel,
     } = await db.getReviewAssignmentById(reviewAssignmentId)
 
-    const otherReviewAssignments = await db.getMatchingReviewAssignments(
-      reviewAssignmentId,
-      applicationId,
-      stageNumber,
-      reviewLevel
-    )
+    console.log('trigger', trigger)
```
CarlosNZ, line 26: Remove?

src/components/actions/getApplicationData.ts
```diff
@@ -40,11 +41,19 @@ export const getApplicationData = async (input: {
   const reviewId =
     input?.reviewId ?? (trigger_payload?.table === 'review' ? trigger_payload?.record_id : null)
 
+  const reviewAssignmentId =
```
CarlosNZ, line 44: Can you update the `ActionApplicationData` type in types.ts -- I don't' think it quite corresponds to this structure now.

src/components/graphQLConnect.ts
```diff
@@ -64,6 +64,33 @@ class GraphQLdb {
...
+            reviewAssignment {
+              isLocked
+            }
+          }
+        }
+      }
+      `,
+      { reviewAssignmentId }
+    )
+    return data.reviews.nodes[0] || null
```
CarlosNZ, line 91:
```suggestion
    return data?.reviews?.nodes[0] || null
```
Necessary?

CarlosNZ (approved):
Wow, this is pretty epic. Thanks for getting it sorted.

Couple of tiny comments.

I'll test out the front-end now.

database/insertData/dev/07_review_assignments.js
nmadruga: Yes, I think that should be opk to do. I was only afraid we could need to start from scratch again. But I think that keeping previous versions of snapshots is safe. So yes, I can just delete these folders previously used to insert using js files.

plugins/action_update_review_assignments_status/src/updateReviewAssignmentsStatus.ts
nmadruga: Ok

database/insertData/dev/07_review_assignments.js
nmadruga: Ok - in order to keep `insertData.ts` file (which can be used as part of `snapshotCLI` insert) I had to keep the folder `insertData/_common` in, so deleting all of the others...

CarlosNZ: I'm not sure we want to delete them all quite yet, just that we don't need to keep maintaining them. I think there's some stuff in Feature Showcase I'd like to get at some point, so good to have that still around.

nmadruga:
Trying to update this with latest changes on `develop`... Taking longer than expected.

src/components/actions/getApplicationData.ts
nmadruga: Nothing changed in the returned values. Is just allowing the action to run when received from event on `reviewAssignment` table, that's all.

database/insertData/dev/07_review_assignments.js
nmadruga: Oh.... ok. Reverting

PR #591, opened 2021-09-24
Localisation prototype

CarlosNZ:
Back-end  for Localisation prototype. Just has the routes and existing localisation files.

See front-end for more details: https://github.com/openmsupply/application-manager-web-app/pull/1004


andreievg:
looks good thanks !

localisation/languages.json
```diff
@@ -0,0 +1,16 @@
+[
+  {
+    "languageName": "English",
+    "description": "NZ English (default)",
+    "code": "default",
```
nmadruga, line 5: Is this something defined in here? Or this is not in use yet? I thought that the default language would be something defined per project.

src/components/localisation/routes.ts
```diff
@@ -0,0 +1,18 @@
+// import languages from '../../../localisation/languages.json'
```
nmadruga, line 1: Can be removed?

src/components/localisation/routes.ts
```diff
@@ -0,0 +1,18 @@
...
+  reply.sendFile('languages.json', path.join(getAppEntryPointDir(), localisationsFolder))
+}
+
+export const routeGetLanguageFile = async (request: any, reply: any) => {
+  const { code } = request.params
+  const stringsFile = path.join(code, 'strings.json')
+  reply.sendFile(stringsFile, path.join(getAppEntryPointDir(), localisationsFolder))
+}
+
+export const updateLanguageFiles = async (request: any, reply: any) => {}
```
nmadruga, line 18: Hmmm is there a plan for this one? Not sure why e would use an update here?

nmadruga (commented):
Just comments for now - will have a look on front-end

src/components/localisation/routes.ts
CarlosNZ: Because at some point we need a mechanism to write language files back to the server, so this was just meant as a placeholder until we have a spec for how that will work. I figured we'd build the API for it before the UI.

localisation/languages.json
CarlosNZ: "default" just means -- use the "master" strings stored in the front-end, which will always be English.

But I've just realised I haven't made "disabling" the default actually work (disabling others does), so I should fix that (in front end)

CarlosNZ: But no matter what, the "default" will always be the fallback when strings are missing. (unless we implement a cascading fallback system like I commented in the front-end PR, but that's actually not that trivial to do in an efficient way)

CarlosNZ: Okay, I've given this a bit more thought, and I think this is how is should work. See if you agree:

- the "default" is the front-end strings, and these are ONLY used when keys are missing from Language files
- so even an "English" language file will still be fetched from server, even if it has identical content to the "default" set in front-end (though we can over-ride specific terms if we want to)
- that way "English" won't be "special" in any way (other than being the fallback), and will just be another editable language file on the server.
- So there won't be a "default" language code, "default" will just be "fallback".

CarlosNZ: Done 92d277f3  (and 0471f59e in front-end)

Makes much more sense now, thanks for pointing it out.

src/components/localisation/routes.ts
CarlosNZ: Done

localisation/languages.json
CarlosNZ: It'll start with whatever language is set as the default in preferences. It's currently in `config.ts` in the front-end, but I've moved it to a common "preferences.json" in the [later PR](#595 )

But you can still switch language from the login screen before you even start.

nmadruga (approved):
All good! Great work!

PR #593, opened 2021-09-26
Documentation for Outcome configuration

CarlosNZ:
Explains the API, and how to configure outcome displays

- also removed old outcome tables from schema that were marked for deletion


documentation/Outcomes-Display.md
```diff
@@ -0,0 +1,427 @@
...
+- `priority`: when multiple layouts match the request (i.e. user has permissions for more than one Layout), the returned columns will be a union of the columns specified in both layouts. However,there will still only be one "Title" and "Header column" returned and each layout may have a different one, so the layout with the highest priority will be used. In general you'd want to give the more "restrcited" permission the higher priority, but most of the time you won't need to consider this as the Title and Header would often be the same. Default value: `1`.
+
+Okay, so if you're just wanting to display fields directly taken from the outcome table in question, and the formatting requirements are all "simple" types (text, number, boolean, Date) the `outcomes_display` table is all you need. However, if you want to return columns with more complex data (such as a list of ingredients, or a query to another table) or require non-default formatting, you'll need to define these in the `outcome_display_column_definition` table.
+
+### `outcome_display_column_definition` table
```
nmadruga, line 229: I'm a little lost in the structure here (maybe a diagram would help). Let me check this is accurate:
- `outcome_table` is defining the `displayDefinitions` which can be defining a title per column  for each column displayed in the list of entities to be displayed for this outcome.
- `outcomes_display ` specify how the column header  to show in a column when displaying one entity using `detail_view_header_column`
  just be using the field name (as my query before - which can be displayed as  Camel case with spaces...).
- `outcome_display_ column_definition` can also define more fine grained header to show in a column when displaying one entity.

documentation/Outcomes-Display.md
```diff
@@ -0,0 +1,427 @@
...
+
+As mentioned, this table is for defining how to return columns with specifications beyond that which can be defined in the `outcomes_display` table.
+
+The input fields are as follows:
+
+- `table_name` (as above)
+- `column_name`: must match with one of the column names specified in the inclusion lists above.
+- `title`: title to show for the field name. If not specified, it will just be the column_name in "Start Case".
+- `element_type_plugin_code`: if the value being returned is a whole application response, then it will parsed and displayed using a front-end element plugin (SummaryView). This is where you specify the code of the plugin required. If `null`, value won't be sent to a SummaryView element for display.
```
nmadruga, line 238:
```suggestion
- `element_type_plugin_code`: if the value being returned is a whole application response, then it will be parsed and displayed using a front-end element plugin (SummaryView). This is where you specify the code of the plugin required. If `null`, value won't be sent to a SummaryView element for display.
```

documentation/Outcomes-Display.md
```diff
@@ -0,0 +1,427 @@
...
+- `element_parameters`: if a plugin is specified above, then it will require input parameters. This value should be an object containing all the parameters required for display. Values of parameters can be literals or evaluator queries, and will be evaluated by the front-end in SummaryView just like a normal form element.
+- `additional_formatting`: a custom object specifiying additional formatting definitons. Detailed explanation below.
+- `value_expression`: if `null`, the returned value will just be the value of the field from the table in question. However, if you are defining a custom column, or you want to over-ride the value of the field, this field should contain an evaluator expression. The "object" passed into the evaluator for "objectProperties" operator is the current record (item) represented as key-value pairs. So, for example, if you wanted to return a custom field from the "user" table that combined both first and last names (column name: `fullName`), your `value_expression` query would be:
```
nmadruga, line 241:
```suggestion
- `value_expression`: if `null`, the returned value will just be the value of the field from the table in question. However, if you are defining a custom column, or you want to over-ride the value of the field, this field should contain an evaluator expression. The "object" passed into the evaluator for "objectProperties" operator is the current record (item) represented as key-value pairs. 
   - For example, if you wanted to return a custom field from the "user" table that combined both first and last names (column name: `fullName`), your `value_expression` query would be:  
```

documentation/Outcomes-Display.md
```diff
@@ -0,0 +1,427 @@
...
+      "children": ["firstName"]
+    },
+    {
+      "operator": "objectProperties",
+      "children": [ "lastName"]
+    }
+  ]
+}
+```  
+`firstName` and `lastName` are native fields on the "user" table, and are at the root level of the passed-in object here. Note that this object contains *all* fields, not just the ones being returned, so you can extract values from fields not being returned, such as in this case (where you wouldn't want to return `firstName` and `lastName` fields as well as a `fullName` column.)
```
nmadruga, line 258:
```suggestion
`firstName` and `lastName` are native fields on the "user" table, and are at the root level of the passed-in object here. Note that this object contains *all* fields, not just the ones being returned on `outcome_display. table_view_include_columns` definition, so you can extract values from fields not being returned, such as in this case (where you wouldn't want to return `firstName` and `lastName` fields as well as a `fullName` column.)
```

documentation/Outcomes-Display.md
```diff
@@ -0,0 +1,427 @@
...
+
+### Additional formatting
```
nmadruga, line 260: Not sure to which table this relates to...

documentation/Outcomes-Display.md
```diff
@@ -0,0 +1,427 @@
...
+- `permissionNames`: an array of permissions names that the request requires (via JWT) in order to see this Layout. Default is an empty array, and both that and `null` allow full access (i.e. no restrictions)
+- `table_view_include_columns`: an array of fields to return for the "table" endpoint. Should be field names written in camelCase. All values should match with either the name of a field in the table being queried, or the name of a "custom" column defined in `outcome_display_column_definition` (see below).  
+Some things to be aware of:
+  - `null` or empty array means return ALL fields
+  - if you've named a "custom" field in here and yet still want to return all other fields, you can specify this by including the special value "..." as an item, which has the same effect as the "...rest" parameter in javascript.
+- `table_view_exclude_columns`: an array of columns to exclude. Basically, it takes all the columns collected from the "include" list, and removes any named here.
+- `detail_view_include_columns` / `detail_view_exclude_columns`: exactly the same as the "table_view" fields, except applies to the fields that show up in the Details view
```
nmadruga, line 222: Wait a minute. Can't you just send an empty array to defined you would like to include all columns and then exclude the ones you don't want to display?

Because in the example bellow (starting in line 283) you are listing both the columns to include and the columns to exclude. I think the best usage would be define either one or the other...

documentation/Outcomes-Display.md
```diff
@@ -0,0 +1,427 @@
...
+    - members (custom column)
+- In Details view, the page Header will be the value from the `name` field (which is why it made sense to exclude it from the list of fields to show)
+- We also are allowed to see a list of Linked Applications in Details view
+
+We now define three columns with custom definitons in `outcome_display_column_definition`: `logoUrl`, `registrationDocumentation` and `members`. That means that the other two will just use their field name as column title, and the values in those fields will be shown as plain text.
+
+Let's look at the custom column definitions one by one:
+
+**logoUrl**
+
```
nmadruga, line 333:
```suggestion
`outcome_display_column_definition`
```

Just so it's easier to track which is this referring to

documentation/Outcomes-Display.md
```diff
@@ -0,0 +1,427 @@
...
+            "responses.thisResponse"
+            ],
+            "operator": "objectProperties"
+        }
+        ],
+        "operator": "+"
+    },
+    "size": "tiny",
+    "altText": "No logo supplied"
+    },
```
nmadruga, line 361:
```suggestion
"elementParameters": {
        "url": {
            "children": [
            {
                "children": [
                "applicationData.config.serverREST"
                ],
                "operator": "objectProperties"
            },
            {
                "children": [
                "responses.thisResponse"
                ],
                "operator": "objectProperties"
            }
            ],
            "operator": "+"
        },
        "size": "tiny",
        "altText": "No logo supplied"
    },
```

documentation/Outcomes-Display.md
```diff
@@ -0,0 +1,427 @@
...
+    "valueExpression": {}
+}
+```
+- It will displayed using the "imageDisplay" plugin
+- The `elementParameters` define the url for the logo, namely: a combination of the server host and the value from the current response (the response object being the "value" returned in the `logoUrl` field, which is just a text path to the file).
+- No "additionalFormatting" -- it's all handled by the "imageDisplay" plugin
+- No "valueExpression" -- we just want the data straight from the `logoUrl` field
+
+**registrationDocumentation**
+
```
nmadruga, line 372: `outcome_display_column_definition`

documentation/Outcomes-Display.md
```diff
@@ -0,0 +1,427 @@
...
+    "elementTypePluginCode": "fileUpload",
+    "elementParameters": {},
+    "additionalFormatting": null,
+    "valueExpression": null
+}
+```
+Pretty straightforward, it just uses the "fileUpload" plugin to display, and the value in this field is the whole response object required for the file display.
+
+**members**
+
```
nmadruga, line 388: `outcome_display_column_definition`

nmadruga (changes requested):
Thanks for adding this docs. Very helpful! A couple of suggestions : )

documentation/Outcomes-Display.md
```diff
@@ -0,0 +1,427 @@
...
+        },
+        ...etc.
+    ],
+    "totalCount": 13
+}
+```
+
+Note that each table row is returned as both an array of values *and* a single "item" object with key-value pairs.
+
+`totalCount` is used for pagination purposes (since only one page of data is returned at a time)
```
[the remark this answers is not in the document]
CarlosNZ: The full result -- it's the "count" field on all "plural" GraphQL queries

documentation/Outcomes-Display.md
```diff
@@ -0,0 +1,427 @@
...
+
+The following fields are also configurable in "outcomes_display" layouts:
+
+- `title`: the name of the "outcome" that shows in the "Outcomes" menu and as the Header of the table (e.g. "Users"). Default is just the tableName converted to plural Title Case, e.g. `user` => "Users"
+- `permissionNames`: an array of permissions names that the request requires (via JWT) in order to see this Layout. Default is an empty array, and both that and `null` allow full access (i.e. no restrictions)
+- `table_view_include_columns`: an array of fields to return for the "table" endpoint. Should be field names written in camelCase. All values should match with either the name of a field in the table being queried, or the name of a "custom" column defined in `outcome_display_column_definition` (see below).  
+Some things to be aware of:
+  - `null` or empty array means return ALL fields
+  - if you've named a "custom" field in here and yet still want to return all other fields, you can specify this by including the special value "..." as an item, which has the same effect as the "...rest" parameter in javascript.
```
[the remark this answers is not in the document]
CarlosNZ: There's some examples at the bottom that demonstrate this.

documentation/Outcomes-Display.md
```diff
@@ -0,0 +1,427 @@
...
+- `table_view_exclude_columns`: an array of columns to exclude. Basically, it takes all the columns collected from the "include" list, and removes any named here.
+- `detail_view_include_columns` / `detail_view_exclude_columns`: exactly the same as the "table_view" fields, except applies to the fields that show up in the Details view
+- `detail_view_header_column`: name of the field/column whose value should be used as the Header display in Details view. If not specified, the table name will just be used.
```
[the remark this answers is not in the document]
CarlosNZ: Yes.

documentation/Outcomes-Display.md
CarlosNZ: You just use whatever is easiest for what you're doing. If you're wanting all columns but one, then use empty array and add an exclusion. But if you're only including a few columns from a lot, then add an include list.

In the example you mentioned, the table columns only show three fields, so there is an include list and no exclusions.

In the details view, I want everything *except* `id`, `name` and `isSystemOrg`, so they go in the exclusion list. However, I also want to change the order of `logoUrl` (so it shows first) and add a special column called "members", so they need to go in the inclusion list (along with `...` to append all other columns in between)

documentation/Outcomes-Display.md
CarlosNZ: I'm not entirely clear what you're referring to. Are you perhaps getting mixed up between the two definition tables, and the structure of the returned data? I'm not sure which one you're asking about here.

documentation/Outcomes-Display.md
CarlosNZ: ??

documentation/Outcomes-Display.md
CarlosNZ: It's a pretty standard white-list / black-list kind of thing.

documentation/Outcomes-Display.md
CarlosNZ: I see, you mean at the start of the object block.

I just tried it and it looked weird, so have added a clarifying note in the "Let's look at..."

nmadruga (approved):
Thanks for the explanations Carl. I think this is good. Like a said just a simple relational diagram here would be very helpful. Please add you you can, otherwise looks great.

documentation/Outcomes-Display.md
CarlosNZ: Can you explain which part you're having trouble with and I'll try and clarify it, but I honestly can't think what a diagram here would even look like -- it's (almost) a single table of fields, and I've tried hard to start at the simplest possible cases and work up to more complexity with examples.

Is it the configuration, or the output data structure you're confused about?

Issue #594, opened 2021-09-27
Config files -- export with snapshots


CarlosNZ:
I realise that it's currently not easy to change anything in the configuration file(s) (such as default language), as it's part of the build.

Would be good if config files (from front and back end) were (optionally) exported with snapshots, so they could be edited separately, then re-imported. (A snapshot could be ONLY a config file, in some cases).

Similarly, would it make sense to just have a single config file, rather then separate front and back end ones?

Why don't we just put the config file in back-end, and have the front-end "require" it remotely?

CarlosNZ:
While we're at it, should export all the content in "localisations" folder too, since we don't have a UI for it yet.

CarlosNZ:
What I'm doing first, the easy bit

- separate "preferences" (i.e. things users might actually want to change) and "config" (i.e settings required for the build)
- "config" stuff remains in existing config.ts files (front and back)
- "preference" stuff all goes in a common (back-end) preferences.json file, separated into "server" and "web" objects
- back-end config.ts imports prefs and continues as before
- front-end fetches prefs from server (/get-prefs endpoint) right at the top (before LanguageProvider), makes prefs available via `usePrefs` hook (Most things are still in "config" (server addreses etc. Main  thing we want in prefs is default language))
  - have re-configured Localisation a bit so that "languageOptions" list is fetched as part of the same request as prefs, so they'll be passed into LanguageProvider ready to go

This will make it easy to export the "preferences.json" file along with snapshots

PR #595, opened 2021-09-27
#594 (1) -- consolidate configs and prefs

CarlosNZ:
First part of #594, as outlined in [this comment](https://github.com/openmsupply/application-manager-server/issues/594#issuecomment-928235451)

Requires [front-end #1011](https://github.com/openmsupply/application-manager-web-app/pull/1011)


src/components/permissions/routes.ts
```diff
@@ -116,11 +119,23 @@ const routeVerification = async (request: any, reply: any) => {
...
 
+// Serve prefs to front-end
+const routeGetPrefs = async (request: any, reply: any) => {
+  const prefs = JSON.parse(
+    readFileSync(path.join(getAppEntryPointDir(), '../preferences.json'), 'utf8')
+  )
+  const languageOptions = JSON.parse(
+    readFileSync(path.join(getAppEntryPointDir(), '../localisation/languages.json'), 'utf8')
+  )
+  reply.send({ preferences: prefs.web, languageOptions })
```
nmadruga, line 130: Hmm I see so this would get passed in the first iteration between server and front-end app?

CarlosNZ: Yeah, so the front-end starts off by loading prefs and languageOptions before it does anything else.

It already did that for languageOptions, I've just updated it so it gets prefs at the same time.

CarlosNZ:
Oh, one thing I forgot to mention @nmadruga -- I've moved the JWT secret out of the config file and moved it into the `.env` file along with SMTP password and WEB_HOST (so it's not shared in the repo and elsewhere).

We'll probably need to update the build script/instructions a little bit.

nmadruga:
> Oh, one thing I forgot to mention @nmadruga -- I've moved the JWT secret out of the config file and moved it into the `.env` file along with SMTP password and WEB_HOST (so it's not shared in the repo and elsewhere).
> 
> We'll probably need to update the build script/instructions a little bit.

Ok - that's good. Have you added in an entry on Bitwarden? I think we need to add this to our docs straight away - so don't need to deal with that on next time we just need to release a version.

CarlosNZ:
> Ok - that's good. Have you added in an entry on Bitwarden? I think we need to add this to our docs straight away - so don't need to deal with that on next time we just need to release a version.

We don't need any particular key to be saved. We can just make one up each time we restart the server (which will mean current JWTs will no longer be valid).

preferences.json
```diff
@@ -0,0 +1,13 @@
...
+  "server": {
+    "thumbnailMaxWidth": 300,
+    "thumbnailMaxHeight": 300,
+    "hoursSchedule": [
+      0, 1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11, 12, 13, 14, 15, 16, 17, 18, 19, 20, 21, 22, 23
+    ]
+  },
+  "web": {
+    "paginationPresets": [2, 5, 10, 20, 50],
+    "defaultLanguageCode": "en_nz"
```
nmadruga, line 11: Ok, the default is defined here... I think this answers one question I had while reviewing the Front-end PR. 😄

src/components/permissions/routes.ts
nmadruga: Cool

nmadruga:
One thing that I noticed, not sure if it is expected behavior - I changed to a different language before login and then selected on user to login in (admin) after I logged in the language changed back to English. Shouldn't it stayed with my selected one?

CarlosNZ:
> One thing that I noticed, not sure if it is expected behavior - I changed to a different language before login and then selected on user to login in (admin) after I logged in the language changed back to English. Shouldn't it stayed with my selected one?

Currently, when logging out, the local storage gets completely wiped, so the login screen will go back to the default language.

I can make it so that it only deletes the JWT and leaves the language preference untouched. I was slightly reluctant to do that, since in the past we've had problems with Apollo cache stuff hanging around in local storage and messing things up, so it was safer to fully delete the cache.

What I might do is delete the cache as it does, but then immediate re-write just the language setting back. Effectively -- delete everything BUT the language setting. That sound okay?

CarlosNZ:
> What I might do is delete the cache as it does, but then immediate re-write just the language setting back. Effectively -- delete everything BUT the language setting. That sound okay?

Okay, done that: 33df78ec
(I also created a tiny `/logout` route which I've been wanting to do for ages. Sometimes the app get itself stuck in some weird state with nothing on screen, so can just manually go to /logout url to reset. )

nmadruga (approved):
Yes, I was a bit confused with the change on languages before. And thanks for adding the strings to core_templates, that will be useful to have there.

PR #596, opened 2021-09-28
#594 Export / Import localisation and prefs with snapshots

### LINKED ISSUE #594: Config files -- export with snapshots

CarlosNZ:
Fixes #594 

- Additional options parameters to specify if "localisation" folder and "preferences" should be exported/imported with snapshot
- Update use/take snapshot scripts to export/import them
- Add base localisations and prefs to "basic_snapshot"

So we'll need to remember that when we want to change language settings or prefs, to update them in "basic_snapshot", not in the project root, as they'll be overwritten next time "basic_snapshot" is loaded. (Unless we intentionally *exclude* them from basic_snapshot so the project root versions are the "master" versions? -- what do you think is best?)


nmadruga:
> * Additional options parameters to specify if "localisation" folder and "preferences" should be exported/imported with snapshot

Very nice. I was missing that definetelly!

nmadruga:
> So we'll need to remember that when we want to change language settings or prefs, to update them in "basic_snapshot", not in the project root, as they'll be overwritten next time "basic_snapshot" is loaded. (Unless we intentionally _exclude_ them from basic_snapshot so the project root versions are the "master" versions? -- what do you think is best?)

I think something like this needs to be documented in the wiki somewhere. Maybe in the document where we talk about snapshots (the one that needs updating)

CarlosNZ:
> I think something like this needs to be documented in the wiki somewhere. Maybe in the document where we talk about snapshots (the one that needs updating)

Sure, but which option do you prefer?

nmadruga (approved):
Looks good. I can see the files beiing exported and imported correctly.
I'll try another one without exporting those to see how it imports, but so far so good.

I just have one sort of unrelated question.

The languagesFile imported looks like this:
Screen Shot 2021-10-05 at 3 24 22 PM

But when II go select the language is giving me 2 options:
Screen Shot 2021-10-05 at 3 24 03 PM
Screen Shot 2021-10-05 at 3 23 46 PM

CarlosNZ:
> But when II go select the language is giving me 2 options:
Screen Shot 2021-10-05 at 3 27 24 PM

😉

nmadruga:
Cool. Thanks! All looks good to me

PR #597, opened 2021-09-28
REST JWT endpoint example

andreievg:
i've tested with postman but not with front end, see comment in front end PR.

Much easier to see actual changes if `hide whitespace changes` is on

### DIFF
--- a/src/server.ts
+++ b/src/server.ts
@@ -1,9 +1,8 @@
-import fastify from 'fastify'
+import fastify, { FastifyPluginCallback, FastifyReply } from 'fastify'
 import fastifyStatic from 'fastify-static'
 import fastifyMultipart from 'fastify-multipart'
 import fastifyCors from 'fastify-cors'
 import path from 'path'
-import { loadActionPlugins } from './components/pluginsConnect'
 import {
   routeUserInfo,
   routeLogin,
@@ -27,12 +26,13 @@ import config from './config'
 import lookupTableRoutes from './lookup-table/routes'
 import snapshotRoutes from './components/snapshots/routes'
 import { routeLocalisations, routeGetLanguageFile } from './components/localisation/routes'
+import { extractJWTfromHeader, getTokenData } from './components/permissions/loginHelpers'
 require('dotenv').config()
 
 // Fastify server
 
 const startServer = async () => {
-  await loadActionPlugins() // Connects to Database and listens for Triggers
+  // await loadActionPlugins() // Connects to Database and listens for Triggers
 
   createFilesFolder()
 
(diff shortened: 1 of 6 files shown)


CarlosNZ:
Thanks, Andrei, I'll try and follow up with this in the morn.

src/server.ts
```diff
@@ -27,12 +26,13 @@ import config from './config'
...
 import snapshotRoutes from './components/snapshots/routes'
 import { routeLocalisations, routeGetLanguageFile } from './components/localisation/routes'
+import { extractJWTfromHeader, getTokenData } from './components/permissions/loginHelpers'
 require('dotenv').config()
 
 // Fastify server
 
 const startServer = async () => {
-  await loadActionPlugins() // Connects to Database and listens for Triggers
+  // await loadActionPlugins() // Connects to Database and listens for Triggers
```
CarlosNZ, line 35: @andreievg -- I assume this was just so it would load quicker and you intended to put it back?

src/server.ts
```diff
@@ -47,100 +47,141 @@ const startServer = async () => {
...
-  // Snapshot routes
-  server.register(snapshotRoutes, { prefix: '/snapshot' })
+    // File download endpoint (get by unique ID)
+    server.get('/file', async function (request: any, reply: any) {
+      const { uid, thumbnail } = request.query
+      const { original_filename, file_path, thumbnail_path } = await getFilePath(
+        uid,
+        thumbnail === 'true'
+      )
+      // TO-DO Check for permission to access file
```
CarlosNZ, line 103: @andreievg -- do you think uniqueID is enough to keep files accessible by people who shouldn't, or do we need to add some user-based checks?

It's a bit tricky if we do, cos some files will need to be widely accessible (such as org logos, which show up in Outcomes), but some should only be accessed by applicants and reviewer's reviewing their applications.

src/server.ts
```diff
@@ -47,100 +47,141 @@ const startServer = async () => {
...
-      const isUnique = await DBConnect.isUnique(tableName, fieldName, value)
-      reply.send({
-        unique: isUnique,
-        message: '',
-      })
-    } catch (err) {
-      reply.send({ unique: false, message: err.message })
-    }
-  })
+  server.register(api, { prefix: '/api' })
```
CarlosNZ, line 184: Just wondering -- what's the advantage of all routes starting with `/api`? Convention?

CarlosNZ:
Have pulled this PR into new branch and made new "final" PR, so closing this one now: https://github.com/openmsupply/application-manager-server/pull/607

That one is basically just this PR with a few small additions.

PR #600, opened 2021-09-29
#590 Row restrictions on Outcomes

### LINKED ISSUE #590: Outcomes -- row filtering for new tables
As discussed with @andreievg  and briefly mentioned [here](#589), plan is to extend the Outcomes API to also handle restricting by row based on an (optional) filter specified in OutcomeDisplay config. This is to make row-level control simpler for NEW outcome tables -- existing "user" and "organisation" tables will (should?) be handle by existing row-level policies.

I will describe the solution in more detail here when I've worked it out, but basically it'll involve an additional field in "outcome_display" to define a filter so that results can be restricted by user or orgid (or anything really).

CarlosNZ:
Fix #590 

Requires [front-end PR #1013](https://github.com/openmsupply/application-manager-web-app/pull/1013)

Adds the ability to restrict outcome displays by row (in addition to existing row-level permissions) by adding an addition "row_restrictions" field to "outcome_display" table. This field takes an (optional) GraphQL Filter object. The variables for "userId" and "orgId" are supported, but must be quoted. An example restrictions object:

```
{
  "userOrganisations": {
    "some": {
      "organisationId": {
        "equalTo": "$orgId"
      }
    }
  }
}
```
This is used on the "Users" display to limit the results to only show users in the current user's current organisation. A similar filter is used to limit the list of Organisations to only ones the current user belongs to.

I've also added a "description" field to outcome_displays, as I thought some further explanation might be necessary to clarify what each one does. This field is intended for devs and configurers, not to be shown to normal users.

For testing and snapshot, see front-end PR. 

Note: requires `yarn install` as I've imported one new package.


nmadruga (approved):
Looks great. Thanks!

The only bit that I'm not sure is this: wouldn't be better to show a 404 page?
Screen Shot 2021-10-06 at 2 26 25 PM

nmadruga:
> > The only bit that I'm not sure is this: wouldn't be better to show a 404 page?
> 
> Yeah, I guess so -- our generic 'Page not found' page?
> 
> Do we need to distinguish between "doesn't exist" and "unauthorized" though? I'm just thinking of a scenario -- one user copies a link to an item from their browser and sends it to another user. The other user opens it, and gets the 404 page. This might be a bit confusing as it's just that they're not allowed to see it, not that it's missing.
> 
> I guess we've been doing similar with applications, right? You just get 404 page if unauthorized to view, so I guess we should be consistent. Thanks, I'll update it before I merge.

Yes, I get your point. And I think this subject has been raised before. What I would suggest is to open a new issue so we can discuss that idea and pass throught more people - if we agree that it should be this way then we would have it for the whole system.

nmadruga:
Also thinking about something else here. I don't really like using the ids to display these outcomes items... Which is also a problem for Reviews. I think for security reasons is better to mask that with some sort of generic serial - that gets generated per entity and is also unique. What do you think? I'll can add a separated issue about that.

CarlosNZ:
> Yes, I get your point. And I think this subject has been raised before. What I would suggest is to open a new issue so we can discuss that idea and pass throught more people - if we agree that it should be this way then we would have it for the whole system.

Okay, I'll just use the 404 page here and make an issue to think about it (not that's it's particularly high priority at the moment ;)  )

CarlosNZ:
> Also thinking about something else here. I don't really like using the ids to display these outcomes items... Which is also a problem for Reviews. I think for security reasons is better to mask that with some sort of generic serial - that gets generated per entity and is also unique. What do you think? I'll can add a separated issue about that.

Yes I agree, but we did discuss that quite a bit earlier on (for reviews that is), and Andrei was fairly adamant it was fine using IDs.  So yeah, I would be happy to change it but  a) it would be a fairly substantial job to do properly;  and b) probably not a high priority, all things considered.

Definitely falls into "nice to have" for now.

CarlosNZ:
> Yes, I get your point. And I think this subject has been raised before. What I would suggest is to open a new issue so we can discuss that idea and pass throught more people - if we agree that it should be this way then we would have it for the whole system.

Okay, @nmadruga -- how about this? 64e3f703

I've updated the NoMatch/404 page to be more consistent with other pages (it's basically the same as Submission and Verification pages layout), and I've given it optional "header" and "message" props (with generic defaults, now localised).

So with the outcome errors, it'll have the same message but look like this:
Screen Shot 2021-10-06 at 7 35 01 PM

Generic 404 page is:

Screen Shot 2021-10-06 at 7 49 14 PM

PR #601, opened 2021-09-29
Core_templates snapshot v1

nmadruga:
### Linked PR:
- Front-end https://github.com/openmsupply/application-manager-web-app/pull/1014

Basically updates the initial version of our public snapshots. Renamed from `basic_snapshot` to `core_templates` and a list of all Core-actions currently used in the `Sample template` (and others) to be used as a base for new templates created. 

**Note**: I have also removed the `basic_snapshots` from our private repo, and updated `dev` current to open - it needed some changes to work with new outcomes changes... Not sure what we do about `laos` one, it should be out-of-date now.


CarlosNZ (approved):
Great, let's get this merged ASAP and then we can re-base our existing PRs.

PR #602, opened 2021-09-29
#993F Delete snapshot endpoint

CarlosNZ:
So we can delete snapshots in the UI.

Required for [front-end #998](https://github.com/openmsupply/application-manager-web-app/pull/998)


nmadruga (approved):
LGTM

PR #603, opened 2021-09-29
#update action generate doc to use localised options

nmadruga:
Missing change done on the old Portuguese-localisation PR...

Used to get a generated certificate using Carbone localisation (use to translate dates and currency)


CarlosNZ (approved):
Looks good 👍

PR #607, opened 2021-10-04
#458 Authentication for REST endpoints

### LINKED ISSUE #458: Authentication for REST endpoints
Currently all endpoints are open access. Need to have an authentication system to limit access, based on JWT and permissions.

Place holder issue -- exact solution to be discussed

CarlosNZ:
Fix #458 

Basically just takes [Andrei's existing PR](#597) and makes a few additional tweaks:

- Update docs with all current endpoints, divided into Public / Authenticated / Admin
- Update from latest develop (with slightly different endpoints)
- Move `/check-unique` endpoint function into "routes" file to keep main `server.ts` file cleaner

Now need to update front-end to make sure it's using correct endpoint prefixes.

### DIFF
--- a/documentation/API.md
+++ b/documentation/API.md
@@ -9,7 +9,7 @@ The back-end currently has two server instances which are launched to handle inc
 
 ---
 
-### Postgraphile server API:
+## Postgraphile server API:
 
 `http://localhost:5000/graphql`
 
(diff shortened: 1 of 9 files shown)


documentation/API.md
```diff
@@ -9,7 +9,7 @@ The back-end currently has two server instances which are launched to handle inc
 
 ---
 
-### Postgraphile server API:
+## Postgraphile server API:
 
 `http://localhost:5000/graphql`
```
nmadruga, line 14: Would be helpful to add the `production` path too

documentation/API.md
```diff
@@ -124,6 +81,92 @@ Basic login endpoint that will also return (on success):
...
 
+#### Get preferences endpoint:
+
+GET: `/get-prefs`
+
+Provides current preferences and language options for the current installation. Called by front-end at the very start of loading in order to configure the environment according to system settings.
+
+#### Get language endpoint:
+
+GET: `/language/<langauge-code>`
```
nmadruga, line 92:
```suggestion
GET: `/language/<language-code>`
```

nmadruga: Example would be a good thing to also add here

nmadruga: Also - tried getting for `en_dtac` and it wasn't there, but the return is 404. Should it be some other error maybe?

documentation/API.md
```diff
@@ -124,6 +81,92 @@ Basic login endpoint that will also return (on success):
...
+
+Sets the verification record at `<uniqueid>` to `true` and triggers configured Action. Used for verification via email, etc. See [Create Verification Action](List-of-Action-plugins.md) for more info.
+
+#### File download endpoint:
+
+GET: `/file?id=<uniqueId>`
+
+Usage: `GET` request with file database id as a URL query parameter.
+
+This is a public endpoint, but files all have a long uniqueId which should prevent unauthorized access.s
```
nmadruga, line 108:
```suggestion
This is a public endpoint, but files all have a long uniqueId which should prevent unauthorized access.
```

nmadruga: Tried using this endpoint to get the file existing in `core_templates` and got this error:
Screen Shot 2021-10-06 at 3 25 02 PM

 In the server was showing: 
relation "permissions_all" does not exist
Cannot parse JWT
Cannot parse JWT

path: `http://localhost:8080/api/public/file?id=CylhAzxRhSX_QjtArq3bi`

documentation/API.md
```diff
@@ -124,6 +81,92 @@ Basic login endpoint that will also return (on success):
...
+
+They require a valid JWT supplied as a Bearer token in the authorization header of the request:
+
+```
+Authorization: Bearer <token>
+```
+
+#### File upload endpoint:
+
+POST: `/upload`
```
nmadruga, line 122: Nice! Tried without passing the JWT and got this message:
Screen Shot 2021-10-06 at 3 27 29 PM

documentation/API.md
```diff
@@ -124,6 +81,92 @@ Basic login endpoint that will also return (on success):
...
+
+Usage: `POST` request with file(s) in the request `body` form-data:  
+`key: "file" value: <File(s)>`
+
+URL query paramter fields (optional):
+
+- `user_id`
+- `application_serial`
+- `application_response_id`
```
nmadruga, line 131: How about `template_id` isn't that one of the optional?

src/config.ts
```diff
@@ -27,7 +27,7 @@ const config: { [key: string]: any } = {
   // In production postgraphile is started with -q and -i /postgraphile/...
   nodeModulesFolder:
     process.env.NODE_ENV === 'production' ? '../../node_modules' : '../node_modules',
-  jwtSecret: process.env.JWT_SECRET,
+  jwtSecret: process.env.JWT_SECRET || 'devSecret',
```
nmadruga, line 30: Maybe needs to be removed?

nmadruga (commented):
To be continued (halfway through reviewing front-end now).

documentation/API.md
CarlosNZ: Yeah, there's actually a few others, have added now: 977b0244

src/config.ts
CarlosNZ: No, I put this in to stop the server crashing if no secret was supplied in .env. This default is useful for development, cos it means all our existing JWTs keep working (for example in test suites, and Postman collections). But we should always be adding a secure randomly generated secret key whenever we update and re-start the server, so there's no problem leaving this "default" one out in plain view as it should never be used on a live server.

nmadruga: Cool. Thanks for explaning this

documentation/API.md
nmadruga: Great. Seems ok, I thought some specific message as: this language isn't included, but I guess 404 is alright.

You still have the typo to correct on this one, please.

PR #608, opened 2021-10-05
#1007B applicationData endpoint for TemplateBuilder

CarlosNZ:
Just adds additional endpoint `/admin/get-application-data` so TemplateBuilder's Actions editor can properly evaluate parameters in the UI.

(Required for https://github.com/openmsupply/application-manager-web-app/issues/1007 -- front-end PR coming soon)


nmadruga (approved):
LGTM

Does what is supposed to. Check that requires JWT correctly and works with front-end while logged in as Admin.

PR #610, opened 2021-10-06
#379 Authentication in Evaluator

### LINKED ISSUE #379: Evaluator: find a (secure) way to provide authentication in API/GraphQL requests
At least JWTs, so we can write queries to our own endpoints!

CarlosNZ:
Fix #379 

Still need to update front-end to use this mechanism, but this includes:
- changes required to evaluator to handle authentication/headers
- updated test suite to use this mechanism and demonstrate custom header queries
  - Test suite now requires "evaluator_test" snapshot and an additional secrets file -- see documentation for details (@nmadruga I'll send you my testSecrets.json file on Telegram)
- updated documentation to explain the above in detail
- have also updated the GUI query builder app to allow inserting a JWT into evaluator parameters

Please see the new "Authentication" and "Testing" sections of "Query-Syntax.md" doc for detailed explanation.

The primary mechanism for providing authentication headers is pretty simple -- just add an additional "headers" field to the evaluator parameters. There is also a mechanism for providing specific/ dynamic authentication within the query itself. This is more complicated, and I'll admit it's not the most elegant way to format this info, but I think it's acceptable for a couple of reasons:
1. It needed to be backwards compatible with all existing API/GraphQL queries, so couldn't change the underlying child node structure
2. I don't actually think we need to use it right now. I'm pretty sure all front-end API queries can just use the "parameters" approach to add current user's JWT to all evaluator queries, but I've provided this option just in case I'm wrong about that.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -60,3 +60,6 @@ src/modules/expression-evaluator/expression-evaluate-gui/src/expression-evaluato
 #Snapshots working folder
 database/_snapshots
 database/snapshot_path_aliases.json
+
+#Other secrets
+testSecrets.json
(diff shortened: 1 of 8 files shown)


nmadruga (approved):
Looks great. Thanks for doing this, Carl

Tested running the test and evaluator-expression-gui and seems to be working well.

PR #612, opened 2021-10-06
#611 Fix for templates going to DRAFT on snapshot load

### LINKED ISSUE #611: Importing Snapshot sets templates to "DRAFT" if more than one version.
As reported [here](https://github.com/openmsupply/application-manager-server/issues/506#issue-968361408)

CarlosNZ:
Fixes #611 

Turns out this was due to the `public.set_template_to_draft` function in `template.sql` being a bit over-eager. Was supposed to make duplicated templates as DRAFT, but didn't really consider re-importing them. Have just added another condition that means it will only do this if there are no other versions with 'AVAILABLE' status.

**Additional:**

I've also added another function in here that will automatically set other "AVAILABLE" version status to "DISABLED" when you enable another version. This is consistent with what we do with "is_current" in stage/status history, and when "Outcome" is changed from "PENDING", etc., so it makes sense to enforce this at the database level. This means the front-end will be able to set a template to "AVAILABLE" without having to go and disable another version first. (Front-end PR https://github.com/openmsupply/application-manager-web-app/pull/1025)

### DIFF
--- a/database/buildSchema/07_template.sql
+++ b/database/buildSchema/07_template.sql
@@ -47,6 +47,7 @@ $template_event$
 LANGUAGE plpgsql;
 
 -- FUNCTION to make sure duplicated templates have 'DRAFT' status
+--   but only if there are no other versions with 'AVAILABLE' status
 CREATE OR REPLACE FUNCTION public.set_template_to_draft ()
     RETURNS TRIGGER
     AS $template_event$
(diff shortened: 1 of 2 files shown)


database/buildSchema/07_template.sql
```diff
@@ -47,6 +47,7 @@ $template_event$
 LANGUAGE plpgsql;
 
 -- FUNCTION to make sure duplicated templates have 'DRAFT' status
+--   but only if there are no other versions with 'AVAILABLE' status
```
nmadruga, line 50: No sure if I understood this phrase. Shouldn't be like this instead:
> -- but only if there is another version with 'AVAILABLE' status

nmadruga (approved):
That is an awesome fix! I looked for this solution for about 2 hours the other day and coudn't find it. Good job! 

I wonder if it would be also good to check: #965

I can have a look on that in the afternoon and I'll post here.

PR #618, opened 2021-11-09
Additional build fixes

CarlosNZ:
This will need to be tested properly. I can confirm that the build process should work fine and the docker image runs locally okay. There's still some weirdness with files not showing up correctly and some other things which I'll look into and maybe push to this PR too.

**Update**: When running the docker image locally, the core_templates snapshot hasn't fully loaded properly, although it's there enough to log in. So I can manually import a snapshot (including core_templates) and load that which works. But now I get this error when trying to start a new "User Registration"
Screen Shot 2021-11-10 at 11 54 08 AM

Inspecting the network logs, it's failing when trying to access `http://localhost:3000/postgraphile/graphql/api?dev=getTemplates` This doesn't seem right, but not sure why it's requesting that.

### DIFF
--- a/docker/run.sh
+++ b/docker/run.sh
@@ -5,7 +5,7 @@
 docker run \
     -ti \
     -p 3000:3000 \
-    -e 'SMTP_PASSWORD=smtp_password_goes_here' \
+    -e 'SMTP_PASSWORD=smtp_password_goes_here' -e 'WEB_HOST=http://localhost:3000' \
     testbuild
 
 # -ti -> interactive (connect to shell on startup)
--- a/documentation/Demo-Server-Guide.md
+++ b/documentation/Demo-Server-Guide.md
@@ -69,6 +69,7 @@ docker push msupplyfoundation/mflow-demo:front-demo-19-08-2021_back-demo-19-08-2
   - name
   - port 8000
   - WEB_HOST url
+  - JWT_SECRET (this can be any random string, but should be a decent length, say > 24 alphanumeric characters. There's no need to record this key anywhere, as it can change anytime -- that just means existing JWTs become invalid, so users will need to re-login)
 
   The system will be launched with “basic_snapshot” data. Upload and run a new snapshot as required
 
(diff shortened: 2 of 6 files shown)


docker/run.sh
```diff
@@ -5,7 +5,7 @@
 docker run \
     -ti \
     -p 3000:3000 \
-    -e 'SMTP_PASSWORD=smtp_password_goes_here' \
+    -e 'SMTP_PASSWORD=smtp_password_goes_here' -e 'WEB_HOST=http://localhost:3000' \
```
nmadruga, line 8: Is this always the case? Well, I think that's only used to test locally right?

utils/build_all.sh
```diff
@@ -11,7 +11,7 @@ cp -R database/buildSchema ./build/database/
 cp -R database/insertData ./build/database/
 cp -R database/snapshotOptions ./build/database/
 cp -R database/core_templates ./build/database/
-echo 'copyling localisation files'
+echo 'copying localisation files'
```
nmadruga, line 14: Oh noe! And I locked to that for so long. Didn't see the typo! Good catch

documentation/Demo-Server-Guide.md
```diff
@@ -160,6 +160,7 @@ sudo chown 472 grafana_on_port_8009
 export TAG='front-demo-19-08-2021_back-demo-19-08-2021_pg-12_node-14'
 export SMTP_SECRET='add_smtp_secret_here'
 export WEB_URL='https://irims-demo.msupply.org:<replace port>'
+export JWT_SECRET='random private key'
```
nmadruga, line 163: Is this where we set the new JWT for this server? Maybe worth mentioning how to generate this JWT and use here. Do we need to keep this saved somewhere for future use? Like a Bitwarden entry?

nmadruga (changes requested):
Just some thoughts

nmadruga (changes requested):
Hi, sorry I was supposed to have sent this earlier to you

docker/run.sh
CarlosNZ: Yes, exactly, you'll still need to change it for the server. This is just for running the docker image locally

documentation/Demo-Server-Guide.md
CarlosNZ: I don't think it really matters, we can just make up a new one every time we launch a new instance (or even restart the server). The only time we need to keep this value is if we want existing JWTs to remain valid (like in our testing environment, or with existing Postman collections, say)

But yes, have added a note 084b0c65

nmadruga (approved):

PR #619, opened 2021-11-10
Add updates to Demo-Server-Guide for launching instance with new datafile

nmadruga:
Problem was that when the exiting instance was running the commands to stop it were not clear and took me some time to find out.


CarlosNZ (approved):
Cool, thanks.

PR #621, opened 2021-11-11
#1029F add field is interactive to template

nmadruga:
To be used with Front-end [PR](https://github.com/openmsupply/application-manager-web-app/pull/1031)

### Changes
- Add a new optional field `interactive` in template table to allow reviewers to send the List of Questions to applicants.
- Add example to **core_template** snapshot on template code: `OrgJoin`.


database/buildSchema/07_template.sql
```diff
@@ -11,13 +11,15 @@ CREATE TABLE public.template (
     name_plural varchar,
     code varchar NOT NULL,
     is_linear boolean DEFAULT TRUE,
+    is_interactive boolean DEFAULT TRUE,
```
CarlosNZ, line 14: I wonder if there's a better name than `is_interactive`, it doesn't seem that clear what this would be. I'm not sure, maybe "allow_applicant_interaction" or something? Not a big deal, up to you.

nmadruga: I kind of like the is_interactive as more general... I think maybe could be kept like that, and if we dislike it we change later?

CarlosNZ: Yeah, it was the "general-ness" that makes it unclear, when in fact this is for a very specific use case. Are you imagining that this field would be used for determining other things too?

nmadruga: Not for now... What do you think about is_applicant_interactive?

CarlosNZ: Yeah, that's okay. I think it's the "interactive" term that is somewhat confusing, as the forms are pretty "interactive" in their normal state. It's more about whether the applicant can make changes. Could just have "can_applicant_make_changes" and keep it very specific and clear?

But I dunno, don't want to get stuck on this, it's fine whatever. You choose what you think is best and move on.

PR #623, opened 2021-11-17
#622 Handle either type of empty row restriction (null or {})

### LINKED ISSUE #622: Small problem with Outcome row-restriction defaults
The default for the "row_restrictions" field is `{}` (rather than `NULL`), which is causing a bug with blank filter objects.

Could change the default, but probably best to fix it so it handles either correctly.

CarlosNZ:
Fix #622 

For background, the filter `{ id: { isNull: false } }` is just a dummy filter that is required, or else GraphQL throws an error for "empty filter object".

Previously the logic looked for `null` in row restrictions and puts this dummy filter in if so. Now it also checks for empty object `{}` too (which is the default value for the row_restrictions field)

### DIFF
--- a/src/components/outcomes/helpers.ts
+++ b/src/components/outcomes/helpers.ts
@@ -129,7 +129,10 @@ const getFilters = (
   orgId: number | undefined
 ): object => {
   // We're only interested in the highest priority restrictions
-  const restrictions = outcomes[0].rowRestrictions || { id: { isNull: false } }
+  const restrictions =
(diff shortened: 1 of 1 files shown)


nmadruga (approved):
No sure how to test this, but the change looks very straight forward

CarlosNZ:
> No sure how to test this, but the change looks very straight forward

Thanks. If you really want to test:

- In `develop`, change one of the "row restrictions" fields in "outcome_display" that is currently `null` to `{}`
- Visit this Outcome in the UI
- Note the error: `Problem with Outcome Table query`
- Switch to this branch and try again -- it should display correctly

I'll merge this at the end of today :)

Issue #620, opened 2021-11-11
Bug: Failed verification code with underscore at last character.


nmadruga:
When clicked the link on email it didn't add the _ and the link was failing. Copy and paste worked: http://localhost:3000/verify?uid=AiQfv12pBEE23gUTX3lfYMP_

CarlosNZ:
Ah, good catch, although this is actually an issue with specific email client's parsing on plain text links, as the template doesn't currently specify it as a link. It was fine in the email client when sending that link as plain text from Gmail web app to Apple Mail, for example.

Simplest immediate fix would be to just update the "send Email" action in the template to wrap the url in actual markdown link tags.

CarlosNZ:
> Simplest immediate fix would be to just update the "send Email" action in the template to wrap the url in actual markdown link tags.

Although that will need me to finally fix this: https://github.com/openmsupply/application-manager-server/issues/423

CarlosNZ:
Okay, with [this fix](https://github.com/openmsupply/application-manager-server/pull/624), this problem can be resolved just making the verification link an actual Markdown link, rather than leaving it to email clients to automatically parse.

So instead of:
```
Hi %1,

please confirm your user account registration by clicking (or copy-pasting) the following link:

%2%3
```

We can use:
```

Hi %1,

please confirm your user account registration by clicking (or copy-pasting) the following link:

[%2%3](%2%3)
```

CarlosNZ:
@nmadruga -- are you happy with this solution? If so, we just need to update the core templates then we can close this.

nmadruga:
> are you happy with this solution? If so, we just need to update the core templates then we can close this

Yes. I think that's a good solution for now. We need to keep the loop of wuick fixes for best results. So looks great. 

- Nice if you can add this to all templates in `core_templates`
- Add a note about this on the Template Builder docs - looks pretty empty now, but I think we need to start writing this small tips to help when we finally write the full on documentation for it.

PR #624, opened 2021-11-17
#423 Evaluator -- String substitution improvements

### LINKED ISSUE #423: Evaluator improvement -- stringSubstitution should be able to handle duplicate parameter substitutions
In other words, the following substitution:
```
{
 operator: "stringSubstitution",
 children: [
  "%1 is the same as %1 but not %2",
  "this",
  "that"
 ]
}
```
should return "this is the same as this but not that", but currently it returns "this is the same as that but not"

CarlosNZ:
Fix #423 

Been meaning to fix this for a while, but now that it will be needed to solve #620 properly, have quickly done it.

Means that now a substitution parameter can be used more than once in a string, for example a url with a link:
```
To verify, please visit: [%1](%1)
```

Added a new test to handle this case. (Remember to use the "evaluator_test" snapshot to run the test suite)


src/generated/graphql.ts
```diff
@@ -638,10 +638,14 @@ export type Application = Node & {
...
   /** Reads and enables pagination through a set of `Verification`. */
   verifications: VerificationsConnection;
-  /** Reads and enables pagination through a set of `OrganisationApplicationJoin`. */
-  organisationApplicationJoins: OrganisationApplicationJoinsConnection;
   /** Reads and enables pagination through a set of `UserApplicationJoin`. */
   userApplicationJoins: UserApplicationJoinsConnection;
+  /** Reads and enables pagination through a set of `OrganisationApplicationJoin`. */
+  organisationApplicationJoins: OrganisationApplicationJoinsConnection;
+  /** Reads and enables pagination through a set of `LicenceApplicationJoin`. */
+  licenceApplicationJoins: LicenceApplicationJoinsConnection;
```
nmadruga, line 646: A couple of additions on the graphQL generated file. You probably will need to run core_templates and re-commit this to clean up from other specific snapshots

nmadruga (changes requested):
As mentioned in application-manager-app/issues/#620:

> * Nice if you can add this to all templates in `core_templates`
> * Add a note about this on the Template Builder docs - looks pretty empty now, but I think we need to start writing this small tips to help when we finally write the full on documentation for it.

src/generated/graphql.ts
CarlosNZ: Yeah, I usually regenerate this file in develop after a significant merge

CarlosNZ:
> Nice if you can add this to all templates in `core_templates`

Done! fc93370b
Note that these links won't show correctly until the published evaluator package is upgraded (which I'll do as soon as this is approved)

> Add a note about this on the Template Builder docs - looks pretty empty now, but I think we need to start writing this small tips to help when we finally write the full on documentation for it.

I don't think there's anything to document here -- we should never have been doing it that way to begin with. The natural thing to do would be to put links inside Markdown links. If you don't and you rely on native email client link parsing, you're asking for problems. But the problem is with the email clients, not our system, and the original problem was a very edge case even if you do it that way.

nmadruga:
> > Add a note about this on the Template Builder docs - looks pretty empty now, but I think we need to start writing this small tips to help when we finally write the full on documentation for it.
> 
> I don't think there's anything to document here -- we should never have been doing it that way to begin with. The natural thing to do would be to put links inside Markdown links. If you don't and you rely on native email client link parsing, you're asking for problems. But the problem is with the email clients, not our system, and the original problem was a very edge case even if you do it that way.

Yes, my thinking was just when we do a hand-over to external users (which was asked) we have a road-map for then to follow. Documenting this kind of tip would really help us not having to deal with fixes required for problems that were just created by not using the specific required configuration. And there is no way to know if it's not written somewhere.

CarlosNZ:
> Yes, my thinking was just when we do a hand-over to external users (which was asked) we have a road-map for then to follow. Documenting this kind of tip would really help us not having to deal with fixes required for problems that were just created by not using the specific required configuration. And there is no way to know if it's not written somewhere.

Yeah, the template builder docs are in the front-end anyway (and I can't actually see them anywhere tbh), so outside the scope of this issue I'd say. Why don't you add a quick note to the front-end and make that the PR for [this one](https://github.com/openmsupply/application-manager-server/issues/620)?

nmadruga:
@CarlosNZ Can you add something to the Template Builder docs, please? I think just one line would be enough, is more as a note for us not to forget about adding this to the official docs sometime soon.

CarlosNZ:
@nmadruga added a note: https://github.com/openmsupply/application-manager-web-app/pull/1050

nmadruga (approved):

PR #630, opened 2021-11-25
#1046F fix grant revoke permission actions

nmadruga:
Use front-end PR: https://github.com/openmsupply/application-manager-web-app/pull/1059

### Changes
- Add type cast to make sure correct functions are called
- Update revokePermisison for user and org to be deleted (permission_join only) to keep this simple  

### Test
Using latest **demo_angola** snapshot called: demo_angola_2021_11_25 with the template grantUserPermissions updated 

Question: Do I add something on docs about that change on the revokePermisison action? Probably should right?

### DIFF
--- a/plugins/action_revoke_permissions/src/databaseMethods.ts
+++ b/plugins/action_revoke_permissions/src/databaseMethods.ts
@@ -23,10 +23,15 @@ const databaseMethods = (DBConnect: any) => ({
   revokePermissionFromUserOrg: async (
     username: string,
     org: string | number,
-    permissionNames: string[]
+    permissionNames: string[],
+    isRemovingPermissions: boolean
   ) => {
     const text = `
-      UPDATE permission_join SET is_active = false
+      ${
+        isRemovingPermissions
+          ? 'DELETE FROM permission_join '
+          : 'UPDATE permission_join SET is_active = false '
+      } 
       WHERE user_id = (
         SELECT id from "user" WHERE username = $1
       ) AND organisation_id = ${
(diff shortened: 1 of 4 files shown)


plugins/action_revoke_permissions/src/databaseMethods.ts
```diff
@@ -23,10 +23,15 @@ const databaseMethods = (DBConnect: any) => ({
   revokePermissionFromUserOrg: async (
     username: string,
     org: string | number,
-    permissionNames: string[]
+    permissionNames: string[],
+    isRemovingPermissions: boolean
```
CarlosNZ, line 27:
```suggestion
    isRemovingPermissions: boolean = true
```
Might be good to add default for now.

plugins/action_revoke_permissions/src/revokePermissions.ts
```diff
@@ -6,14 +6,16 @@ const revokePermissions = async ({ applicationData, parameters, DBConnect }: Act
   const db = databaseMethods(DBConnect)
   // Don't specify orgName/Id default because we might be targeting a permission_join with no user (even if application has organisation)
   const { username = applicationData?.username, orgName, orgId, permissionNames } = parameters
+  const isRemovingPermission = true
```
CarlosNZ, line 9:
```suggestion
  const { username = applicationData?.username, orgName, orgId, permissionNames, isRemovingPermission = true } = parameters
```
If you move this value into parameters (but with default), it'll do exactly the same thing but allows us to do it the other way if/when we need to without updating this again.

CarlosNZ (approved):
Sweet, just a couple of small suggestions.

> Question: Do I add something on docs about that change on the revokePermisison action? Probably should right?

Yeah, that'd be good. If you just add the `isRemovingPermissions` parameter to it, then explain what that does (and a clarification that doing it the other way has no effect currently)

CarlosNZ (approved):

PR #631, opened 2021-11-26
#allow user permissions to list org permissions

nmadruga:
Needed to list permissions for company before adding an user - CreateInternalUser template (version 2)

### DIFF
--- a/src/components/permissions/routes.ts
+++ b/src/components/permissions/routes.ts
@@ -112,13 +112,13 @@ const routeUserPermissions = async (request: any, reply: any) => {
     console.log('userExistingPermissions', userExistingPermissions)
 
  grantedPermissions = Array.from(new Set(userExistingPermissions.map((p) => p.permissionName))).sort()
-    availablePermissions = Object.values(templatePermissionRows)
-      .filter(({ permissionName }) => !grantedPermissions.includes(permissionName))
-      .reduce((availablePermissions, { permissionName }) => {
-        if (!availablePermissions.includes(permissionName))
-          availablePermissions.push(permissionName)
-        return availablePermissions
-      }, [])
+    availablePermissions = Array.from(
+      new Set(
+        Object.values(templatePermissionRows)
+          .filter(({ permissionName }) => !grantedPermissions.includes(permissionName))
+          .map((p) => p.permissionName)
+      )
+    ).sort()
   } else {
     // Get permissions for organisation without association with as user
     const orgExistingPermissions = await databaseConnect.getOrgTemplatePermissions(isSystemOrg)
--- a/src/components/permissions/routes.ts
+++ b/src/components/permissions/routes.ts
@@ -122,13 +122,7 @@ const routeUserPermissions = async (request: any, reply: any) => {
   } else {
     // Get permissions for organisation without association with as user
     const orgExistingPermissions = await databaseConnect.getOrgTemplatePermissions(isSystemOrg)
-    availablePermissions = orgExistingPermissions.reduce(
-      (availablePermissions, { permissionName }) => {
-        if (!grantedPermissions.includes(permissionName)) availablePermissions.push(permissionName)
-        return availablePermissions
-      },
-      []
-    )
+  availablePermissions = Array.from(new Set(orgExistingPermissions.map((p) => p.perm))).sort()
   }
 
   // Store array of object per permissionNames with properties and an array of templateCodes
--- a/src/components/permissions/routes.ts
+++ b/src/components/permissions/routes.ts
@@ -150,9 +150,7 @@ const routeUserPermissions = async (request: any, reply: any) => {
   )
 
   return reply.send({
-    templatePermissions: templatePermissions.sort(({ name: aName }, { name: bName }) =>
-      aName.localeCompare(bName)
-    ),
+    templatePermissions,
     grantedPermissions,
     availablePermissions,
   })
--- a/src/components/permissions/routes.ts
+++ b/src/components/permissions/routes.ts
@@ -101,7 +101,7 @@ const routeUserPermissions = async (request: any, reply: any) => {
   const templatePermissionRows = await databaseConnect.getTemplatePermissions(isSystemOrg)
   const userExistingPermissions = username
     ? await databaseConnect.getUserTemplatePermissions(username, orgId)
-    : await databaseConnect.getOrgTemplatePermissions(orgId)
+    : await databaseConnect.getOrgTemplatePermissions(isSystemOrg)
 
   const grantedPermissions = userExistingPermissions.reduce(
     (grantedPermissions, { permissionName }) => {
--- a/src/components/permissions/routes.ts
+++ b/src/components/permissions/routes.ts
@@ -2,7 +2,7 @@ import databaseConnect from '../databaseConnect'
 import { getUserInfo } from './loginHelpers'
 import { updateRowPolicies } from './rowLevelPolicyHelpers'
 import bcrypt from 'bcrypt'
-import { UserOrg } from '../../types'
+import { UserOrg, PermissionDetails } from '../../types'
 import path from 'path'
 import { readFileSync } from 'fs'
 import { startCase } from 'lodash'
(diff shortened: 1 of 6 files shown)


nmadruga:
Changes you suggested done and tested @CarlosNZ

CarlosNZ (commented):
Cool, my only concern is that now the fields "grantedPermissions" and "isUserGranted" are somewhat misleading, since they're not necessarily granted for that org, they just *could* be.

In terms of this template, we don't want the boxes to start of all checked (which they would do with this currently), so the `isUserGranted` value should probably be `false` in this case (or `null` but `false` would be easier to use with the template?)

Similarly, the "grantedPermission" should be `[]` and they should all be in `availablePermissions` instead?

CarlosNZ:
> Changes you suggested done and tested @CarlosNZ

I'm not seeing any changes, and I'm still getting `isUserGranted: true` for all permissions when no orgId. Did you forget to push perhaps?

nmadruga:
> > Changes you suggested done and tested @CarlosNZ
> 
> I'm not seeing any changes, and I'm still getting `isUserGranted: true` for all permissions when no orgId. Did you forget to push perhaps?

Sorry for the confusion. The message you quoted was **prior** to the suggestion on changing isUserGranted, and I haven't worked on this change yet. Sorry it has been 4 days, but that was counting weekend... I'll try to work on this today in the evening.

CarlosNZ:
One other small improvement that might be useful here is to return the list of permissions sorted alphabetically. The list seems a bit randomly ordered at the moment (presumably ordered by ID):
Screen Shot 2021-12-01 at 9 56 24 PM

nmadruga:
@CarlosNZ changes done to:
- When querying for `orgId` only (not passing `username`) in templatePermissions array set `isUserGranted` to false and revert from setting permisisonNames in `availablePermissions` instead of `grantedPermissions`.
- Order by permissionsName on query

src/components/permissions/routes.ts
```diff
@@ -99,20 +99,46 @@ const routeUserPermissions = async (request: any, reply: any) => {
...
+    )
+    console.log('userExistingPermissions', userExistingPermissions)
+
+    grantedPermissions = userExistingPermissions.reduce(
+      (grantedPermissions, { permissionName }) => {
+        if (!grantedPermissions.includes(permissionName)) grantedPermissions.push(permissionName)
+        return grantedPermissions
+      },
+      []
+    )
```
CarlosNZ, line 120:
```suggestion
 grantedPermissions = Array.from(new Set(userExistingPermissions.map((p) => p.permissionName))).sort()
```
Personally I find this easier to understand (and using Sets to generate lists of unique items is more efficient than searching through the same array over and over again)
Note also the additional `.sort()` these lists weren't coming back in alphabetical order

src/components/permissions/routes.ts
```diff
@@ -99,20 +99,46 @@ const routeUserPermissions = async (request: any, reply: any) => {
...
+    availablePermissions = Object.values(templatePermissionRows)
+      .filter(({ permissionName }) => !grantedPermissions.includes(permissionName))
+      .reduce((availablePermissions, { permissionName }) => {
+        if (!availablePermissions.includes(permissionName))
+          availablePermissions.push(permissionName)
+        return availablePermissions
+      }, [])
```
CarlosNZ, line 127:
```suggestion
    availablePermissions = Array.from(
      new Set(
        Object.values(templatePermissionRows)
          .filter(({ permissionName }) => !grantedPermissions.includes(permissionName))
          .map((p) => p.permissionName)
      )
    ).sort()
```
Similarly...

src/components/permissions/routes.ts
```diff
@@ -99,20 +99,46 @@ const routeUserPermissions = async (request: any, reply: any) => {
...
+  } else {
+    // Get permissions for organisation without association with as user
+    const orgExistingPermissions = await databaseConnect.getOrgTemplatePermissions(isSystemOrg)
+    availablePermissions = orgExistingPermissions.reduce(
+      (availablePermissions, { permissionName }) => {
+        if (!grantedPermissions.includes(permissionName)) availablePermissions.push(permissionName)
+        return availablePermissions
+      },
+      []
+    )
```
CarlosNZ, line 137:
```suggestion
  availablePermissions = Array.from(new Set(orgExistingPermissions.map((p) => p.perm))).sort()
```
And?

src/components/permissions/routes.ts
```diff
@@ -135,15 +161,10 @@ const routeUserPermissions = async (request: any, reply: any) => {
...
-    .reduce((availablePermissions, { permissionName }) => {
-      if (!availablePermissions.includes(permissionName)) availablePermissions.push(permissionName)
-      return availablePermissions
-    }, [])
-
   return reply.send({
-    templatePermissions,
+    templatePermissions: templatePermissions.sort(({ name: aName }, { name: bName }) =>
+      aName.localeCompare(bName)
+    ),
```
CarlosNZ, line 167:
```suggestion
    templatePermissions,
```
It was really only the two text lists (granted/available) that needed sorting, since they're the ones that will actually appear in the UI (as checkboxes).

src/types.ts
```diff
@@ -203,3 +203,11 @@ export interface Organisation {
 }
 
 export interface UserOrg extends User, Organisation {}
+
```
CarlosNZ, line 206: There's a types file in the "permissions" subfolder with other types used here already in it, so probably makes sense to keep this grouped with them (unless its used elsewhere?)

src/types.ts
```diff
@@ -203,3 +203,11 @@ export interface Organisation {
...
+export interface PermissionDetails {
+  id: number
+  name: string
+  description: string
```
CarlosNZ, line 210: Shouldn't `displayName: string` also be a property on here?

CarlosNZ (changes requested):
Cool, just some minor suggestions, mainly stylistic preference.

src/components/permissions/routes.ts
nmadruga: Cool. That's a good idea : )

src/types.ts
nmadruga: Oh, ok - good point. I'll move to there and also add `displayName` in it

nmadruga:
@CarlosNZ Made all the changes you suggested... Although now the sorting doesn't seem to be working on your latest template. So I think it was using the `templatePermissions` object. It might be better to revert commit b4f77e2

src/types.ts
CarlosNZ: I'm not actually sure how you got away with not having this and not getting a typescript error when you added it to the object 😄

CarlosNZ:
> @CarlosNZ Made all the changes you suggested... Although now the sorting doesn't seem to be working on your latest template. So I think it was using the `templatePermissions` object. It might be better to revert commit [b4f77e2](https://github.com/openmsupply/application-manager-server/commit/b4f77e2f5892c112a7ff550edcac33051ba86b83)

Ah yes, you're right. It's using this field, not the "text" fields, that's right. However, can we make the sort as an `ORDER BY "permissionNameId"` clause on the actual SQL query?

nmadruga:
> > @CarlosNZ Made all the changes you suggested... Although now the sorting doesn't seem to be working on your latest template. So I think it was using the `templatePermissions` object. It might be better to revert commit [b4f77e2](https://github.com/openmsupply/application-manager-server/commit/b4f77e2f5892c112a7ff550edcac33051ba86b83)
> 
> Ah yes, you're right. It's using this field, not the "text" fields, that's right. However, can we make the sort as an `ORDER BY "permissionNameId"` clause on the actual SQL query?

I tried that and it didn't work before. You can try out if you would like. I'm ok with just using the sorting in templatePermisisons

CarlosNZ:
> I tried that and it didn't work before. You can try out if you would like. I'm ok with just using the sorting in templatePermisisons

Sure, just do it as a JS .sort(), that's fine

CarlosNZ:
> I tried that and it didn't work before. You can try out if you would like. I'm ok with just using the sorting in templatePermisisons

Sorry, I just did it -- `SELECT * FROM permissions_all
      WHERE  "isSystemOrgPermission" = $1
      ORDER BY "permissionName"
      ` seems to work fine :)

CarlosNZ (approved):
Good to go! Thanks for the changes. :)

PR #632, opened 2021-11-29
#429B update action to regenerate review assignments

### LINKED ISSUE #429: Generate assignmentJoins for current user on all available applications after user is associated to permission to review
Run async the existing action `generateReviewAssignment` for each application and the current stage/level to generate the reviewAssignmentJoins

#### Create NEW "Parent" action "refreshReviewAssignments":

- takes "userId" as parameter
- searches for:
  1. all applications with review assignments for this user
  2. all applications that the user COULD have assignments for (i.e. based on permissions)
- from that, build a list of application Ids
- iterates over the application Ids, calling existing "generateReviewAssignments" action for each one sequentially
- The above iteration should be run sequentially, but this parent action can (maybe optionally) be run Async so it doesn't hold up the submission trigger and cause a time-out

##### Changes to EXISTING "generateReviewAssignments" Action:

- if only an "application Id" is provided (i.e. not a reviewId), then the first thing the action should do is lookup the highest review stage/level for this application from the review_assignments table. 
  - If there are none, it's assumed to be an application submission and the action proceeds the same as it does now
  - if there *are* some, then it's assumed that this is a "regeneration" and it should run for the **current** (i.e. highest) review stage/level and not the **next** one which would be the case when supplied with a reviewId (as in a review submission)
- Further down, the action needs to compare the newly calculated reviewAssigments with the existing ones, and DELETE any from the database that are no longer correct

### Considerations

- Do we need to have a flag somewhere ensuring that multiple instances of this Action can't run simultaneously? Could it cause problems if it did? If so, how do we do this? (Perhaps it checks for the entries in "action_queue" with "actionCode = generateAllReviewAssignments" AND status = PENDING, then doesn't continue until there are none left, or else it times out after one minute)

nmadruga:
Partly fixes #429

### Changes
- When `generateReviewAssignment` runs now it can be either to **generate** reviewAssigmnents (as previously) or **update** existing reviewAssignments after permissionNames were joined/unjoined to reviewers (users).
- Correctly list the review assignments that were updated/created after granting the same permissionName required for an application/stage/level. And also start listing the review assigments (and `reviewAssignmentAssignerJoins` for assignments) 
that were deleted after revoking a permissionName that was required for an application/stage/level.
- New tests to cover new cases added to `plugins/action_generate_review_assignment_records/src/generateReviewAssignments.test.ts`


CarlosNZ:
Looks good, but because this action has incredibly convoluted code, it's hard to figure out if it's all working as in should. 

In particular, I'm wondering about this case, from our agreed solution:
> if there are some, then it's assumed that this is a "regeneration" and it should run for the current (i.e. highest) review stage/level and not the next one which would be the case when supplied with a reviewId (as in a review submission)

I can't figure out where this is being done. It looks like the "existingReviewAssignments" fetch is occuring after the stage/level calculation is being done, but wouldn't that be needed first for the above situation? My apologies if I'm just not understanding this correctly.

nmadruga:
> I can't figure out where this is being done. It looks like the "existingReviewAssignments" fetch is occurring after the stage/level calculation is being done, but wouldn't that be needed first for the above situation? My apologies if I'm just not understanding this correctly.

I think you are right Carl, I haven't actually added this code yet. I was working with the tests, and it seemed to be creating the new reviewAssignment on the current stage, but I wasn't sure how. I'll have a better look into this now.

nmadruga:
Ok - @CarlosNZ I have done the changes needed - I didn't have to delete ReviewAssignmentAssigner records in the end... When I deleted the ReviewAssignments these were also being removed already. So that was the problem.

I'm starting to work on a diagram, but it would be good to get this one tested today and an action added to the template to grantUserPermissions (running after the granted/revoked permissions) happened. Also for when an internal user is added. I can start on that if you would like.

plugins/action_generate_review_assignment_records/src/generateReviewAssignments.test.ts
```diff
@@ -112,6 +112,7 @@ test('Test: Submit Application ID#4001 - Stage 1 (Last level)', () => {
         reviewAssignmentIds: [1005, 1062, 1063, 1064, 1065],
         reviewAssignmentAssignerJoins: [],
         reviewAssignmentAssignerJoinIds: [],
+        removedAssignmentIds: [],
```
CarlosNZ, line 115: Not really important, but I notice some of the existing tests have a "templateId" parameter, which is not required. Might be best to get rid of it?

plugins/action_generate_review_assignment_records/src/generateReviewAssignments.ts
```diff
@@ -25,6 +26,7 @@ async function generateReviewAssignments({
   // Get application/reviewId from applicationData if not provided in parameters
   const applicationId = parameters?.applicationId ?? applicationData?.applicationId
   const reviewId = parameters?.reviewId ?? applicationData?.reviewData?.reviewId
+  const isRegenaration = parameters?.isRegenaration ?? false
```
CarlosNZ, line 29: Spelling: `isRegeneration`

CarlosNZ:
Screen Shot 2021-12-01 at 1 25 08 PM

You were asking about the `isReview` parameter. It's for when you're submitting a reviewId but you want to generate assignments for the first level anyway (so it's acts like you're submitting an application).

Considering that, you can probably consolidate `isReview` and `isRegeneration` into one parameter, since they almost do the same thing. Will help when I can see the exact logic written out, but hopefully we can get rid of one of them.

CarlosNZ:
Okay, so I'm having trouble getting the re-generation of "current" level assignments to work.

Testing with "dev" snapshot, using "ReviewTest" template. Got an application to Stage 2, level 2, all assignments created correctly along the way.

I can delete some of the Level 1 assignments and they get regenerated fine. But I can't get any of the level 2 assignments to regenerate, no matter what I set `isRegeration` to.

Using `/run-action` endpoint with the following parameters:
```
{
    "actionCode": "generateReviewAssignments",
    "applicationId": 4019,
    "parameters": {
        "applicationId": 4019,
        "isRegenaration": false
    }
}
```

It doesn't matter if I switch `isRegenation` to true or false, it only re-creates the Level 1 assignments.

Am I missing something?

plugins/action_generate_review_assignment_records/src/generateReviewAssignments.ts
nmadruga: Thanks!

plugins/action_generate_review_assignment_records/src/generateReviewAssignments.test.ts
nmadruga: Hmm. Sure. I'll take it off now : )

nmadruga:
@CarlosNZ Phew, finally done - and docs updated. Please have another look and see what you think.

Basically changed to what was agreed yesterday: 
- Removed `isReview` flag no longer required
- Rewritten code to make it clearar so sorting what **type** of generation at the start and calling helper functions to deal with each case - required since now one of the cases would be running a loop.
- New logic for `isRegeneration` to generate reviewAssignments for **all** levels from 1 to current (based on existing reviewAssignments) to be creating or deleting the required as matching with reviewers with permissions to REVIEW.

plugins/action_generate_review_assignment_records/src/generateReviewAssignments.ts
```diff
@@ -19,168 +21,299 @@ async function generateReviewAssignments({
...
+  console.log(
+    `Application ${applicationId} stage ${stageNumber}, current level ${currentReviewLevel}`
+  )
 
-      // Save review_assignment records to database
-      const reviewAssignmentIds = await db.addReviewAssignments(
-        Object.values(reviewAssignments) as ReviewAssignment[]
-      )
+  // Create array with levels [1,2..,N]
+  const arrayLevels = Array.from({ length: currentReviewLevel }, (v, k) => k + 1)
```
CarlosNZ, line 190: Fancy! 😉

CarlosNZ (changes requested):
Great -- that's a lot of work you did yesterday, but it's looking really good and everything seems to work in a way that makes sense (to me at least!) but without being overly complicated.

Couple of very minor suggestions.

We will need to keep an eye on this in the next couple of weeks, cos it is pretty complicated and there might still be unexpected issues, but I think it's pretty much good to go!

CarlosNZ:
Oh, one more thing I think this might need:

You raised a good point in [this comment](https://github.com/openmsupply/application-manager-server/pull/633#discussion_r759794184) about applications in "DRAFT" status. However if we include the "DRAFT" ones, then this Action will generate review_assignments for applications that haven't been submitted yet.

So what we should probably do is, in _this_ Action, do a check right at the top:
- if status == DRAFT AND there are no existing review_assignments, then exit immediately.

Do you think that will take care of it, or can you think of a better way? I could put this check in my Action and so we won't even pass those applications to this action, but I think it would be safer in here, as even if somehow a "DRAFT" application gets passed into this action accidentally, it still won't create invalid review assignments.

(Just looking at your code, you could move the `currentReviewLevel = (await db.getLastReviewLevel` up  to the top of the main function in order to do this check -- we already know status in applicationData -- then pass in the "current/highest" level into `generateForAllLevelsUntilCurrentLevel` as a parameter?)

plugins/action_generate_review_assignment_records/src/generateReviewAssignments.ts
```diff
@@ -19,168 +21,299 @@ async function generateReviewAssignments({
...
-      const reviewAssignmentAssignerJoinIds = await db.addReviewAssignmentAssignerJoins(
-        reviewAssignmentAssignerJoins
+        level,
+        stageId,
+        stageHistoryTimeCreated,
+        templateId,
+        numReviewLevels
       )
+    )
+  ).then((values) => values.map((reviewResult) => result.output.levels.push(reviewResult)))
```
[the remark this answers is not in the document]
nmadruga: Thanks! That's sounds better.

plugins/action_generate_review_assignment_records/src/generateReviewAssignments.ts
```diff
@@ -19,168 +21,299 @@ async function generateReviewAssignments({
...
-        nextLevelReviewers,
-        applicationId,
-        levelNumber: nextReviewLevel,
-        isLastLevel,
-        isLastStage,
-        stageId,
-        stageNumber,
-        timeStageCreated: stageHistoryTimeCreated,
-      })
+const generateForAllLevelsUntilCurrentLevel = async (
```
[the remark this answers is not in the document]
CarlosNZ: 1fbdfbf7

plugins/action_generate_review_assignment_records/src/generateReviewAssignments.ts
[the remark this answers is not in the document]
CarlosNZ: 64a2555c

CarlosNZ (approved):

nmadruga:
@CarlosNZ merging this now and going to merge `develop` into `main`.

PR #633, opened 2021-11-29
#429A - new Action to regenerate user(s)' review assignments

CarlosNZ:
~~Still got some work to do on this (mainly around error handling, output shape and docs), but just want to get this Draft up so @nmadruga can use it to work with her changes to the existing generateReviewAssignments action.~~
(Will make additional tweaks once #632 is finalised, if necessary)

Easiest way to test it is to call it from the `/run-action` endpoint, so you can quickly fire it repeatedly with different userIds. Body JSON for query would be something like:
```
{
    "actionCode": "refreshReviewAssignments",
    "parameters": {
        "userId": [2]
    }
}
```
Note: "userId" parameter can be *either* a number *or* an array of numbers (for multiple users at once). Also -- if no "userId" is provided, it will refresh ALL active applications (so we can run this Action globally if we want)


nmadruga (changes requested):
The new action isn't showing in the list of actions to choose on the Template Builder. Could you have a look on this please? I was trying to add this to a latest version of demo_angola to run after grant/revoke one user's permissions : )
Screen Shot 2021-11-30 at 4 27 35 PM

CarlosNZ:
> The new action isn't showing in the list of actions to choose on the Template Builder. Could you have a look on this please? I was trying to add this to a latest version of demo_angola to run after grant/revoke one user's permissions : )

It's there after a full refresh:
Screen Shot 2021-11-30 at 4 33 47 PM

plugins/action_generate_review_assignment_records/plugin.json
```diff
@@ -3,15 +3,9 @@
...
-  "output_properties": [
-    "reviewAssignmentIds",
-    "reviewAssignmentAssignerJoins",
-    "reviewAssignmentAssignerJoinIds",
-    "currentReviewLevel",
-    "nextReviewLevel"
-  ],
+  "required_parameters": ["userId"],
+  "optional_parameters": [],
+  "output_properties": ["updatedApplications", "reviewAssignments"],
```
nmadruga, line 8: Oh, I think that was the wrong action updated. No?

plugins/action_refresh_review_assignments/src/databaseMethods.ts
```diff
@@ -0,0 +1,58 @@
...
+  },
+  getApplicationsFromUserPermissions: async (userIds: number[]) => {
+    const text = `SELECT DISTINCT app.application_id
+      FROM permissions_all perm
+      JOIN application_stage_status_latest app
+      ON perm."templateId" = app.template_id
+      WHERE "userId" = ANY($1)
+      AND "permissionType" = ANY('{REVIEW, ASSIGN}')
+      AND app.outcome = 'PENDING'
+      AND app.status <> 'DRAFT'
```
nmadruga, line 25:
```suggestion
      AND app.status <> 'DRAFT'
```

Wondering about this, if the application is back for the Applicant to **make changes** then the application woudn't be listed as one of the ones required to update the reviewAssignment?

plugins/action_refresh_review_assignments/src/refreshReviewAssignments.ts
```diff
@@ -0,0 +1,68 @@
...
+  applicationData,
+  DBConnect,
+}: ActionPluginInput) {
+  const db = databaseMethods(DBConnect)
+
+  // If no userId parameter, we will update ALL active applications
+  const shouldRefreshAll = !parameters?.userId
+
+  // Can handle input of either single userId or an array of userIds
+  const userIds = Array.isArray(parameters?.userId) ? parameters.userId : [parameters.userId]
```
nmadruga, line 17: Maybe it would throw an error here when parameters.userId is undefined?

nmadruga (commented):
Just a few comments - can be changed on PR as both have been merged now to make merging the full work together easier : )

plugins/action_refresh_review_assignments/src/databaseMethods.ts
CarlosNZ: Yeah, good point. Was just concerned about what would happen if you tried to generate review assignments for an application that had never been submitted yet. I think your action would create them, wouldn't it (which wouldn't be good)?

plugins/action_refresh_review_assignments/src/refreshReviewAssignments.ts
CarlosNZ: Okay, I've changed it to put the `?.` operator in all references here. 6a41f29b

plugins/action_generate_review_assignment_records/plugin.json
CarlosNZ: Yes, sorry. When I merged this into your branch I used your one instead of this to resolve the conflict, so should be all good over there.

plugins/action_refresh_review_assignments/src/databaseMethods.ts
CarlosNZ: @nmadruga see my suggestion for dealing with this: https://github.com/openmsupply/application-manager-server/pull/632#issuecomment-984213967

CarlosNZ:
Closing this now, it's been merged to #632 . Will do any further changes in there.

PR #634, opened 2021-11-30
Remove NPX from postgraphile launch

CarlosNZ:
Just realised this was still in here, whereas the `postgraphile_launch.sh` file doesn't. Changed for consistency.


nmadruga (approved):
LGTM

PR #635, opened 2021-11-30
Org-only permissions

CarlosNZ:
I think this is mostly what's required for this to work. 

When fetching template permissions that go into the "userInfo" object and associated JWT, it will also fetch permissions that are associated with just the organisation and have `null` in the "user_id" field (of permission_join).

Test it by manually changing one of the entries in `permission_join` to have `null` in the user_id, and notice that it should allow access for any user belonging to that org. Should work for applicants AND reviewers (not that we'd use if for reviewers much).

### DIFF
--- a/src/components/databaseConnect.ts
+++ b/src/components/databaseConnect.ts
@@ -108,6 +108,8 @@ class DBConnect {
 
   public getUserTemplatePermissions = PostgresDB.getUserTemplatePermissions
 
+  public getOrgTemplatePermissions = PostgresDB.getOrgTemplatePermissions
+
   public getAllPermissions = PostgresDB.getAllPermissions
 
   public getTemplatePermissions = PostgresDB.getTemplatePermissions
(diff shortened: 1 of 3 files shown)


nmadruga (commented):
Ok - the code looks pretty simple. So no problem with that. I'm going to test now, wondering which would be the best snapshot to test this `core_action`? `dev`? 

Maybe a good moment to update [this docs](https://github.com/openmsupply/application-manager-web-app/wiki/Actions-on-Applications#applicant-home-page-or-list):
> Note: For now the front-end allows a user to see any application, not only restricted by user permission. Later, when restrictions apply, only users with appropriate permissions will be able to see other users' application statuses.

nmadruga (approved):
Ok - that's all good!

Also checked that the `join company` permission was being granted to user/org so I think we can sort that out also only with changes to config - in order to achieve the result expected for UserRegistration. + Join company, that **should not** list application to Join another company while user is logged as a current company.

PR #636, opened 2021-12-07
Update plugin configs and docs

CarlosNZ:
I realised when I was investigation [this issue](https://github.com/openmsupply/application-manager-web-app/issues/1072) that several of our back-end plugins had out-of-date or incorrect parameters listed in their `plugin.json` configs, which is why the "parameters" UI in the front-end template builder was showing incorrect parameters for some Actions.

So I've updated that and also made sure the docs were correct for each one.

And I realised I hadn't previously written any documentation for the "scheduleAction" plugin, so I did that quickly too.


documentation/List-of-Action-plugins.md
```diff
@@ -535,6 +543,34 @@ The output object `notification` contains all the fields from the notification r
...
+
+Each scheduled event can be saved with an `eventCode` -- this is used by Actions that are triggered by this event to determine *which* action should be fired for any given `ON_SCHEDULE` trigger on each template type. Every Action defined for each template has an optional `scheduledActionCode` field, which can be used to match specific events. If no event code is provided, then *every* action for that template type with an `ON_SCHEDULE` trigger will be executed.
+
+By default, when an event is saved, the `outputCumulative` object from the `scheduleAction` action (including outputs of previous actions in the sequence) is stored in the `data` field, but additional data can be added to this as well. Then when the scheduled action runs, it gets this data passed in as `outputCumulative`, so it has access to the output info of previous actions -- this means it effectively can continue as though it were part of the same sequence of actions, but with a "pause" due the the schedule.
```
nmadruga, line 567:
```suggestion
By default, when an event is saved, the `outputCumulative` object from the `scheduleAction` action (including outputs of previous actions in the sequence) is stored in the `data` field, but additional data can be added to this as well. Then when the scheduled action runs, it gets this data passed in as `outputCumulative` -- this means it effectively can continue as though it were part of the same sequence of actions, but with a "pause" due the schedule.
```

documentation/List-of-Action-plugins.md
```diff
@@ -535,6 +543,34 @@ The output object `notification` contains all the fields from the notification r
...
+
+The `cancel` parameter is a way to prevent a previously scheduled event from occurring. Any existing event that has matching `applicationId` and `eventCode` will be set to inactive without ever firing. In practice, though, targeting an event by `applicationId` is often not feasible, so the preferred way to cancel a scheduled event is to just apply an appropriate Condition to the subsequent action -- so the event is still triggered, but the matching action won't occur if the condition is not met (e.g. don't expire a product if registration has been renewed)
```
nmadruga, line 569:
```suggestion
The `cancel` parameter is a way to prevent a previously scheduled event from occurring. Using the action to run with the cancel will find existing event that has matching `applicationId` and `eventCode` will be set to inactive without ever firing. In practice, though, targeting an event by `applicationId` is often not feasible, so the preferred way to cancel a scheduled event is to just apply an appropriate Condition to the subsequent action -- so the event is still triggered, but the matching action won't occur if the condition is not met (e.g. don't expire a product if registration has been renewed)
```

nmadruga (approved):
Cool addition. It will be super helpful when listing action son Template Builder!

Just a few suggestions on schedule_action

CarlosNZ:
Cool, thanks. Changes made.

PR #639, opened 2021-12-10
Finally! (?) fix missing FDA logo with fresh Docker-compose instance

CarlosNZ:
@andreievg  showed me what I needed to do to finally fix this properly and have it work with docker-compose.

Same fix as before, but now it happens in the "entry" script rather than the setting up "database.sh" script, which gets overwritten once the docker-compose volume is attached.

Testing and working with docker-compose (with fresh database reset) on instance 50006.


nmadruga (approved):
LGTM

PR #640, opened 2021-12-13
Bug fix -- Outcomes endpoints

CarlosNZ:
Discovered a problem when table names with multiple words (e.g. `template_action`) were used -- there were problems caused by conversion between camelCase and snake_case that weren't apparent when using only single-word table names. 

Anyway, this fixes that problem. Use Postman to test.


nmadruga (approved):
Looks good - I think we can't spent too long on testing now. If you saw this and it's fixed I trust you.

PR #643, opened 2021-12-16
Outcomes -- allow org-only permissions

CarlosNZ:
Based on chat with @jagoje on Telegram:
Screen Shot 2021-12-16 at 2 32 22 PM

I realised that it wouldn't work with org-only permissions without a small tweak to the `getUserOrgPermissionNames` query to also consider `userId IS NULL` permission joins.


nmadruga (approved):
LGTM

Issue #645, opened 2022-01-20
Database changes for Internal Notes tab


CarlosNZ:
Part of front-end issue: https://github.com/openmsupply/application-manager-web-app/issues/1079

- New table “application_note”
  Fields:
  - application_id (FK)
  - user_id (FK)
  - org_id (FK)
  - timestamp
  - comment (markdown text)
  - files (JSON, same format as FileUpload response)
  What permissions do we need for this table?

PR #646, opened 2022-01-20
#645 #1082F Changes for Review tabs

CarlosNZ:
Implements #645 and the back-end requirements for [front-end #1082](https://github.com/openmsupply/application-manager-web-app/issues/1082)

See those issues for the specifics of what's been done.

One thing to note: I have **removed** the "application_section" table from the schema, since we've not been using it -- it's basically an orphan table. So I've just put the new table in place of that one in the build schema scripts.

To test:
- Load new schema with `yarn database_init`
- Upload this carbone template using the `/upload` endpoint and note the unique ID returned: [doc_action_demo_additional_data.odt](https://github.com/openmsupply/application-manager-server/files/7902207/doc_action_demo_additional_data.odt)
- Call the `/run-action` endpoint with the following body JSON (but with the "docTemplateId" replaced with the unique Id from the last step):
  ```
  {
      "actionCode": "generateDoc",
      "parameters": {
          "docTemplateId": "lL1NoI4RGI631y8krbhwL",
          "additionalData": {
              "firstName": "Carl",
              "lastName": "Smith",
              "applicationSerial": "XYZ-1234",
              "status": "ALL GOOD",
              "stage": "ASSESSMENT"
          },
          "description": "This is a demo file",
          "isOutputDoc": true
      }
  }
  ```
  The PDF should be generated and the "file" table will have the "description" and "is_output_doc" fields written as specified (try changing them above and check the results are correct)

---

**UPDATE**: I've also made some additional tweaks to the `/upload` endpoint -- now if you try and upload something with the same unique-id as an existing file, it'll update the file details. This way we can make edits to the "description" (or other) fields using the same endpoint, which will be handy for the adjustments to the fileUpload plugin and/or editing comments in the Internal Notes tab. 0d904d17


CarlosNZ:
FYI -- there should be no problem merging this before the front-end work is done -- it won't conflict with any existing functionality

CarlosNZ:
> I'm still trying to undestand the goal of this one. Should I be looking with the Front-end PR? Is that still on progress or would it be added later? That is for when the generated documents starts to show up in the UI for the Documents tab, but we would also need to update the generation of specific documents to use the new flags.

Sorry, I chucked in some extra stuff here not directly connected. 

The main thing: just the database changes.

However, I was also playing around with the file upload plugin (in the front-end) and I realised we'll need a way to "update" existing files' "description" field -- for when reviewers upload files to the "Notes" tab and want to make comments.. So I've added that functionality to the "/upload" endpoint. It's not perfect (you currently have to include a file in the query otherwise it fails), but it gives us the basic back-end functionality we'll need when we make the Notes tab -- even if we don't end up using the FileUpload component. If you want to test it, see the "Testing" notes above.

Make sense?

CarlosNZ:
Can this be merged now, @nmadruga ? (There's no breaking changes affecting front end at all)

nmadruga:
Yes, all good to merge this

nmadruga (approved):

PR #651, opened 2022-02-03
Add some new actions and a certificate to core_templates snapshot

nmadruga:
Please test:
- [ ] Create new Applicant user
- [ ] Create new Sample Templates application
- [ ] Approve (as Admin Admin)
- [ ] Check that a certificate is generated and is showing on Documents tab
- [ ] Check that the expire runs after 3 days (change the expiry time to test using minutes if best)


CarlosNZ:
Since the git diff on snapshot.json is not particularly helpful here, can you just quickly say which templates you changed?

> Add some new actions and a certificate to core_templates snapshot

Any actions added that should be recorded in the "List of Core Actions" docs?

CarlosNZ:
> * Create new Applicant user

Failed to grant the "applyGeneral" permission. I think this is because we now need to explicitly have the `orgName: null` parameter in the grantPermissions action, which we added recently to prevent accidental over-granting when org failed.

From action_queue log: `Invalid or missing orgId or orgName`

> * Create new Sample Templates application

Seems fine ✅

> *Approve (as Admin Admin)

When I'm logged in as Admin, I see the submitted application (from my new user) from an application POV in Dashboard and List. No option to assign or do anything else.
Screen Shot 2022-02-08 at 4 00 43 PM

CarlosNZ (changes requested):
See my longer comment

nmadruga:
> > * Create new Applicant user
> 
> Failed to grant the "applyGeneral" permission. I think this is because we now need to explicitly have the `orgName: null` parameter in the grantPermissions action, which we added recently to prevent accidental over-granting when org failed.
> 
> From action_queue log: `Invalid or missing orgId or orgName`
> 
> > * Create new Sample Templates application
> 
> Seems fine ✅
> 
> > *Approve (as Admin Admin)
> 
> When I'm logged in as Admin, I see the submitted application (from my new user) from an application POV in Dashboard and List. No option to assign or do anything else. Screen Shot 2022-02-08 at 4 00 43 PM

Yes. I noticed that too, I have done the fix in another branch and will upload the template here  🤞

nmadruga:
> 
> > * Create new Sample Templates application
> 
> Seems fine ✅
> 
> > *Approve (as Admin Admin)
> 
> When I'm logged in as Admin, I see the submitted application (from my new user) from an application POV in Dashboard and List. No option to assign or do anything else. Screen Shot 2022-02-08 at 4 00 43 PM

I noticed the same in the current `core_snapshot` when I tried the Join company application = Admin had only the option to click and view the Review home page. It needs investigation.

> When I'm logged in as Admin, I see the submitted application (from my new user) from an application POV in Dashboard and List. No option to assign or do anything else.
> Screen Shot 2022-02-08 at 4 00 43 PM

I think that's fixed now. Admin admin has permission to `apply` so check if URL is not showing the view for user=applicant and I have also changed the permission to review in that template to `reviewScreening` since the one used before `reviewGeneral` wasn't granted to that user.

nmadruga:
I have also fixed the problem with granting user permission without an org as you noted. But now I saw there is a problem to modify record which I haven't found out what is causing it. If you have time please see if you find the reason, it would be really good to have this core_snapshot updated with most recent changes

CarlosNZ:
> But now I saw there is a problem to modify record which I haven't found out what is causing it.

I think it's to do with this:
Screen Shot 2022-02-09 at 4 54 52 PM

You're trying to get "orgId" from applicationData, but there isn't one, as this form is filled out before the org is created.
You'll probably need to get "orgId" from the outputCumulative from the previous "modifyRecord" which created the org in the database.

I'll see if I can fix it.

CarlosNZ:
Screen Shot 2022-02-09 at 4 53 36 PM

The approval email for Sample Template is quite out of date:
- mFlow -> conforma
- Company name not showing up, probably for the same reason as the previous comment
- {Regulatory Authority} ?

CarlosNZ:
> You're trying to get "orgId" from applicationData, but there isn't one, as this form is filled out before the org is created. You'll probably need to get "orgId" from the outputCumulative from the previous "modifyRecord" which created the org in the database.

This was indeed the problem, but this field shouldn't have actually caused it to fail. The failing one was the GraphQL query to get the registration value based on the orgID -- with missing orgID, gql failed.

Have fixed these issues (including the outdated email notification) and pushed a new version to this branch.

Will check out the rest later.

nmadruga:
> Screen Shot 2022-02-09 at 4 53 36 PM
> 
> The approval email for Sample Template is quite out of date:
> 
> * mFlow -> conforma
> * Company name not showing up, probably for the same reason as the previous comment
> * {Regulatory Authority} ?

Looking good now : ) I added a new version of Core Template: version 5
Screen Shot 2022-02-10 at 8 25 02 PM

Issue #652, opened 2022-02-04
File deletion: write trigger to automatically delete the file when DB record is deleted


CarlosNZ:
This would be handy. Similar to "Action" triggers, we just make a new Node script to listen to PostGres trigger on File table.

Trigger: AFTER DELETE on TABLE 'file'
Node script: listen for trigger notification, delete the actual file

PR #653, opened 2022-02-04
#1079F #652 Back-end changes for Notes tab

CarlosNZ:
Back-end changes required for Notes tab (front-end issue https://github.com/openmsupply/application-manager-web-app/issues/1079)

Front-end PR: https://github.com/openmsupply/application-manager-web-app/pull/1102

Also included #652 as it was pretty simple and makes deleting comments with files much nicer.

### Summary of changes
- removed "files" field from "application_note" table and added "application_note_id" FK link from files table. This allows us to have multiple files linked to one note (and GraphQL follows the link quite nicely)
- add "application_node_id" to file upload route, so we can insert this info when posting notes with files
- add trigger to "file" table to notify server on file deletions
- add new listener to postgresConnect to listen for aforementioned file deletions
- function to delete files based on above listener

### Also:
- changed "cleanupFiles" Action to no longer delete the actual files, since this new trigger will take care of it
- added documentation for "cleanupFiles" Action, since we missed it before! 

Front-end PR coming shortly...


CarlosNZ:
Just realised that when the files use the "generic" thumbnail icons, these files are being deleted as well which means they don't show up next time they're required.

Will fix. 😳

CarlosNZ:
> Will fix. 😳

Done! 6af3f560

nmadruga (approved):
Looks good. Checking with front-end now

PR #654, opened 2022-02-09
#1095F Check triggers endpoint

CarlosNZ:
Back-end requirements for [front-end issue #1095](https://github.com/openmsupply/application-manager-web-app/issues/1095), as outlined in this comment: https://github.com/openmsupply/application-manager-web-app/issues/1095#issuecomment-1032311757

Creates endpoint `/check-triggers` to return the trigger status of ALL triggers associated with a given application.

PR #656, opened 2022-02-10
Add new field to keep assigned section in ReviewAssignment

### LINKED ISSUE #655: Add field to keep assignedSections in the reviewAssignment table
This change would be great to solve some issues found related to the existing need to be deducing which sections are assigned to a reviewer in the UI. There is only 1 reviewAssignment per reviewer/application and all sections assigned are linked to it, but there is no simple way to find which ones: currently we need to did it from `assignment.reviewQuestionsAssignment` array and find the `element.section` that are part of the assignment.

### Current issue
There are at least 2 issues that could be resolved more easily by adding the array of current assigned sections to this record. The issues are described in more details on https://github.com/openmsupply/application-manager-web-app/issues/1110, https://github.com/openmsupply/application-manager-web-app/issues/1109 and https://github.com/openmsupply/application-manager-web-app/issues/1081.

#### Related issue 1
The problem described in issue [Front-end#1110](https://github.com/openmsupply/application-manager-web-app/issues/1100)
is basically that from the 3 sections (see image) only Section 1 & 2 are really assigned to the current reviewer (admin) to start.
Note that Section 3 is already submitted by Samanta reviewer, so it doesn't belong to Admin. But all sections are showing the **start** action erroneously... And after clicking on the start this leads to a major problem which doesn't allow this user to submit the review because the App is expecting all sections to be filled.
Screen Shot 2022-02-11 at 10 22 56 AM

We only have one ReviewAssignment linked to all the assigned questions to a user (see this console.log with more details). That's what is currently received for example in `ReviewSectionRowAction` to display the action to the Reviewer on the Assignments tab UI:
Screen Shot 2022-02-11 at 10 23 53 AM

##### Related issue 2
The last 3 bugs described in issue [Front-end#1081](https://github.com/openmsupply/application-manager-web-app/issues/1081) description are also caused by the same problem. One of the bugs is exactly the same as above - just described at a different time, and the next ones are consequences of this same problem on the ReviewPage.

In each section on the ReviewPage it is also only checking that if is an Action START associated, without checking for which sections, so we would need to also check the sections, which would be much simpler if it was only an array on the structure.

##### Related issue 3
The other but [Front-end#1109](https://github.com/openmsupply/application-manager-web-app/issues/1109) is also cause because we currently just UNASSIGN all sections even when the re-assignment was supposed to remove one of the section of a reviewer. It would be good after this change to set the un-assignment to either say UNASSIGN all or just remove some sections/questions of the assignment.

nmadruga:
Fixes #655 

Based on PR #654 

Needs to use with Front-end [PR#655B](https://github.com/openmsupply/application-manager-web-app/pull/1112)

### Changes
- Add new field `assignedSections`
- (wip) New trigger to run on Update that checks if `assignedSections` are empty but status is still ASSIGNED


CarlosNZ (dismissed):
LGTM!

You doing the trigger in a different PR, obviously?

database/buildSchema/28_review_assignment.sql
```diff
@@ -31,6 +31,7 @@ CREATE TABLE public.review_assignment (
     application_id integer REFERENCES public.application (id) ON DELETE CASCADE NOT NULL,
     template_id integer GENERATED ALWAYS AS (public.review_assignment_template_id (application_id)) STORED REFERENCES public.template (id) ON DELETE CASCADE,
     allowed_sections varchar[] DEFAULT NULL,
+    assigned_sections varchar[] DEFAULT array[]::varchar[],
```
CarlosNZ, line 34:
```suggestion
    assigned_sections varchar[] DEFAULT ARRAY[] ::varchar[],
```

This is what the linter auto-formats it to.

database/buildSchema/28_review_assignment.sql
```diff
@@ -42,13 +45,36 @@ CREATE TABLE public.review_assignment (
...
+        UPDATE
+            public.review_assignment
+        SET
+            status = 'AVAILABLE'
+        WHERE
+            id = NEW.id;
+        RETURN NULL;
+    END;
+$review_assignment_event$
+LANGUAGE plpgsql;
```
CarlosNZ, line 62: Our assignment mutations *should* be setting this correctly, right? So in theory this shouldn't need to execute.

Or are you just changing the sections in the front-end when un-assigning and leaving the "status" field to this trigger function?

database/buildSchema/28_review_assignment.sql
```diff
@@ -42,13 +45,36 @@ CREATE TABLE public.review_assignment (
...
+-- TRIGGER (Listener) on review_assignment table: To update trigger
 CREATE TRIGGER review_assignment_trigger
     AFTER INSERT OR UPDATE OF trigger ON public.review_assignment
     FOR EACH ROW
     WHEN (NEW.trigger IS NOT NULL AND NEW.trigger <> 'PROCESSING' AND NEW.trigger <> 'ERROR')
     EXECUTE FUNCTION public.add_event_to_trigger_queue ();
 
+-- TRIGGER (Listener) on review_assignment table: To set status accordingly to assignedSections
+CREATE TRIGGER review_assignment_trigger2
+    AFTER UPDATE OF trigger ON public.review_assignment
```
CarlosNZ, line 73:
```suggestion
    AFTER UPDATE OF assigned_sections ON public.review_assignment
```

Wrong target column -- this fixes it.

CarlosNZ (changes requested):

database/buildSchema/28_review_assignment.sql
nmadruga: Sorry - this work is partially done. II would like to give this a better solution in a PR to fix [this](https://github.com/openmsupply/application-manager-web-app/issues/1109) Front-end issue. But yes, the idea is that the Front-end would send the reassigned sections to the reassign mutation and keep the status as ASSIGNED if there is still one. Or set the status to AVAILABLE when all sections have been un-assigned. But since this is a front-end solution I thought it would be good to add a safe guard here.
There is a problem still with the trigger, because the front-end would be setting trigger: ON_REVIEW_UNASSIGN when if encouters a fully unassignment reviewAssignment, but this safe-guard won't do it, so it wouldn't trigger an unassignment correctly.

Issue #642, opened 2021-12-15
Database migration code


CarlosNZ:
Placeholder -- need to discuss

We need some kind of migration code between versions to handle schema changes. When we just upgrade an instance but don't change the snapshot data, there needs to be some migration script to run and implement any schema changes.

CarlosNZ:
To summarise what we discussed, will add more here once I've started doing it and have a better handle on some of the considerations / use cases:

- TS script to run "node-postgres" queries in a linear, version by version sequence
- should run after snapshot data is loaded but before server (re)starts
- Use Nicole's front-end changes from the other day as example use case

PR #657, opened 2022-02-14
#642 Part 1 - versioning and auto release script

CarlosNZ:
In order to properly implement #642, we need to have the versioning be consistent between front-and-back end, github releases/tags and each of the "package.json" files, so this automates that process:

- source of truth for the current version number is the back-end package.json file
- we bump the version and automatically create a tag using the new `yarn release` script
- the script bumps the back-end (according to the type of release (patch, minor, major, prerelease)) and then sets the front-end to the same version and makes tags for both repos
- internally it mainly just uses the built-it [`yarn version`](https://classic.yarnpkg.com/en/docs/cli/version), which bumps the version and creates the tag
- Have reset the version back to `0.1.0`, so we should stick to proper semantic versioning going forward

Requires front-end PR https://github.com/openmsupply/application-manager-web-app/pull/1114
Will also need to run `yarn install` (for the "[readline-sync](https://www.npmjs.com/package/readline-sync)" user-prompt package)

Have updated documentation as well, so I'll just copy and paste from Demo-Server-Guide here:

> ## Prep
> 
> - ...
> - Ensure your back-end `.env` has a variable called `FRONT_END_PATH` with the *full* path to the front-end repo on your local system.  
>   e.g. `FRONT_END_PATH='/Users/<username>/GitHub/conforma/application-manager-web-app/'`
> - Ensure your local Docker installation is logged in with the **msupplyfoundation** account (Password in Bitwarden)
> 
> ## Bump the version and generate a tag.
> 
> - The `version` field in `package.json` is the source of truth for the current version number. Increases to version numbers should be done according to [semantic versioning](https://semver.org/), including "pre-release" versions (for testing purposes, for example).
> - Bump the version and create a tag by running the script `yarn release <version-type>`, where `version-type` is one of "--major", "--minor", "--patch" or "--prerelease". ("--pre-release" is the default if not specified.) This will also apply the new version number to the front-end (on whatever branch is currently checked out, so it should normally be `develop` or `main`). The tag is auto-generated by the `yarn version` command (run as part of this release script), and is just `v<version-num>` (e.g. `v0.1.5-2`)
> - Note: it is not necessary to create a "release" on Github for each tag unless we need to make a particular version publicly downloadable, but one can easily be created based on the current tag/version on Github if required.
> 
> ## Build (and upload to Docker hub)
> 
> `yarn dockerise <tag-name> [push]`
> 
> (If you don't specify any parameters, it will build the current `develop` branch.)
> 
> Note that the previous `yarn release` script will offer to start this build process automatically with the tag you've just created.
> ...

### Testing 

For testing purposes don't worry about making a bunch of tags. It's easy to delete them by using:
- `git tag -d <tag-name(s)>` (for local tags)
- `git push --delete origin <tag-name(s)>` (for remote tags)

You'll need to manually reset the version number in the "package.json" files though.

Would also suggest making a new branch from this one (and in front-end too), so we don't end up with a bunch of deleted tag commits in the main commit history (alternatively you can just squash them into one commit)

### Note:
- I've tested the script itself fairly well, but I haven't let the Docker build fully run. There's a chance that there could be an issue when the built version tries to access the version number from "package.json", so would be good to check that before we merge this.
- You'll note that there's a slightly hacky workaround to get the newly generated tag back into the Docker build script -- the Node script writes the tag name to a temp file, which is read by the 'release.sh' shell script and passed on to the Docker script. I couldn't find any other way to do this nicely. I could spawn the Docker script directly from the Node script, but it doesn't then print it's progress nicely to the same console window, which is pretty necessary in such a long-running process. Tried piping the output of the Node script or capturing it in a shell variable, but this messed up the interactivity of the Node script (console output was out of sync). This seemed like the most straightforward solution in the end.
- There's a change to the "database_info.sql" table in preparation for the migration script (will store current version in database, so the server can tell when it's changed and should run migration), but that's the only thing I've done towards that. Proper migration script will be in next PR.


CarlosNZ:
> * I've tested the script itself fairly well, but I haven't let the Docker build fully run. There's a chance that there could be an issue when the built version tries to access the version number from "package.json", so would be good to check that before we merge this.

Ran the full build, and ran the Docker image locally -- all seems good, correct version number shows up in the Footer 👍

nmadruga (approved):
Looks Good Carl, I haven't seem anything that requires attention. Seems to be very precise on what it should be doing.

PR #659, opened 2022-02-15
#642 Migration script

### LINKED ISSUE #642: Database migration code

CarlosNZ:
Fix #642 

Please see the added documentation (`Data Migration.md`) for how to use.

Requires `yarn install` (to load the [semver parser)](https://www.npmjs.com/package/semver))

I have highlighted one potential problem that will still exist when loading a snapshot that wants to load data into an incompatible schema -- the migration script will *not* handle that for reasons I explain in the documentation. I have also created an issue for further discussion about how to handle it.

The migration code included for v0.2.0 covers the recent changes Nicole made to the "review_assignment" table, and updating the data accordingly.


database/migration/migrateData.ts
```diff
@@ -0,0 +1,87 @@
...
+const migrateData = async () => {
+  let databaseVersion: string
+
+  try {
+    databaseVersion = (await DB.getDatabaseVersion()).value
+    // No migration if database version matches current version, but we still
+    // proceed if this is a manual migration
+    if (semverCompare(databaseVersion, version) >= 0 && !isManualMigration) return
+  } catch (err) {
+    // Nothing version in database yet, so run all migration
```
nmadruga, line 19: "No version in database yet"?

documentation/Data Migration.md
```diff
@@ -0,0 +1,53 @@
...
+This process is handled by the "migrateData.ts" script, which runs automatically every time the server (re-)starts, or it can be run manually in the dev environment.
+
+## 1. Upgrading an existing server in place
+
+After [upgrading the server](Demo-Server-Guide.md), when the server restarts, the migration process is automatic. The script checks the version of the app saved in the database, compares it against the current (new) version number, and acts accordingly. If the saved version matches the current version, nothing happens. Each migration block has a version number check on it, and will run if the database version is less than this version number. So even when upgrading a server by several versions at once, the migration script will run in sequence for every previous version that has been skipped over.
+
+For live servers, data migration after an upgrade should be an invisible process.
+
+## 2. Migrating data with snapshots
+
```
nmadruga, line 20: Maybe good to add here the existing limitation that `migration` doesn't run after importing the snapshot using the UI

database/migration/migrateData.ts
CarlosNZ: 😳

src/components/snapshots/takeSnapshot.ts
```diff
@@ -65,6 +68,12 @@ const takeSnapshot: SnapshotOperation = async ({
     // Copy prefs
     if (options?.includePrefs) execSync(`cp '${PREFERENCES_FILE}' '${newSnapshotFolder}'`)
 
+    // Save snapshot info (version, timestamp, etc)
+    await fs.writeFile(
+      path.join(newSnapshotFolder, `${INFO_FILE_NAME}.json`),
+      JSON.stringify(getSnapshotInfo(), null, ' ')
+    )
```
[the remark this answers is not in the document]
CarlosNZ: Can you think of any other data it might be good to include in the info.json file?

documentation/Data Migration.md
CarlosNZ: I've just simplified that whole paragraph in the docs to something much more generic -- there is a link to the issue if more detail is required. 1af63272

Though we should be fixing this soon, so will update docs properly then.

nmadruga (changes requested):
Just one code simplification

database/migration/migrateData.ts
```diff
@@ -0,0 +1,87 @@
...
+            })
+          currentReviewAssignment = { id: ra.id, assignedSections: new Set([ra.code]) }
+        }
+      })
+      // Don't forget about the last one
+      reviewAssignmentsWithSections.push({
+        id: currentReviewAssignment.id,
+        assignedSections: [...currentReviewAssignment.assignedSections] as string[],
+      })
+      await DB.addAssignedSections(reviewAssignmentsWithSections)
```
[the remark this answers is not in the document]
CarlosNZ: Okay, thanks. I've made a change that works now: 9480ddf7

Basically, you were saying that it's easier to build up an indexed object than an array of objects, cos you don't need to keep track of where you are in the array. Good point, have simplified it now, although there were a couple of caveats (such as object keys needing be strings). 

But looks nicer and working well, so thanks :)

PR #660, opened 2022-02-18
Fix setting assignedSections to null when status is AVAILABLE

nmadruga:
[Front-end PR](https://github.com/openmsupply/application-manager-web-app/pull/1124) in progress

### DIFF
--- a/database/buildSchema/28_review_assignment.sql
+++ b/database/buildSchema/28_review_assignment.sql
@@ -33,7 +33,7 @@ CREATE TABLE public.review_assignment (
     application_id integer REFERENCES public.application (id) ON DELETE CASCADE NOT NULL,
     template_id integer GENERATED ALWAYS AS (public.review_assignment_template_id (application_id)) STORED REFERENCES public.template (id) ON DELETE CASCADE,
     allowed_sections varchar[] DEFAULT NULL,
-    assigned_sections varchar[] DEFAULT NULL,
+    assigned_sections varchar[] DEFAULT array[]::varchar[] NOT NULL,
     TRIGGER public.trigger,
     time_updated timestamptz DEFAULT CURRENT_TIMESTAMP,
     level_number integer,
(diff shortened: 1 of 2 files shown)


CarlosNZ (commented):
I think we should probably leave it as non-nullable, default `[]`. The only reason we were changing it is so we could mutate it from the front-end, but we're using the trigger instead so no need. If it's NULL we don't need the trigger, if we're using the trigger its easier to deal with as [] rather than null.

Non-nullables are so much less hassle to deal with in the front-end, don't you think?

CarlosNZ:
Also -- need to add migration code (can be part of 0.2.0 script, since we haven't released that yet)

nmadruga:
Yes, Carl. Thanks for inputs. I agree. Will be changing this after stand-up meeting : )

database/buildSchema/28_review_assignment.sql
```diff
@@ -68,12 +68,12 @@ CREATE TRIGGER review_assignment_trigger
     WHEN (NEW.trigger IS NOT NULL AND NEW.trigger <> 'PROCESSING' AND NEW.trigger <> 'ERROR')
     EXECUTE FUNCTION public.add_event_to_trigger_queue ();
 
--- TRIGGER (Listener) on review_assignment table: To set status accordingly to assignedSections
+-- TRIGGER (Listener) on review_assignment table: Set assignedSections to Null when changing status to AVAILABLE
```
CarlosNZ, line 71:
```suggestion
-- TRIGGER (Listener) on review_assignment table: Set assignedSections to [] when changing status to AVAILABLE
```

database/buildSchema/28_review_assignment.sql
```diff
@@ -68,12 +68,12 @@ CREATE TRIGGER review_assignment_trigger
...
 CREATE TRIGGER review_assignment_trigger2
     AFTER UPDATE OF trigger ON public.review_assignment
```
CarlosNZ, line 73:
```suggestion
    AFTER UPDATE OF status ON public.review_assignment
```

I thought we already fixed this? 🤔

CarlosNZ (changes requested):
Not sure how that incorrect field ended up in develop. Anyway, the trigger works once that's fixed.

CarlosNZ (approved):
Great thanks

Issue #658, opened 2022-02-15
Handle snapshot migration properly


CarlosNZ:
Even though we have a [data migration script](https://github.com/openmsupply/application-manager-server/pull/659), this still won't handle the case when loading a snapshot with an incompatible schema, since the new schema is reloaded from the current version and the snapshot contains mutations for an outdated schema.

I can think of two possible solutions to this:

1. Export all the database .sql initialisation scripts along with the snapshot. Then when importing snapshot, if "databaseInit = true", we initialise using these files instead of the current ones. Then the migration script will change the database when the server restarts.
2. We "lock" the .sql files (don't change them anymore), and then only do database changes via the migration code

My preference is for option 1, at least for now. Once we have versions in production, perhaps option 2 will be best.

_Edited to add summary of solutions_

CarlosNZ:
Another possibility is if we no longer make modifications to our sql scripts -- we "lock" them and do all further schema changes using migration code. I think that would take care of the above problem, need to think about it some more. 🤔

CarlosNZ:
As a very first step, we should store the current version number when we save a snapshot. Maybe add it as a field in options.json?

CarlosNZ:
So, to summarise the possible solutions:

1. Export all the database .sql initialisation scripts along with the snapshot. Then when importing snapshot, if "databaseInit = true", we initialise using these files instead of the current ones. Then the migration script will change the database when the server restarts
2. We "lock" the .sql files (don't change them anymore), and then only do database changes via the migration code

I think I'm in favour of option 1 at the moment, since the database is still undergoing significant changes. I think once we hit version 1.0.0 though we should probably shift to option 2.

CarlosNZ:
One more point -- when loading a snapshot from the browser, the server doesn't restart. So we'll also need to remember to make the snapshot import script run the migration routine after loading.

PR #661, opened 2022-02-20
#658 Snapshot migration

CarlosNZ:
Implements #658, using Solution 1 outlined in that issue. It's not a huge PR so can change if there's a problem with this approach.

Seems to be working okay though.

- exported .sql script is just one file (concatenation of all script files). Found a couple of missing ";" in existing scripts as a result
- manually calls the "migrate" script at the end (which deals with [this comment](https://github.com/openmsupply/application-manager-server/issues/658#issuecomment-1042277402))
- updated the "database_init" command so it loads the snapshot via the "useSnapshot.ts" script (which is should have been doing anyway, since it has a few extra little things)
- modified "initialise_database.sh" so take an optional parameter -- the name of the ".sql" script file to initialise with. If the snapshot being loaded has a script file, the filepath is passed to this script by "useSnapshot", if not it just runs the default .sql scripts
- updated some of the "execSync" calls to add " { stdio: 'inherit' } " option -- this prints the subshell's console output to main console (was previously not shown)
- tweak "migrateData":
  - add "CREATE TABLE system_info"
  - a generic "changeSchema" database method for schema changing queries (no variables), with appropriate error handling (doesn't throw, just console.log)

### Testing

- Try testing both using the "database_init" command line as well as via the "Snapshots" ui in browser
- Try loading old snapshots and note that the schema is migrated after loading (it won't work with snapshots that have incompatible schemas though, i.e. trying to insert to a field/table that no longer exists -- we can't do anything about these, will need to be manually updated if there are any and we still need to keep them)
- Export a snapshot. Change something in the schema (e.g. rename a field, delete a table). Reload snapshot, and note the schema will be put back to how it was when the snapshot was saved
- Change something in the main database .sql scripts. Load a snapshot that has a build script saved with it and note that the snapshot's schema is loaded, not the modified one (obviously we would have migration code to bring the snapshot schema up-to-date with the main schema script)
- Note that after any snapshot load, the "system_info" table should a) exist; and b) have a newly-created record with the current version in it
- Anything else? Find any thing that breaks or doesn't behave as expected?

**Edit**: TO-DO:

- ~~as per @andreievg 's comment -- I'm going to add a safeguard to forbid loading a snapshot with a higher version number than the current installation.~~ Done 35112aa1
- ~~add migration code for trigger/function modified [here](https://github.com/openmsupply/application-manager-server/pull/660)~~ Done 949bc078

### DIFF
--- a/src/components/snapshots/takeSnapshot.ts
+++ b/src/components/snapshots/takeSnapshot.ts
@@ -29,7 +29,6 @@ import {
 import { getBaseFiles, getDirectoryFromPath } from './useSnapshot'
 import config from '../../config'
 import { DateTime } from 'luxon'
-import { SchemaColumnsOrderBy } from '../../generated/graphql'
 const asyncRimRaf = promisify(rimraf)
 
 const takeSnapshot: SnapshotOperation = async ({
@@ -71,14 +70,11 @@ const takeSnapshot: SnapshotOperation = async ({
     // Copy prefs
     if (options?.includePrefs) execSync(`cp '${PREFERENCES_FILE}' '${newSnapshotFolder}'`)
 
-    // Copy schema build scripts
+    // Copy schema build script
     if (options?.shouldReInitialise)
       execSync(
         `cat ${DATABASE_FOLDER}/buildSchema/*.sql >> ${newSnapshotFolder}/${SCHEMA_FILE_NAME}.sql`
       )
-    // execSync(
-    //   `cp -r '${DATABASE_FOLDER}/buildSchema' '${newSnapshotFolder}/${SCHEMA_FOLDER_NAME}'`
-    // )
 
     // Save snapshot info (version, timestamp, etc)
     await fs.writeFile(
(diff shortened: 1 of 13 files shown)


src/components/snapshots/takeSnapshot.ts
```diff
@@ -68,6 +70,12 @@ const takeSnapshot: SnapshotOperation = async ({
     // Copy prefs
     if (options?.includePrefs) execSync(`cp '${PREFERENCES_FILE}' '${newSnapshotFolder}'`)
 
+    // Copy schema build script
+    if (options?.shouldReInitialise)
+      execSync(
+        `cat ${DATABASE_FOLDER}/buildSchema/*.sql >> ${newSnapshotFolder}/${SCHEMA_FILE_NAME}.sql`
```
andreievg, line 76: nice

andreievg (approved):
The only thing I can think of that could go wrong, is higher version of snapshot then server.

CarlosNZ:
> The only thing I can think of that could go wrong, is higher version of snapshot then server.

Yeah, good point. We should probably display some kind of warning if attempting to load a snapshot with a higher version number than current installation. Not that it should really happen very often (if at all). Or perhaps it would be safer to just not allow it at all. If the user really wants to proceed and they know it's not going to cause problems, then they can always manually edit the version no. in the snapshot database.

Thanks for taking a look. 👍

PR #662, opened 2022-02-21
#1109F Trigger/Function for deleting review question assignments

CarlosNZ:
Required for [front-end issue #1109](https://github.com/openmsupply/application-manager-web-app/issues/1109) (PR https://github.com/openmsupply/application-manager-web-app/pull/1128)

The front-end had no knowledge of the state of existing review question assignments for items on the "Assignment" tab, so rather than write a bunch of additional front-end queries to fetch this info and then process it, it is simpler to leave the question_assignment handling to database triggers & functions.

There was a slight headache, in that the `delete_question_assignments` function's main query didn't work when "assigned_sections" was an empty array, so had to have slightly different logic for that case. Basically:
- trigger function runs when "assigned_sections" on "review_assignment" changes (either from the front-end, or reset via the existing review_assignment trigger)
- If sections remain assigned, this function deletes question assignments relating to the remaining sections
- If no sections remain (we check for "status = AVAILABLE" instead), ALL review question assignments are deleted

Also here:
- Added a generated column to "review_question_assignment" to show the template section code
- Update migration script (with the above plus a couple of refinements to existing)
- Update documentation regarding data/schema migration


nmadruga:
> * If no sections remain (we check for "status = AVAILABLE" instead), ALL review question assignments are deleted

I was thinking about this one... This will change slightly the logic that we added previously. Basically we wanted to keep the state of a previous review done by a reviewer when they get un-assigned from a review. Just in case it is re-assigned to them again iit would be keeping their previous reviewResponses.

I was lookiing for where this was explained in the Wiki, but could only find this:

> [Re-Assignment](https://github.com/openmsupply/application-manager-web-app/wiki/Assignment-Process#re-assignment)
> A GraphQL mutation reassignReviewAssignment runs on the front-end when one assigner:
> 
> selects re-assign in the drop down
> selects the new reviewer in drop down for a section (or all sections).
> review_assignment.trigger for previous assignment is set to ON_REVIEW_UNASSIGN.
> review_assignment.trigger for new assignment is set to ON_REVIEW_REASSIGN.
> Changes status of previous review_assignment to Available and flag is_locked to True.
> Changes status of new review_assignment to Assigned.
> Sets assignerId of current user in the new assignment.
> Adds review_question_assignment (for each question on the section assigned) to review_assignment.
> Other updates done on the server-side:
> If an existing review was connected to previous assignment will change the review.status from Draft to Discontinued.
> **If an existing review was connected to new assignment will change the review.status from Discontinued to Draft.**

nmadruga (commented):
Looks good. I also tested using the snapshot `__just_before_assignment`, and saw the questions being correctly removed from one assignment and created on the new assignment to another user. And I was thinking about the idea of `DISCUTINUED` status. It seems it's going to be very hard to keep both logic. IIf we want to do half re-assignment I don't think we can also keep the ability of **continuing** a previously assigned (and unassigned) review questions. So I'm happy with just changing the docs on the front end so we remove the logic for DISCONTINUE a review.

CarlosNZ:
Okay, have added significant changes to update some of the functions to handle the removal of review_question_assignments and add migration code. Can be tested with https://github.com/openmsupply/application-manager-web-app/pull/1128

database/buildSchema/34_review_status_history.sql
```diff
@@ -92,18 +92,42 @@ CREATE FUNCTION public.submitted_assigned_questions_count (app_id int, stage_id
...
+        FROM
+            TEMPLATE
+        WHERE
+            id = (
+                SELECT
+                    template_id
+                FROM
+                    application
+                WHERE
+                    id = 236))
```
nmadruga, line 129: Something left behind here?

database/migration/migrateData.ts
```diff
@@ -57,24 +58,76 @@ const migrateData = async () => {
...
+    AND ra.stage_id = $2
+    AND ra.level_number = $3
+    AND ra.status = 'ASSIGNED'
+    AND te.category = 'QUESTION'
+    AND rsh.status = 'SUBMITTED'
+    AND te.template_code = (
+        SELECT code FROM TEMPLATE
+        WHERE id = (
+          SELECT template_id FROM application
+          WHERE id = 236))
```
nmadruga, line 126: Again something that might be left behind?

plugins/action_update_review_status/src/databaseMethods.ts
```diff
@@ -55,9 +55,17 @@ const databaseMethods = (DBConnect: any) => ({
...
-    WHERE review_assignment_id = $1
+    SELECT DISTINCT(te.id) AS "templateElementId"
+    FROM template_element te JOIN template_section ts
+    ON te.section_id = ts.id
+    JOIN (
+    		SELECT id, application_id, template_id, UNNEST(assigned_sections) as section_code
+    		FROM review_assignment
+    	) ra 
+    ON ra.template_id = ts.template_id  
+    WHERE ra.id = 28
```
nmadruga, line 66: Another hardcoded id

nmadruga (changes requested):
Some hardcoded ids, I'm confused with what needs to be included and what is temp code that you were using to test with this particular example

CarlosNZ:
> Some hardcoded ids, I'm confused with what needs to be included and what is temp code that you were using to test with this particular example

Ooops, sorry. 😳

Fixed now dcdba919

nmadruga (approved):

PR #665, opened 2022-02-22
#664 Fix user permissions endpoint

### LINKED ISSUE #664: user-permissions endpoint -- not respecting orgId parameter
Needs further investigation -- seems to be giving the same results no matter what the orgId is set to.

Affecting "grantPermissions" template

Also -- one other thing it needs: Should be able to supply "orgId=null" as a parameter, not just omit orgId. Probably should also accept "orgId=0" as well -- this would make it easier to build queries from templates.

CarlosNZ:
Fix #664 

A bit of a subtle change here, but makes a big difference in how the "grantPermissions" template works.

- Firstly, the `OR "isUserCategory" = true` clause in "getTemplatePermissions" was giving false data when used in this endpoint (although it was doing the expected thing when used in the /login endpoints). It was always returning that whatever permission name is associated with the "User" category is granted to the user/org even when it wasn't. I've added an additional parameter to "getTemplatePermissions" which allows turning this on or off -- it's only useful in the login context.
- Better handling of null/undefined values for username or orgId. These parameters can either be omitted, or accept `orgId=null` / `orgId=0` and `username=""` as the same thing. Useful for the grantPermissions template when we need to conditionally supply these values, but can't turn the parameters "off" completely.

### DIFF
--- a/documentation/API.md
+++ b/documentation/API.md
@@ -201,7 +201,8 @@ The same as login endpoint, without the success field
 GET: `/user-permissions?username=<username>&orgId=<orgId>`
 
 End point to get **another** user's granted permissions + all existing permissions on templates for a given organisation.
-Intended for use in template to view/edit antoher user's permissions -- client supplies `username` and `orgId`.
+Intended for use in template to view/edit another user's permissions -- client supplies `username` and `orgId`.
+Either `username` or `orgId` can be omitted and the result will returned for that user without an org, or for that org without a user. For `orgId`, the values `null` or `0` are equivalent to omitting `orgId` and for `username` an empty string `""` can be used. This is useful in cases such as a template query where you're supplying an `orgId` and `username` parameter, but can't "turn off" the parameters when you want to omit either of them -- just use `orgId=null` or `username=""` to achieve the same thing.
 
 #### Check Triggers
 
--- a/documentation/API.md
+++ b/documentation/API.md
@@ -200,9 +200,13 @@ The same as login endpoint, without the success field
 
 GET: `/user-permissions?username=<username>&orgId=<orgId>`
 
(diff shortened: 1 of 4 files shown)


documentation/API.md
```diff
@@ -201,7 +201,8 @@ The same as login endpoint, without the success field
 GET: `/user-permissions?username=<username>&orgId=<orgId>`
 
 End point to get **another** user's granted permissions + all existing permissions on templates for a given organisation.
-Intended for use in template to view/edit antoher user's permissions -- client supplies `username` and `orgId`.
+Intended for use in template to view/edit another user's permissions -- client supplies `username` and `orgId`.
+Either `username` or `orgId` can be omitted and the result will returned for that user without an org, or for that org without a user. For `orgId`, the values `null` or `0` are equivalent to omitting `orgId` and for `username` an empty string `""` can be used. This is useful in cases such as a template query where you're supplying an `orgId` and `username` parameter, but can't "turn off" the parameters when you want to omit either of them -- just use `orgId=null` or `username=""` to achieve the same thing.
```
nmadruga, line 205:
```suggestion
Either `username` or `orgId` can be omitted and the result returned will be either: 
  * for that user without an org; or 
  * for that org without a user. 
  
For `orgId`, the values `null` or `0` are equivalent to omitting `orgId` and for `username` an empty string `""` can be used. This is useful in cases such as a template query where you're supplying an `orgId` and `username` parameter, but can't "turn off" the parameters when you want to omit either of them -- just use `orgId=null` or `username=""` to achieve the same thing.
```

nmadruga (approved):
Looks good. Haven't tested though

documentation/API.md
CarlosNZ: Thanks 4e65545d

PR #669, opened 2022-02-24
#488 #908F Evaluator array access

### LINKED ISSUE #488: evaluator: extend "type" parameter to coerce all node outputs to specified type
We already have the "type" parameter to specify the output type when there is potential ambiguity (like when adding arrays, for example).

Would be useful to extend this to (attempt to) coerce outputs into the type specified in "type". For example, we sometimes need to return a number as a string (and vice versa), or take a value from an array.

CarlosNZ:
Fix #488 (and front-end [#908](https://github.com/openmsupply/application-manager-web-app/issues/908))

Finally got round to adding this. Arrays can now be access via index, and output types can be coerced into other types. Will make certain queries a bit less hassle in templates.

All previous behaviour is preserved, and I've updated the docs to explain a bit more detail.

Have added a few new tests to check the new behaviour. All previous tests still pass.

### DIFF
--- a/documentation/Query-Syntax.md
+++ b/documentation/Query-Syntax.md
@@ -95,7 +95,7 @@ This expression would return the number **8** when evaluated. If `children` are 
 
 ## type
 
-The `type` property is optional in most cases, but some operators will return their results in a different format depending on the `type` (e.g. pgSQL -- see below). For *all* operators, the output will also be attempted to convert it to to the specified `type`. This can be useful, for example, when a number is required from a string output. Note: type conversion follows Javascript type conversion rules using `Number(), String(), Boolean()`.
+The `type` property is optional in most cases, but some operators will return their results in a different format depending on the `type` (e.g. pgSQL -- see below). For *all* operators, the evaluator will also attempt to convert it to to the specified `type`. This can be useful, for example, when a number is required from a string output. Note: type conversion follows Javascript type conversion rules using `Number(), String(), Boolean()`.
 
 Valid values:
 
--- a/documentation/Query-Syntax.md
+++ b/documentation/Query-Syntax.md
@@ -485,7 +485,7 @@ i.e `{user: "user"}`
 
 Tree structure:
 
-[[images/query-syntax-example-1.png|Example 1 tree diagram]]
+![Example 1 tree diagram](images/query-syntax-example-1.png)
 
 ```
 {
@@ -511,7 +511,7 @@ _Note: this query uses PostGres (pgSQL) operator -- in the app, we’ll probably
 
 Tree structure:
 
-[[images/query-syntax-example-2.png|Example 2 tree diagram]]
+![Example 2 tree diagram](images/query-syntax-example-2.png)
 
 ```
 {
@@ -552,7 +552,7 @@ Tree structure:
 
 Tree structure:
 
-[[images/query-syntax-example-3b.png|Example 3 tree diagram]]
+![Example 3 tree diagram](images/query-syntax-example-3b.png)
 
 ```
 {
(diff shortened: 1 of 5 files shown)


documentation/Query-Syntax.md
```diff
@@ -209,6 +207,19 @@ application = {
...
+{ orgs: [{id: 1, name: "Org 1"}, {id: 2, name: "Org 2"}]}
+```
+The following property strings will return these results:
+```
+"orgs" -> [{id: 1, name: "Org 1"}, {id: 2, name: "Org 2"}]
+"orgs.name" -> [ "Org 1", "Org 2" ]
+"orgs[0]" -> {id: 1, name: "Org 1"}
+"orgs[1].id" -> 2
+
+```
```
nmadruga, line 221: Very nice!

documentation/Query-Syntax.md
```diff
@@ -347,7 +358,7 @@ This expression queries our `/login` endpoint to check a user's credentials, and
 
 ### Authentication
 
-The `GET`, `POST`, and `GraphQL` operators may require authentication for the endpoints they are querying, such as our own [API](API.md). This can be achieved in a couple of different ways.
+The `GET`, `POST`, and `GraphQL` operators may require authentication for the endpoints they are querying, such as our own [API](API). This can be achieved in a couple of different ways.
```
nmadruga, line 361: Maybe wrong?

documentation/Query-Syntax.md
```diff
@@ -474,7 +485,7 @@ i.e `{user: "user"}`
 
 Tree structure:
 
-![Example 1 tree diagram](images/query-syntax-example-1.png)
+[[images/query-syntax-example-1.png|Example 1 tree diagram]]
```
nmadruga, line 488: Not sure if that's how this should be showing? Also when I click it it's not finding....
Screen Shot 2022-03-03 at 1 33 05 PM

documentation/Query-Syntax.md
```diff
@@ -670,7 +672,7 @@ There is a [Jest](https://jestjs.io/) test suite for the expression evaluator in
 
 However, for the tests to work you'll need two things:
 
-1. Load the [snapshot](Snapshots.md) called `evaluator_test` from the private templates repo (in `/dev/snapshots`)
+1. Load the [snapshot](Snapshots) called `evaluator_test` from the private templates repo (in `/dev/snapshots`)
```
nmadruga, line 675: Also check if `.md` isn't required

nmadruga (commented):
Not sure which snapshot to use, I was trying with `dev/current` and there are some differences so I'm getting some tests failing. I would recommend renaming the one with tests working to be the new `current` and rename the old current to `dev_2022_03_03` to keep the pattern we were using before on keeping the `dev` test snpashot in pair with latest changes

documentation/Query-Syntax.md
CarlosNZ: Yeah, I noticed this too -- the auto-formatter has been changing a lot of the way it formats links lately. I checked some of the existing links on Github wiki that have been changed to this format, and they seem fine, so I think it's okay. Will definitely keep an eye on once we publish the next wiki update!

CarlosNZ:
> Not sure which snapshot to use, I was trying with `dev/current` and there are some differences so I'm getting some tests failing. I would recommend renaming the one with tests working to be the new `current` and rename the old current to `dev_2022_03_03` to keep the pattern we were using before on keeping the `dev` test snpashot in pair with latest changes

Same snapshot as usual: `/dev/evaluator_test` :)

documentation/Query-Syntax.md
CarlosNZ: Yeah, that's super annoying, I'm not sure why these got auto-formatted like that. Have fixed the image links now, will be vigilant if it happens again. 

ed34da9c

nmadruga (approved):

PR #671, opened 2022-02-24
#668 modifyRecord improvements & #1125F

### LINKED ISSUE #668: ModifyRecord action -- nice to have -- pass an shorthand of field names and response to be converted into database fields
This would be much faster than having to create a whole parameter mapping for each field in the Action config.

CarlosNZ:
Fix #668

Also solves the issue raised by [front-end #1125](https://github.com/openmsupply/application-manager-web-app/issues/1125) -- we don't need to make changes to that plugin, the problem was that ISO dates weren't being interpreted as postgres "date" types, so I've fixed that here.

Basically, it'll be much easier to fill in "modifyRecord" actions in the template builder now -- instead of painstakingly filling out a dozen (or more!) separate parameters, each with their own evaluator expression to fetch data from form responses, we can now do it all in one with a simple mapping from form responses to data record values. (See docs and original issue for more details)

Easiest way to test is to use the `/run-action` endpoint, with something like this as the body JSON:
```
{
    "actionCode": "modifyRecord",
    "applicationId": 19,
    "parameters": {
        "tableName": "test",
        "start_date": "2022-01-23",
        "company_id" : 4,
        "license_type": {"three": 3},
        "data": {
            "first_name": "responses.Q1.text",
            "last_name": "responses.Q2.text",
            "type": "responses.Q5",
            "new_date": "responses.date.date.start"
        },
        "shouldCreateJoinTable": false
    } 
}
```
Change the "applicationID" and the "responses..." values to match an application you're working with. When run, a record will be created in a new "test" table with all the above fields, and the ones in the "data" object pulled straight from application responses.

Also note the data types are now correct for dates, as long as they are in one of the standard ISO formats (such as "start_date" above -- a postGres "date" field will be created)

### DIFF
--- a/documentation/List-of-Action-plugins.md
+++ b/documentation/List-of-Action-plugins.md
@@ -8,6 +8,7 @@
   - [Increment Stage](#increment-stage)
   - [Change Status](#change-status)
   - [Modify Record](#modify-record)
+    - [How the record is built](#how-the-record-is-built)
   - [Generate Text String](#generate-text-string)
     - [Parameters summary](#parameters-summary)
       - [String parameters](#string-parameters)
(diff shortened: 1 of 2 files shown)


nmadruga (approved):
Great addition this will definetelly help us during the Template building process.

My only question here is about the type definition. In case a different type if defined for a parameter, what would happen in the UI? Do we have any prevention of this happening before running the action? I imagine no, but would iit be a good time to create an issue to sort this?

CarlosNZ:
> My only question here is about the type definition. In case a different type if defined for a parameter, what would happen in the UI? Do we have any prevention of this happening before running the action? I imagine no, but would iit be a good time to create an issue to sort this?

Not sure exactly what you're saying, but -- the data type that is created in the database is based on whatever was first sent when creating that field. So the "type" setting in the UI will either *define* that, or it'll make no difference. We don't currently have a way to ensure that *subsequent* record entries will have the correct type, but that's not something we can really control, based on the way we've chosen to have "outcome" tables created. However, unless you mess around with the template output types *after* the outcome table has been auto-generated, it should always keep sending the same type for the same field.

Does that answer your question?

PR #672, opened 2022-02-24
#670 Evaluator -- better / global fallback parameter

### LINKED ISSUE #670: Evaluator: Better fallbacks / dev mode
A common template element is a list of options generated by a Database query, using the result of a previous question as a parameter. Unfortunately, when the earlier question is empty, the database query throws an error and returns the ugly "GraphQL - Can't resolve node" message.

These error messages are quite useful when building and debugging queries, but not good for using in a form. 

Some of the operators have a "fallback" node, but it's not universal or consistent.

I suggest adding a new top-level parameter (along with `operator`, `children` and `type`) called `fallback`, which is a _global_ fallback for *any* error in the entire query, and it can be any type, not just a string. This would make it easier to, for example, return an empty array for a list of options instead of an error string, which crashes the plugin.

When building queries / debugging, we can leave this parameter out, and then we'll get the current, more informative error strings, but then apply the fallback once it's working properly.

CarlosNZ:
Fix #670

More improvements to evaluator -- allows an overall fallback for every node. See docs and test cases for specific details and examples.

I have turned most of the standard "messages" into actual errors that are thrown, since its helpful if they bubble up and can be caught as required by using a fallback at whatever level is most suitable.

Fully backwards compatible -- all existing test cases still work (except for one which has changed from an error string to an actual thrown/caught error), and new test cases added which aim to capture the full range of use cases.

Note: the large diff in `evaluateExpression` is mainly due to the entire main block being placed inside a try/catch block -- the inner contents haven't changed.

Easiest way to test is to just play round with some queries in the Evaluator GUI dev app. `yarn gui`


nmadruga (commented):
Looks good on initial look. Still needs testing

PR #674, opened 2022-03-08
Fixes for tests on new dev/current snapshot

nmadruga:
Have updated the dev/current snapshot to:

- Be available on the repo `application-maganer-templates`
- Have latest changes on `reviewAssignment` adding correct `assigned_sections`
- Removing the use of `reviewQuestionAssignments` from tests
- Ran the action to  `refreshReviewAssignments`
- Fixed tests on **grantPermissions**
- Fixed tests on **modifyRecord** 
- Fixed tests on **generateReviewAssignments** 
- Fixed tests on **updateReviewVisibility**  
- Fixed tests on **updateReviewStatuses**  
- Fixed tests on **trimResponses** 
- Remove Valerio user - transformed into **John Smith**


CarlosNZ:
Cool, thanks for sorting this. I'm trying to run every test suite I can find using your latest "dev/current" snapshot:

- `./server.test` -- FAIL (missing config, should be easy fix)
- modifyRecord -- 1 failed, 8 passed, 
- trimResposnes -- PASS
- utilityFunctions --  PASS
- updateReviewStatuses -- Throws error (still trying to reference "review_question_assignment")
- generateReviewAssignments -- PASS!
- permissions.test --  4 failed, 5 passed
- grantPermissions -- PASS
- updateReviewVisibility -- PASS

For the ones that are quick fixes, we should probably fix them in this PR, you think? (That's just server, modifyRecord and permissions)

Also -- do you think we should rename the snapshot to "testing", so it's clear? 

Once you've done that, I'll go and consolidate the evaluator tests into that snapshot so all test cant then run from the same snapshot.

nmadruga:
> For the ones that are quick fixes, we should probably fix them in this PR, you think? (That's just server, modifyRecord and permissions)
> 
> Also -- do you think we should rename the snapshot to "testing", so it's clear?
> 
> Once you've done that, I'll go and consolidate the evaluator tests into that snapshot so all test cant then run from the same snapshot.

Thanks for looking at this Carl. I was still fixing a few tests... Please check `updateReviewStatuses` again. The one failing in `modifyRecord` I'm not sure how to fix and the ones in permissions I was not sure what needs fixing and was leaving as is. If you can help with that one would be awesome!

Not sure what is the `server` test you mentioned...

CarlosNZ:
> 
> Thanks for looking at this Carl. I was still fixing a few tests... Please check `updateReviewStatuses` again. The one failing in `modifyRecord` I'm not sure how to fix and the ones in permissions I was not sure what needs fixing and was leaving as is. If you can help with that one would be awesome!
> 
> Not sure what is the `server` test you mentioned...

Okay, I'll take a quick look at those two.

> Not sure... do you want me to rename `dev` to `testing`? I think dev is a good name for this. No?

Well `dev` is fairly generic and we have all kinds of snapshots in that folder, most of them single-use or only intended for temporary use. "Current" is fairly meaningless. The "testing" snapshot should probably be in `/dev/testing`, wouldn't you think? If it's intended for testing, the word "test" or "testing" should probably be in there somewhere. I wouldn't have known which snapshot to use if you hadn't mentioned it.

nmadruga:
> Well `dev` is fairly generic and we have all kinds of snapshots in that folder, most of them single-use or only intended for temporary use. "Current" is fairly meaningless. The "testing" snapshot should probably be in `/dev/testing`, wouldn't you think? If it's intended for testing, the word "test" or "testing" should probably be in there somewhere. I wouldn't have known which snapshot to use if you hadn't mentioned it.

I see your point. But for me it's quite confusing keeping all these different versions for different uses. And I like the pattern of keeping which is the `current` one in each snapshot folder. We have tried to do that so far, and I was fairly happy with this pattern. I feel lost when we start changing that. Although I understand your point, I think the **main** use of the `dev` snapshot is for testing...

PR #700, opened 2022-03-30
699 Fix build problem

CarlosNZ:
Took a wee while to figure out the correct exclusion pattern for the test files.

Anyway, seems to be working well now, have built and run Docker image locally.


src/defaults.ini
```diff
@@ -1,15 +0,0 @@
-
```
CarlosNZ, line 1: This file has been there since the beginning and we've never used it -- we've gone in a different direction for our config files. 

So it's gone! 💨

nmadruga (approved):
Perfect. Thanks

PR #701, opened 2022-03-30
#667 Docker extra volumes & localisations fix

### LINKED ISSUE #667: Docker -- localisations folder not in data container
We have the files, database and snapshots in the "data" container so they don't get reset when the main container is restarted (docker-compose). 

We also need to put "localisations" folder in there too, cos that's currently getting reset each time.

CarlosNZ:
Fix #667

We'll need to test this carefully on the demo server when we first install it -- probably start with just one instance.

There's two separate issues here:
- fix for why the localisations weren't changing with snapshot installation
- isolate localisations and prefs in docker volumes so they don't get over-written when upgrading the app

Note also that I've tidied up volume folder structure so we don't have a million separate folders in `/demo-server` -- each set of volume folders will be contained within a `conforma_on_port_${PORT_APP}`  folder.

Docs updated to reflect changes.


docker/Dockerfile
```diff
@@ -53,6 +53,9 @@ RUN echo "deb https://dl.yarnpkg.com/debian/ stable main" | tee /etc/apt/sources
 RUN apt-get update
 RUN apt-get install yarn -y
 
+# Uncomment following line to force Docker to skip cache from here on:
+# ADD "https://www.random.org/cgi-bin/randbyte?nbytes=10&format=h" skipcache
+
```
CarlosNZ, line 58: This is how we can force the build to not use the cached repo, but we can still get the benefit of having all the previous installation stuff cached.

documentation/Demo-Server-Guide.md
```diff
@@ -127,34 +124,54 @@ Logs are in `/var/log/nginx`
...
 
 ```bash
-mkdir app_snapshots_on_port_8000
-mkdir app_postgres_on_port_8000
+mkdir conforma_on_port_8000 \
+  conforma_on_port_8000/postgres \
+  conforma_on_port_8000/snapshots \
+  conforma_on_port_8000/files \
+  conforma_on_port_8000/localisation \
+  conforma_on_port_8000/preferences
```
CarlosNZ, line 132: We'll need to re-do this on the demo server once. But do it on one instance and make sure it is stable before changing the others.

src/components/permissions/routes.ts
```diff
@@ -198,7 +198,7 @@ const routeVerification = async (request: any, reply: any) => {
 // Serve prefs to front-end
 const routeGetPrefs = async (request: any, reply: any) => {
   const prefs = JSON.parse(
-    readFileSync(path.join(getAppEntryPointDir(), '../preferences.json'), 'utf8')
+    readFileSync(path.join(getAppEntryPointDir(), '../preferences/preferences.json'), 'utf8')
```
CarlosNZ, line 201: Note that the prefs file is now inside a `preferences` folder -- this is the easiest way to put the prefs in a docker volume.

src/components/snapshots/useSnapshot.ts
```diff
@@ -87,7 +87,7 @@ const useSnapshot: SnapshotOperation = async ({
     // Import localisations
     if (options?.includeLocalisation) {
       try {
-        execSync(`cp -r  '${snapshotFolder}/localisation/' '${LOCALISATION_FOLDER}/' `)
+        execSync(`cp -r  '${snapshotFolder}/localisation/.' '${LOCALISATION_FOLDER}' `)
```
CarlosNZ, line 90: This was the bugger that stumped me for ages. In develop the localisation files were being restored just fine, but on the server they were ending up nested a level deep. Subtle differences in the behaviour of different versions of `cp` I guess 🤷‍♂️

docker/demo_server/docker-compose.yml
```diff
@@ -13,9 +13,11 @@ services:
...
       JWT_SECRET: '${JWT_SECRET}'
     volumes:
-      - ./app_postgres_on_port_${PORT_APP}:/var/lib/postgresql/12/main/
-      - ./app_snapshots_on_port_${PORT_APP}:/usr/src/application-manager-server/build/database/_snapshots
-      - ./app_file_on_port_${PORT_APP}:/usr/src/application-manager-server/build/files
+      - ./conforma_on_port_${PORT_APP}/postgres:/var/lib/postgresql/12/main/
+      - ./conforma_on_port_${PORT_APP}/snapshots:/usr/src/application-manager-server/build/database/_snapshots
+      - ./conforma_on_port_${PORT_APP}/files:/usr/src/application-manager-server/build/files
+      - ./conforma_on_port_${PORT_APP}/localisation:/usr/src/application-manager-server/build/localisations
+      - ./conforma_on_port_${PORT_APP}/preferences:/usr/src/application-manager-server/build/preferences
```
CarlosNZ, line 20: @andreievg -- just checking with you that this folder structure will be okay for volumes? 

Still separate volumes for each folder, but grouped by installation to make it a bit tidier.

src/components/snapshots/useSnapshot.ts
nmadruga: I noticed this on my local copy while loading new snapshots! Good fix : )

nmadruga (commented):
Building now to test v0.2.0-2

PR #705, opened 2022-04-04
#702 Update Notification Action to handle cc / bcc

### LINKED ISSUE #702: Change for additional email addresses to be included as BCC recipients
Edit additional recipients of emails to be blind copied (Bcc'd) instead of being direct recipients

CarlosNZ:
Fix #702

This completes the changes required to the "sendNotification" Action in order to put email address(es) in the bcc (or cc) fields, but will need to be tried out with actual template configurations to see how easy it is to build the email arrays or fields.

For consistency, I've renamed the "email" field to "to" (the new ones being "cc" and "bcc"), but I've kept "email" as a valid alias for "to", so we don't need to change any existing templates.

It now checks inputs for valid email addresses and only sends to those, so if an evaluator query in a template generates an invalid result (e.g. null or empty array), the other email addresses will still work fine.

Note that in the case of *no* valid email addresses, the notification record is still created and the Action is considered a "SUCCESS", but there will be a warning in the output "error" field


nmadruga:
> For consistency, I've renamed the "email" field to "to" (the new ones being "cc" and "bcc"), but I've kept "email" as a valid alias for "to", so we don't need to change any existing templates.

Cool - good thinking to keep the email as a valiid option : )

CarlosNZ:
Did you have a chance to look at this yet, @nmadruga ?

nmadruga:
> Did you have a chance to look at this yet, @nmadruga ?

Yes. I was just going to test something else, but it's been to hard for me to focus on things... But I have looked before and agree with code. All good

nmadruga (approved):

PR #707, opened 2022-04-05
#706 Snapshot import fix

### LINKED ISSUE #706: Snapshot import (on SOME systems)
I'm having trouble loading any snapshots that have additional tables (i.e. that weren't in the base schema, e.g. licence, lookup_tables, etc.)

I have figured out what the problem is -- after running the initialisation and schema_diff scripts, postgraphile's "watch" is not getting the updated schema reloaded quick enough before the insert snapshot data script starts, and so causes an error when it first comes across an insertion into a table it doesn't know exists yet.

This is only happening on my M1 Mac, it's still fine on my older Intel Macbook. 

Any idea what to about it, @andreievg ? I can't seem to find a way to make the script wait until the new schema has been detected. And it's weird that it only has this problem on some systems.

CarlosNZ:
Fix #706 

I hate having to add such a hacky thing as a time delay, but I can't see any other way to fix this without rethinking how we handle monitoring the schema in postgraphile without using `--watch`

But it works (for me on new M1), so 🤷‍♂️


nmadruga (approved):
Cool - seems safe

Issue #692, opened 2022-03-23
New company registration - Template basics


jagoje:

jagoje:
### User applies to register a new company - U2
**User story:**
Only external users, single review level by ARMED

Step 1: User starts application to register a new business.
Step 2: For the fields business name and business address, the system should cross check that these are not existing in the system. If they are existing, warning text should appear to prompt the applicant to submit a ‘join business application’.
Step 3: Applicant continues to fill out the application form.
Step 4: The applicant submits the application and receives a confirmation email (template provided) 
Step 5: ARMED staff member with permission to approve business applications are able to ‘self-assign’ application.
Step 6: ARMED staff member reviews the application. The outcomes are:
Approve - Email sent, company is registered with applicant as a member, this information is included in the outcome table.
Reject - Email sent, company is not registered, no information is included in the outcome table.
Send request for information to applicant - Email sent to applicant, applicant is able to log in and amend application, application is resubmitted to the assigned reviewer for continued evaluation.

Granted permissions: User should now be added to the company and able to access template P1.

**Template information:**
Template/Form name - Novo registro comercial
Section name/s - Section 1:Informações da Empresa

<img width="299" alt="image" src="https://user-images.githubusercontent.com/88009219/159804783-362edb2d-6fe4-40a0-9e55-6d8394e44a89.png">

nmadruga:
## Portuguese

Template name: Formulário para adicionar empresa

nmadruga:
### Start message in portuguse

## Favor solicitar os seguintes documentos durante o registro:
- TBD

nmadruga:
> Step 2: For the fields business name and business address, the system should cross check that these are not existing in the system. If they are existing, warning text should appear to prompt the applicant to submit a ‘join business application’.

TODO: Should use validation with endPoint checkUnique associated to existing company (external) that already have the same address

nmadruga:
<img width="922" alt="Screen Shot 2022-04-04 at 3 35 09 PM" src="https://user-images.githubusercontent.com/16461988/161469603-f60c961c-f7e8-4fd4-9a19-1f5f04362465.png">
## Applicant

**permission**: applyGeneral

### Stage
**Name**: Aprovação, **description**: O formulário para criar novo registro deverá ser aprovado pela agência reguladora

#### Level 1
**Name**: Revisão, **permission**: reviewOrgRego, self-assignable: ✔️ 
**Assigner permission**: assignGeneral (Not required)

[...]

PR #708, opened 2022-04-05
#692 endpoint to server current date

nmadruga:
Just add new end-point to retrieve the server year. This could be improved to also allow returning either:
- Day
- Month
- Year
- Full-date
- Time

Ideas on how we can do this? Just using a simple enum?

### DIFF
--- a/documentation/API.md
+++ b/documentation/API.md
@@ -133,13 +133,21 @@ URL query paramter fields (all optional):
 - `user_id`
 - `application_serial` (for associating files with their applications)
 - `application_response_id` (for specifying what response a file belongs to, useful for the file "clean-up" action)
-- `unique_id` (the randomly generated one should normally be sufficient. If this is specifed and the id already exists in the database, the file record will be *updated* with the new file data.)
+- `unique_id` (the randomly generated one should normally be sufficient. If this is specifed and the id already exists in the database, the file record will be _updated_ with the new file data.)
 - `template_id` (to associate a file with a template, for example a carbone template doc)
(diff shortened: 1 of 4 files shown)


CarlosNZ:
Just quickly -- what's the purpose of this, as opposed to using browser's local time?

Also -- seems way too specific. If we're going to go to the trouble of creating an endpoint to get server time, we should just have a single `/get-time` endpoint which returns the whole date and time as an object with `year`, `month` fields, etc:
```
{
 year:.. 
 month:...
 day: ...
 hour: ...
 minute: ...
 second: ...
 (and maybe):
 timezone:...
 ISOString: ...
 timestamp: <unix timestamp>
}
```
I don't see any reason to break different date/time components into separate endpoints .

> Ideas on how we can do this? Just using a simple enum?

CarlosNZ:
> Also -- seems way too specific. If we're going to go to the trouble of creating an endpoint to get server time, we should just have a single `/get-time` endpoint which returns the whole date and time as an object with `year`, `month` fields, etc:

If you don't want to return all those fields right now, that's fine, but please can we keep this extensible so we don't end up with a whole lot of different endpoints for different parts of the date/time. It would be ridiculous to make three different queries to three different endpoints just to get the server date in years, months and days.

So for now, can we just use `/get-time` and return maybe just year/month/day/hour/minute/second?

It should just be a matter of creating a new Luxon object then mapping its fields to the return object, just as simple as what you've already done.

CarlosNZ:
Okay, ignore all that, now that I've seen your use case -- we can do this quite easily without creating a new endpoint at all. We just need to add a "getYear" function to "evaluatorFunctions". I'll make a quick PR for that now.

PR #709, opened 2022-04-06
#692 Evaluator functions -- getYear

CarlosNZ:
Adds a simple `getYear` function to "evaluatorFunctions" which can be called from evaluator expressions.

Note, these are only available in the back-end currently. We're not passing any functions into front-end query evaluations.

Also -- realised I hadn't actually documented this operator, so have updated that.


nmadruga (approved):
LGTM

PR #712, opened 2022-04-07
#1141F adding missing pt translations

nmadruga:
Adding missing strings in Portuguese : D

To be review in combination with Front-end PR: https://github.com/openmsupply/application-manager-web-app/pull/1145


localisation/languages.json
```diff
@@ -4,7 +4,7 @@
     "description": "NZ English",
     "code": "en_nz",
     "flag": "🇳🇿",
-    "enabled": true
+    "enabled": false
```
CarlosNZ, line 7: This shouldn't be changed in the default languages.json.

Any changes should come from snapshot.

CarlosNZ (approved):
I'll just have to trust that the Portugese is all good, but please revert the languages.json file before merging.

PR #713, opened 2022-04-07
#710 Docker volume improvements, and fresh startup fixes

### LINKED ISSUE #710: Docker -- when loading a new image for the first time using docker-compose, we need to populate the prefs and localisation folder with default data
When we link Volume folders using docker-compose, any contents gets erased.

We want the new Volume folder to be pre-populated with whatever was in the source folders.

This page suggests we need to use "volume" types rather than "bind" types: https://docs.docker.com/storage/volumes/ in order to achieve this.

And this page looks like it explains how to configure the docker-compose.yml file to achieve that: https://docs.docker.com/compose/compose-file/compose-file-v3/

CarlosNZ:
Fix #710 

Sorry, this turned out bigger than I planned, but I think I've finally sorted all the docker and startup quirks that have been bugging us (I know, I've said that before!). And bear in mind half the changed files are just because I relocated the "constants.ts" file

Two main changes here:

- Docker compose now uses "named volumes" rather than "bind mounts", as described here: https://docs.docker.com/storage/volumes/. The main advantage of this is that the volumes get created automatically if they don't exist, but it's also the recommend approach.
- I finally (mostly) figured out all the logic for where things get put during start-up and when, and how. It's quite confusing. So basically, I've removed "localisations" from the repo (in .gitignore) and it gets created o server startup using core_templates if its missing, along with files and prefs. This check is done whenever a snapshot is loaded or saved too. The tricky thing was trying to come up with a solution that will always fully load the core_templates on a fresh install that will work for both the Docker build and a stand-along server. I think this achieves that now.

I'll annotate the diff files with some more inline comments to explain.

FYI -- this build is currently running at https://irims-demo.msupply.org:50008/ so feel free to have a poke around on the server to start/stop it, see the file structure internally etc.

Also -- still need to update docs to reflect this. Will wait until your docs PR is merged and go from there.


database/insert_data.sh
```diff
@@ -1,9 +1,5 @@
 #!/bin/bash
 
-#copy folder with snapshot_basic for inital Application manager setup
-mkdir ./database/_snapshots
-cp -rf ./database/core_templates ./database/_snapshots
```
CarlosNZ, line 5: This is all done as part of `createDefaultDataFolders()` which runs on server startup as well as snapshot take/use scripts

database/snapshotCLI.ts
```diff
@@ -1,6 +1,6 @@
 import takeSnapshot from '../src/components/snapshots/takeSnapshot'
 import useSnapshot from '../src/components/snapshots/useSnapshot'
-import { DEFAULT_SNAPSHOT_NAME, DEFAULT_OPTIONS_NAME } from '../src/components/snapshots/constants'
+import { DEFAULT_SNAPSHOT_NAME, DEFAULT_OPTIONS_NAME } from '../src/constants'
```
CarlosNZ, line 3: I moved the "constants.ts" file out of the snapshot area, because we want to use it in more contexts than just snapshots.

docker/database.sh
```diff
@@ -23,3 +23,8 @@ echo '--- RUNNING POST INSTALL'
 
 echo '--- COPY CLEAN DATABASE TO BE USED IF NO VOLUMES ARE MOUNTED'
 cp -R /var/lib/postgresql/12/main/ ./fresh_db
+
+# We end up with extraneous file folders in the repo root, so delete them now
+rm -r files
+rm -r localisation
+rm -r preferences
```
CarlosNZ, line 30: Deleting unnecessary folders from the Docker container makes it less confusing to figure out which folders are being used for what.

docker/demo_server/docker-compose.yml
```diff
@@ -13,15 +13,23 @@ services:
...
-      - ./conforma_on_port_${PORT_APP}/postgres:/var/lib/postgresql/12/main/
-      - ./conforma_on_port_${PORT_APP}/snapshots:/usr/src/application-manager-server/build/database/_snapshots
-      - ./conforma_on_port_${PORT_APP}/files:/usr/src/application-manager-server/build/files
-      - ./conforma_on_port_${PORT_APP}/localisation:/usr/src/application-manager-server/build/localisation
-      - ./conforma_on_port_${PORT_APP}/preferences:/usr/src/application-manager-server/build/preferences
+      - postgres:/var/lib/postgresql/12/main/
+      - snapshots:/usr/src/application-manager-server/build/database/_snapshots
+      - files:/usr/src/application-manager-server/build/files
+      - localisation:/usr/src/application-manager-server/build/localisation
+      - preferences:/usr/src/application-manager-server/build/preferences
```
CarlosNZ, line 20: These are the new "volumes". Docker automatically adds the instance name to them, so don't need the port variable any more.

The volumes are stored in `/var/lib/docker/volumes` in the container, but I've made a symlink to this folder in the usual `/demo_server` folder where the current volumes are kept (on the demo server)

docker/dockerise.sh
```diff
@@ -12,7 +12,11 @@ PUSH=${2:-nopush} # Default won't push to Docker hub
 NODE_VERSION='14'
 POSTGRES_VERSION='12'
 
-IMAGE_TAG="build-${BRANCH_NAME}_$(date +"%Y-%m-%d")_pg-${POSTGRES_VERSION}_node-${NODE_VERSION}"
+# Generate a random ID so Images built on same day with same branch
+# have a unique name
+RANDOM_ID=$(openssl rand -hex 3)
+
+IMAGE_TAG="build-${BRANCH_NAME}_$(date +"%Y-%m-%d")_${RANDOM_ID}"
```
CarlosNZ, line 19: I thought it was dumb and a waste of valuable screen real estate to have `_pg-12_node-14` on the end of every tag name. But I've added a random 6-char string instead -- this is so if we make multiple builds from the same branch on the same day they will get unique names (something I've done quite a bit while testing this, and it makes things a lot less confusing)

docker/entry.sh
```diff
@@ -1,23 +1,5 @@
...
-
-if [[ ! -f ./build/files ]]; then
-   echo '---'
-   echo 'no files present, will pull from core-templates'
-   echo '---'
-   cp -r ./build/database/core_templates/files ./build
-   cp -r ./build/database/core_templates/localisation ./build
-   cp ./build/database/core_templates/preferences.json ./build
-fi
-
```
CarlosNZ, line 20: This was messing things up, and we don't need it any more. The files part is now handled by `createDefaultDataFolders()`, and the database part is already done when it loads the core snapshot before first launch (in Docker)

localisation/en_dtac/strings.json
```diff
@@ -1,10 +0,0 @@
-{
-  "LABEL_REVIEW_APPROVE": "Mark as reviewed",
-  "LABEL_REVIEW_RESSUBMIT": "Request clarification",
-  "LABEL_REVIEW_DECICION_CONFORM": "Reviewed",
-  "LABEL_REVIEW_DECISION_NON_CONFORM": "Clarification required",
-  "DECISION_CONFORM": "Move application to the next stage",
-  "DECISION_NON_CONFORM": "Mark application as unsuccessful",
-  "OUTCOME_APPROVED": "Granted!",
-  "OUTCOME_REJECTED": "Unsuccessful this time"
-}
```
CarlosNZ, line 10: As mentioned `/localisation` removed from repo and is restored locally when starting up or loading a snapshot. This keeps it behaving the same as "files" and "snapshots".

src/components/files/createDefaultFolders.ts
```diff
@@ -0,0 +1,57 @@
...
+  GENERIC_THUMBNAILS_SOURCE_FOLDER,
+  DATABASE_FOLDER,
+  BASE_SNAPSHOT_NAME,
+} from '../../constants'
+import fs from 'fs'
+import path from 'path'
+import { execSync } from 'child_process'
+import { makeFolder } from '../utilityFunctions'
+
+export function createDefaultDataFolders() {
```
CarlosNZ, line 22: This is the major "catch-all" function, which combines a lot of things that were done piecemeal elsewhere. It ensures that whenever the server starts or a snapshot is saved or loaded, there will always be a valid "files", "snapshots", "localisations" and "preferences" folders with (at the very least) contents from core_templates.

src/components/files/createDefaultFolders.ts
```diff
@@ -0,0 +1,57 @@
...
+
+  // If localisation folder is missing, we assume it's a fresh install and
+  // fetch them (and preferences) from core templates
+  try {
+    if (
+      !fs.existsSync(LOCALISATION_FOLDER) ||
+      !fs.existsSync(path.join(LOCALISATION_FOLDER, 'languages.json'))
+    ) {
+      makeFolder(LOCALISATION_FOLDER, 'Restoring LOCALISATIONS and PREFERENCES')
+      execSync(`cp -r '${BASE_SNAPSHOT_FOLDER}/localisation/.' '${LOCALISATION_FOLDER}'`)
```
CarlosNZ, line 51: I can't actually remove "preferences" from the repo root like I did with "localisations" because nothing will run without a prefs file. However, the preferences will get replaced with the "core_templates" one whenever "localisations" is restored -- that way we can be sure that whatever is in "core_templates"  is the "official" source of truth and we can change the main one around without worrying about changing defaults.

src/components/files/fileHandler.ts
```diff
@@ -16,32 +16,6 @@ interface HttpQueryParameters {
...
-        fs.copyFile(
-          path.join(getAppEntryPointDir(), imagesFolder, 'generic_file_thumbnails', file),
-          path.join(getAppEntryPointDir(), filesFolder, genericThumbnailsFolderName, file),
-          () => {}
-        )
-      )
-    }
-  )
-}
-
```
CarlosNZ, line 44: This functionality incorporated into `createDefaultDataFolders()`

src/components/snapshots/takeSnapshot.ts
```diff
@@ -25,18 +25,22 @@ import {
...
 
 const takeSnapshot: SnapshotOperation = async ({
   snapshotName = DEFAULT_SNAPSHOT_NAME,
   optionsName = DEFAULT_OPTIONS_NAME,
   options: inOptions,
   extraOptions = {},
 }) => {
+  // Ensure relevant folders exist
+  createDefaultDataFolders()
+
```
CarlosNZ, line 43: To prevent errors in case the data folders don't exist yet. (This shouldn't happen, it's just a failsafe)

src/components/utilityFunctions.ts
```diff
@@ -32,8 +32,11 @@ export const combineRequestParams = (request: any, outputCase: OutputCase | null
 type OutputCase = 'snake' | 'camel'
 
 // Create folder if it doesn't already exist
-export const makeFolder = (folderPath: string) => {
-  if (!fs.existsSync(folderPath)) fs.mkdirSync(folderPath)
+export const makeFolder = (folderPath: string, message?: string) => {
+  if (!fs.existsSync(folderPath)) {
+    message && console.log(message)
+    fs.mkdirSync(folderPath)
```
CarlosNZ, line 38: Just adding an optional "message" -- allows us to only print a console message if the folder is being created for the first time (won't show on subsequent runs)

src/constants.ts
```diff
@@ -12,15 +13,24 @@ export const INFO_FILE_NAME = 'info'
 export const SCHEMA_FILE_NAME = 'schema_init'
 export const PG_DIFF_CONFIG_FILE_NAME = 'pg-diff-config'
 export const PG_SCHEMA_DIFF_FILE_NAME = 'schema_diff'
+export const PREFERENCES_FILE_NAME = config.preferencesFileName
 
 export const ROOT_FOLDER = path.join(getAppEntryPointDir(), '../')
 export const DATABASE_FOLDER = path.join(getAppEntryPointDir(), config.databaseFolder)
 export const SNAPSHOT_FOLDER = path.join(DATABASE_FOLDER, SNAPSHOT_SUBFOLDER)
 export const SNAPSHOT_OPTIONS_FOLDER = path.join(DATABASE_FOLDER, OPTIONS_SUBFOLDER)
```
CarlosNZ, line 21: This file is still a bit of a mess and I'm not super happy with how we've ended up with multiple "config" type files, but it's too big a job to do a complete overhaul right now, so just sticking with current pattern here.

utils/build_all.sh
```diff
@@ -11,10 +11,6 @@ cp -R database/buildSchema ./build/database/
 cp -R database/insertData ./build/database/
 cp -R database/snapshotOptions ./build/database/
 cp -R database/core_templates ./build/database/
-echo 'copying localisation files'
-cp -R localisation ./build/localisation/
-# _snapshots needs to be created manually because not in repo (.gitignore)
-mkdir ./build/database/_snapshots
```
CarlosNZ, line 17: No longer required because of `createDefaultDataFolders()`

package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "application-manager-server",
-  "version": "0.2.0-3",
+  "version": "0.2.0-7",
```
CarlosNZ, line 3: Sorry, I should have just used branches rather than tags to make these builds, will slow down the version bumping in future.

nmadruga:
> * I've removed "localisations" from the repo (in .gitignore) and it gets created o server startup using core_templates if its missing, along with files and prefs

Ok - I have copied all my new strings to `application-manager-server/database/core_templates/localisation/pt_an/strings.json` so this would also copy the newly translated ones

docker/demo_server/docker-compose.yml
nmadruga: Hmmm is this documented somewhere?

CarlosNZ: I plan to update the docs, was waiting for your docs PR to be in. Will do soon.

localisation/en_dtac/strings.json
nmadruga: Oh... Good point. I have added a change on the keys to be in core_templates also - along with copying latest version of lcoalisation in pt_an to be there also - straight commited to develop

CarlosNZ:
@nmadruga I've update the documentation now -- much more thorough explanation of what's going on with docker-compose and volumes, as well as some general tidy-ups for clarity and structure.

66785b98

CarlosNZ:
Oh, there's still something fishy going on in this when I load snapshots on the demo server. Will need to investigate further -- but probably won't sort it today! :(

documentation/Demo-Server-Guide.md
```diff
@@ -96,15 +99,14 @@ _To see the actual Docker commands that are constructued, please inspect the fil
   - port 8000
   - WEB_HOST url
   - JWT_SECRET (this can be any random string, but should be a decent length, say > 24 alphanumeric characters. There's no need to record this key anywhere, as it can change anytime -- that just means existing JWTs become invalid, so users will need to re-login)
-    The system will be launched with “core_snapshot” data. Upload and run a new snapshot as required
+  
+  The system will be launched with “core_snapshot” data. Upload and load a new snapshot as required. 
 
-## JUMP to section for a basic server upgrade
-
-- [docker-compose](#docker-compose)
```
nmadruga, line 103: Can you keep this please?

CarlosNZ: It didn't make sense what it refers to, and what do we mean by "basic server upgrade"? 

I still have the link to the docker-compose section in the same place, but with a bit more clarification what it actually means:

>   Note that this `docker run` procedure resets the container every time it is restarted, so any data changes are wiped out. In order to avoid this, we need persist the data using Docker volumes. See the [docker-compose](#docker-compose) section below for information regarding this.

nmadruga (approved):

docker/demo_server/docker-compose.yml
```diff
@@ -13,15 +13,24 @@ services:
...
+      - grafana:/var/lib/grafana
+
+volumes:
+  postgres:
+  snapshots:
+  files:
+  localisation:
+  preferences:
+  grafana:
+
```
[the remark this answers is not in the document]
CarlosNZ: Nooooo, this is where the volume names are defined.

See https://docs.docker.com/compose/compose-file/compose-file-v3/#volume-configuration-reference

documentation/Demo-Server-Guide.md
```diff
@@ -159,86 +159,70 @@ Logs are in `/var/log/nginx`
 
 ## docker-compose
 
-docker-compose will complain if directories are not present, create them if needed (they will persist when restarting image, even if `docker-compose down` was run) - use `docker compose down` alternativelly.
```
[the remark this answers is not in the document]
CarlosNZ: > You removed the part where I'm expliciting that we need to run using `docker compose`

Where is this? That's not what the highlighted bit here is about, this is just removing  information that is no longer correct or relevant:

> docker-compose will complain if directories are not present, create them if needed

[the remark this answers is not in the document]
nmadruga: Just the "- use `docker compose down` alternativelly." part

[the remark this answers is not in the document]
nmadruga: Which is actually not required here, but for the `docker-compose up` it would actually be. `docker compose up`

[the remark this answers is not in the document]
CarlosNZ: > Just the "- use `docker compose down` alternativelly." part

That's further down. This section is divided into:

- What docker compose does 
- Commands for launching
- Commands for stopping (i.e. docker-compose down)

> but for the docker-compose up it would actually be. docker compose up
This is a bit confusing -- looks like at the moment, different commands are required on different systems: https://stackoverflow.com/a/66516826

Probably best to just leave it with the one that works on our demo server (docker-compose), and add a note somewhere that docker-compose might need to be replaced with docker compose, would you say?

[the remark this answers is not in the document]
nmadruga: Yes, indeed. I'll add a line with that in here. I'm planning to upgrade the server now. Would be ok for me to merge this one in?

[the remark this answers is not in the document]
CarlosNZ: Yes, but you'll need to do some fiddling with the folders on the server when you first load it up. Do you want me to talk you through that?

PR #718, opened 2022-04-11
Add new functions to get current date on server

nmadruga:
Used for U2-OrgRegistration emails

### DIFF
--- a/src/components/actions/evaluatorFunctions.ts
+++ b/src/components/actions/evaluatorFunctions.ts
@@ -5,6 +5,12 @@ import { DateTime, Duration } from 'luxon'
 
 const generateExpiry = (duration: Duration) => DateTime.now().plus(duration).toJSDate()
 
+const getDay = () => new Date().getDate()
+
+const getMonth = () => new Date().getMonth()
+
 const getYear = () => new Date().getFullYear()
 
-export default { generateExpiry, getYear }
+const getDate = () => `${getDay()}-${getMonth()}-${getYear()}`
+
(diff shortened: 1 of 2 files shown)


src/components/actions/evaluatorFunctions.ts
```diff
@@ -5,6 +5,12 @@ import { DateTime, Duration } from 'luxon'
...
 const generateExpiry = (duration: Duration) => DateTime.now().plus(duration).toJSDate()
 
+const getDay = () => new Date().getDate()
+
+const getMonth = () => new Date().getMonth()
+
 const getYear = () => new Date().getFullYear()
 
-export default { generateExpiry, getYear }
+const getDate = () => `${getDay()}-${getMonth()}-${getYear()}`
```
CarlosNZ, line 14: I don't really like how this is an over-specific and unique formatting just called "getDate". If we need a different date format for something else, the name "getDate" would be already taken and it would get very messy fast.

Suggest you use Luxon formatting for any type of date format:
```
import {DateTime} from 'luxon'
...
const getFormattedDate = (formatString: string) => DateTime.now().toFormat(formatString)
```

Then in the template query when you call it, just use `getFormatedDate('DD-MM-YYYY')` (for example)

Note: that format string is not the exact one you need, you''ll need to look that up in Luxon reference and test it in the template builder. But for this PR you don't have to worry about it, as the string can be worked out in the template.

CarlosNZ (changes requested):

PR #719, opened 2022-04-13
1033B Lookup table fixes

CarlosNZ:
This is back-end branch required for front-end [PR #1150](https://github.com/openmsupply/application-manager-web-app/pull/1150)

*Finally* got a fix for that annoying random error when uploading certain files.

Still not sure exactly why it happens, seems to be some kind of bug with Fastify or its file-handling plugin, but it just spits an error on certain files for no apparent reason. Anyway, when I isolated it to that, I remembered that I'd had a similar problem with the file upload endpoint. So all I've done here is change this endpoint to handle files the same way that one does and it seems to work. Basically, we process an iterable of `request.files()` rather than a single `request.file()`.  Don't fully understand what's going on here, but it seems to be working!

Edit: I've put the full-size "IngredientsList.csv" file in `/demo_angola/files` in templates repo -- that's the file that was mostly causing problems previously.

### DIFF
--- a/src/components/files/createDefaultFolders.ts
+++ b/src/components/files/createDefaultFolders.ts
@@ -32,10 +32,10 @@ export function createDefaultDataFolders() {
     if (!fs.existsSync(FILES_FOLDER)) {
       makeFolder(FILES_FOLDER, 'Creating FILES folder')
       execSync(`cp -r '${BASE_SNAPSHOT_FOLDER}/files/.' '${FILES_FOLDER}'`)
-      // Restore generic thumbnails, they get wiped out during snapshot loading
-      makeFolder(path.join(GENERIC_THUMBNAILS_FOLDER))
-      execSync(`cp -r '${GENERIC_THUMBNAILS_SOURCE_FOLDER}/.' '${GENERIC_THUMBNAILS_FOLDER}'`)
     }
+    // Restore generic thumbnails, they get wiped out during snapshot loading
+    makeFolder(path.join(GENERIC_THUMBNAILS_FOLDER))
+    execSync(`cp -r '${GENERIC_THUMBNAILS_SOURCE_FOLDER}/.' '${GENERIC_THUMBNAILS_FOLDER}'`)
   } catch {
     console.log('\nProblem creating FILES folder\n')
   }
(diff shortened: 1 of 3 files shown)


src/lookup-table/controllers/ImportCsv.controller.ts
```diff
@@ -3,29 +3,31 @@ import { parseStream } from 'fast-csv'
 import { LookupTableService } from '../services'
 
 const ImportCsvController = async (request: FastifyRequest, reply: FastifyReply) => {
-  const data = await request.file()
-  const tableNameLabel: string = (data.fields.tableName as any).value
-  const lookupTableService = await LookupTableService({ tableNameLabel })
+  const data = await request.files()
+  const { tableName } = request.query as any
```
CarlosNZ, line 7: Because of the "files" change, I'm now sending the "tableName" as url query rather than part of formData. Front-end updated accordingly.

src/components/files/createDefaultFolders.ts
```diff
@@ -32,10 +32,10 @@ export function createDefaultDataFolders() {
     if (!fs.existsSync(FILES_FOLDER)) {
       makeFolder(FILES_FOLDER, 'Creating FILES folder')
       execSync(`cp -r '${BASE_SNAPSHOT_FOLDER}/files/.' '${FILES_FOLDER}'`)
-      // Restore generic thumbnails, they get wiped out during snapshot loading
-      makeFolder(path.join(GENERIC_THUMBNAILS_FOLDER))
-      execSync(`cp -r '${GENERIC_THUMBNAILS_SOURCE_FOLDER}/.' '${GENERIC_THUMBNAILS_FOLDER}'`)
     }
+    // Restore generic thumbnails, they get wiped out during snapshot loading
+    makeFolder(path.join(GENERIC_THUMBNAILS_FOLDER))
+    execSync(`cp -r '${GENERIC_THUMBNAILS_SOURCE_FOLDER}/.' '${GENERIC_THUMBNAILS_FOLDER}'`)
```
CarlosNZ, line 38: Unrelated, but I just realised this shouldn't be inside the conditional. Oops.

PR #743, opened 2022-05-05
#742 Update user permissions endpoint

### LINKED ISSUE #742: /user-permissions endpoint only returns permissions that are already associated with templates
Which means that if a permission_name has been newly created and not associated with a template, it's not in the list of available permissions (which affects grantPermissions template)

Should be a matter of making one of the JOINs in the "permissions_all" view a LEFT JOIN

CarlosNZ:
Closes #742 

No changes to the actual endpoint code, it's all about just making sure the "permissions_all" View includes permission_names that don't have associated templates (via a LEFT JOIN)

Also fixed some erroneous migration code that had slipped through untested.

### DIFF
--- a/database/migration/migrateData.ts
+++ b/database/migration/migrateData.ts
@@ -216,7 +216,9 @@ const migrateData = async () => {
     // New field description returned in permissions_all
     console.log(' - Add permission_all VIEW field: description')
 
-    await DB.changeSchema(`CREATE OR UPDATE VIEW permissions_all AS (
+    await DB.changeSchema(`
+    DROP VIEW IF EXISTS permissions_all;
+    CREATE OR REPLACE VIEW permissions_all AS (
     SELECT
         "user".username AS "username",
         organisation.name AS "orgName",
--- a/database/migration/migrateData.ts
+++ b/database/migration/migrateData.ts
@@ -34,8 +34,8 @@ const migrateData = async () => {
   if (databaseVersionLessThan('0.1.0')) {
     console.log('Migrating to 0.1.0...')
     // Create "system_info" table
-    console.log(' - Add system_info TABLE');
-    
+    console.log(' - Add system_info TABLE')
+
     await DB.changeSchema(`
       CREATE TABLE public.system_info (
         id serial PRIMARY KEY,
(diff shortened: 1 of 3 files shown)


database/migration/migrateData.ts
```diff
@@ -34,8 +34,8 @@ const migrateData = async () => {
   if (databaseVersionLessThan('0.1.0')) {
     console.log('Migrating to 0.1.0...')
     // Create "system_info" table
-    console.log(' - Add system_info TABLE');
-    
+    console.log(' - Add system_info TABLE')
+
```
CarlosNZ, line 38: Someone didn't have their prettier config enabled!

database/migration/migrateData.ts
```diff
@@ -206,11 +212,13 @@ const migrateData = async () => {
       RETURN NEW; END;
       $$ LANGUAGE plpgsql;`
     )
-  
-  // New field description returned in permissions_all
-  console.log(' - Add permission_all VIEW field: description');
-  
-  await DB.changeSchema(`CREATE OR UPDATE VIEW permissions_all AS (
```
CarlosNZ, line 213: This wasn't correct. Should be "CREATE OR REPLACE VIEW". But that didn't work either, as adding columns in the middle of view wasn't allowed, so have to do "DROP VIEW" first.

Also -- this obviously hadn't been tested since it reported an obvious error when running `yarn migrate`. Need to remember to always thoroughly test migration code!

nmadruga: Yes... My bad, I did this without being able to do a proper check. I had it changed on the schema_init for the snapshot, so probably why I didn't see the error coming up. I hadn't tested with `yarn migrate` though. It would be better to improve the READ-ME page to list all these cases for development, since it's an open-source project we don't want other people running into the same problems.

nmadruga (approved):
👍

database/migration/migrateData.ts
CarlosNZ: Will do!

CarlosNZ: Done! ce996962

PR #744, opened 2022-05-06
#740 Shorthand mapping for generateDoc action

### LINKED ISSUE #740: Improvement: Add "shorthand" style field mapping for additionalData in generateDoc action
In the [modifyRecord](https://github.com/openmsupply/application-manager-server/wiki/List-of-Action-plugins#modify-record) action, we have implemented a "shorthand" mapping to make it quicker and easier to map many responses straight to output fields.

We should do the same for generateDoc so we don't have to have an elaborate "buildObject" operator to pass in additionalData from responses/

CarlosNZ:
Fix #740 

Implement shorthand mapping for "data" parameter which works exactly the same as it does in "modifyRecord".

Updated docs for clearer explanation

There is a dev template saved in the templates repo called "740_generate_doc_shorthand" which you can test:

- Template Builder -> Form 18 - Pharmacist Registration (Full) -> Actions -> DEV_TEST -> generateDoc:
Screen Shot 2022-05-06 at 12 35 39 PM
- Then you can execute this action by running SQL: `UPDATE application SET trigger = 'DEV_TEST' where id = 107;`
- The output PDF with fields correctly mapped is generated in `/files/S-YFW-0001


documentation/List-of-Action-plugins.md
```diff
@@ -538,24 +535,32 @@ Generates a PDF file based on a [Carbone](https://carbone.io/api-reference.html)
...
 
 `docTemplateId` specifies the uniqueId of the carbone template file (from the "file" table) and `options` optional can define a localisatiion to be used for dates and currency formatting.
 
-The data used by the action primarily comes from `applicationData` and `outputCumulative`, which are flattened/spread into a combined object to the carbone processer. However, if you wish to supply extra data, this can be added within an `additionalData` parameter. It gets sent to the carbone processer's "data" field like so:
+The data used by the action primarily comes from `applicationData` and `outputCumulative`, which are flattened/spread into a combined object to the carbone processer. Extra data (such as from individual responses, or other expressions) can be provided in one of two ways:
+
+- `data`: this parameter is an object in which you can define a simple mapping between field names required by Carbone and fields on `applicationData`. This is the same as the mapping available in `modifyRecord`, so please [see that plugin](#modify-record) info for more detail and examples.
```
nmadruga, line 550:
```suggestion
- `data`: this parameter is an object in which you can define a simple mapping between field names required by Carbone and fields on `applicationData`. This is the same as the mapping available in `modifyRecord`, so please [see modify-record plugin](#modify-record) info for more detail and examples.
```

Just better to have what is is instead of **that**

nmadruga (approved):
Seems like a good improvement

PR #745, opened 2022-05-07
Review list actions only returns active applications

CarlosNZ:
As per @adamdewey comments in Telegram chat:
Screen Shot 2022-05-08 at 11 35 20 AM

This fix just filters reviewer actions so only active (i.e. Outcome = "PENDING") applications are considered.


nmadruga (approved):
Looks great. Although I'm not sure why the migration didn't run correctly on one of the snapshots we currently have on version 0.2.0-something... Is this because it would only run if the snapshot was on a previous version?

CarlosNZ:
> Is this because it would only run if the snapshot was on a previous version?

If you can point to the example, I'll take a look at it.

PR #746, opened 2022-05-08
#735 Trigger payload fix

### LINKED ISSUE #735: Problem with scheduled actions: trigger payload too long
Noticed an error when some scheduled actions were run:
```
(node:23247) UnhandledPromiseRejectionWarning: error: payload string too long
    at Parser.parseErrorMessage (/Users/carl/GitHub/conforma/application-manager-server/node_modules/pg-protocol/src/parser.ts:357:11)
```

Turns out the "data" object is too large to send via a trigger, due to this limitation:

https://stackoverflow.com/questions/41057130/postgresql-error-payload-string-too-long

So we'll need to find a different way to retrieve that data than passing it as part of the trigger notification.

CarlosNZ:
Fix #735 

Removes elements from `pg_notify` functions that can possibly exceed byte-length limit and fetches them separately.

(Prevents error that you see when launching with the current Angola snapshot from scheduled actions)


nmadruga (approved):
Not sure how to test. Shouldn't be harmful. Would this be fixing this error?
Screen Shot 2022-05-23 at 2 17 48 PM

CarlosNZ:
> Not sure how to test. Shouldn't be harmful. Would this be fixing this error? Screen Shot 2022-05-23 at 2 17 48 PM

Yes

CarlosNZ:
> Not sure how to test

I've checked it pretty thoroughly, but the main thing is to ensure the trigger_queue and action_queue tables get populated with all the same, correct payload info as before -- they now get it via a separate query rather than passed through the notify payload.

nmadruga:
All good by me, it would take me too long to test this, since I'm not sure. But I trust that you checked that already, so good to go

PR #748, opened 2022-05-08
Localisation endpoints

CarlosNZ:
Endpoints for language install/uninstall/enable/disable:

- POST: `/admin/enable-language?code=<languageCode>&enabled=<true/false>`
- POST: `/admin/install-language`
- POST: `/admin/remove-language?code=<languageCode>`

See docs (API.md) for details.

Used by `/admin/localisations` UI in front-end ([PR #1190](https://github.com/openmsupply/application-manager-web-app/pull/1190))


documentation/API.md
```diff
@@ -8,6 +8,38 @@ The back-end currently has two server instances which are launched to handle inc
...
+      - [Check Triggers](#check-triggers)
+      - [Generate PDF](#generate-pdf)
+    - [Outcomes](#outcomes)
+    - [Admin only endpoints](#admin-only-endpoints)
+      - [Update row level policies](#update-row-level-policies)
+      - [Run Action](#run-action)
+      - [Manage localisations](#manage-localisations)
+      - [Snapshot endpoints](#snapshot-endpoints)
+      - [Lookup table endpoints](#lookup-table-endpoints)
+
```
nmadruga, line 41: 👏 Bravo!

documentation/API.md
```diff
@@ -396,6 +428,44 @@ returns:
...
+- To install a a new language into the system.
+
+**Input parameters** (as body JSON) example:
+```
+{
+    "language": {
+        "languageName": "Portugese",
+        "description": "Portugese translation for Angola users",
+        "code": "pt_an",
+        "flag": "🇦🇴",
```
nmadruga, line 448: Two points here: 
1 - Not sure if good to add this information here. It might be sensitive? Might be worth to use a less common example... Perhaps Tetum - so we keep the same approach as in mSupply mobile?
2 - Portugese -> Portuguese

documentation/API.md
```diff
@@ -396,6 +428,44 @@ returns:
...
+        ...
+    }
+}
+```
+
+**Returns**:
+```
+{
+    "success": true,
+    "message": "Language installed: Portugese / pt_an"
```
nmadruga, line 463: To be updated also

src/components/permissions/routes.ts
```diff
@@ -200,9 +201,7 @@ const routeGetPrefs = async (request: any, reply: any) => {
   const prefs = JSON.parse(
     readFileSync(path.join(getAppEntryPointDir(), '../preferences/preferences.json'), 'utf8')
   )
-  const languageOptions = JSON.parse(
-    readFileSync(path.join(getAppEntryPointDir(), '../localisation/languages.json'), 'utf8')
-  )
+  const languageOptions = readLanguageOptions()
```
nmadruga, line 204: Nice one : )

src/components/localisation/routes.ts
```diff
@@ -1,13 +1,181 @@
...
+
+  const changedLanguage = {
+    ...languageOptions[index],
+    enabled:
+      enabled === 'true' ? true : enabled === 'false' ? false : !languageOptions[index].enabled,
+  }
+  languageOptions[index] = changedLanguage
+
+  if (languageOptions.filter((language) => language.enabled).length === 0)
+    return reply.send({ success: false, message: 'At least one language must remain enabled' })
```
nmadruga, line 75: Do we need to localise these strings? I'm thinking about a local Admin person using another default language

nmadruga (commented):
- Can the default language be changed using the UI?
- Forgot to mention the GET `all-languages` in API docs :)

I'm going to review the Front-end PR now also

documentation/API.md
CarlosNZ: Yeah, good point. I changed the example to just a generic "Portuguese" with Brazil flag. No mention of Angola now.

3bca97b1

> 2 - Portugese -> Portuguese

😲

I've been doing it wrong my whole life 🤦‍♂️

src/components/localisation/routes.ts
CarlosNZ: I've been avoiding trying to localise server strings, cos it seems like a bit of a headache. We'll need a general solution at some point.

My thinking here was that anyone doing language management would need to be familiar with English anyway, since all the default strings are in English and you can't really make sense of the exported CSV without knowing English.

CarlosNZ: But the front-end ones are localised anyway

nmadruga (approved):
Thanks for changes. 🎉

src/components/localisation/routes.ts
nmadruga: That's a good point. I didn't realise it was back-end, far point.

documentation/API.md
nmadruga: 😆 Oh that's funny!

PR #749, opened 2022-05-09
1075F Part 1: Rename outcomes

CarlosNZ:
This just lays the groundwork for the work to be done in [#1075](https://github.com/openmsupply/application-manager-web-app/issues/1075) -- it simply renames the outcome tables and endpoints to use the more helpful "data" and "data displays"

Required [front-end PR #1191](https://github.com/openmsupply/application-manager-web-app/pull/1191)

Note: you'll probably need to manually run `yarn migrate` since it won't detect a version change


documentation/Data-Display.md
```diff
@@ -1,38 +1,34 @@
...
   - [API](#api)
-    - [`/outcomes`](#outcomes)
-    - [`/outcomes/table/<tableName>?<queries>`](#outcomestabletablenamequeries)
-      - [Query parameters](#query-parameters)
-    - [`/outcomes/table/<tableName>/item/<id>`](#outcomestabletablenameitemid)
   - [Configuration](#configuration)
-    - [`outcomes_display` table](#outcomes_display-table)
-    - [`outcome_display_column_definition` table](#outcome_display_column_definition-table)
+    - [`data_displays` table](#data_display-table)
+    - [`data_display_column_definition` table](#data_display_column_definition-table)
```
nmadruga, line 9: Hmmmm a little annoyed with this name. Couldn't the endpoint be data-tables instead?
data-displays is a bit strange. I'm sure we talked about how to name this before and had a better name

documentation/Data-Display.md
```diff
@@ -1,38 +1,34 @@
...
+Display of (custom) **Data** tables (e.g. Products, Users, Orgs) (or any table, strictly speaking) can be configured for viewing in the UI by setting entries in the `data_display` and `data_display_column_definition` database tables. This allows us to set:
 
 - who is allowed to see them
 - what columns to show different types of user
 - "special" columns, which show the result of an [evaluator](Query-Syntax.md) expression, so can show some fields combined, or queries to another table, for example
 - how columns should be formatted in the UI
 
-The data is served to the front-end via the `/outcomes` API end-points, and serves data based on the request's JWT header, so we can ensure that no user can receive data that they haven't been configured to be allowed to see.
+The data is served to the front-end via the `/data-displays` API end-points, and serves data based on the request's JWT header, so we can ensure that no user can receive data that they haven't been configured to be allowed to see.
```
nmadruga, line 21: If update, remember also to update the docs

documentation/Data-Display.md
```diff
@@ -1,38 +1,34 @@
...
 
-The database's row-level permissions ensure that only allowed rows are returned to the user, and the logic in the "outcomes" code returns allowed columns. (NB: We will also be adding some row-level logic to the outcomes API for additional configuration -- see [this issue](https://github.com/openmsupply/application-manager-server/issues/590))
+The database's row-level permissions ensure that only allowed rows are returned to the user, and the logic in the "outcomes" code returns allowed columns. We also support custom "rowRestrictions" for limiting the individual records a given user is allowed to see (see below).
```
nmadruga, line 23: There is a missing Outcomes left here

database/insertData/_common/04_data_displays.js
```diff
@@ -76,7 +76,7 @@ exports.queries = [
               }
             ]
           }
-          outcomeDisplayTablesUsingId: {
+          dataDisplayTablesUsingId: {
```
nmadruga, line 79: This could be a good timing to remove the inserts... Not sure if it is a good idea to keep this. We don't use this anymore...

src/components/data_display/helpers.ts
```diff
@@ -126,11 +126,7 @@ export const buildAllColumnDefinitions = async ({
   }
 }
 
-const getFilters = (
-  outcomes: OutcomeDisplay[],
-  userId: number,
-  orgId: number | undefined
-): object => {
+const getFilters = (outcomes: DataDisplay[], userId: number, orgId: number | undefined): object => {
```
nmadruga, line 129:
```suggestion
const getFilters = (outcomes: DataDisplay[], userId: number, orgId: number | undefined): object => {
```

Maybe good to change the variable as well?

nmadruga (commented):
The idea is nice, but jsut using this moment to say: can we have another endpint name? `data-displays` seems weird to me, can we vote on something?

documentation/Data-Display.md
CarlosNZ: Sure, the only other name I could think of (that we may have discussed?) is "data_views". "Display"/ "View" is what it is -- it's very important that we don't call it "data tables", since this is not a list of data tables, it's a list of different ways of *viewing* (or displaying) what's _in_ the data tables.

What do you think of "data_views"?

src/server.ts
```diff
@@ -119,9 +123,9 @@ const startServer = async () => {
     server.post('/login-org', routeLoginOrg)
     server.post('/create-hash', routeCreateHash)
     server.post('/generate-pdf', routeGeneratePDF)
-    server.get('/outcomes', routeOutcomes)
-    server.get('/outcomes/table/:tableName', routeOutcomesTable)
-    server.get('/outcomes/table/:tableName/item/:id', routeOutcomesDetail)
+    server.get('/data-displays', routeDataDisplays)
+    server.get('/data-displays/table/:tableName', routeDataDisplayTable)
+    server.get('/data-displays/table/:tableName/item/:id', routeDataDisplayDetail)
```
nmadruga, line 128: I'm confused. The front-end is using `data` and it works! But here is clearly set as `data-displays` 😮

nmadruga: Ok - confusion sorted. I see that Front-end is setting the route as `data` and this is just to get the definition to how the data gets displayed. 👍

nmadruga (approved):
Hi again. This looks good to me to be merged now : )

Note that I have changed the keys on `database/core_templates/localisation/pt_an/strings.json` and I'm planning to do the same (using the very cool new Languages UI page) to update the ones on our new Config work - to be recorded in the private repo. So it will all matched with the newly released version once we upgrade it too.

database/insertData/_common/04_data_displays.js
CarlosNZ: Yeah, it probably is, I was just a little nervous to get rid of them completely.

nmadruga:
I don't know for you but for me is very soothing to see `data_view` everywhere. Thanks, Carl!

Just a reminder. Do we need to update the `core_templates` with what's included here after this is included? I have updated the core_templates on branch `localisations` to have the "new" strings in Portuguese and languages setup, so recommend doing the same for this merged after the localisations. : )

CarlosNZ:
> Just a reminder. Do we need to update the `core_templates` with what's included here after this is included? I have updated the core_templates on branch `localisations` to have the "new" strings in Portuguese and languages setup, so recommend doing the same for this merged after the localisations. : )

Yes we should. But if it works as expected, it should just be a matter of loading the core templates with this version active, then re-exporting them. Will do it once these are all merged to develop.

PR #750, opened 2022-05-09
1075F Part 2: Migrate lookup tables

CarlosNZ:
Requires [front-end PR #1192](https://github.com/openmsupply/application-manager-web-app/pull/1192)

Updates modifyRecord action to follow table naming standards, and updates lookup tables to follow new convention.

Now any new table that is not part of the core *must* adhere to the following rules, which is enforced by both lookup tables and modifyRecord:
- prefixed "data_table_"
- snake_case
- singular name
- entry listed in "data_table" table

See inline comments for further details

Note: you'll probably need to manually run `yarn migrate` since it won't detect a version change


.postgraphilerc.js
```diff
@@ -1,6 +1,7 @@
 require('dotenv').config()
 module.exports = {
   options: {
+    watch: true,
```
CarlosNZ, line 4: Note: this is still not working properly, not sure why. If you create a new lookup table, then try and view it, it will fail until you stop and restart the pg server. But it's broken on develop too, so need to fix this.

database/buildSchema/44_data_display.sql
```diff
@@ -29,3 +29,12 @@ CREATE TABLE data_display_column_definition (
     UNIQUE (table_name, column_name)
 );
 
+-- Table for cataloguing all "data" tables, including lookup tables
+CREATE TABLE data_table (
+    id serial PRIMARY KEY,
+    table_name varchar NOT NULL UNIQUE,
+    name varchar,
+    field_map jsonb,
```
CarlosNZ, line 37: I'm not sure if we need "field_map", but I've kept it here for now. We can remove it later if we need to. It's not used for custom data tables, and I'm not sure how it's used in lookup tables, if at all.

database/migration/databaseMethods.ts
```diff
@@ -77,6 +77,34 @@ const databaseMethods = {
...
+      const result = await DBConnect.query({
+        text,
+        values: [tableName, name, fieldMap, isLookupTable],
+      })
+      return result.rows[0]
+    } catch (err) {
+      // Table record already exists, probably
+      console.log(err.message)
+    }
+  },
```
CarlosNZ, line 107: For migrating existing lookup tables to new format

plugins/action_modify_record/src/modifyRecord.ts
```diff
@@ -15,6 +23,8 @@ const modifyRecord: ActionPluginType = async ({ parameters, applicationData, DBC
     ...record
   } = parameters
 
+  const tableNameProper = getValidTableName(tableName)
```
CarlosNZ, line 26: We now strictly enforce table names. This also means we can give tables "pretty" names and use them if we like. So we can call our Products table "Registered Products" for example, and the created table with be called "data_table_registered_products", but we can keep using "Registered Products" happily in modifyRecord action, etc.

plugins/action_modify_record/src/modifyRecord.ts
```diff
@@ -2,7 +2,15 @@ import { ActionQueueStatus } from '../../../src/generated/graphql'
...
-import { mapValues, get } from 'lodash'
+import { mapValues, get, snakeCase } from 'lodash'
+import { singular } from 'pluralize'
+
+// This will be prepended to NEW table created if not already present
+export const DATA_TABLE_PREFIX = 'data_table_'
+
+// These are the only tables in the system that we allow to be mutated with this
+// plugin. All other names will have "data_table_" prepended.
+const ALLOWED_TABLE_NAMES = ['user', 'organisation']
```
CarlosNZ, line 13: This allows us to write to the "user" and "organisation" tables. But it also prevents writing to any other "core" table, which is a good thing. (Can add to this exception list later, if required)

src/lookup-table/services/LookupTable.service.ts
```diff
@@ -24,15 +25,15 @@ const LookupTableService = async (props: LookupTableServiceProps) => {
...
+  } else if (props.name) {
+    name = props.name
   }
 
   // Exported Methods
   const getAllRowsForTable = async () => await lookupTableModel.getAllRowsForTable(structure)
 
   const createTable = async () => {
-    tableName = toSnakeCase(tableNameLabel)
+    tableName = toSnakeCase(singular(name))
```
CarlosNZ, line 36: Singular table names enforced in back-end. It really does mess things up (because of Postgraphile inflections) if we give a table a plural name, so it's pretty important to do this.

src/lookup-table/services/LookupTable.service.ts
```diff
@@ -169,7 +170,10 @@ const LookupTableService = async (props: LookupTableServiceProps) => {
 
   const compareFieldMaps = () => {
     fieldMaps = [...dbFieldMap, ...fieldMaps].filter(
-      ((set) => (obj: any) => (set.has(obj.label) ? false : set.add(obj.label)))(new Set())
+      (
+        (set) => (obj: any) =>
+          set.has(obj.label) ? false : set.add(obj.label)
+      )(new Set())
```
CarlosNZ, line 176: Auto-formatting only

database/buildSchema/44_data_display.sql
```diff
@@ -29,3 +29,12 @@ CREATE TABLE data_display_column_definition (
     UNIQUE (table_name, column_name)
 );
 
+-- Table for cataloguing all "data" tables, including lookup tables
+CREATE TABLE data_table (
+    id serial PRIMARY KEY,
+    table_name varchar NOT NULL UNIQUE,
+    name varchar,
```
CarlosNZ, line 36: Just to clarify -- "table_name" is the actual Postgres table name (without the "data_table_" prefix), whereas "name" is the non-converted name.

E.g. if we enter "List of Ingredients" as the "name", the "table_name" will be "list_of_ingredient"

nmadruga (commented):
Carl, I think you broke `omsupply-bot` 😜 

I'm planning to review this tomorrow. Sorry ran out of time today...

database/buildSchema/44_data_display.sql
```diff
@@ -29,3 +29,12 @@ CREATE TABLE data_view_column_definition (
     UNIQUE (table_name, column_name)
 );
 
+-- Table for cataloguing all "data" tables, including lookup tables
```
nmadruga, line 32: Please rename this file to `43_data_table` if possible.

database/buildSchema/44_data_display.sql
nmadruga: Should we rename `name` to `description` or `display_name`. I think it would be clearer

plugins/action_modify_record/src/modifyRecord.ts
nmadruga: Should we also include `product` at this point? We were considering doing that. Just because is such a core thing, we could add the table with the minimum and add to this list to allow modifications

plugins/action_modify_record/src/modifyRecord.ts
nmadruga: > happily in modifyRecord action

is the table name unique?

src/lookup-table/models/LookupTable.model.ts
```diff
@@ -64,26 +64,26 @@ const LookupTableModel = () => {
...
   }
 
   const createTable = async ({
-    name: tableName,
+    tableName,
     fieldMap: fieldMaps,
   }: LookupTableBase): Promise<boolean> => {
     try {
-      const text = `CREATE TABLE lookup_table_${tableName}
+      const text = `CREATE TABLE data_table_${tableName}
```
nmadruga, line 86: Should we also receive the prefix as param?

nmadruga (commented):
Ok - just had a quick look. Hoping to have another look after reviewing the Front-end.

plugins/action_modify_record/src/modifyRecord.ts
CarlosNZ: > is the table name unique?

What do you mean? It behaves as it currently does -- if the table name exists it adds a new record to it, if the table doesn't exist it starts a new one.

plugins/action_modify_record/src/modifyRecord.ts
CarlosNZ: > Should we also include `product` at this point? We were considering doing that. Just because is such a core thing, we could add the table with the minimum and add to this list to allow modifications

I think not. We haven't installed anywhere yet, so it's easy enough to update our current snapshots (we'll have to do it manually) to match this style. (Also, we'll need to add a record to the "data_table" table anyway, and that expects the prefix to NOT be in the name -- that's how sworup set it up.)

So I think it's better to just bite the bullet and change existing data tables (product, license, etc) before we get to far into installing somewhere.

database/buildSchema/44_data_display.sql
CarlosNZ: Yeah, I think "display_name" seems appropriate. I'll update it (will need to change front-end too).

Good suggestion, I wasn't particularly happy with it.

CarlosNZ:
> Carl, I think you broke `omsupply-bot` 😜

Ha ha -- the bot hates me!

I think Mark has fixed it now ;)

database/buildSchema/44_data_display.sql
CarlosNZ: You mean 'data_view'? 😜

database/buildSchema/44_data_display.sql
CarlosNZ: Okay, it's used in the front-end to build the dynamic query for querying the lookup tables. So it's fine, we don't need to use it for regular "data" tables, just lookup tables.

database/buildSchema/44_data_display.sql
CarlosNZ: Here: 49a3ca01

Front-end updated too.

database/buildSchema/44_data_display.sql
CarlosNZ: Done.

src/lookup-table/models/LookupTable.model.ts
CarlosNZ: Yes, we probably should. I'll do it now.

src/lookup-table/services/LookupTable.service.ts
```diff
@@ -76,8 +77,8 @@ const LookupTableService = async (props: LookupTableServiceProps) => {
   }
 
   const updateTable = async () => {
-    tableName = structure.name
-    tableNameLabel = structure.label
+    tableName = structure.tableName
+    // displayName = structure.displayName
```
nmadruga, line 81: Why this is commented?

nmadruga (approved):
Look great. Just one comment lef!

src/lookup-table/services/LookupTable.service.ts
CarlosNZ: Done a6c79e9d

Thanks for reviews :)

PR #758, opened 2022-05-26
Update core_templates

CarlosNZ:
@nmadruga Let me know if this clashes with any changes you've made.

All I've done here is load core templates into the latest develop, then re-export it so it is update with the latest schema changes (and removed a few dead files that shouldn't have been there).


CarlosNZ:
Closing -- core templates updated on `main` and `develop` in latest round of updates.

PR #762, opened 2022-05-27
#761 Data views fix

### LINKED ISSUE #761: Don't require "dataTable" prefix in data_view definitions
This would be consistent with other areas.

Also -- check what happens with the created join tables (eg product_application_join) and make it consistent

CarlosNZ:
Fix #761

Data view entries do **NOT** require the prefix in the table name column -- it will automatically be added (and we check for the non-prefix version as well), but if you do add the prefix it should still work.

You'll need to rename the application join table to match then name of the entity table it's linked to (e.g. data_table_product -> data_table_product_application_join) AND update the column name to have the prefix included (i.e. product_id -> data_table_product_id)

### DIFF
--- a/src/components/data_display/helpers.ts
+++ b/src/components/data_display/helpers.ts
@@ -1,5 +1,5 @@
 import DBConnect from '../databaseConnect'
-import { objectKeysToCamelCase } from '../utilityFunctions'
+import { objectKeysToCamelCase, capitaliseFirstLetter } from '../utilityFunctions'
 import evaluateExpression from '@openmsupply/expression-evaluator'
 import fetch from 'node-fetch'
 import { camelCase, snakeCase, startCase } from 'lodash'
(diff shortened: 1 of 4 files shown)


src/components/data_display/helpers.ts
```diff
@@ -54,7 +55,11 @@ export const buildAllColumnDefinitions = async ({
   // Look up allowed Data views
   const dataTables = (await DBConnect.getAllTableNames()).map((tableName) => camelCase(tableName))
 
-  if (!dataTables.includes(tableName)) throw new Error(`Invalid table name: ${tableName}`)
+  const matchingTableName = dataTables.find(
+    (table) => table === dataTablePrefix + capitaliseFirstLetter(tableName) || table === tableName
+  )
+
+  if (!matchingTableName) throw new Error(`Invalid table name: ${tableName}`)
```
CarlosNZ, line 62: Checks for *either* with or without prefix. We now pass along the matched table name rather than the requested table name, so whichever one was found (prefixed or non-prefixed) will continue to be used for remaining computation.

This means that "user" and "organisation" tables will continue to work as expected.

nmadruga (approved):
LGTM

Issue #751, opened 2022-05-10
TO-DO: Change repo names


CarlosNZ:
Should be fairly straightforward, I think.

Suggest:

- conforma-server
- conforma-web-app

Happy with that @nmadruga ?

PR #763, opened 2022-05-27
751 Rename repo

CarlosNZ:
#751 Repo is renamed, these changes just replace all the text within the app.

The only ones that will actually affect functionality (I think) are the docker scripts, will need to run a build and deploy to check that it's all good.


docker/Dockerfile
```diff
@@ -60,11 +60,11 @@ RUN apt-get install yarn -y
 ARG SERVER_BRANCH=master
 ARG WEB_APP_BRANCH=master
```
nmadruga, line 61: We don't have a `master` branch... I know this is changed based on the current build, but the default should be `main` maybe.

docker/run.sh
```diff
@@ -18,13 +18,13 @@ docker run \
...
 # Some common container paths
 
 # nginx config: /etc/nginx/conf.d/
 # nginx log: /var/log/nginx/
 # postgresql /etc/postgresql/12/main/
 # postgresql log: /var/log/postgresql/
-# back end snapshots: /usr/src/application_manager_server/database/snapshots/
-# back end and postgraphile log: /var/log/application_manager
+# back end snapshots: /usr/src/conforma_server/database/snapshots/
+# back end and postgraphile log: /var/log/conforma
```
nmadruga, line 30: Would be good to test this with the instance for testing prior to updating our production server - even though we save the snapshot locally before re-uploading (and running a migration), I'm afraid it would fail due to changes on folders (and not getting the existing volumes

documentation/Demo-Server-Guide.md
```diff
@@ -227,8 +227,8 @@ To remove all currently unused volumes at once:
 
 ```bash
 # don't need bash inside contiainer for this
-sudo docker exec -ti conforma-on-8000_app_1 cat /var/log/application_manager/server.log
-sudo docker exec -ti conforma-on-8000_app_1 cat /var/log/application_manager/graphile.log
+sudo docker exec -ti conforma-on-8000_app_1 cat /var/log/conforma/server.log
+sudo docker exec -ti conforma-on-8000_app_1 cat /var/log/conforma/graphile.log
```
nmadruga, line 231: This is using new volume

documentation/Triggers-and-Actions.md
```diff
@@ -4,14 +4,15 @@
...
 
-- [Overview of Trigger and Action system](#overview-of-trigger-and-action-system)
-  - [List of available Triggers](#list-of-available-triggers)
-  - [Database tables that have a TRIGGER field](#database-tables-that-have-a-trigger-field-that-is-listened-to)
-  - [List of Core-actions available in initial configuration](https://github.com/openmsupply/application-manager-web-app/wiki/List-Core-Actions)
-- [Actions](#actions)
-- [Action parameters](#action-parameters)
-- [Passing information to Actions](#passing-information-to-actions)
-- [Running actions via REST endpoint](#running-actions-via-rest-endpoint)
+- [Triggers & Actions](#triggers--actions)
```
nmadruga, line 7:
```suggestion
- [Triggers & Actions](#triggers-actions)
```

nmadruga (changes requested):
Note: You changed the `image` name on `docker-compose` to be only confoma, but in `dockerise` on line 7 is still referred as: `IMAGE_NAME='conforma-demo'`

docker/Dockerfile
CarlosNZ: Thanks, changed it. Although I don't think this is actually doing anything since we normally build from a different branch or tag and it seems to work fine. 🤷‍♂️

CarlosNZ:
> Note: You changed the `image` name on `docker-compose` to be only confoma, but in `dockerise` on line 7 is still referred as: `IMAGE_NAME='conforma-demo'`

Cool, thanks. Fixed it.

docker/run.sh
CarlosNZ: I was planning on testing it with the 50004 instance as Jess wants the Fiji demo to go on there.

documentation/Demo-Server-Guide.md
CarlosNZ: Yeah, makes sense right?

documentation/Triggers-and-Actions.md
CarlosNZ: This is auto-generated and it's technically correct since the link it jumps to has an `&` in the middle.

But it's weird to have a link to the top of the page that we're already at, so I've just removed it altogether and indented the TOC back one. e2e62e25

nmadruga (approved):

PR #773, opened 2022-06-30
#768 Dummy action for Alias and alias functionality

### LINKED ISSUE #768: New dummy action for aliasing template actions
Part of "Preview Document" Epic: https://github.com/openmsupply/conforma-web-app/issues/1203

This action is really a "dummy" action -- it only exists to define a reference to another action and pass through some (optional) parameters, we never actually run it. We almost don't even need to create an "Action" plugin package for it, the logic happens all in the `triggersAndActions` module.

Here's what I'm thinking:
We designate a certain actionCode as a "special" alias code -- say `alias` or `actionAlias`. Then in `processTrigger`, in the bit where we fetch the array of actions (`getActionsByTemplateId`) -- if any actions match this alias actionCode, then we simply swap that item in the action array for the one that is being referenced, and override any parameter expressions with ones coming from the alias action, and the condition. Then everything else just continues as normal.

The front-end will have to have this dummy action in its list of available actions. So we _will_ need to create a dummy Action package in the back-end so it gets registered in the list of available actions, but it can probably just have plugin.json in it and nothing else.

Also:
Front-end:
- Template builder UI should allow setting the existing "code" field in Template Actions

Database:
- enforce "code" field being unique per template_id on "template_action" table (if not already done)

Question: presumably the condition expression for the swapped-out action should come from the alias? Or the alias condition takes priority, but if none is specified then we use the original action's condition expression?

CarlosNZ:
Fix #768 

Implemented as described in [issue](#768)

To test: there is a snapshot in the templates repo: `/dev/snapshots/AliasActionTesting`

If you load this, you'll see there is two aliased actions in the "Aliased Actions" template, with the DEV_TEST trigger. One has a condition set as "true" (i.e. will always run), the other doesn't so will run as per the condition of the original action. 

Trigger them both with:
```
UPDATE application SET trigger = 'DEV_TEST' where id = 239;
```

The two "sendNotification" actions will be called via the aliases. One will execute, but *won't* send an email, as the `sendEmail` parameter has been over-ridden to `false` (You'll be able to see the notification it created in the "notification" table). The other won't execute due to "Condition not met".

### DIFF
--- a/src/components/actions/helpers.ts
+++ b/src/components/actions/helpers.ts
@@ -1,0 +1,30 @@
+import { ActionInTemplate } from '../../types'
+import DBConnect from '../databaseConnect'
+
+export const swapOutAliasedActions = async (templateId: number, action: ActionInTemplate) => {
+  // Fetch aliased action from database
+  const {
+    condition,
+    parameter_queries: { code, condition: altCondition, ...overrideParams },
+  } = action
+  if (!code) throw new Error('Missing code in aliased action')
+
+  const aliasedAction = await DBConnect.getSingleTemplateAction(templateId, code)
+
+  if (!aliasedAction) throw new Error('No Action matching alias')
+
+  // Override condition if specified
+  // It would make most sense for actual condition field to be the one we look
+  // at for overriding the aliased action's condition, but because this has
+  // default "true" and can never be null, then we'd always get "true" as the
+  // condition. So we only consider the condition field if it's something other
+  // than "true". And if we actually *want* it to be "true", then we can
+  // override it further by specifying a "condition" field in parameters
+  if (condition !== true) aliasedAction.condition = condition
+  if (altCondition !== undefined) aliasedAction.condition = altCondition
+
+  // Override parameters
+  aliasedAction.parameter_queries = { ...aliasedAction.parameter_queries, ...overrideParams }
+
+  return aliasedAction
+}
--- a/src/components/actions/triggersAndActions.ts
+++ b/src/components/actions/triggersAndActions.ts
@@ -23,6 +23,7 @@ import { ActionQueueStatus, TriggerQueueStatus } from '../../generated/graphql'
 import scheduler from 'node-schedule'
 import config from '../../config'
 import { DateTime } from 'luxon'
+import { swapOutAliasedActions } from './helpers'
 
 // Dev configs
 const showApplicationDataLog = false
@@ -76,17 +77,19 @@ export async function processTrigger(payload: TriggerPayload) {
   const templateId = await DBConnect.getTemplateIdFromTrigger(payload.table, payload.record_id)
 
   // Get Actions from matching Template (and match templateActionCode if applicable)
-  const actions = await (
-    await DBConnect.getActionsByTemplateId(templateId, trigger)
-  ).filter((action) => {
-    if (!event_code) return true
-    else return action.event_code === event_code
-  })
+  const actions = (await DBConnect.getActionsByTemplateId(templateId, trigger))
+    .filter((action) => {
+      if (!event_code) return true
+      else return action.event_code === event_code
+    })
+    .map((action) => (action.code !== 'alias' ? action : swapOutAliasedActions(templateId, action)))
+
+  const resolvedActions = await Promise.all(actions)
 
   // Separate into Sequential and Async actions
   const actionsSequential: ActionSequential[] = []
   const actionsAsync: ActionInTemplate[] = []
-  for (const action of actions) {
+  for (const action of resolvedActions) {
     if (action.sequence) actionsSequential.push(action as ActionSequential)
     else actionsAsync.push(action)
   }
--- a/plugins/action_alias/plugin.json
+++ b/plugins/action_alias/plugin.json
@@ -1,0 +1,24 @@
+{
+  "type": "action_plugin",
+  "code": "alias",
+  "name": "Alias Action",
+  "description": "Dummy plugin to facilitate aliasing of template actions",
+  "required_parameters": ["code"],
+  "optional_parameters": ["condition", "...anyPluginFields"],
+  "output_properties": [],
+  "info": {
+    "author": {
+      "name": "The mSupply Foundation",
+      "url": "http://msupply.foundation"
+    },
+    "keywords": ["action", "update"],
+    "logos": {
+      "small": "",
+      "large": ""
+    },
+    "links": [],
+    "screenshots": [],
+    "version": "1.0.0",
+    "updated": "2022-06-30"
+  }
+}
--- a/plugins/action_alias/src/index.ts
+++ b/plugins/action_alias/src/index.ts
@@ -1,0 +1,4 @@
+// Dummy file so the system thinks this plugin exists and template_elements can
+// be created for it. But all the alias logic is handled internally as part of
+// the "processTrigger" function
+export {}
--- a/src/components/databaseConnect.ts
+++ b/src/components/databaseConnect.ts
@@ -44,6 +44,8 @@ class DBConnect {
 
   public getActionsByTemplateId = PostgresDB.getActionsByTemplateId
 
+  public getSingleTemplateAction = PostgresDB.getSingleTemplateAction
+
   public updateActionPlugin = PostgresDB.updateActionPlugin
 
   public updateTriggerQueueStatus = PostgresDB.updateTriggerQueueStatus
--- a/src/components/postgresConnect.ts
+++ b/src/components/postgresConnect.ts
@@ -419,6 +419,25 @@ class PostgresDB {
     }
   }
 
+  public getSingleTemplateAction = async (
+    templateId: number,
+    code: string
+  ): Promise<ActionInTemplate> => {
+    const text = `
+      SELECT action_plugin.code, action_plugin.path, action_plugin.name, trigger, template_action.event_code AS event_code, sequence, condition, parameter_queries 
+      FROM template 
+      JOIN template_action ON template.id = template_action.template_id 
+      JOIN action_plugin ON template_action.action_code = action_plugin.code 
+      WHERE template_id = $1 AND template_action.code = $2
+    `
+    try {
+      const result = await this.query({ text, values: [templateId, code] })
+      return result.rows[0] as ActionInTemplate
+    } catch (err) {
+      throw err
+    }
+  }
+
   public updateActionPlugin = async (plugin: ActionPlugin): Promise<boolean> => {
     const setMapping = Object.keys(plugin).map((key, index) => `${key} = $${index + 1}`)
     const text = `UPDATE action_plugin SET ${setMapping.join(',')} WHERE code = $${


database/migration/migrateData.ts
```diff
@@ -380,6 +380,18 @@ const migrateData = async () => {
...
 
+  // v0.3.1
+  if (databaseVersionLessThan('0.3.1')) {
+    console.log('Migrating to v0.3.1...')
+
+    // Enforce code must be unique per template in template_action table
+    console.log(' - Creating unique index for code/template in template_action')
+    await DB.changeSchema(`
+      CREATE UNIQUE INDEX IF NOT EXISTS unique_template_action_code
+      ON template_action (code, template_id)
```
andreievg, line 391: Just something I am thinking about, it's probably good idea to save version at the end of each migration (in case migration fails, we at least save version up to that last migration)

CarlosNZ: You mean the version of the app, or the version of where we migrated to?

We currently save the database version in the "system_info" table, and that's what we compare to the current app version (from package.json) on start-up. Is that what you mean?

documentation/List-of-Action-plugins.md
```diff
@@ -658,6 +659,26 @@ Can supply *either* `applicationId` or `applicationSerial`, although this can be
...
+| `code`\*                                 | `output as per the aliased action` |
+| `condition`                              |                                    |
+| `...any other aliased action parameters` |                                    |
+
+To be able to reference an existing template action, it needs to have a unique (per template) code in its template_action record.
+
+Then this alias action just passes in the code of the action it's referencing and the trigger processor will swap it out for that one.
+
+By default the action runs with the condition and parameters it has defined, but these can be overridden by the alias. To override the condition, you can either set the "condition" field directly, or supply a "condition" parameter (which takes priority). The one problem with the native "condition" field, though, is that it is always `true` by default, so we wouldn't want to always override with that. Therefore, the "condition" field only overrides the aliased actions condition if it is something other than `true`. (If you want it to be `true` (i.e. always run, then set `condition: true` in parameters))
```
andreievg, line 680: This is not 100% clear, an example would be great. 

>  Therefore, the "condition" field only overrides the aliased actions condition if it is something other than `true`. (If you want it to be `true` (i.e. always run, then set `condition: true` in parameters))

Like this one, to me it read, `condition overrides aliased condition if value other then true, to alway run set condition: true` (which seems to contradict)

andreievg: Ohh i think my confusion is between `field` and `parameter`, maybe we can differentiate somehow

src/components/actions/helpers.ts
```diff
@@ -0,0 +1,30 @@
...
+    condition,
+    parameter_queries: { code, condition: altCondition, ...overrideParams },
+  } = action
+  if (!code) throw new Error('Missing code in aliased action')
+
+  const aliasedAction = await DBConnect.getSingleTemplateAction(templateId, code)
+
+  if (!aliasedAction) throw new Error('No Action matching alias')
+
+  // Override condition if specified
```
andreievg, line 16: Hmm, I think a boolean would be more explanatory. `shouldOverrideCondition` as a parameter, would signify that we should use the `filed` condition of alias instead of the aliased one.

documentation/List-of-Action-plugins.md
andreievg: See below i've have a suggestion

documentation/List-of-Action-plugins.md
```diff
@@ -658,6 +659,26 @@ Can supply *either* `applicationId` or `applicationSerial`, although this can be
 
 Note: There is a database trigger/postgres listener to automatically delete files when their database record is deleted, so we only need to delete the records, not the files themselves.
 
+### Aliasing existing template actions
+
+Finally, we have this "special" action which allows us to create an alias to an existing template action. The use for this is if you have configured a complex action with a lot of parameters or queries, and you wish to run it with slight differences for a document preview (for example), you can just set an alias to the original action instead of configuring and maintaining the same thing twice within the template (LINK TO PREVIEW DOCUMENTATION WHEN ITS DONE)
```
andreievg, line 664: Can we add a quick term glossary, that `alias` means this action and `aliased` means the action it is referring too, sorry I got a little bit confused about the termss

documentation/List-of-Action-plugins.md
```diff
@@ -658,6 +659,26 @@ Can supply *either* `applicationId` or `applicationSerial`, although this can be
...
+
+In fact, this is not really an action at all -- it's just a dummy plugin that makes the action library and front-end (and therefore template_actions) see it as an available action. But all the logic is handled internally as part of the `processTrigger` functionality -- if it sees an "alias" action, the function simply swaps it out with the template action it is referencing.
+
+- _Action Code:_ **`alias`**
+
+| Input parameters<br />(\*required) <br/> | Output properties                  |
+| ---------------------------------------- | ---------------------------------- |
+| `code`\*                                 | `output as per the aliased action` |
```
andreievg, line 672: hmm, I wonder if there is a short but more explanatory term for this 'codeOfActionToRun` lol

src/components/actions/triggersAndActions.ts
```diff
@@ -76,17 +77,19 @@ export async function processTrigger(payload: TriggerPayload) {
...
-    else return action.event_code === event_code
-  })
+  const actions = (await DBConnect.getActionsByTemplateId(templateId, trigger))
+    .filter((action) => {
+      if (!event_code) return true
+      else return action.event_code === event_code
+    })
+    .map((action) => (action.code !== 'alias' ? action : swapOutAliasedActions(templateId, action)))
+
+  const resolvedActions = await Promise.all(actions)
```
andreievg, line 87: nice, simple and direct. (maybe a comment about Promise.all, that swapOutAliasedActions returns a promise that might needs to be resolved)

andreievg (commented):
Thanks,

Logic looks good, and docs too, have a couple of suggestions though

src/components/actions/helpers.ts
CarlosNZ: I dunno, much of the time it's better to leave out the condition and let the original action's condition be considered.
Then if you want to change it, just provide a new condition parameter -- I don't see why you'd want to also have to provide a boolean flag as well. The condition parameter only ever exists when `shouldOverrideCondition` would be `true` -- it's implicitly true.

My thinking was that these parameters all work like when spreading objects and having overrides. Like:

```
overrides = {option1: true, option2: "A"}

baseSettings = {option1: false, option2: "B", option3: 99, option4: "happy"}
```
Then `activeSettings = {...baseSettings, ...overrides}`

I use this pattern a lot, it seems nice and self-explanatory to me. I don't like a bunch of extraneous compulsory, hard-to-remember flags unless they're unavoidable.

I agree that I could be clearer in the docs though -- I'll try and improve the explanation.

documentation/List-of-Action-plugins.md
CarlosNZ: The field in `template_action` is called `code` though, so was trying to make it match. I'd be reluctant to change the name of the field, as it's intended to be a unique identifier that could be used for many different purposes -- anything that needs to refer to a specific template_action.

`codeOfActionToRun` doesn't quite work either, as it's not clear you're talking about the *specific* template_action instance as opposed to the "actionCode" which defines the action type. We could probably have been a little clearer overall though, cos there is a little inconsistency in how the term "code" is used elsewhere lol.

src/components/actions/triggersAndActions.ts
CarlosNZ: This used to be a "for...of" loop, which loops synchronously. Swapping it for a .filter and .map makes them all async, hence the Promise.all() afterwards.

So the Promise.all is due to the async looping, not the swapOutAliasedActions function specifically.

Added a comment though: 6a54efe4

src/components/actions/helpers.ts
CarlosNZ: I see your confusion probably came from this:
> To override the condition, you can either set the "condition" field directly, or supply a "condition" parameter (which takes priority). The one problem with the native "condition" field, though, is that it is always `true` by default, so we wouldn't want to always override with that. Therefore, the "condition" field only overrides the aliased actions condition if it is something other than `true`. (If you want it to be `true` (i.e. always run, then set `condition: true` in parameters))

Okay, I think I can see the point of the "shouldOverrideCondition" parameter -- but you should only need it in the rare situation where the alias has `condition=true` AND you want that to over-ride the aliased action.

So to summarise:
- default behaviour: original (aliased) action condition takes priority
- if the alias has a condition that is anything other than `true`, then *that* overrides the original action
- if the alias has `condition=TRUE` AND we *want* this to take priority, then we need to add the `shouldOverrideCondition` flag as a parameter, but we shouldn't need it at all in those other two situations.

Sound okay?

(Another way would be to make the default condition `null` rather than `true`, but I think that creates more problems than it solves 😆)

CarlosNZ: Okay, I've updated the docs to reflect the above: 5b22f6fc

See if that makes sense and, if so, I'll update the actual code to reflect it.

documentation/List-of-Action-plugins.md
CarlosNZ: In my latest doc update (5b22f6fc) I've tried to keep to "alias" for the new action, and "original action" for the one it references. This okay?

src/components/actions/helpers.ts
CarlosNZ: It's a simple code change: 4fbe57d8

I might have to update my snapshot for "PreviewAction" to reflect this, I'll check before you get to it ;)

database/migration/migrateData.ts
andreievg: the version that migration block is for:
```
if (databaseVersionLessThan('0.3.1')) {
    console.log('Migrating to v0.3.1...')

    // Enforce code must be unique per template in template_action table
    console.log(' - Creating unique index for code/template in template_action')
    await DB.changeSchema(`
      CREATE UNIQUE INDEX IF NOT EXISTS unique_template_action_code
      ON template_action (code, template_id)
   `)

    // Here -> 
    await DB.setDatabaseVersion('0.3.1')
  }
```

documentation/List-of-Action-plugins.md
andreievg: thanks

database/migration/migrateData.ts
andreievg: Just food for thought btw, not required change

documentation/List-of-Action-plugins.md
andreievg: Sounds good

src/components/actions/helpers.ts
andreievg: yeah, sorry if my PR reviews are a bit late and quite a few upstream changed need to be done because of it, i've pmed saying it's probably best to just do the changes to the latest pr in the stack to avoid conflicts etc..

andreievg (approved):
Thanks, looks good

PR #774, opened 2022-06-30
768(2) re-organise trigger/action code

CarlosNZ:
No actual changes here, just re-organising the triggersAndActions code to be a bit more modular and manageable -- been meaning to do this for a while.

Would have included with base PR (#773) but that would have made it difficult to review.


andreievg:
> Would have included with base PR (https://github.com/openmsupply/conforma-server/pull/773) but that would have made it difficult to review. 

Thanks for that consideration

src/components/actions/processTrigger.ts
```diff
@@ -0,0 +1,101 @@
+import { ActionInTemplate, TriggerPayload, ActionSequential } from '../../types'
+import DBConnect from '../databaseConnect'
+import { actionLibrary } from '../pluginsConnect'
+import { EvaluatorNode } from '@openmsupply/expression-evaluator/lib/types'
+import { executeAction } from './executeAction'
+import { ActionQueueStatus, TriggerQueueStatus } from '../../generated/graphql'
+import { swapOutAliasedActions } from './helpers'
+
+// Dev config
+const showActionOutcomeLog = false
```
andreievg, line 10: process.env ?

andreievg: as in maybe we should have like a `debug` env variable to toggle this ?

src/components/actions/processTrigger.ts
```diff
@@ -0,0 +1,101 @@
...
+
+  const resolvedActions = await Promise.all(actions)
+
+  // Separate into Sequential and Async actions
+  const actionsSequential: ActionSequential[] = []
+  const actionsAsync: ActionInTemplate[] = []
+  for (const action of resolvedActions) {
+    if (action.sequence) actionsSequential.push(action as ActionSequential)
+    else actionsAsync.push(action)
+  }
```
andreievg, line 33: Personal stylistic preference

```suggestion
// Separate into Sequential and Async actions
  const actionsSequential = resolvedActions.filter(
    ({ sequence }) => !!sequence
  ) as ActionSequential[]
  const actionsAsync = resolvedActions.filter(({ sequence }) => !sequence)
```

src/components/actions/processTrigger.ts
CarlosNZ: Maybe, but you'd want to have some fine-grained control over what gets printed. Wouldn't want to enable `debug` and have a whole lot of big objects scroll past when you're only interested in one of them.

But yes, in general, sounds like a good idea.

src/components/actions/processTrigger.ts
```diff
@@ -0,0 +1,101 @@
...
+        parameter_queries: action.parameter_queries,
+        trigger_payload: action.trigger_payload,
+      }
+      const result = await executeAction(actionPayload, actionLibrary, {
+        outputCumulative,
+      })
+      outputCumulative = { ...outputCumulative, ...result.output }
+      // Debug helper console.log to inspect action outputs:
+      if (showActionOutcomeLog) console.log('outputCumulative:', outputCumulative)
+      if (result.status === ActionQueueStatus.Fail) console.log(result.error_log)
```
andreievg, line 91: should we be setting actionFailed = action.action_code here ?

src/components/actions/executeAction.ts
```diff
@@ -0,0 +1,98 @@
...
+  // Evaluate condition
+  let condition
+  try {
+    condition = await evaluateExpression(
+      payload.condition_expression as EvaluatorNode,
+      evaluatorParams
+    )
+  } catch (err) {
+    console.log('>> Error evaluating condition for action:', payload.code)
+    return await DBConnect.executedActionStatusUpdate({
```
andreievg, line 44: hmm since you are refactoring this, passing through payload through to return of db method is not that clear. I think it's better in two steps:

```js
let actionResult = {
      error_log: 'Problem evaluating condition: ' + err,
      parameters_evaluated: null,
      output: null,
      id: payload.id,
    };

 await DBConnect.executedActionStatusUpdate(actionResult)
 return actionResult
```

src/components/actions/executeAction.ts
```diff
@@ -0,0 +1,98 @@
...
+      status: ActionQueueStatus.Fail,
+      error_log: 'Problem evaluating condition: ' + err,
+      parameters_evaluated: null,
+      output: null,
+      id: payload.id,
+    })
+  }
+
+  if (condition) {
```
andreievg, line 53: could do early return here if !condition

src/components/actions/processTrigger.ts
```diff
@@ -0,0 +1,101 @@
...
+          : ActionQueueStatus.Queued,
+    })
+  }
+  await DBConnect.updateTriggerQueueStatus({
+    status: TriggerQueueStatus.ActionsDispatched,
+    id: trigger_id,
+  })
+
+  // Get sequential Actions from database
+  const actionsToExecute = await DBConnect.getActionsProcessing(templateId)
```
andreievg, line 58: I've just been looking at this with fresh mind, and would be great to have a little comment here that asynchronous (queued) actions are initiated via pg notify

andreievg:
> Would have included with base PR (https://github.com/openmsupply/conforma-server/pull/773) but that would have made it difficult to review. 

Thanks for that consideration

andreievg:
I was looking at how async actions are triggered:

Add to queue:

https://github.com/openmsupply/conforma-server/blob/bd8fb087a2b6207d7d2ac22a9b2f575a0f59c79e/src/components/actions/processTrigger.ts#L37-L51

Notify

https://github.com/openmsupply/conforma-server/blob/bd8fb087a2b6207d7d2ac22a9b2f575a0f59c79e/database/buildSchema/25_action_queue.sql#L29-L40

Listen to notify

https://github.com/openmsupply/conforma-server/blob/bd8fb087a2b6207d7d2ac22a9b2f575a0f59c79e/src/components/postgresConnect.ts#L55-L64

And then I was a little bit confused about outputCumulative in processTrigger being extracted from data and then used in executeAction, but the same not happening in above, but realised it still happens because because it's in data which is passed on to executeAction in the above.

src/components/actions/processTrigger.ts
CarlosNZ: Even cleaner now that we can remove the `as`: 5af9f91b

src/components/actions/processTrigger.ts
CarlosNZ: Hmm.. yes, strictly we probably should. However I'm slightly reluctant to change it now as I don't want to inadvertently break an existing template. 

From what I remember from debugging the action queue, most of the time an error is thrown, so it takes the "catch" route and writes actionFailed to true (therefore preventing subsequent actions from running). I'm basing that on the observation that I pretty much always see `Action cancelled due to failure of previous sequential action` in the error log of the action_queue table after a failed action.

However, I'll keep an eye on this as I do the next lot of configs and see if I'm correct -- and if I'm wrong whether there'd be any negative consequences. Then I'll come back and change this once I'm sure it's all good.

src/components/actions/executeAction.ts
CarlosNZ: I think the reason I did it like that (it's been a while!) was so that if there is a database error, then that will get thrown back up to the caller instead of the action result. 

But I guess the database method would throw an error which would still get thrown from the catch block, wouldn't it?

So yeah, okay, will simplify:  918b9ba7

(BTW -- I seem to remember picking up this `return await` syntax from you in the first place 😉 ):
https://github.com/openmsupply/conforma-server/blob/47b013ede13d29ca10f3f71a03e911191245dfc9/database/updateRowPolicies.ts#L9-L11

src/components/actions/executeAction.ts
CarlosNZ: Sure: b8d192c0

Haven't tested, but seems like it should be identical behaviour.

src/components/actions/processTrigger.ts
CarlosNZ: Yeah good point, I've found myself wondering where the heck they get dealt with when looking through this before lol 850ae0fa

CarlosNZ:
> And then I was a little bit confused about outputCumulative in processTrigger being extracted from data and then used in executeAction, but the same not happening in above, but realised it still happens because because it's in data which is passed on to executeAction in the above.

Yeah, this was clearer before, but we had that problem a while back when "data" can be too big for pg_notify, so had to not pass it through and re-fetch it in listener method instead, which is a bit harder to follow.

src/components/actions/processTrigger.ts
andreievg: Ok, i take your point about existing config might fail, but silently ignored error might be worse, as the side effects are only observed way further up the track (and at this stage it's hard to find the root).

src/components/actions/executeAction.ts
andreievg: return await syntax is fine imo, in the signPromise it return a new thing, but in the executeAction, when return await executedActionStatusUpdate, it's just passing through the parameter it received, which cause a little bit confusion for me.

> But I guess the database method would throw an error which would still get thrown from the catch block, wouldn't it?

Yep, it would propagate either way

andreievg:
Thanks for changes

PR #775, opened 2022-07-01
#769 /preview-action endpoint

### LINKED ISSUE #769: Preview action endpoint and handling
Part of Epic https://github.com/openmsupply/conforma-web-app/issues/1203 

Agreed solution: https://github.com/openmsupply/conforma-web-app/issues/1203#issuecomment-1169449868

- Create endpoint `/preview-action`
- Create additional Trigger value "ON_PREVIEW" (don't forget migration code)
- Handle processing as per agreed solution and return data

### LINKED ISSUE #776: Add "Name" and/or "Description" parameter for "generateDoc"
Part of Epic https://github.com/openmsupply/conforma-web-app/issues/1203

This would be useful for making the Document previews look a bit nicer. With current values, we can only show the filename, but if they had a name and/or description, this would be informative to the reviewer (and probably applicant at some point).

Edit -- there's already a "description" field in the file table and in the generateDoc action, so all we need to do is ensure the description field is included in the action output, then include that in the returned display data from the preview endpoint

CarlosNZ:
Fix #769 

Implements as per the [agreed solution](https://github.com/openmsupply/conforma-web-app/issues/1203#issuecomment-1169449868)

See inline comments for additional details.

Edit: this now includes the file "description" value, which closes #776 as well

### To test:

Load snapshot in template repo: `/dev/snapshots/previewDocsTesting_2022-07-01`

Have changed the "DEV_TEST" triggers in the last PR to "ON_PREVIEW" triggers and added a document that gets generated.

You can see the resulting data returned (and doc created) when either approving or rejecting using Postman:

POST request: `http://localhost:8080/api/preview-actions` (JWT authorization as normal)

To reject, use this body JSON:
```
{
    "reviewId": 3,
    "previewData": {
        "outcome": "REJECTED"
    }
}
```

To approve, use this:
```
{
    "reviewId": 3,
    "previewData": {
        "outcome": "APPROVED"
    }
}
```

This overrides the current "outcome" value (PENDING) with the simulated one, which is the condition on which either the approve or reject notifications go out.


src/components/actions/executeAction.ts
```diff
@@ -17,10 +23,11 @@ const graphQLEndpoint = config.graphQLendpoint
...
   payload: ActionPayload,
   actionLibrary: ActionLibrary,
-  additionalObjects: any = {}
+  additionalObjects: any = {},
+  previewData?: Partial<ActionApplicationData>
 ): Promise<ActionQueueExecutePayload> {
-  // Get fresh applicationData for each Action
-  const applicationData = await getApplicationData({ payload })
+  // Get fresh applicationData for each Action, and inject previewData if present
+  const applicationData = merge(await getApplicationData({ payload }), previewData)
```
CarlosNZ, line 30: Need to use lodash merge not just spread operators, cause we will sometimes need to override deep properties without obliterating the other properties in the same deep object -- spread operators only go one level deep.

src/components/actions/helpers.ts
```diff
@@ -44,7 +46,52 @@ export const swapOutAliasedActions = async (templateId: number, action: ActionIn
   if (altCondition !== undefined) aliasedAction.condition = altCondition
 
   // Override parameters
-  aliasedAction.parameter_queries = { ...aliasedAction.parameter_queries, ...overrideParams }
+  aliasedAction.parameter_queries = merge(aliasedAction.parameter_queries, overrideParams)
```
CarlosNZ, line 49: Same here.

There's still a problem when you get a mixture of evaluated and pre-evaluated properties in the same object, but that's a pretty edge case. Something to bear in mind though.

src/components/files/documentGenerate.ts
```diff
@@ -14,7 +14,7 @@ import { render, RenderCallback, RenderOptions } from 'carbone'
 
 const appRootFolder = getAppEntryPointDir()
 const filesFolder = config.filesFolder
-const PDF_THUMBNAIL = 'noun_PDF_3283219.png'
+const PDF_THUMBNAIL = `${config.genericThumbnailsFolderName}/noun_PDF_3283219.png`
```
CarlosNZ, line 17: Oops -- think this has been broken for a while. 😳

CarlosNZ:
FYI -- I've been finding the Carbone/LibreOffice PDF conversion *very* flaky on my local system. More than half the time it'll timeout without finishing. Eventually it'll start working for subsequent executions, and then it'll be fine again until the server is restarted.

Wonder if it's an Mac M1 thing, will have to see how it goes on the demo server. Let me know how it is for you, @andreievg

src/components/actions/helpers.ts
andreievg: hmm, are you sure that's desired behaviour, don't we want to override the whole parameter ?

src/components/actions/executeAction.ts
andreievg: I think this is kind of dangerous, would expect previewData to override the whole object, otherwise it's hard to think about permutations of combination of object (and no way to actually override the whole object if needed). What would be a use case for overriding something that's deep nested ?

CarlosNZ: In applicationData.reviewData.reviewDecision -- we want to set reviewDecision.decision without destroying the whole "reviewData" object within applicationData. This is the exact case I came across that prompted me to make this change.

CarlosNZ: You can still override the whole object if you set "applicationData.reviewData" in override props, but this allows us to surgically target a specific deeply-nested property if we need to without destroying its sibling properties (but we still can if we want to as well)

CarlosNZ: Also, in "generateDoc", for the LOQ document, we need to pass in a few extra "additionalData" props in without removing the existing ones.

It's demonstrated in the front-end of this PR, with the associated snapshot.

Just out of interest, why would you ever NOT want to do this? In which use case would you want to pass in a deeply nested property path and have it destroy all the parent property objects on the way down?

src/components/actions/helpers.ts
```diff
@@ -44,7 +46,60 @@ export const swapOutAliasedActions = async (templateId: number, action: ActionIn
...
+  status: ActionQueueStatus
+  displayString: string
+  text?: string // for email string
+  fileId?: string // for generating url in front-end
+  errorLog: string | null // if Action failed
+}
+
+// Convert the full action output into a simplified format that can be easily
+// displayed by the front-end Preview module
+export const createDisplayData = (actionsOutput: ActionResult[]): ActionResultDisplayData[] => {
```
andreievg, line 72: feels like previewActionRoute is a good place for this one

src/components/actions/helpers.ts
```diff
@@ -44,7 +46,60 @@ export const swapOutAliasedActions = async (templateId: number, action: ActionIn
   if (altCondition !== undefined) aliasedAction.condition = altCondition
 
   // Override parameters
-  aliasedAction.parameter_queries = { ...aliasedAction.parameter_queries, ...overrideParams }
+  aliasedAction.parameter_queries = merge(aliasedAction.parameter_queries, overrideParams)
 
   return aliasedAction
 }
+
+interface ActionResult {
```
andreievg, line 54: hmm maybe `processTrigger(..): ActionResult[]`

src/components/actions/executeAction.ts
andreievg: > Just out of interest, why would you ever NOT want to do this? In which use case would you want to pass in a deeply nested property path and have it destroy all the parent property objects on the way down?

This makes sense (you are right for the previewData override, it makes sense to override in a nested way)
I think I was more thinking about the other example with parameter_queries override, like what happens when evaluator expressions get merged

src/types.ts
```diff
@@ -170,13 +170,14 @@ export interface FileGetPayload {
...
 export interface TriggerPayload {
-  trigger_id: number
+  trigger_id: number | null
   trigger: Trigger
   table: string
   record_id: number
   application_id?: number
   event_code?: string
   data?: { [key: string]: any }
+  previewData?: Partial<ActionApplicationData>
```
andreievg, line 180: I would call it applicationDataOverride, I think it's a bit more clear

src/components/actions/processTrigger.ts
```diff
@@ -82,20 +87,38 @@ export async function processTrigger(payload: TriggerPayload) {
...
+        actionPayload,
+        actionLibrary,
+        {
+          outputCumulative,
+        },
+        previewData
+      )
+
       outputCumulative = { ...outputCumulative, ...result.output }
+      if (result.status !== ActionQueueStatus.ConditionNotMet)
```
andreievg, line 100: Hmm, i think this should really be filtered out at the consumption end, it's still actionOutput

andreievg: Also I've noticed the FAIL ones are coming through to displayData return

andreievg:
>FYI -- I've been finding the Carbone/LibreOffice PDF conversion very flaky on my local system. More than half the time it'll timeout without finishing. Eventually it'll start working for subsequent executions, and then it'll be fine again until the server is restarted.

> Wonder if it's an Mac M1 thing, will have to see how it goes on the demo server. Let me know how it is for you, @andreievg

Yes, very slow, try [LibreOffice_7.4.0.1_MacOS_aarch64.dmg](http://downloadarchive.documentfoundation.org/libreoffice/old/7.4.0.1/mac/aarch64/LibreOffice_7.4.0.1_MacOS_aarch64.dmg) in http://downloadarchive.documentfoundation.org/libreoffice/old/7.4.0.1/mac/aarch64/ 

It worked instantly

src/components/actions/helpers.ts
CarlosNZ: Yeah, sure: 1ef72368

src/components/actions/helpers.ts
CarlosNZ: Yup, you mean adding return type to processTrigger:
```
export async function processTrigger(payload: TriggerPayload): Promise<ActionResult[]> {
...
```

Moved "ActionResult" to a shared type and added it to processTrigger: 96294828

src/types.ts
CarlosNZ: Yup sure: 51dcdd72

This will definitely break the snapshot used for the front-end though, so you'll need to rename the parameter in there if I don't get to it first.

src/components/actions/processTrigger.ts
CarlosNZ: > Hmm, i think this should really be filtered out at the consumption end, it's still actionOutput

No, cos those actions never run, so they don't have any valid or useful output. I don't see the point in this info being returned to the front end. For example, when previewing a decision, there's three different actions, one for each of APPROVED, REJECTED and LOQ. 2 out of 3 never run as they don't meet the condition, so the front-end preview should only be interested in the one that *did* run.

> Also I've noticed the FAIL ones are coming through to displayData return

Yeah, that was intentional. The front-end displays an error message for the failed ones.

src/types.ts
CarlosNZ: Didn't break the snapshot, just had to change the parameter name in front-end. Good to go now.

src/components/actions/helpers.ts
CarlosNZ: I made an issue to consider this specific case (https://github.com/openmsupply/conforma-server/issues/795), but I don't it will cause problems currently, as we're only passing "static" parameters to the preview endpoint / alias action.

andreievg (approved):

PR #777, opened 2022-07-04
#770 Clean up preview docs

### LINKED ISSUE #770: Clean up used preview documents
Part of Epic: https://github.com/openmsupply/conforma-web-app/issues/1203

As per [agreed solution](https://github.com/openmsupply/conforma-web-app/issues/1203#issuecomment-1169449868):

> Obviously this process will generate a large amount of "garbage" in terms of preview docs that are never seen again. Suggestion is to just add an additional field "is_preview" to the file table and mark it "TRUE" for all these preview docs. (ideally we'd have some way to make this happen automatically -- perhaps the endpoint always injects the parameter?)
Then we just add a function that runs on a schedule (once per day) and deletes any documents that are marked "is_preview" (and are more than, say, 1 hour old -- we don't want to delete docs that are haven't been downloaded yet, in case the reviewer is doing previews when the cleanup function runs)

Note -- we already have a database listener to detect when file records are deleted and then automatically delete the actual file. So all we need to do for this is delete the records. One improvement that could be made to this though -- it deletes the files, but not the folder, so we end up with a pile of empty folders lying around. Would be good if the scheduled script (above) could also take care of mopping up empty folders.

CarlosNZ:
Fix #770 

Adds function and scheduler to clean up the "is_preview" docs.

Tries to read schedule from prefs, but defaults to once per day if not specified (and only deletes docs more that 2 hours old).


src/components/actions/helpers.ts
```diff
@@ -46,7 +46,12 @@ export const swapOutAliasedAction = async (templateId: number, action: ActionInT
   if (altCondition !== undefined) aliasedAction.condition = altCondition
 
   // Override parameters
-  aliasedAction.parameter_queries = merge(aliasedAction.parameter_queries, overrideParams)
+  aliasedAction.parameter_queries = merge(
+    aliasedAction.parameter_queries,
+    // All docs generated through here should be preview docs/non-output
+    { isPreviewDoc: true, isOutputDoc: false },
```
CarlosNZ, line 52: This just saves the template configurer having to remember to set these two parameters for every single preview doc.

src/components/actions/runActionRoute.ts
```diff
@@ -15,6 +15,10 @@ export const routeRunAction = async (request: any, reply: any) => {
...
 
 export const routeGetApplicationData = async (request: any, reply: any) => {
-  const { applicationId } = combineRequestParams(request, 'camel')
-  reply.send(await getApplicationData({ applicationId: Number(applicationId) }))
+  const { applicationId, reviewId } = combineRequestParams(request, 'camel')
+  const appDataParams: { applicationId: number; reviewId?: number } = {
+    applicationId: Number(applicationId),
+  }
+  if (reviewId) appDataParams.reviewId = Number(reviewId)
+  reply.send(await getApplicationData(appDataParams))
```
CarlosNZ, line 23: Not strictly part of this issue, but it just brings the "generateApplicationData" endpoint up to date with the actual function, where a "reviewId" can also be supplied. -- was useful in dev.

src/components/actions/helpers.ts
```diff
@@ -46,7 +46,12 @@ export const swapOutAliasedAction = async (templateId: number, action: ActionInT
   if (altCondition !== undefined) aliasedAction.condition = altCondition
 
   // Override parameters
-  aliasedAction.parameter_queries = merge(aliasedAction.parameter_queries, overrideParams)
+  aliasedAction.parameter_queries = merge(
+    aliasedAction.parameter_queries,
+    // All docs generated through here should be preview docs/non-output
+    { toBeDeleted: true, isOutputDoc: false },
+    overrideParams
```
andreievg, line 53: I think this should be in the preview endpoint

andreievg: `{ toBeDeleted: true, isOutputDoc: false }` this one I mean

src/components/files/deleteFiles.ts
```diff
@@ -24,6 +24,11 @@ export const deleteFile = async (file: FileDetail) => {
     if (!thumbnailPath.match(genericThumbnailsFolderName))
       await fsPromises.unlink(path.join(filesPath, thumbnailPath))
     console.log(`File deleted: ${originalFilename}`)
+
+    // Also delete folder if it's now empty
+    const dir = path.dirname(filePath)
+    if (fs.readdirSync(path.join(filesPath, dir)).length === 0)
```
andreievg, line 30: I think apart from snapshot things, we should used fsPromise and await

plugins/action_generate_document/src/generateDoc.ts
```diff
@@ -32,6 +33,7 @@ const generateDoc: ActionPluginType = async ({
       applicationSerial,
       description,
       isOutputDoc,
+      toBeDeleted,
```
andreievg, line 36: On second thought, I think ideally we should have 'expiry' on file (timestamps field), which is set on creation, vs toBeDeleted

src/components/actions/helpers.ts
andreievg: never mind, i guess its ok to have here

src/components/actions/runActionRoute.ts
andreievg, line 23: Is this only used by template builder ? (I would imagine this was added to see full application data for review via postman ?)

CarlosNZ: > Is this only used by template builder?

It's not even used by the template builder, it's purely a dev tool (and I think it should stay that way). We can make more specific endpoints (such as "Preview" and "Expire") for actual tasks we need the app to do. The "/run-action" endpoint is a very blunt instrument and shouldn't be widely available.

> (I would imagine this was added to see full application data for review via postman ?)

Yes, exactly.

src/components/files/deleteFiles.ts
CarlosNZ: Okay, I changed to this: dc696156

So this should be exactly the same functionality?
```ts
    const dir = path.dirname(filePath)
    if ((await fsPromises.readdir(path.join(filesPath, dir))).length === 0)
      await fsPromises.rmdir(path.join(filesPath, dir))
  } catch (err) {
    console.log(err)
```

andreievg: yep, thanks

andreievg (approved):

src/components/actions/runActionRoute.ts
andreievg: > It's not even used by the template builder

There was a link to it from application builder, i think it's used for application data in actions

PR #778, opened 2022-07-05
File upload -- allow description to be edited without reuploading whole file

CarlosNZ:
Slightly off track, but I think it will actually be useful for https://github.com/openmsupply/conforma-web-app/issues/1177 when we need the admin to be able to add descriptions to uploaded files.

Front-end PR: https://github.com/openmsupply/conforma-web-app/pull/1227


src/components/files/fileHandler.ts
```diff
@@ -62,6 +64,24 @@ export async function saveFiles(data: any, queryParams: HttpQueryParameters) {
         mimetype: file.mimetype,
       })
     }
+    // For updating file description without re-uploading file
+  } catch (err) {
+    // If no file provided, we'll get this 406 error, so we continue.
+    if (err?.statusCode !== 406) throw err
```
andreievg, line 70: can we add a new endpoint for this please. it looks like a good use case for it

andreievg (commented):
Hmm, I wonder how we would restrict ability to edit file description, in terms of permissions. By user_id ? That would work I think, in that case file graphql can be used, and in row level policies, we enable only update, and only certain fields, when user matches

andreievg (approved):

Issue #780, opened 2022-07-05
Update modifyRecord to take in multiple records


CarlosNZ:
Needed to modify multiple files at once in https://github.com/openmsupply/conforma-web-app/issues/1177

Should work the same way it currently does, we allow a "records" parameter which would contain an array of records to modify.

We would also need a "keyMap" parameter so we can map the "records" fields to database table fields (since we can't manipulate them individually in advance).

Within the plugin, if "records" exists, we just iterate over them and call "modifyRecord" recursively on each one.

PR #781, opened 2022-07-07
#779 #780 Help/Reference docs

### LINKED ISSUE #779: Schema changes for reference docs
Back-end requirements for https://github.com/openmsupply/conforma-web-app/issues/1177

As per [suggested solution](https://github.com/openmsupply/conforma-web-app/issues/1177#issuecomment-1174475544)

- add field "is_reference_doc" to file table

CarlosNZ:
Fix #779 #780

Back-end changes required to make this functionality work.

- add ref doc fields to "file" table (internal/external)
- add postgres function/trigger to automatically delete ref doc files that are no longer in use (as reference docs)
- change "modifyRecord" action to handle multiple records at once by iterating whole action recursively over the input array

Test with [front-end PR #1228](https://github.com/openmsupply/conforma-web-app/pull/1228) (and use the template provided there to manage reference docs)

---

**Edit**: Also updated the test suite for modifyRecord to test the new functionality. Can run with `yarn test modify` -- requires fresh `yarn database_init`


database/buildSchema/35_file.sql
```diff
@@ -36,3 +38,26 @@ CREATE TRIGGER file_deletion
...
+        to_be_deleted = TRUE
+    WHERE
+        id = NEW.id;
+    RETURN NULL;
+END;
+$file_event$
+LANGUAGE plpgsql;
+
+-- TRIGGER to execute above function when files no longer reference
+CREATE TRIGGER file_no_longer_reference
```
andreievg, line 58: Hmm, as I was reading this, I was contemplating the complexity repercussions of indirection, and was wondering if it's easier to follow if this was a seperate action, instead of using modifyRecord. We could also reduce complexity of template configuration (even though it would require double upload if document is shared between internal and external, i think general UI would be simpler to use and configurations would also be simpler)

i.e. 

## Template

Just two file upload form elements, one for internal, and another for external (with default values from graphql file request)

## Action 

updateReferenceDocuments
two parameters, for codes of form elements for internal and external file.

Action internally insert/updates or deletes relevant file entries.

plugins/action_modify_record/src/modifyRecord.test.ts
```diff
@@ -260,6 +273,169 @@ test('Test: Check creating of application join record', () => {
...
+    })
+  })
+})
+
+test('Test: add multiple records to multiple tables, using keyMap', () => {
+  return modifyRecord({
+    parameters: {
+      tableName: 'test2',
+      records: [
+        { tableName: 'test3', fullname: 'New Record', amount: 5, isCompleted: true },
```
andreievg, line 365: Is this really needed ? Being able to edit multiple tables in one action, i can see potential use case for editing multiple records in a table, but for multiple tables ? 
If it's just ability to updateMultiple records in one table, then `updateMultipleRecords` would be a bit simpler

plugins/action_modify_record/src/modifyRecord.ts
```diff
@@ -78,6 +87,70 @@ const modifyRecord: ActionPluginType = async ({ parameters, applicationData, DBC
...
+  parameters,
+  applicationData,
+  DBConnect,
+}) => {
+  let status = ActionQueueStatus.Success
+  const errors: string[] = []
+  const {
+    tableName,
+    matchField,
+    matchValue,
```
andreievg, line 102: It seems a little bit weird to update multiple records matched by the same value

andreievg (commented):
Although the logic seems sound, would be good to either:
* Remove trigger and add another action for just referenceDocs as per comment below
Or
* Add modifyMultipleRecords action vs adding more conditions to modifyRecords action (can share logic between the two)

CarlosNZ:
> Add modifyMultipleRecords action vs adding more conditions to modifyRecords action (can share logic between the two)

This seems like the better option to me. I'll change to using that for multiple records.

plugins/action_modify_record/src/modifyRecord.test.ts
CarlosNZ: Sure, there's not really an obvious use case for updating/adding to multiple tables in a single action, but is there any downside to it being possible? Like, I would have to do *more* work to not allow it.

We should probably have a test case that doesn't touch multiple tables though, so I'll add that.

plugins/action_modify_record/src/modifyRecord.ts
CarlosNZ: Yeah, again -- not really likely in practice, but don't see the need to actively prevent it either.

CarlosNZ:
Okay, new "modifyMultipleRecords" action created, and tests updated. Tests are the same as they were, with same output, but just changed to call the new "multiple" action.

e67d7aae

plugins/action_modify_multiple_records/src/modifyMultipleRecords.ts
```diff
@@ -0,0 +1,73 @@
...
+
+  // Main parameters can be provided to each individual record, but most
+  // parameters would usually be the same for all of them, in which case common
+  // parameter values can be provided at the top-level. Per-record parameters
+  // will take precedence in the case that both are provided.
+  const multipleRecords = records.map((record: { [key: string]: any }) => {
+    const newRecord = keyMap ? constructMappedRecords(record, keyMap) : { ...record }
+    newRecord.tableName = record.tableName ?? tableName
+    newRecord.matchField = record.matchField ?? matchField
+    newRecord.tableName = record.tableName ?? tableName
```
andreievg, line 32: `newRecord.tableName` is duplicated

andreievg (approved):
Thanks looks good, a duplicate in `tableName`, i have some suggestions to simplify things a bit more, but they would require too much migration of existing actions that use modifyRecord, I'll make and issue to explain what i mean

plugins/action_modify_multiple_records/src/modifyMultipleRecords.ts
CarlosNZ: Oops -- thanks. ddec90a8

PR #783, opened 2022-07-12
#782 Allow non-answered elements to be reviewable (or not)

### LINKED ISSUE #782: Back-end changes to allow staff to review un-answered elements
Part of epic https://github.com/openmsupply/conforma-web-app/issues/1199

From [suggested solution:](https://github.com/openmsupply/conforma-web-app/issues/1199#issuecomment-1174445626)

1. Add two new fields to "template_element" table:
- `is_reviewable`, with possible (enum) values of "ALWAYS", "NEVER" and `null` -- default null, which would behave the same as it does currently
- `review_required` -- boolean (default `true` -- same as current behaviour)
The reason why there is two fields rather than an additional "SOMETIMES" value in the first one is that the issue of whether a review is optional is for when an applicant has made a response. Wheras the first one (is_reviewable) is about whether review can interact with a response regardless of whether the applicant has answered it. Thinking that things like names, perhaps, could be not required for review. This second one is less important, but could be useful in some cases.

2. Changes to "trimResponses" logic: when trimming application response (trimResponses action), we also check if an application response is linked to a template_element whose `is_reviewable` value is "ALWAYS", and don't trim those that are.

CarlosNZ:
Closes #782 

Adds field "is_reviewable" to "template_element", with options ALWAYS, NEVER and `null` (default current behaviour)

Use with [Front-end PR #1235](https://github.com/openmsupply/conforma-web-app/pull/1235)


database/buildSchema/16_template_element.sql
```diff
@@ -52,6 +57,8 @@ CREATE TABLE public.template_element (
     validation_message varchar,
     help_text varchar,
     parameters jsonb,
+    is_reviewable public.is_reviewable_status DEFAULT NULL,
+    -- review_required boolean NOT NULL DEFAULT TRUE,
```
CarlosNZ, line 61: I really wanted to add this field to make certain elements *optionally* reviewable, but the impact on the front-end logic was too much to risk right now, as I'm pretty sure it'd introduce unforeseen regressions, so have just left it as a "TO-DO" for now.

database/migration/migrateData.ts
```diff
@@ -431,6 +431,19 @@ const migrateData = async () => {
...
+    console.log(' - Adding reviewability fields to template_element')
+
+    await DB.changeSchema(`
+    CREATE TYPE public.is_reviewable_status AS ENUM (
+      'ALWAYS',
+      'NEVER' );`)
+
+    await DB.changeSchema(`
+      ALTER TABLE template_element
+        ADD COLUMN IF NOT EXISTS is_reviewable public.is_reviewable_status DEFAULT NULL;
```
CarlosNZ, line 444: Thought it best to leave default as NULL as we don't really want to change the functionality of existing templates. For Angola we can just to a bulk database replace if we want their standard to be "ALWAYS".

plugins/action_trim_responses/src/trimResponses.ts
```diff
@@ -52,9 +52,12 @@ const trimResponses: ActionPluginType = async ({ parameters, applicationData, DB
       const previousResponse =
         responseArray.length > 1 ? responseArray[responseArray.length - 2] : null
       if (
-        isEqual(latestResponse.value, previousResponse?.value) ||
-        latestResponse.value === null || // Application Responses
-        latestResponse.value?.decision === null // Review responses
+        (previousResponse !== null && isEqual(latestResponse.value, previousResponse?.value)) ||
+        // Application Responses
+        (latestResponse.value === null &&
+          latestResponse.is_reviewable !== IsReviewableStatus.Always) ||
```
CarlosNZ, line 58: This is the only real bit of additional logic -- if the element is marked as Always, then we don't trim it.

CarlosNZ:
NOTE -- Have discovered an additional bug here -- doesn't quite work for re-reviews after changes requested. Need to investigate further...

andreievg:
I made an example of [test without database](https://github.com/openmsupply/conforma-server/compare/782-schema-changes-for-unanswered-elements...782-example), I've noticed that responses to update might be erroneous (but that was before this PR), i am using `jest` extension which allows running just the selected test

CarlosNZ:
> I made an example of [test without database](https://github.com/openmsupply/conforma-server/compare/782-schema-changes-for-unanswered-elements...782-example)

Cool -- can you merge that branch into here? (though rename the file to just `trimResponses.test.ts`)

> i am using `jest` extension which allows running just the selected test

You can run isolated tests just by specifying in command -- it uses string matching. So can just to `yarn test trim` to run this test for example.

andreievg:
> You can run isolated tests just by specifying in command -- it uses string matching. So can just to yarn test trim to run this test for example.

Yep. Jest extension also allows a quick debug on a specific test.

andreievg (approved):

Issue #787, opened 2022-07-13
Add "unexpire" event to timeline logic (DB function)


CarlosNZ:
Part of Epic https://github.com/openmsupply/conforma-web-app/issues/1225

See [agreed solution](https://github.com/openmsupply/conforma-web-app/issues/1225#issuecomment-1178466161)

CarlosNZ:
Might be closed by #788 but need to wait till front-end done to be sure.

PR #788, opened 2022-07-13
#784 Database function for outcome change

### LINKED ISSUE #784: DB Function to handle changing Outcome back to PENDING

- When outcome changes back to PENDING:
  - "is_active" (on application table) set to True
  - new "application_status_history" which sets status to the previous value (before COMPLETED)

CarlosNZ:
Fix #784 

~~I think this is probably also enough for #787 but won't know for sure until I do the front-end timeline adjustments.~~ Update: Not quite -- #792 

To test:
 - Manually change an application "outcome" field from PENDING to one of the "done" ones (Approved, Rejected, Withdrawn, Expired). Note the "is_active" field gets changed to "false" and the application_status_history has a new "COMPLETED" entry -- this is already existing behaviour
 - Change the "outcome" back to PENDING, then note the reversal of the above. In particular, the status will be reverted (i.e. a new history record) to whatever it was before it was COMPLETED.


database/buildSchema/18_application.sql
```diff
@@ -55,3 +55,49 @@ CREATE TRIGGER outcome_trigger
...
+                WHERE
+                    application_id = NEW.id
+                    AND is_current = TRUE),
+                (
+                    SELECT
+                        status
+                    FROM
+                        application_status_history
+                    WHERE
+                        time_created = (
```
CarlosNZ, line 84: Assuming that the "previous" record will be the one with the most recent timestamp that is currently `is_current = false`

database/buildSchema/18_application.sql
```diff
@@ -55,3 +55,49 @@ CREATE TRIGGER outcome_trigger
...
+                            SELECT
+                                MAX(time_created)
+                            FROM
+                                application_status_history
+                            WHERE
+                                is_current = FALSE
+                                AND application_id = NEW.id)));
```
andreievg, line 91: doesn't look like there is application_id on `application_status_history`. Probably need to use `application_stage_status_all`

CarlosNZ: Yeah, there is -- it's an auto-computed column -- see `CREATE OR REPLACE FUNCTION public.application_status_history_application_id` in `21_application_status_history.sql`

Screen Shot 2022-08-02 at 11 29 40 AM

andreievg: oops, thanks

andreievg (approved):
Thanks

PR #790, opened 2022-07-14
#785 extend-application endpoint

### LINKED ISSUE #785: Create extend-application endpoint (and ON EXTEND trigger)

CarlosNZ:
Fix #785 

Implements /extend-application endpoint.

To test, please use snapshot `conforma-templates/dev/snapshots/appDeadlineTest_2022_07_16`

When you apply as "jane_smith", then review as "basic_reviewer", a deadline is set when review is submitted (it's only 10 mins). When the applicant re-submits, the deadline is cancelled.

You can extend the deadline (either before or after it has expired) by using the endpoint with postman:

`http://localhost:8080/api/extend-application`

with Body JSON something like:
```
{
    "applicationId": 258,
    "eventCode": "appDeadline",
    "extensionTime": {
        "years": 5
    }
}
```

The application will be set back to "PENDING" and be editable by the applicant again.


database/buildSchema/05_trigger_queue.sql
```diff
@@ -34,7 +35,7 @@ CREATE TABLE public.trigger_queue (
     record_id int,
     event_code varchar,
     data jsonb,
-    timestamp timestamptz,
+    timestamp timestamptz DEFAULT CURRENT_TIMESTAMP,
```
CarlosNZ, line 38: This was an oversight -- should have had a default all along

src/components/actions/executeAction.ts
```diff
@@ -25,10 +25,10 @@ export async function executeAction(
   payload: ActionPayload,
   actionLibrary: ActionLibrary,
   additionalObjects: any = {},
-  previewData?: Partial<ActionApplicationData>
+  applicationDataExtra?: Partial<ActionApplicationData>
```
CarlosNZ, line 28: Not really related to this PR (was going to be initially), but makes sense to give this field a more generic name since it could be used by more than just the "preview" endpoint. (Have kept the name "previewData" *within* the "preview" endpoint, it just passes on the more generic name to the processTrigger function.

src/components/actions/extendApplicationRoute.ts
```diff
@@ -0,0 +1,63 @@
...
+import { DateTime } from 'luxon'
+import DBConnect from '../databaseConnect'
+import { getPermissionNamesFromJWT } from '../data_display/helpers'
+
+export const routeExtendApplication = async (request: any, reply: any) => {
+  const { applicationId, eventCode, extensionTime, data } = combineRequestParams(request, 'camel')
+
+  // Check permissions first -- currently we just check user has ANY Reviewing
+  // permissions for current template.
+  // TO-DO: define specific permission names required for extending deadlines
```
CarlosNZ, line 12: I would quite like to add this, as I think we'd want more specificity over who can extend. Will see how much time we have in the Angola budget when we get further along.

src/components/actions/extendApplicationRoute.ts
```diff
@@ -0,0 +1,63 @@
...
+  const { permissionNames: userPermissions } = await getPermissionNamesFromJWT(request)
+
+  const templatePermissions = (
+    await DBConnect.getTemplatePermissionsFromApplication(Number(applicationId))
```
CarlosNZ, line 16: Coercing `applicationId` to Number here, as it could come in on query parameters (no just body JSON), in which case it is always a string. The node-postgres module doesn't care, but GraphQL very much does.

CarlosNZ, line 16: Coercing `applicationId` to Number here, as it could come in on query parameters (no just body JSON), in which case it is always a string. The node-postgres module doesn't care, but Postgraphile very much does.

src/components/actions/extendApplicationRoute.ts
```diff
@@ -0,0 +1,63 @@
...
+    const extensionResult = await DBConnect.updateScheduledEventTime(
+      applicationId,
+      eventCode,
+      scheduledTime
+    )
+
+    if (extensionResult.length === 0)
+      return reply.send({ success: false, message: 'No matching event found' })
+
+    // Add trigger event directly to trigger_queue so we can include eventCode
```
CarlosNZ, line 49: Could have gone one of three ways here:
- "simulate" a trigger by sending directly to "processTrigger" the same as the "preview" endpoint does
- set a real trigger on the "application" table and have it act as normal
- set a trigger record directly, and then the database listeners proceed as normal

Went with the last option, as it was the easiest way to get the eventCode (and other possible data) passed through while keeping action behaviour as "standard" as possible

src/components/actions/extendApplicationRoute.ts
```diff
@@ -0,0 +1,67 @@
...
+      (permission: any) =>
+        permission?.permissionName?.permissionPolicy?.type === PermissionPolicyType.Review
+    )
+    .map((permission: any) => permission?.permissionName?.name)
+
+  const hasPermission = userPermissions.some((permission) =>
+    templatePermissions.includes(permission)
+  )
+
+  if (!hasPermission) return reply.send({ success: false, message: 'Unauthorized' })
```
andreievg, line 28: I think it might be better to use `permissions_all` view ? We probably want to also filter by org

CarlosNZ: This does ultimately fetch from `permissions_all` and includes "org" restrictions.

`hasPermissions` checks `userPermissions`, which comes from `getPermissionNamesFromJWT()`, which calls database function `getUserOrgPermissionNames`, which looks up the `permissions_all` table :)

```
SELECT "permissionNameId" as id,
      "permissionName" FROM permissions_all
      WHERE ${userMatch}
      AND ${orgMatch}`
```

src/components/actions/extendApplicationRoute.ts
andreievg: hmm i noticed that in front end we do `reload`, which usually waits for application trigger to be reset. When we add trigger directly, do we set trigger on application ? (is it possible that ui won't update with correct 'new' application deadline if processing triggers take a bit longer ?)

CarlosNZ: If I remember correctly, the front-end doesn't call "reload" until after it receives a valid response from this endpoint. And this endpoint doesn't send response until everything is completed.

src/components/actions/extendApplicationRoute.ts
andreievg: ok, great, thanks for clarifying, I might try and reduce/consolidate this in row permission work, I just saw that permissions_all also has template_id, and thought it would be more direct to use it

src/components/actions/extendApplicationRoute.ts
```diff
@@ -0,0 +1,67 @@
...
+
+export const routeExtendApplication = async (request: any, reply: any) => {
+  const { applicationId, eventCode, extensionTime, data } = combineRequestParams(request, 'camel')
+
+  // Check permissions first -- currently we just check user has ANY Reviewing
+  // permissions for current template.
+  // TO-DO: define specific permission names required for extending deadlines
+  const { permissionNames: userPermissions } = await getPermissionNamesFromJWT(request)
+
+  const templatePermissions = (
```
andreievg, line 15: technically permissionNames

andreievg:
@CarlosNZ, related to previous PR, do you thing he trigger `WHEN (NEW.outcome = 'PENDING' AND OLD.outcome <> 'PENDING')` might fire during application creation ? Maybe good idea to make default value = 'PENDING'

CarlosNZ:
> @CarlosNZ, related to previous PR, do you thing he trigger `WHEN (NEW.outcome = 'PENDING' AND OLD.outcome <> 'PENDING')` might fire during application creation ? Maybe good idea to make default value = 'PENDING'

Okay, have set outcome to default = 'PENDING' : 9a2109db

Front-end already has a outcome fallback value = "PENDING" for the create application mutation, so we've got double cover now.

src/components/actions/extendApplicationRoute.ts
CarlosNZ: It's the permissions per template though, not just the names. So it's strictly "template_permission_names", but since we use "templatePermissions" in the front-end and in the JWT, this seems consistent I thought.

src/components/actions/extendApplicationRoute.ts
andreievg: ok, so small chance that it will still show expired, maybe it's good to make an issue to take note of that. I would assume we can add another trigger_queue_status 'completed' and await for that (or error) before sending response to api call

andreievg:
> Front-end already has a outcome fallback value = "PENDING" for the create application mutation, so we've got double cover now.

Cools, I would remove the 'pending' mutation when doing row level policies btw (or ability for anyone to edit outcome on applications

andreievg:
This looks good, took a little bit of time to figure it out, I made this diagram to help me think of use cases:
AFDC1A7C-F8C1-4F73-ABBA-2D7376399445

andreievg (approved):

src/components/actions/extendApplicationRoute.ts
CarlosNZ: > I would assume we can add another trigger_queue_status 'completed' and await for that (or error) before sending response to api call

Actually, I guess it wouldn't be that hard to just add a bit extra here to:
- wait for "trigger_queue" event status to be "ACTIONS_DISPATCHED"
- wait for the actions added to the "action_queue" that match this trigger_queue id to become no longer "PROCESSING"

I'll do that now, shouldn't be too tricky.

CarlosNZ: Okay @andreievg -- here you go:
089dc9d6

Made a generic "waitForDatabaseValue" method which can be re-used for any other "waiting" events we need.

CarlosNZ:
@andreievg -- going to merge this now so I start getting my PR chain a bit tidier.

But if you want to have a look at the latest commit to add "waiting for trigger", here it is: 089dc9d

Can address any problems with it in a separate issue, or one of the other PRs.

PR #791, opened 2022-07-15
#786 Add applicant_deadline column to application_list

### LINKED ISSUE #786: Add computed "deadline" column to application list view

Needs to look up "trigger_schedule" table and find any "active" items that have eventCode "appDeadline"

CarlosNZ:
Fix #786 

Just change to the application_list function to show "applicant_deadline" column, plus migration code to migrate existing schema.


database/buildSchema/41_application_list_view.sql
```diff
@@ -62,6 +65,9 @@ CREATE FUNCTION application_list (userid int DEFAULT 0)
     LEFT JOIN assignment_list (stage_status.stage_id) ON app.id = assignment_list.application_id
     LEFT JOIN review_list (stage_status.stage_id, $1) ON app.id = review_list.application_id
     LEFT JOIN assigner_list (stage_status.stage_id, $1) ON app.id = assigner_list.application_id
+    LEFT JOIN trigger_schedule ts ON app.id = ts.application_id
```
andreievg, line 68: If we misconfigure actions and there are multiple rows with that event_code, application list would duplicates, just something to keep in mind (all the other joins are meant to return just one row)

andreievg (approved):
A little bit concerned about the future performance of this (not really related to your PR), would be awesome to benchmark it sometime, just food for thought.

Looks good thanks (apart from the relocation of application_deadline to last element in select (as per your discovery)

database/buildSchema/41_application_list_view.sql
CarlosNZ: No, I have a joint unique constraint on the trigger_schedule table for "application_id" and "event_code", so that can't happen ;)

```
CREATE UNIQUE INDEX unique_application_event ON trigger_schedule (application_id, event_code);
```

CarlosNZ:
> A little bit concerned about the future performance of this (not really related to your PR), would be awesome to benchmark it sometime, just food for thought.

Yeah, I was wondering that too. Adding some indexes might help, but yeah, need to keep an eye on it.

PR #792, opened 2022-07-18
#787 Update activity_log functions to record deadline extensions

### LINKED ISSUE #787: Add "unexpire" event to timeline logic (DB function)

CarlosNZ:
Fix #787

Updates database functions/trigger so that deadline extension events get logged. Required for front-end Timeline.

Also added a "editor_user_id" field to "trigger_schedule" which is filled in when running the "extend-application" endpoint. This is so we can log the user that made the extension (which is captured in the activity log and displayed in the front-end Timeline)


src/components/actions/evaluatorFunctions.ts
```diff
@@ -5,7 +5,10 @@ import { DateTime, Duration } from 'luxon'
 
 const generateExpiry = (duration: Duration) => DateTime.now().plus(duration).toJSDate()
 
-const getYear = () => new Date().getFullYear()
+// getYear() => "2022"
+// getYear("short") => "22"
+const getYear = (type?: 'short'): string =>
+  type === 'short' ? String(new Date().getFullYear()).slice(2) : String(new Date().getFullYear())
```
CarlosNZ, line 11: Off-track, but needed for one of the demo configs

andreievg: ohh that's where it's added, thanks

database/buildSchema/45_activity_log.sql
```diff
@@ -125,6 +126,31 @@ CREATE TRIGGER outcome_update_activity_trigger
...
+END;
+$application_event$
+LANGUAGE plpgsql;
+
+DROP TRIGGER IF EXISTS deadline_extension_activity_trigger ON public.application;
+
+CREATE TRIGGER deadline_extension_activity_trigger
+    AFTER UPDATE ON public.trigger_schedule
+    FOR EACH ROW
+    WHEN (NEW.time_scheduled > OLD.time_scheduled AND NEW.event_code = 'applicantDeadline')
```
andreievg, line 151: would this fire when time_scheduled is updated by subsequent LOQ ?

database/migration/migrateData.ts
```diff
@@ -496,11 +496,48 @@ const migrateData = async () => {
...
+
     console.log(' - Adding applicant_deadline to application_list')
 
+    // We need to drop and re-create the application_list_shape table so the
+    // column order gets preserved, otherwise the subsequent application_list
+    // function will fail
     await DB.changeSchema(`
-      ALTER TABLE application_list_shape
-      ADD COLUMN IF NOT EXISTS applicant_deadline timestamptz;
+      DROP TABLE IF EXISTS application_list_shape CASCADE;
```
andreievg, line 512: probably should be in the next statement. Also would this be allowed, since application list uses it ? Should we drop application list first ?

andreievg (commented):
I would manually add activity_log in `extendApplicationRoute`. Personally would only really use triggers for things that bypass back end (i.e. postgraphile mutations)

database/buildSchema/45_activity_log.sql
CarlosNZ: Okay, yeah good point. I guess if I add the activity_log addition to the extendApplicationRoute like you suggested below, that'll take care of this problem too.

database/migration/migrateData.ts
CarlosNZ: The `CASCADE` forces the application_list() function to also be dropped. And then they both get recreated in the next bit.

database/buildSchema/45_activity_log.sql
CarlosNZ: Okay, this seems to work to handle that case: d184f5c3

(Using the "editor_user_id" column more effectively)

andreievg (approved):
Thanks! Looks good

Issue #800, opened 2022-07-28
Serial/Sequence numbers. not updating for dynamic tables when loading snapshot


andreievg:
We have [this script](https://github.com/openmsupply/conforma-server/blob/build-2022-07-27/database/reset_serial.sql) that runs after snapshot was loaded, and it worked quite nicely until we've updated to this [version pg-diff-cli](https://github.com/openmsupply/conforma-server/blob/84ee2286a5c03a859af347736412985bcd2fd5df/package.json#L93)

In the latest version of pg-diff-cli the following, instead of:

```
CREATE TABLE IF NOT EXISTS "public"."organisation_application_join" (
	"id" serial NOT NULL
```

it does

```
CREATE TABLE IF NOT EXISTS "public"."organisation_application_join" (
	"id" int4 NOT NULL DEFAULT nextval('organisation_application_join_id_seq'::regclass) 
```

The difference between the two is minimum, but the script that's used to reset sequences does not capture the second, because there is some dependency (column ownership of the sequence) that are not created.

This is replicated by loading snapshot (with organisation_appication_joins) on `build-2022-07-27` branch, and trying to insert new `organisation_application_join` which will result in error thrown on unique constraint.

Proposed solution is to create a new script for updating serials

PR #801, opened 2022-07-28
800-Update-serial-reset-sql

andreievg:
closes: #800 

Description as per comments

### DIFF
--- a/database/reset_serial.sql
+++ b/database/reset_serial.sql
@@ -1,19 +1,57 @@
--- https://wiki.postgresql.org/wiki/Fixing_Sequences
-SELECT
-    'SELECT SETVAL(' || quote_literal(quote_ident(PGT.schemaname) || '.' || quote_ident(S.relname)) || ', COALESCE(MAX(' || quote_ident(C.attname) || '), 1) ) FROM ' || quote_ident(PGT.schemaname) || '.' || quote_ident(T.relname) || ';'
-FROM
-    pg_class AS S,
-    pg_depend AS D,
-    pg_class AS T,
-    pg_attribute AS C,
-    pg_tables AS PGT
-WHERE
-    S.relkind = 'S'
-    AND S.oid = D.objid
-    AND D.refobjid = T.oid
-    AND D.refobjid = C.attrelid
-    AND D.refobjsubid = C.attnum
-    AND T.relname = PGT.tablename
-ORDER BY
-    S.relname;
+-- When a field is created as `SERIAL` postgres adds a seqeunce for it, and adds default values for that field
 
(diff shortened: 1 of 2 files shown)


andreievg:
This PR also updates version of pg-diff-cli to 2.4, which is in latest build branch, we wanted `COMMENT ON` statements to persist for postgraphile.

CarlosNZ:
Hi @andreievg, thanks for this, but I can't get it to work.

I downloaded latest snapshot from 50000 instance, which has been having problems with the `organisation_application_join` table and the `data_table_product` table, amongst others.

I copied and pasted and ran this script as per your PR:
```
SELECT 'SELECT SETVAL(' || quote_literal(sequence_number.relname) || ', COALESCE(MAX(' || quote_ident(table_and_column.column_name) || '), 1)) FROM ' || quote_ident(table_and_column.table_name) || ';'
FROM information_schema.columns AS table_and_column
JOIN pg_class AS sequence_number ON table_and_column.column_default LIKE 'nextval(' || quote_literal(sequence_number.relname) || '::regclass)'
WHERE sequence_number.relkind = 'S'
ORDER BY table_and_column.table_name
```

And it appear to run okay. But when I try and manually add records to those tables (without specifying the id), I still get the same error, e.g.
```
ERROR:  duplicate key value violates unique constraint "organisation_application_join_pkey"
DETAIL:  Key (id)=(3) already exists.
```

andreievg:
@CarlosNZ did you run the script that this script generates ? (i.e. the script is meant to generate SQL script, [this .sh](https://github.com/openmsupply/conforma-server/blob/develop/database/update_serials.sh) saves generated script in temp file and runs it, useSnapshot runs that `.sh`)

CarlosNZ (approved):
Okay, I tested as part of a snapshot restore, and it seems to work nicely. Thanks for sorting that.

One thing I don't quite understand -- when I run that script manually in SQL, all it does is *generate* a bunch of statements such as `SELECT SETVAL('data_table_product_id_seq', COALESCE(MAX(id), 1)) FROM data_table_product;`, but it doesn't actually *run* them, so the serial reset hasn't occurred. I can run each statement manually and it works though.

So how does this script get fully executed when called as part of the snapshot script?

CarlosNZ:
> @CarlosNZ did you run the script that this script generates ? (i.e. the script is meant to generate SQL script, [this .sh](https://github.com/openmsupply/conforma-server/blob/develop/database/update_serials.sh) saves generated script in temp file and runs it, useSnapshot runs that `.sh`)

Ha ha timing -- looks like you answered my question at the exact time I just asked it. Nice :)

PR #804, opened 2022-08-05
Fix core templates and generated graphql

andreievg:
Had problems with running on dev:
* `yarn migrated` or `dev` didn't run (had type issue which i fixed manually in generate)
* after `yarn database_init`, failed to start server with the same error

What I did
* manually fixed generated graphql ts file
* ran yarn migrate
* start server with yarn dev
* run yarn database_init 
* again fixed generated graphql ts file
* took current snapshot
* copied current snapshot to core_templates
(now yarn database_init works correctly)

It turns out that package.json wasn't on 0.3.1:
* update version in package.json
* snapshot take and move to database folder
* re-tested


CarlosNZ (approved):
Thanks. Merge it in!

PR #808, opened 2022-08-09
#806 Fix template export bloat

### LINKED ISSUE #806: Bug: template export is saving a bunch of unnecessary files with it
The output of the template export seems to be exporting all files that are in the root of the files folder.

It should only export files that are linked to the template in question via the `template_id` field in the `file` table.

The problem is that if you export a template from one snapshot into another, it brings a bunch of "dead" (i.e. unlinked) files with it that will never be removed from the files base folder.

Also -- check that they get imported correctly after the fix.

CarlosNZ:
Fix #806 

Now when exporting templates, only the linked files are exported, not the "baseFiles" in the root files folder.


nmadruga (approved):
LGTM

PR #810, opened 2022-08-10
include new caseSensitive variable to prevent caseissues with passwor…

fergie-nz:
…d reset


src/components/postgresConnect.ts
```diff
@@ -656,8 +656,15 @@ class PostgresDB {
...
-    const text = `SELECT COUNT(*) FROM "${table}" WHERE LOWER(${field}) = LOWER($1)`
+  public isUnique = async (
+    table: string,
+    field: string,
+    value: string,
+    caseSensitive: boolean
+  ): Promise<boolean> => {
+    const text = caseSensitive
+      ? `SELECT COUNT(*) FROM "${table}" WHERE(${field}) =($1)`
+      : `SELECT COUNT(*) FROM "${table}" WHERE LOWER(${field}) = LOWER($1)`
```
CarlosNZ, line 667:
```suggestion
    const text = caseSensitive
      ? `SELECT COUNT(*) FROM "${table}" WHERE LOWER(${field}) = LOWER($1)`
      : `SELECT COUNT(*) FROM "${table}" WHERE ${field} = $1`
```
Your one's round the wrong way -- we want "caseSensitive" (i.e. the default) to use the existing query -- only when "caseSensitve = false" should it use the non-casesensitive query.

You also don't need the `( )` round the values in the new one -- they were just for the `LOWER()` function.

CarlosNZ (changes requested):
Sweet as, looking good. Just that one little note to fix.

src/components/postgresConnect.ts
fergie-nz: I see it now,

The double negatives were playing tricks on my brain.

fergie-nz: I'm thinking now it should be the other way.

Typical calling for unique is not caseSensitive. So false should be standard. This particular call is true caseSensitive. So I want to change both files to the reverse.

What do you think?

CarlosNZ: Yeah, I see what you mean. Default should be NOT case-sensitive. Cool, as long as we don't have to change existing implementations other than the reset-password form.

Other option is to just rename the parameter to "caseInsensitive" and leave the default as `true`. Up to you.

CarlosNZ:
Also, one other thing while I remember. We generally follow the convention of naming our branches off the issue number we're addressing. So this one would be something like `810-fix-case-sensitive-unique-check`

And then in the PR, start the subject line with something similar, e.g. **#810 Fix case-sensitive Unique check**\

Check out the other PRs for comparison.

Cheers.

fergie-nz:
> Also, one other thing while I remember. We generally follow the convention of naming our branches off the issue number we're addressing. So this one would be something like `810-fix-case-sensitive-unique-check`
> 
> And then in the PR, start the subject line with something similar, e.g. **`#810 Fix case-sensitive Unique check`**
> 
> Check out the other PRs for comparison.
> 
> Cheers.

That's great, thanks Carl

src/components/permissions/routes.ts
```diff
@@ -211,7 +211,7 @@ const routeGetPrefs = async (request: any, reply: any) => {
 
 // Unique name/email/organisation/other check
 const routecheckUnique = async (request: any, reply: any) => {
-  const { type, value, table, field } = request.query
+  const { type, value, table, field, caseInSensitive } = request.query
```
CarlosNZ, line 214: That's great, just one super-nitpicky thing: I think it should be `caseInsensitive` since "insensitive" is a single unbroken work in standard English.

fergie-nz: Of course, that's great. Great attention to detail.

database/core_templates/snapshot.json
```diff
@@ -7503,7 +7534,90 @@
...
+    "status": "DRAFT",
+    "prevStatus": null
+   },
+   "id": 7,
+   "recordId": 18,
+   "table": "application_status_history",
+   "timestamp": "2022-08-11T10:15:21.717953+12:00",
+   "type": "STATUS",
+   "value": "DRAFT"
+  }
```
CarlosNZ, line 7619: I think we can delete this stuff. Just put it back to an `"activityLog": []` like it was before. (Just manually edit this file)

CarlosNZ (approved):
That's great, thanks a lot. 

If you could just make that tiny tweak to the snapshot file, then merge this in.

Cheers.

database/core_templates/snapshot.json
fergie-nz: Nice - I've removed those to '"activityLog": [ ]'.

I see maybe that it isn't viewable in github because it thinks of it as a massive delete + massive addition

CarlosNZ: It's because the auto-formatter changed the whole file to match Prettier formatting. 

There is an option so "Save without formatting" that is quite useful in these cases.

fergie-nz: I see, good as gold, I'll be aware of that in future

PR #815, opened 2022-08-12
1270F Admin differentiation

CarlosNZ:
Back-end changes required for front end-issue https://github.com/openmsupply/conforma-web-app/issues/1270

Front-end PR: https://github.com/openmsupply/conforma-web-app/pull/1283

- Adds a new UI location -- "MANAGEMENT" -- appears as new Menu bar item
- Schema migration to add the above
- user-info/login endpoints also return an array of permission names (needed to determine if user has "management" permissions, but will probably come in handy elsewhere)

### DIFF
--- a/database/buildSchema/06_template_category.sql
+++ b/database/buildSchema/06_template_category.sql
@@ -5,7 +5,8 @@ CREATE TYPE public.ui_location AS ENUM (
     'DASHBOARD',
     'LIST',
     'USER',
-    'ADMIN'
+    'ADMIN',
+    'MANAGEMENT'
 );
 
 CREATE TABLE public.template_category (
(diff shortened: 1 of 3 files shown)


nmadruga (approved):
Just a little confused over `permissionNames` is that just checking (in front-end) for permissions name that has the same strings as what was called that permission granted to this user? Not sure if this could be exploited.

CarlosNZ:
> Just a little confused over `permissionNames` is that just checking (in front-end) for permissions name that has the same strings as what was called that permission granted to this user? Not sure if this could be exploited.

Basically returns the same information (a simplified version) as the "user-permissions" endpoint, but included it in the "user-info" return so we can populate user state without making multiple requests to multiple endpoints.

PR #818, opened 2022-08-24
#1303F - Part1 & Part2 - Return reviewer list action when changes required

nmadruga:
Front-end PR: https://github.com/openmsupply/conforma-web-app/pull/1311

Testing: Please use the latest snapshot in our shared folder: https://drive.google.com/drive/folders/1pAhRoqqVNrWggEYEtlZ8T5ae2cWU_MDP 

Please, remember to run `yarn migrate` on the snapshot you are working with to get the changes applied. 

Testing:
- Login as `jess` and see there is already an application for _Cadastramento de entidade_ to review
- Self assign => Check that in the applications list the action showing "Start review"
- Start the review, add one 'Non conform' then submit as _Request further information_ (LOQ) => In the list the action now is "Awaiting response" 

Basically changes database for:
- Fixes front-end as requested by Jess to allow Reviewer to see "action" when the application has been submitted to Applicant for LOQ. 
- Add migration code for version 0.4.4 (to be release) - @CarlosNZ I wasn't sure if that's how this is done - for the next release
- Update core_templates with new Pt strings & after running the migration code


database/buildSchema/40_reviewer_action_list.sql
```diff
@@ -37,6 +38,11 @@ CREATE FUNCTION review_list (stageid int, reviewerid int)
             AND (review = NULL
             OR is_locked = FALSE)) != 0 THEN
             'SELF_ASSIGN'
+        WHEN COUNT(*) FILTER (WHERE (appstatus = 'CHANGES_REQUIRED'
+            OR appstatus = 'DRAFT')
+            AND review_assignment.status = 'ASSIGNED'
+            AND review_status_history.status = 'SUBMITTED') != 0 THEN
+            'AWAITING_RESPONSE'
```
CarlosNZ, line 45: Your editor had messed up the whitespace formatting in the sql files. Have pushed a fix so I can see what's actually changed: 1dbb13ff

Can you check your settings so we keep this consistent?

database/core_templates/localisation/pt_an/strings.json
```diff
@@ -1,6 +1,7 @@
 {
-  "_APP_NAME": "conforma",
+  "_APP_NAME": "Sirema",
```
CarlosNZ, line 2: FYI -- this string is never actually shown in the app.

database/core_templates/schema_diff.sql
```diff
@@ -0,0 +1,592 @@
...
+
+
+CREATE SEQUENCE IF NOT EXISTS "public"."user_application_join_id_seq" 
+	INCREMENT BY 1 
+	MINVALUE 1
+	MAXVALUE 2147483647
+	START WITH 1
+	CACHE 1
+	NO CYCLE;
+
```
CarlosNZ, line 16: The core_templates should not have a schema_diff.sql, as the snapshot should have the exact same schema as the default sql init files. If you can't get rid of it and have it still work, let me know and I'll take a look at it -- I dealt with this problem a couple of weeks ago.

database/core_templates/schema_init.sql
```diff
@@ -1784,60 +1784,66 @@ CREATE TYPE public.reviewer_action AS ENUM (
...
-            'CONTINUE_REVIEW'
-        WHEN COUNT(*) FILTER (WHERE review_assignment.status = 'ASSIGNED'
-            AND review_assignment.is_final_decision = TRUE
-            AND review_assignment.is_last_stage = TRUE
-            AND review = NULL) != 0 THEN
-            'MAKE_DECISION'
-        WHEN COUNT(*) FILTER (WHERE review_assignment.status = 'ASSIGNED'
-            AND review.id IS NULL) != 0 THEN
-            'START_REVIEW'
-        WHEN COUNT(*) FILTER (WHERE review_assignment.status = 'AVAILABLE'
```
CarlosNZ, line 1813: Unnecessary whitespace changes. I've fixed it in the main sql files, so this should be fixed when you re-export core_templates.

CarlosNZ (commented):
Looks good. I've just made a few quick comments (and fixed the whitespace formatting) -- will test it properly in the morning. 

Can you provide your snapshot that shows the different reviewer actions, please?

database/buildSchema/40_reviewer_action_list.sql
nmadruga: Ah, I don't think it was my editor. I did this change because the spacing looked odd to me. But yes, you would have to set the option "Remove space changes" to see the diff using Github

database/core_templates/schema_diff.sql
nmadruga: I tried removing this and then regenarating the snapshot (exporting) it always have that. I think the init data isn't set correctly and that's what is causing it.

database/buildSchema/05_trigger_queue.sql
```diff
@@ -11,7 +11,7 @@ CREATE TYPE public.trigger AS ENUM (
     'ON_REVIEW_ASSIGN',
     'ON_REVIEW_UNASSIGN',
     'ON_REVIEW_REASSIGN',
-    'ON_REVIEW_SELF_ASSIGN',
+    'ON_REVIEW_SELF_ASSIGN', -- Not used (not removed for consitency with previous snapshots)
```
CarlosNZ, line 14: You can safely remove this, as long as you write migration code to deprecate the value. Old snapshots will load with the database script they were saved with, so will have this value included. Then they'll run migration to remove it. (Migration code will need to also *change* any existing to "ON_REVIEW_ASSIGN" otherwise it won't let you remove the value)

database/core_templates/localisation/languages.json
```diff
@@ -20,4 +20,4 @@
     "flag": "🇦🇺",
     "enabled": false
```
CarlosNZ, line 21: Shouldn't be `false` on core_templates

database/migration/migrateData.ts
```diff
@@ -643,6 +643,136 @@ const migrateData = async () => {
...
+            assigners,
+            reviewers,
+            reviewer_action,
+            assigner_action,
+            -- CASE WHEN is_fully_assigned_level_1 IS NULL THEN
+            --     FALSE
+            -- ELSE
+            --     is_fully_assigned_level_1
+            -- END,
+            -- assigned_questions_level_1,
```
CarlosNZ, line 754:
```suggestion
```
I'd remove the commented code from migration. Can leave it the original script.

database/migration/migrateData.ts
```diff
@@ -643,6 +643,136 @@ const migrateData = async () => {
...
+        LEFT JOIN assigner_list (stage_status.stage_id, $1) ON app.id = assigner_list.application_id
+        LEFT JOIN trigger_schedule ts ON app.id = ts.application_id
+            AND ts.is_active = TRUE
+            AND ts.event_code = 'applicantDeadline'
+    WHERE
+        app.is_config = FALSE
+  $$
+  LANGUAGE sql
+  STABLE;`)
+  }
```
CarlosNZ, line 775: I'm just going to trust that these functions are 100% identical to the ones in the sql script. ;)

plugins/action_update_review_assignments_status/src/updateReviewAssignmentsStatus.ts
```diff
@@ -33,14 +33,13 @@ async function updateReviewAssignmentsStatus({
         isSelfAssignable
       )
 
-      return otherSelfAssignments.map(({ id }: ReviewAssignment) => ({
+      return otherSelfAssignments.map(({ id, assignedSections }: ReviewAssignment) => ({
         id,
-        isLocked,
+        isLocked: isLocked || assignedSections.length > 0,
```
CarlosNZ, line 38: Nice 👍🏼

CarlosNZ (commented):
This looks great. I'm happy to approve as far as the main functionality goes, I'd just like to get the issues with the core_templates and the migration code to remove the enum values sorted, so I'll do that and push to this branch.

database/migration/migrateData.ts
```diff
@@ -643,6 +643,229 @@ const migrateData = async () => {
...
+          USING TRIGGER::text::public.trigger;
+      
+      CREATE TRIGGER verification_trigger
+          AFTER INSERT OR UPDATE OF trigger ON public.verification
+          FOR EACH ROW
+          WHEN (NEW.trigger IS NOT NULL AND NEW.trigger <> 'PROCESSING' AND NEW.trigger <> 'ERROR')
+          EXECUTE FUNCTION public.add_event_to_trigger_queue ();
+      
+      DROP TYPE public.trigger_old;    
+    `)
```
CarlosNZ, line 750: A lot of code just to remove a couple of enum values! 🙄

src/components/snapshots/takeSnapshot.ts
```diff
@@ -55,7 +55,7 @@ const takeSnapshot: SnapshotOperation = async ({
     // Write snapshot and config to folder
     await fs.writeFile(
       path.join(newSnapshotFolder, `${SNAPSHOT_FILE_NAME}.json`),
-      JSON.stringify(snapshotObject, null, ' ')
+      JSON.stringify(snapshotObject, null, 2)
```
CarlosNZ, line 58: This is my attempt to make the json format saved by the snapshot exporter the same as our editor auto-formatter. It's not exactly the same, but it's closer, I think.

CarlosNZ (approved):
Okay, I've added that migration code, and updated the core_templates to be closer to the original (the changes in snapshot.json are most whitespace, as the last version of it had manual edits/auto-formatting).

Once you're happy with it, please merge :)

database/migration/migrateData.ts
nmadruga: Wow 😮

nmadruga: The only comment I have is on `trigger_action` the ones for ON_REVIEW_SELF_ASSIGN should be removed instead of replaced. But I guess it won't be a big problem to have the ON_REVIEW_ASSIGN duplicated, as it will when replacing the action

nmadruga:
> Okay, I've added that migration code, and updated the core_templates to be closer to the original (the changes in snapshot.json are most whitespace, as the last version of it had manual edits/auto-formatting).
> 
> Once you're happy with it, please merge :)

Thanks Carl! That is awesome. I'm going to have a look now and merge if it's seems all accordingly to what I was imagining too

database/buildSchema/16_template_element.sql
```diff
@@ -6,11 +6,7 @@ CREATE TYPE public.template_element_category AS ENUM (
 
 CREATE TYPE public.is_reviewable_status AS ENUM (
     'ALWAYS',
-    'NEVER',
-    'OPTIONAL_IF_NO_RESPONSE'
-    -- TO-DO:
-    -- 'OPTIONAL_IF_RESPONSE',
-    -- 'OPTIONAL' (the above two combined)
+    'NEVER'
```
nmadruga, line 9: I think this was overwritten by mistake!

nmadruga (commented):
Also checking the generated graphql.ts I can see there are removals for `review_list` which shouldn't be happening. I'm going to check that, it might be just a miss to re-generate after the migration code runs

PR #823, opened 2022-08-27
#822 Exclude prefs from repo

### LINKED ISSUE #822: Find a way to not include preferences in repo
It's far too easy to overwrite, and it's tricky keeping track of which is the correct one when switching branches.

Suggest something like:
- exclude "preferences" folder from repo
- when starting server, if "preferences" file doesn't exist, copy it from core_templates

CarlosNZ:
Fix #822 

This should prevent the problem we've had a few times when someone accidentally overwrote the snapshot preferences with the default one.

This completely excludes the preferences folder from the repo (like localisations, files, snapshots), so it won't keep changing when switching branches. It's slightly different to those other folders though, because we can't just restore the folder on server startup, as the server needs the preferences file before it can do anything (I think this must be why we hadn't already done this). 

So I just made a little script that checks for preferences and copies them from core_templates if they don't exist. This runs before server.ts (in nodemon) when in dev mode, and before compilation in the build script (build_all.sh)

Have also tested it with a Docker build and seems all good.


nmadruga (approved):
I haven't tested on Docker build... (I'll have a go on it now)

But tested in dev mode looks good : )

PR #824, opened 2022-08-29
Fix problem to show application in DRAFT (not yet submitted) in list to Reviewer

### LINKED ISSUE #677: Update permission policy - assignBasic
Problem noticed by Jess while testing. Some users were able to view application in DRAFT before then have been submitted for the first time. Investigating it was found that the problem was in `assignBasic` permission policy. 

Users that were both reviewer and assigner were being able to view more applications. The rowLevel permissions set for ReviewBasic and ReviewAdvanced, to only list applications that the user is a possible Reviewer (in other words already have a `reviewAssignment` and therefor has been submitted)

nmadruga:
Fixes #677

Just changing to permissionPolicy in **core_templates** - check change in Issue comment

Basically the problem was found (kudos to @andreievg) that the reviewer which was being able to view applications not submitted - had also the permission to assign... Then we checked that although the permissions to list application set for `reviewBasic` permissionPolicy was correct, the `assignBasic` was too open for applications. And the change done made it also restrict applications to be visible only when they have been submitted (once). It should still show after LOQ is sent.


CarlosNZ (approved):
LGTM! Thanks for figuring this out.

PR #826, opened 2022-08-30
Evaluator functions -- couple of date functions

CarlosNZ:
Couple of date functions which make it easier to get the current time. "modifyRecord" should be able to parse both JSDate and ISO string, but JSDate would be safer, so use that when possible.


nmadruga (approved):
Looks good. Is this required for any latest update to the Snapshot?

CarlosNZ:
> Looks good. Is this required for any latest update to the Snapshot?

I haven't added it to snapshot yet, but it will be next time I update it.

PR #828, opened 2022-08-31
#827 Fix for stray data in localisation (and prefs) folder

### LINKED ISSUE #827: Fix the nested "localisation" folder that keeps appearing in the snapshot
I'm sure I've fixed this several times, but it seems to keep happening, but only periodically. Hmmmm.

CarlosNZ:
Fix #827 

Finally worked out what's going on here! Every time I tried to replicate it, I couldn't -- and yet the annoying, nested "localisations" folder kept appearing inside the snapshot folder... *sometimes*.

Screen Shot 2022-08-31 at 3 51 52 PM

Turns out, we weren't completely clearing out existing content in the "localisation" and "preferences" in our snapshot script.  And so one of you... 👀 must have had some crap left in your localisation folder from waaay back which was never getting deleted. And so every time _that particular person_ saved a snapshot, this crap came along with it. The reason why I couldn't reproduce it is that my localisations folder was clean...but someone else's wasn't. 🤨

Now that I understand what's going on -- it's a simple fix: just clear out the localisation and preferences folders when we load the snapshot.

We'll still need to clear the crap out of the current snapshot, but at least that'll be the last time (hopefully!!!)

### DIFF
--- a/src/components/snapshots/useSnapshot.ts
+++ b/src/components/snapshots/useSnapshot.ts
@@ -27,6 +27,7 @@ import {
   PREFERENCES_FILE,
   SCHEMA_FILE_NAME,
   INFO_FILE_NAME,
+  PREFERENCES_FOLDER,
 } from '../../constants'
 
 const useSnapshot: SnapshotOperation = async ({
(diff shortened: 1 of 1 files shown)


nmadruga (approved):
Going to merge as well now

PR #831, opened 2022-09-01
#820 improvement review responses part of latest review submission

### LINKED ISSUE #820: Improve how to figure out review_responses latest decisions (to use in LOQ Carbone template)

nmadruga:
Fixes #820 

Requires Front-end PR: [1331](https://github.com/openmsupply/conforma-web-app/pull/1331)

- Adds new field to review_response called `is_latest_review_submission` that should be only set to TRUE once the review is submitted and for the reviewed elements in THIS review. So, it shall be listing ALL review_responses on latest review.
- Creates new Trigger and Function to update other review_responses for same `template_element_id` and `review_id` after new review_response gets created (when the user clicks to Re-review application)

## Testing
Snapshot to use: `dev/snapshots/re-review-showing-correct-LOQ`

After loading it please remember to run `yarn migrate`

Just needs to log in as user `jess` and "Reavaliar" the Review for "[Registo de Medicamento - Medicamento 3](http://localhost:3000/application/MED-PV-20220053/review)" then, you can see the difference of elements you get before re-starting the review and after re-submitting it (when using GraphQL (http://localhost:5000/graphiql):
```
query getDeclineResponses {
  review(id: 9) {
    reviewResponses(filter: { isLatestReviewSubmission: { equalTo: true } }) {
      nodes {
        id
        isLatestReviewSubmission
        decision
        comment
        reviewId
        applicationResponse {
          value
          id
          templateElement {
            parameters
            code
            id
          }
        }
      }
    }
  }
}
````

### DIFF
--- a/database/buildSchema/32_review_response.sql
+++ b/database/buildSchema/32_review_response.sql
@@ -29,10 +29,35 @@ CREATE TABLE public.review_response (
     time_updated timestamptz DEFAULT CURRENT_TIMESTAMP,
     time_submitted timestamptz,
     is_visible_to_applicant boolean DEFAULT FALSE,
+    is_latest_review_submission boolean DEFAULT FALSE,
     template_element_id integer REFERENCES public.template_element ON DELETE CASCADE,
     recommended_applicant_visibility public.review_response_recommended_applicant_visibility DEFAULT 'ORIGINAL_RESPONSE_NOT_VISIBLE_TO_APPLICANT',
     status public.review_response_status DEFAULT 'DRAFT'
 );
+
+-- Function to automatically set previous review_responses 
+-- (for same review & templateElement) as is_latest_review_submitted = false
+CREATE OR REPLACE FUNCTION public.set_previous_review_response ()
+    RETURNS TRIGGER
+    AS $review_response_event$
+BEGIN
+    UPDATE
+        public.review_response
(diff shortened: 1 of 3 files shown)


CarlosNZ (commented):
I don't think you need to get the front-end involved at all. The same trigger can set `is_latest_review_submission = TRUE` on the record being updated, and update all the other ones to FALSE like you're already doing.

The other way would be to make it a computed/generated column that returns TRUE if the review_response is the MAX(time_submitted) out of the review_responses that share the same template_id (and FALSE otherwise).

Either way, there's no need for the front-end to be involved. Considering the purpose of this extra field is for back-end actions, the front-end should have no business even knowing about such a field.

CarlosNZ:
Here you go, I tested it out on a new branch. This will now update *all* the relevant `is_latest_review_submission` values ON SUBMISSION of the review. No need for any front-end involvement:

https://github.com/openmsupply/conforma-server/compare/%23820-improvement-review-responses-part-of-latest-review-submission...820-review-response-tweaks

CarlosNZ (changes requested):
See my comments that I made already.

nmadruga:
Looks great, Carl. Thank you so much! I'll be removing the Front-end PR and once you approve I can merge this one in : )

PR #833, opened 2022-09-01
#830 Return snapshot name to front-end

### LINKED ISSUE #830: Capture "snapshot name" and show in Admin UI
Small improvement that would be nice if someone has a bit of time:

When loading a snapshot, at the end, we should write the name of the snapshot file into the database (use the "system_info" table).

Then in the front-end, for Admin only, show the latest snapshot name in the footer.

CarlosNZ:
Fix #830 

Use front-end: https://github.com/openmsupply/conforma-web-app/pull/1332

This will be pretty handy for reminding ourselves which snapshot we have loaded on various instances. When loading a snapshot, we save the snapshot name in the "system_info" table, then return this info with prefs. In the front-end it can be displayed in the Footer (to Admin only):

Screen Shot 2022-09-02 at 12 24 43 AM

### DIFF
--- a/src/components/databaseConnect.ts
+++ b/src/components/databaseConnect.ts
@@ -156,6 +156,8 @@ class DBConnect {
 
   public getApplicationSections = PostgresDB.getApplicationSections
 
+  public getLatestSnapshotName = PostgresDB.getLatestSnapshotName
+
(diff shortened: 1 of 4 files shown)


nmadruga (approved):
👍

PR #834, opened 2022-09-02
#832 modifyRecord null fix

### LINKED ISSUE #832: Something to investigate -- modifyRecord saving "null" values
Not a high priority, but just something I noticed. The "modifyRecord" action is not supposed to create fields when the incoming value is `null`. However, I noticed that a new product on an empty "product" table had a couple of fields created with `null` values. Not sure how this could happen. Perhaps they were getting "undefined", and the rule only applied to explicitly `null`?

CarlosNZ:
Fix #832 

Simple fix -- see issue for explanation of the problem.

Also updated tests

### DIFF
--- a/plugins/action_modify_record/src/modifyRecord.test.ts
+++ b/plugins/action_modify_record/src/modifyRecord.test.ts
@@ -52,22 +52,24 @@ test('Test: add User to database', () => {
     parameters: { tableName: 'user', ...testUser },
     ...extraParameters,
   }).then((result: any) => {
-    expect(result).toEqual({
-      status: ActionQueueStatus.Success,
-      error_log: '',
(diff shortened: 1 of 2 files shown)


nmadruga (approved):
Looks good. Was this causing any crash or just invalid fields on the table?

CarlosNZ:
> Was this causing any crash or just invalid fields on the table?

Nah, I just noticed some new fields got created in the "product" table with "NULL" values when I approved a new product -- which shouldn't be allowed to happen (modifyRecord ignores nulls). 

At some point we'll probably need to add a `acceptNull` parameter to `modifyRecord`, if we actually want to set a value to `null` -- but we can handle that when we need it.

PR #837, opened 2022-09-06
#1214F warning on reviewer is unassigned

nmadruga:
Front-end PR: https://github.com/openmsupply/conforma-web-app/pull/1338

Please see description of PR in Front-end

### DIFF
--- a/database/buildSchema/45_activity_log.sql
+++ b/database/buildSchema/45_activity_log.sql
@@ -296,14 +296,16 @@ BEGIN
                                 SELECT
                                     title, code, "index"
                                 FROM template_section
-                            WHERE
-                                code = ANY (ARRAY (
-                                        SELECT
-                                            assigned_sections
-                                        FROM review_assignment
-                                    WHERE
-                                        id = assignment_id))
-                                AND template_id = templ_id ORDER BY "index") t), 'level', level_num, 'isLastLevel', is_last_level, 'finalDecision', is_final_decision));
+                                WHERE
+                                    code = ANY (ARRAY (
+                                            SELECT
+                                                assigned_sections
+                                            FROM review_assignment
+                                            WHERE
+                                                id = assignment_id
+                                                and template_id = templ_id
+                                                and assigned_sections <> '{}' ))
+                                    AND template_id = 10 ORDER BY "index") t), 'level', level_num, 'isLastLevel', is_last_level, 'finalDecision', is_final_decision));
     RETURN NEW;
 END;
 $application_event$
@@ -363,8 +365,9 @@ BEGIN
                                         SELECT
                                             assigned_sections
                                         FROM review_assignment
-                                    WHERE
-                                        id = rev_assignment_id))
+                                        WHERE
+                                            id = rev_assignment_id
+                                            and assigned_sections <> '{}' ))
                                 AND template_id = templ_id ORDER BY "index") t)));
     RETURN NULL;
 END;
--- a/database/buildSchema/45_activity_log.sql
+++ b/database/buildSchema/45_activity_log.sql
@@ -296,16 +296,16 @@ BEGIN
                                 SELECT
                                     title, code, "index"
                                 FROM template_section
-                                WHERE
-                                    code = ANY (ARRAY (
-                                            SELECT
-                                                assigned_sections
-                                            FROM review_assignment
-                                            WHERE
-                                                id = assignment_id
-                                                and template_id = templ_id
-                                                and assigned_sections <> '{}' ))
-                                    AND template_id = 10 ORDER BY "index") t), 'level', level_num, 'isLastLevel', is_last_level, 'finalDecision', is_final_decision));
+                            WHERE
+                                code = ANY (ARRAY (
+                                        SELECT
+                                            assigned_sections
+                                        FROM review_assignment
+                                    WHERE
+                                        id = assignment_id
+                                        AND template_id = templ_id
+                                        AND assigned_sections <> '{}'))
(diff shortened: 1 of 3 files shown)


database/buildSchema/45_activity_log.sql
```diff
@@ -296,14 +296,16 @@ BEGIN
...
+                                WHERE
+                                    code = ANY (ARRAY (
+                                            SELECT
+                                                assigned_sections
+                                            FROM review_assignment
+                                            WHERE
+                                                id = assignment_id
+                                                and template_id = templ_id
+                                                and assigned_sections <> '{}' ))
+                                    AND template_id = 10 ORDER BY "index") t), 'level', level_num, 'isLastLevel', is_last_level, 'finalDecision', is_final_decision));
```
CarlosNZ, line 308: I don't think you want to hard-code `template_id = 10` do you?

plugins/action_change_status/src/changeStatus.ts
```diff
@@ -109,11 +109,23 @@ const changeReviewStatus = async (
...
+      returnObject.status = ActionQueueStatus.Success
+      returnObject.error_log = 'Status not changed'
+      returnObject.output = {
+        status: currentStatus.status,
+        statusId: currentStatus.status_history_id,
+        reviewStatusHistoryTimestamp: currentStatus.status_history_time_created,
+      }
+      return returnObject
+    }
+
```
CarlosNZ, line 128: What was the reason for this? When does this action get run with `newStatus === undefined`?

CarlosNZ (changes requested):
Cool, thanks for figuring this out. It looks good, I'll test it properly when I load the front-end. Just that little bit of weirdness with the hard-coded template id to take a look at.

database/buildSchema/45_activity_log.sql
nmadruga: Oops - Nice finding! I changed on the other one but forgot about this

plugins/action_change_status/src/changeStatus.ts
nmadruga: When you don't set the property `newStatus`, I wrongly set it as `status` and got it failing without understanding why!

CarlosNZ: Okay, fair enough.

In that case, this should actually be an error:
```ts
      returnObject.status = ActionQueueStatus.Fail
      returnObject.error_log = 'Missing property: "newStatus"'
     // No returnObject.output
```

We don't want missing properties to silently not work, they should very clearly show an error and stop the action sequence.

database/buildSchema/45_activity_log.sql
```diff
@@ -364,7 +366,8 @@ BEGIN
                                             assigned_sections
                                         FROM review_assignment
                                     WHERE
-                                        id = rev_assignment_id))
+                                        id = rev_assignment_id
+                                        AND assigned_sections <> '{}'))
```
CarlosNZ, line 370: So much clearer and easier to see the diff! (And now I see it's actually a very small change). 

Thanks.

One thing -- you'll need to re-copy and paste this into the migration script.

nmadruga: NP. I'll do this now

nmadruga:
> Okay, fair enough.
> 
> In that case, this should actually be an error:
> 
> ```ts
>       returnObject.status = ActionQueueStatus.Fail
>       returnObject.error_log = 'Missing property: "newStatus"'
>      // No returnObject.output
> ```
> 
> We don't want missing properties to silently not work, they should very clearly show an error and stop the action sequence.

I'll add this in as well

plugins/action_change_status/src/changeStatus.ts
```diff
@@ -42,7 +50,7 @@ const changeApplicationStatus = async (
     status: ActionQueueStatus.Fail,
     error_log: '',
   }
-  console.log(`Changing the Status of Application ${applicationId}...`)
+  console.log(`Changing the Status of Application ${applicationId} to ${newStatus}...`)
```
CarlosNZ, line 53: Can delete this now that you've added it to the top

plugins/action_change_status/src/changeStatus.ts
```diff
@@ -109,7 +117,7 @@ const changeReviewStatus = async (
     status: ActionQueueStatus.Fail,
     error_log: 'uknown error',
   }
-  console.log(`Changing the Status of Review ${reviewId}...`)
+  console.log(`Changing the Status of Review ${reviewId} to ${newStatus}...`)
```
CarlosNZ, line 120: Same -- can delete :)

CarlosNZ (approved):
Great thanks. Just a couple of tiny comments. Merge when you're ready 👍🏼

plugins/action_change_status/src/changeStatus.ts
```diff
@@ -15,6 +15,14 @@ async function changeStatus({
       : parameters?.isReview || applicationData?.action_payload?.trigger_payload?.table === 'review'
   const newStatus = parameters?.newStatus
 
+  console.log(`Changing status of ${isReview ? 'Review' : 'Application'} to ${newStatus}`)
+
+  if (!newStatus)
+    return {
+      status: ActionQueueStatus.Fail,
+      error_log: 'Missing property: "newStatus"',
+    }
```
CarlosNZ, line 24: I actually want to implement a more comprehensive 'required parameter' checker for all the actions. The required parameters are defined in the "plugin.json", so we just need a generic method to check each one right at the top of each action.

But this is a good start for now.

PR #838, opened 2022-09-06
Dev/action trigger testing tool

CarlosNZ:
This is very much a work in progress, but I plan to use it to test and configure actions, so you guys might want to as well. But expect errors, as it's very crude and definitely not intended to be used for anything other than development/testing.

The basic idea is a single endpoint `/api/admin/test-trigger` where we can quickly run any trigger and see the results. It uses the "config" application associated with each template, so shouldn't have to look up any ids or do too much database poking to do basic stuff. You'll need to progress the application in order (i.e. don't try running "ON REVIEW SUBMIT" before any reviews exist), but you can get through all the triggers pretty quickly.

I'll continue to document this as I improve it, and when it's "ready" I'll make a proper developer docs page.

The JSON body syntax is basically:
```
{
    "templateCode": <code>,
    "trigger": <triggerName / alias>,
    <options...>
}
```

As much as possible, it automatically chooses an appropriate review/assignment etc, but you can specify them if you want.

Similarly, it uses the "config" application by default, but you can supply *either* an `applicationId` or `serial` parameter to test with a specific application.

I've allowed some aliases for trigger names to save a little time in typing out the full trigger enum. For example:
```
{
    "templateCode": "P1-ProdRegistrationLong",
    "trigger": "start", // alias for "ON_APPLICATION_CREATE"
}
```

Check out the comments and code in `routes.ts` for more info, and the `helpers.ts` file for the full alias list. (Aliases are case-insensitive BTW)

Example (going through "company registration"):

```
{
    "templateCode": "U2-OrgRegistration",
    "trigger": "create"
}

// returns action output, updates "config" application

{
    "templateCode": "U2-OrgRegistration",
    "trigger": "submit"
}

// Submits application, generates review assignments

{
    "templateCode": "U2-OrgRegistration",
    "trigger": "assign"
}

// Assigns a random reviewer to review, triggers their review assignment record

{
    "templateCode": "U2-OrgRegistration",
    "trigger": "reviewCreate"
}

// Creates review (and decision) for assigned reviewer, runs trigger for that new review record

{
    "templateCode": "U2-OrgRegistration",
    "trigger": "preview"
}

// Preview review decision "CONFORM" (decision not explicitly required  as "CONFORM" is default)

{
    "templateCode": "U2-OrgRegistration",
    "trigger": "reviewSubmit",
     "decision": "LIST_OF_QUESTIONS"
}

// Submits review with "LIST_OF_QUESTIONS" decision, runs trigger accordingly

etc...
```

There's also a "RESET" option, which deletes everything associated with the application except the application (and responses) itself, so you can start again with `ON_APPLICATION_CREATE`:
```
{
    "templateCode": "U2-OrgRegistration",
    "trigger": "reset"
}
```

You mostly need to go through an application in order (as in, how it would flow in a real situation), as some triggers won't work without the appropriate content available (e.g. reviews). But you should be able to re-run most triggers time after time without resetting much.

You can keep an eye on the current stage/status/level etc. by checking the appropriate record in "application_stage_status_latest" and therefore simulate any path through an application.

### Other parameters

You can override `applicationData` at any time by providing an `applicationDataOverride` parameter, the same as the Alias action / preview endpoint takes. This means you can simulate most application states without having to go through all the stages/levels. For example, you can set the "Outcome" to "APPROVED" when testing "reviewSubmit" and it will run the modifyRecord and generateDoc actions as though the application had been approved. (You can do something similar with the "decision" parameter, but that will still have the current stage/review level in applicationData, so this will fast-track that -- or you can override those values directly).

Any applicationData property (even deep ones -- it will be merged in smoothly) can be overridden, but as a shorthand the following parameters can be provided directly (they'll be inserted into applicationDataOverride for you):

- stageNumber
- status
- outcome

---

As I said, very much a work in progress, please let me know how it's working and give suggestions. The idea is to make this a lot faster (and safer) to configure and test actions, so hopefully it will achieve that.

--- 

I also discovered a few pre-existing mistakes and shortcomings in the code, so I've fixed them too. See inline comments for details.


database/buildSchema/35_file.sql
```diff
@@ -5,7 +5,7 @@ CREATE TABLE public.file (
     original_filename varchar NOT NULL,
     user_id integer REFERENCES public.user (id) ON DELETE CASCADE,
     template_id integer REFERENCES public.template (id) ON DELETE CASCADE,
-    application_serial varchar REFERENCES public.application (serial) ON DELETE CASCADE,
+    application_serial varchar REFERENCES public.application (serial) ON DELETE CASCADE ON UPDATE CASCADE,
```
CarlosNZ, line 8: We should have done this a long time ago. Without it, if you change the application serial and there's a file that points to it (via serial), it'll fail with an error. Now it'll update along with it :)

CarlosNZ, line 8: We should have done this a long time ago. Without it, if you change the application serial and there's a file that points to it (via serial), it'll fail with an error. Now it'll update along with it :)

database/migration/migrateData.ts
```diff
@@ -1096,26 +1096,34 @@ const migrateData = async () => {
...
+                  WHERE
+                      id = review_assignment.application_id) = 'PENDING'
+          GROUP BY
+              review_assignment.application_id,
+              review_assignment.level_number;
+
+          $$
+          LANGUAGE sql
+          STABLE;
+    `)
```
CarlosNZ, line 1119: Just tidying up the indentation a little here to be consistent with the rest of it.

database/migration/migrateData.ts
```diff
@@ -1096,26 +1096,34 @@ const migrateData = async () => {
...
+
+    console.log(
+      ' - Change foreign key constraint on file table to allow changing of application serial'
+    )
+    await DB.changeSchema(`
+      ALTER TABLE file DROP CONSTRAINT IF EXISTS file_application_serial_fkey; 
+      ALTER TABLE file ADD CONSTRAINT file_application_serial_fkey FOREIGN KEY (application_serial) REFERENCES application (serial) ON UPDATE CASCADE;
```
CarlosNZ, line 1126: Migration for the above schema change

plugins/action_change_status/src/changeStatus.ts
```diff
@@ -64,11 +64,11 @@ const changeApplicationStatus = async (
         `WARNING: Application ${applicationId} already has status: ${newStatus}. No changes were made.`
       )
       returnObject.status = ActionQueueStatus.Success
-      returnObject.error_log = 'Status not changed'
+      returnObject.error_log = 'Application Status not changed'
```
CarlosNZ, line 67: More helpful to specify Application/Review

plugins/action_change_status/src/changeStatus.ts
```diff
@@ -64,11 +64,11 @@ const changeApplicationStatus = async (
...
       returnObject.output = {
         status: newStatus,
         statusId: current.status_history_id,
-        applicationStatusHistoryTimestamp: current.statu_history_time_created,
+        applicationStatusHistoryTimestamp: current.status_history_time_created,
```
CarlosNZ, line 71: Epic typo! Can't believe this one slipped by for so long. 😆

plugins/action_change_status/src/changeStatus.ts
```diff
@@ -115,24 +115,26 @@ const changeReviewStatus = async (
 ): Promise<ActionPluginOutput> => {
   const returnObject: ActionPluginOutput = {
     status: ActionQueueStatus.Fail,
-    error_log: 'uknown error',
+    error_log: '',
```
CarlosNZ, line 118: Not sure why this was ever in here for default value 🤷

plugins/action_change_status/src/changeStatus.ts
```diff
@@ -115,24 +115,26 @@ const changeReviewStatus = async (
...
         `WARNING: Review ${reviewId} already has status: ${newStatus}. No changes were made.`
       )
       returnObject.status = ActionQueueStatus.Success
-      returnObject.error_log = 'Status not changed'
+      returnObject.error_log = 'Review Status not changed'
       returnObject.output = {
         status: newStatus,
         statusId: currentStatus.status_history_id,
-        reviewStatusHistoryTimestamp: currentStatus.status_history_time_created,
+        reviewStatusHistoryTimestamp: currentStatus.time_created,
```
CarlosNZ, line 137: This was obviously copied from the application version without checking the database script -- the field name is different.

src/components/actions/getApplicationDataRoute.ts
```diff
@@ -0,0 +1,11 @@
...
+import { combineRequestParams } from '../utilityFunctions'
+
+export const routeGetApplicationData = async (request: any, reply: any) => {
+  const { applicationId, reviewId } = combineRequestParams(request, 'camel')
+  const appDataParams: { applicationId: number; reviewId?: number } = {
+    applicationId: Number(applicationId),
+  }
+  if (reviewId) appDataParams.reviewId = Number(reviewId)
+  reply.send(await getApplicationData(appDataParams))
+}
```
CarlosNZ, line 11: I just separated the routes used by the app proper (i.e. `routeGetApplicationData`) from those that are dev tools only (`runAction`, this one).

I really don't want someone to accidentally start using one of the dev routes in the actual app.

src/components/actions/processTrigger.ts
```diff
@@ -111,6 +111,12 @@ export async function processTrigger(payload: TriggerPayload): Promise<ActionRes
         actionFailed = action.action_code
       }
     } catch (err) {
+      actionOutputs.push({
+        action: action.action_code,
+        status: ActionQueueStatus.Fail,
+        output: null,
+        errorLog: err.message,
+      })
```
CarlosNZ, line 119: This was also missing, and is basically the opposite of what I fixed the other day. It was only returning information for failed actions if they returned a "FAIL" status, not for ones that throw an error.

Now it behaves the same for both.

src/components/postgresConnect.ts
```diff
@@ -123,10 +123,14 @@ class PostgresDB {
...
   public getCounter = async (counterName: string, increment = true) => {
     const textSelect = 'SELECT value FROM counter WHERE name = $1;'
+    const textNewCounter = 'INSERT INTO counter (name, value) VALUES($1, 1)'
     const textUpdate = 'UPDATE counter SET value = value + 1 WHERE name = $1;'
     try {
       await this.query({ text: 'BEGIN TRANSACTION;' })
       const result: any = await this.query({ text: textSelect, values: [counterName] })
+      if (result.rows.length === 0) {
+        await this.query({ text: textNewCounter, values: [counterName] })
+      }
```
CarlosNZ, line 133: I was going to use the counter for creating a trigger event id, but it didn't end up working for database constraint reasons. 

However, this seems like a sensible improvement and probably should have been there all along, so I've left it in -- if a counter doesn't exist yet, it'll get created the first time it's queried. (So when creating an action that uses a new counter, we don't have to manually create the database record for it to work)

CarlosNZ:
Marked this no longer "DRAFT", as I've been using it quite a bit and I've ironed out a lot of things. 

No rush to get this merged though, we can still keep using it for testing without it being in the main branch (it's not needed for the build) and improve it further before finalising it.

CarlosNZ:
We can probably use this as a basis for a test suite. Since we seem to cause quite a few accidental regressions and we don't notice them until the next time someone tries to submit or review or whatever an application.

I suggest creating a small set of applications and a jest test script to put them through a whole process using this endpoint. If all the events return the action log they should, then good. But it'd be really easy to catch this kind of unexpected regression. We could even make it part of our build script, or make it a pre-merge requirement or something. Let's discuss.

src/components/actions/dev_tools/databaseMethods.ts
```diff
@@ -0,0 +1,248 @@
...
+    try {
+      const result = await DBConnect.query({ text, values: [assignmentId] })
+      return result.rows[0]
+    } catch (err) {
+      console.log(err.message)
+      throw err
+    }
+  },
+  // Gets all the review assignments for the specified application at the
+  // highest possible stage and review_level
```
nmadruga, line 85: Just wondering if it would be possible to also do partial reviews using this tool. For example, could this check also for Review to be created in the current stage and level if there are un-assigned questions?

nmadruga (commented):
Overall comment - I couldn't see the StartReview function - so it seems this wouldn't create anything related to verifying if questions as are reviewed and that sort of logic. Which is something we perhaps can add later on, but I just wanted to confirm: if we want to also include review_responses should we do this as part of which action?

src/components/actions/dev_tools/routes.ts
```diff
@@ -0,0 +1,264 @@
...
+      finalApplicationData = await getApplicationData({ applicationId, reviewId: revId })
+      break
+
+    case 'ON_REVIEW_ASSIGN':
+      // If no assignment id, select one at random (using latest stage and level) and assign all possible sections to it
+      {
+        const assignment = assignmentId
+          ? await db.getSingleReviewAssignment(assignmentId)
+          : await selectRandomReviewAssignment(applicationId, sectionCodes)
+        if (!assignment.allowedSections) assignment.allowedSections = sectionCodes
```
nmadruga, line 175: This seems as it's already done inside ` selectRandomReviewAssignment`

src/components/actions/dev_tools/routes.ts
```diff
@@ -0,0 +1,264 @@
...
+    applicationDataOverride = {},
+  }: RequestProps = params
+
+  const triggerFull = mapTriggerShortcut(trigger)
+
+  const { configId, configSerial, templateId, sectionCodes } = await db.getConfigApplicationInfo(
+    templateCode
+  )
+
+  if (!configId) reply.send('Invalid template code, or no config application available')
```
nmadruga, line 77: Didn't understand: this would run even if the user defined a different applicationId? Is this to check if the template and config application is defined to avoid other further problems?

src/components/actions/dev_tools/routes.ts
```diff
@@ -0,0 +1,264 @@
...
+    case 'ON_EXTEND':
+      triggerPayload.trigger = Trigger.OnExtend
+      actionsOutput = await processTrigger(triggerPayload)
+      finalApplicationData = await getApplicationData({ applicationId })
+      break
+
+    case 'ON_SCHEDULE':
+      if (!eventCode) return reply.send('eventCode required')
+      const { id, data } = await db.getScheduledEvent(applicationId, eventCode)
+      console.log(id)
```
nmadruga, line 141: To be removed

CarlosNZ:
> I couldn't see the StartReview function - so it seems this wouldn't create anything related to verifying if questions as are reviewed and that sort of logic.

Yes, its' there -- `reviewCreate`, or `ON_REVIEW_CREATE`

> but I just wanted to confirm: if we want to also include review_responses should we do this as part of which action?

It doesn't create review_responses as it doesn't need to. All the triggers work with just application, review, review_assignments.

Just to be clear -- the point of this is just to test the actions that run on specific events, so responses aren't really part of that consideration at all. Except if you want to simulate a certain decision -- in which case you just override applicationData in your body JSON.

src/components/actions/dev_tools/routes.ts
```diff
@@ -0,0 +1,264 @@
...
+      {
+        const assignment = assignmentId
+          ? await db.getSingleReviewAssignment(assignmentId)
+          : await selectRandomReviewAssignment(applicationId, sectionCodes, true)
+        // Create a review record and review_decision record
+        const { id: revId } = await db.createReview(assignment.id)
+        await db.createReviewDecision(revId)
+        // Create one review response
+
+        // TO-DO: Make some review responses and give them dummy values?
```
nmadruga, line 205: Oh here is the response to my question earlier on

src/components/actions/dev_tools/databaseMethods.ts
CarlosNZ: > For example, could this check also for Review to be created in the current stage and level if there are un-assigned questions?

This tool doesn't check anything -- it just simulates events. It's up to you to configure the state of things if you want to test specific scenarios. Which you can do by either creating the applications for real (the slow but probably more reliable way), or simulating it by overriding applicationData in your endpoint request here (fast, but possibly not 100% realistic)

src/components/actions/dev_tools/routes.ts
```diff
@@ -0,0 +1,264 @@
...
+
+        actionsOutput = await processTrigger(triggerPayload)
+        finalApplicationData = await getApplicationData({ applicationId, reviewId: revId })
+      }
+      break
+
+    case 'RESET':
+      await db.resetApplication(applicationId)
+      finalApplicationData = await getApplicationData({ applicationId })
+      break
```
nmadruga, line 246: No clear what this one does : (

src/components/actions/dev_tools/databaseMethods.ts
```diff
@@ -0,0 +1,248 @@
...
+    }
+  },
+  getAppIdFromSerial: async (serial: string) => {
+    const text = `
+      SELECT id FROM application
+      WHERE serial = $1;
+    `
+    try {
+      const result = await DBConnect.query({ text, values: [serial] })
+      console.log('result', result.rows)
```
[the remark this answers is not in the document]
CarlosNZ: 👍 1ff2a684 (for both)

nmadruga (commented):
Ok - just comments so far. Sorry it took me so long to review this, but it will be good to have it included in `develop` so we can start improving our testing of pre-release builds indeed.

I'll have a go on testing this using the latest snapshot we have to also check it's working ok - I'm not so sure just yet how to use this as an automated testing tool. Maybe good to chat after lunch to come up with a plan?

src/components/actions/dev_tools/routes.ts
CarlosNZ: The "allowed sections" field can be NULL, in which case we treat it as "allow all". So this just does that here -- puts all the section codes in if null.

src/components/actions/dev_tools/routes.ts
CarlosNZ: Yes, you can run it for any application you like. By default it'll use the config application (cos that's easier and it always exists), but you can test it for a specific application in a specific state if you want. :)

src/components/actions/dev_tools/routes.ts
CarlosNZ: Yeah, I wasn't sure if it would be needed. I don't think it is. I haven't yet come across a scenario for trigger testing where the presence of real review_responses turned out to actually make a difference. Please let me know if you find one.

src/components/actions/dev_tools/routes.ts
CarlosNZ: Screen Shot 2022-10-14 at 12 48 51 PM

So you can reset and start again. Otherwise you'd only be able to go through it once per application/type. 
(And also serves as something of a garbage collector, since using this tool will leave a bunch of crap in the database)

database/buildSchema/35_file.sql
nmadruga: Cool! I noticed something similar for when we change the Template.code and it doesn't reflect on template_element and other places that use the combination of Template.code and Template.version as reference. It doesn't get updated. Maybe we can also change that now.

src/components/actions/dev_tools/routes.ts
nmadruga: I understood that you have the 2 options. What I was asking was: `this would run even if the user defined a different applicationId?` not really sure why it's checking `configId` even if the user passes another applicationId to use.

src/components/actions/dev_tools/routes.ts
nmadruga: I was thinking about how to check LOQ selected for different review levels. Then it would require to check the review_response and their decision, but I think that's a very specific event check

nmadruga:
Hmm I think this will need a bit of overhead to get merged with your changes on #914. And I'll probably wait for that before I can test this using each event and so we can start building a test suit for one snapshot as agreed on our Catch-up meeting 17/10/2022.

database/buildSchema/35_file.sql
CarlosNZ: That's a separate issue though, as the "template_code" and "version" fields on template_element are GENERATED columns, not foreign key references. And STORED Generated columns get generated when first instantiated, not again.

I'm not sure how often this is a problem. If we're querying through GraphQL it's fine, as we'll drill down through the foreign keys from "template".

Do you know where the "template_code" field in "template_element" is being used?

If it's important, we may have to create a trigger to update it.

CarlosNZ:
> Hmm I think this will need a bit of overhead to get merged with your changes on #914. And I'll probably wait for that before I can test this using each event and so we can start building a test suit for one snapshot as agreed on our Catch-up meeting 17/10/2022.

It shouldn't be too bad -- just a little bit of migration code and one line changed in .sql file. I'll sort that out though, depends which one we merge first.

CarlosNZ:
@nmadruga I've added a mechanism to disable email sending when testing with this tool -- see what you think: f89344bf

It's enabled by default (when using this tool), but can be overridden if you want.

CarlosNZ:
Regarding the other items I was going to check that we discussed:

- **See what happens when trying to create a NEW application**  
  If you try to "reset" or "create" when no config application exists, it just returns "Invalid template code, or no config application available". I think this is fine. There should be a config application for all templates, and if there isn't it's just a matter of loading it in the template builder once. Or you can run the trigger on a specific (already existing) application. There's no point writing additional code to create a new application -- it's done elsewhere and that seems beyond the scope of what this tool should be doing.
- **Check reviewAssignment action output**
  This was fine. We were seeing the result of the "updateReviewAssignmentsStatus" action not the "generateReviewAssignments" one, so it makes sense to see the assignments that got changed (to locked) and not the Assigned one (that was already changed when assigning -- this is the result of that event)

src/components/actions/dev_tools/routes.ts
CarlosNZ: I'll be interested to see if the configs that check review response counts cause problems with this. If they do, I can revisit it.

nmadruga:
> @nmadruga I've added a mechanism to disable email sending when testing with this tool -- see what you think: [f89344b](https://github.com/openmsupply/conforma-server/commit/f89344bfa20412161858be4e58de2979f6148e2b)
> 
> It's enabled by default (when using this tool), but can be overridden if you want.

Looks great. Thanks!

src/components/actions/dev_tools/databaseMethods.ts
```diff
@@ -0,0 +1,250 @@
...
+        WHERE code = $1
+      ) ELSE status = 'AVAILABLE'
+      END
+    )`
+    try {
+      const result = await DBConnect.query({ text, values: [templateCode] })
+      if (result.rows.length > 1)
+        throw new Error(
+          'There is more than one config application for this template type/version. Please remedy this situation.'
+        )
```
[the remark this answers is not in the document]
nmadruga: > I encountered an annoying error where the config application kept changing as I went through. Turns out there was more than one config application for a specific template version and different ones were being returned depending on the current state, which was messing up the progress through each event.

Oh, that would have been confusing!

nmadruga (approved):
We agree that this has been tested a lot on development and it's good to be added in the next build

PR #840, opened 2022-09-07
#839 Re-instate action fail behaviour

### LINKED ISSUE #839: Bug: Action sequence doesn't stop after action FAIL
When an action has a "FAIL" result, it should set the relevant trigger field to `ERROR` and stop subsquent actions in the sequence.

This doesn't seem to be happening.

It's possible I messed this up when I did the action aliases and previews, so need to investigate and re-instate this behaviour.

CarlosNZ:
Fix #839 

Not sure how/when this changed, but have changed it back to how it was -- when an action fails, other actions in the sequence don't run, and the trigger is marked "ERROR".

### DIFF
--- a/src/components/actions/processTrigger.ts
+++ b/src/components/actions/processTrigger.ts
@@ -71,7 +71,7 @@ export async function processTrigger(payload: TriggerPayload): Promise<ActionRes
     if (actionFailed) {
       await DBConnect.executedActionStatusUpdate({
         status: ActionQueueStatus.Fail,
-        error_log: 'Action cancelled due to failure of previous sequential action ' + actionFailed,
+        error_log: 'Action cancelled due to failure of previous sequential action: ' + actionFailed,
         parameters_evaluated: null,
         output: null,
         id: action.id,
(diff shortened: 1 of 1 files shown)


src/components/actions/processTrigger.ts
```diff
@@ -106,7 +106,10 @@ export async function processTrigger(payload: TriggerPayload): Promise<ActionRes
 
       // Debug helper console.log to inspect action outputs:
       if (showActionOutcomeLog) console.log('outputCumulative:', outputCumulative)
-      if (result.status === ActionQueueStatus.Fail) console.log(result.error_log)
+      if (result.status === ActionQueueStatus.Fail) {
+        console.log(result.error_log)
+        actionFailed = action.action_code
+      }
```
CarlosNZ, line 112: It was behaving as expected when the action actually threw an error, but most of the time now we're catching them in the action and returning the error status in the `result`, but this wasn't being acted on when the status is "FAIL".

nmadruga (approved):
Great! Let's build this : )

PR #841, opened 2022-09-07
820-part3 change to get is latest review before submission

nmadruga:
### Previously on #820 
I have added a solution to help with figuring out which review_responses should be included in each iteration of LOQ. Basically we needed a simple way (it seemed impossible or very convoluted to get on a GraphQL query) which reviewed elements belong to the latest list sent. Since we have to:
1. Get the level 1 review responses
2. Only the ones that have been AGREED by consolidator in the latest iteration

So one basic example is
- some review_response was previously set as DECLINED and AGREED (by Consolidator)
- A list of Question sent to Applicant that fixed this issue
- Now this particular review_response is APPROVED and AGREED and should no longer be on a new LOQ sent.

### New approach
Another attempt to improve how we get the reviews to be used in LOQ. Just changing the field (renamed to `is_latest_review` to be set to TRUE when is updated. So while the review is still in DRAFT it would already be collected on the GraphQL query that list which questions will be part of this LOQ.

That's the query that should get the correct list:
```

query getDeclineResponses($reviewId: Int!) {
  reviewResponses(
    filter: {
      and: {
        reviewId: { equalTo: $reviewId }
        isLatestReviewSubmission: { equalTo: true }
        decision: { equalTo: AGREE }
        reviewResponseLink: { decision: { equalTo: DECLINE } }
      }
    }
  ) {
    nodes {
      id
      applicationResponse {
        value
        id
        templateElement {
          parameters
          code
          id
        }
      }
      reviewResponseLink {
        decision
        comment
      }
      templateElementId
    }
  }
}
```
~~One example snapshot will be posted here **soon**~~

[UPDATED 08/09/2022] Added one snapshot to be used for testing on conforma-templates repo: `dev/snapshots/LOQ_reviews_easy_query`. Just log in as user `jess` and check the Preview for "Observações"


database/core_templates/schema_init.sql
```diff
@@ -1363,28 +1363,28 @@ CREATE TABLE public.review_response (
     time_updated timestamptz DEFAULT CURRENT_TIMESTAMP,
     time_submitted timestamptz,
     is_visible_to_applicant boolean DEFAULT FALSE,
-    is_latest_review_submission boolean DEFAULT FALSE,
+    is_latest_review boolean DEFAULT FALSE,
```
CarlosNZ, line 1366: I'd wait until after we merge this before updating core_templates.

CarlosNZ (approved):
Okay, I *think* this is all good, although I haven't been able to test with consolidation, etc. so I'll trust you've tried it out?

Couple of small comments -- in particular the name of that function. 

All good to go after that. Thanks, nice work :)

database/core_templates/schema_init.sql
nmadruga: Maybe just remove then?

CarlosNZ: Yeah I guess

PR #843, opened 2022-09-09
Permission policy of type view

fergie-nz:
fixes [Front-end 524](https://github.com/openmsupply/conforma-web-app/issues/524)


database/migration/migrateData.ts
```diff
@@ -916,149 +916,151 @@ const migrateData = async () => {
...
+                                          WHERE
+                                              id = assignment_id
+                                              AND template_id = templ_id
+                                              AND assigned_sections <> '{}'))
+                                      AND template_id = templ_id ORDER BY "index") t), 'level', level_num, 'isLastLevel', is_last_level, 'finalDecision', is_final_decision));
+          RETURN NEW;
+      END;
+      $application_event$
+      LANGUAGE plpgsql;
+        `)
```
CarlosNZ, line 1007: Did you change this intentionally? If not, can you revert it to how it was so it doesn't show up in the diff?  Thanks.

nmadruga: You can also hide the "white space" changes using this:
Screen Shot 2022-09-09 at 1 08 46 PM

database/migration/migrateData.ts
```diff
@@ -916,149 +916,151 @@ const migrateData = async () => {
       EXECUTE FUNCTION public.set_latest_review_response ();
       `)
 
+    await DB.changeSchema(`ALTER TYPE permission_policy_type ADD VALUE 'VIEW' after 'ASSIGN'`)
+
```
nmadruga, line 920: Please add a console.log about what is the change done in this migration

nmadruga (changes requested):
Fergus, this PR is missing the other changes that Andrei have added to a comparison file to enable this new Permissions to be selected in the Template Builder. Please add those in as well

PR #846, opened 2022-09-13
#845 Fix for template version bug

### LINKED ISSUE #845: Bug: Template Builder failing to upload version of existing template
The bug shown was this: 

Screen Shot 2022-09-13 at 11 27 45 AM

I was trying to upload the **same version** of existing template (but don't think this used to be a problem) with a few changes. 
After the error was displayed there are other versions - note same version number - for the same template, all AVAILABLE:
Screen Shot 2022-09-13 at 11 31 30 AM

And this creates a problem for Applicants when trying to start the application this error was showing:
Screen Shot 2022-09-13 at 11 31 12 AM
Screen Shot 2022-09-13 at 11 31 02 AM

CarlosNZ:
Fix #845 

Just removed `template` table from the list of tables which get triggers disabled on snapshot load.

Please let me know if you can think of any others that might cause problems. (I don't think there are any other triggers that are required during snapshot load)

### DIFF
--- a/src/components/snapshots/triggerTables.ts
+++ b/src/components/snapshots/triggerTables.ts
@@ -22,7 +22,6 @@ export const triggerTables = [
   'application_response',
   'application_stage_history',
   'application_status_history',
-  'template',
   'review',
   'review_decision',
   'review_response',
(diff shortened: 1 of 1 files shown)


nmadruga (commented):
Perfecto!

PR #850, opened 2022-09-14
#849 (1) Disable triggers only for full database re-init

### LINKED ISSUE #849: Import snapshot tweaks for Template import
I kind of forgot that importing a template counts as "loading a snapshot", in that it runs the same script, so there's a couple of consequences that need addressing:

1. change the "disable trigger" script that I altered the other day to _not_ disable when doing a template-only import (and restore "template" to the list of triggerTables)
2. when *duplicating* a template, delete the "snapshot" after import so it doesn't appear in the list (we currently do this from the front-end when *importing* a template, but needs to happen for duplication too)

CarlosNZ:
Fix #849 Part 1:

> 1. change the "disable trigger" script that I altered the other day to _not_ disable when doing a template-only import (and restore "template" to the list of triggerTables)

Part 2 done in front-end [#1344](https://github.com/openmsupply/conforma-web-app/pull/1344)

### DIFF
--- a/src/components/actions/evaluatorFunctions.ts
+++ b/src/components/actions/evaluatorFunctions.ts
@@ -1,5 +1,10 @@
-// Utility functions to extend capabilities of expression evaluator via the
-// "objectFunctions" operator
+/*
+Utility functions to extend capabilities of expression evaluator via the
+"objectFunctions" operator.
+
+Any changes done here should also be replicated in front-end
+"evaluatorFunctions.ts"so they can be simulated in the Template Builder.
+ */
 
 import { DateTime, Duration } from 'luxon'
 
--- a/src/components/snapshots/triggerTables.ts
+++ b/src/components/snapshots/triggerTables.ts
@@ -5,6 +5,8 @@ A list of tables whose triggers should be suppressed during snapshot data
 insert, so they don't create extraneous records in the activity_log, or other
 unwanted behaviours.
 
+Note: we don't disable these for template-only imports, as some of the triggers are actually required e.g. template version management trigger
+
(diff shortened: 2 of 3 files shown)


src/components/actions/evaluatorFunctions.ts
```diff
@@ -1,5 +1,10 @@
-// Utility functions to extend capabilities of expression evaluator via the
-// "objectFunctions" operator
+/*
+Utility functions to extend capabilities of expression evaluator via the
+"objectFunctions" operator.
+
+Any changes done here should also be replicated in front-end
+"evaluatorFunctions.ts"so they can be simulated in the Template Builder.
+ */
```
CarlosNZ, line 7: Not related to this PR, but be meaning to put a note in here about this.

src/components/snapshots/triggerTables.ts
```diff
@@ -31,11 +34,3 @@ export const triggerTables = [
...
   'file',
 ]
-
-/*
-The following tables have triggers, but *shouldn't* be disabled, as
-they're required for the snapshot import process:
-
-- template
-
-*/
```
nmadruga, line 41: I remember this being what fixed the problem with importing templates... Wouldn't this cause the issue again?

nmadruga (approved):
I tested and the problem with "Duplicate" and "Import" of templates didn't happen. So I guess it's all good!

src/components/snapshots/triggerTables.ts
CarlosNZ: No, because we now only disable the triggers for full re-inits, not template imports -- that's the whole point of this PR! ;)

Issue #805, opened 2022-08-09
Extend file cleanup


CarlosNZ:
Currently, whenever a file record is deleted, we have a Postgres trigger that automatically deletes the actual file.

However, over time, we have collected a bunch of stray files in the "files" folder of various snapshots that are no longer in the file table.

I think we should extend our "file clean up" script to include these.

Steps to do:
- in `scheduler.ts`, change the name of "cleanUpPreviewFiles" function to something more generic, like "cleanUpFiles".
- as part of that function (probably call another function), do a recursive crawl through the files folder and check the paths against the "file_path" in the "files" table. If it's not in there, delete the file. 
- Update the snapshot export to only include files that are in the "files" table (it should already do this, not sure why it's not)
- Schema change: add a new field to the "file" table called "file_missing" (default false). When the clean-up script detects a file record that doesn't have a corresponding file, it should mark this `true`.

CarlosNZ:
@fergie-nz you might want to have a look at this, could be useful, but feel free to talk it through first.

CarlosNZ:
> I want to remove the getBaseFiles call from the copyFiles function so it doesn't, by default, copy all of the files existing in the files folder whether or not they have a record on the 'files' table.
> 
> I may be oversimplifying this, as I can't imagine why it would be added in the first place if we only wanted to copy files with a record.

There's a few files in there that need be copied no matter what, namely the `_generic_thumbnails` for non-image file icons. So that's what the "base files" is referring to, I think.

I think you *might* be able to do what you're suggesting and remove the "getBaseFiles", but you'll have to just try it and see. As long as the generic thumbnails get restored when starting up or after loading a snapshot, then it should be fine I think.

PR #855, opened 2022-09-15
#805(2) add to clean up files

fergie-nz:
Checks files for orphaned files and deletes them.

I am not sure how to test this function.


CarlosNZ:
> I am not sure how to test this function.

It's runs on a schedule (cleanupFiles) every day, but you can turn on `const schedulerTestMode = true` at the top of this file to make it run every 30 seconds.

It'd be good to be able to manually run it, too, so it might be a good idea to create a script `yarn cleanup` in `package.json` to manually run the "cleanupFiles" function. (See how it's done for `yarn migrate`)

CarlosNZ (changes requested):
Hey, cool, basic idea is sound. 

Few comments, let me know if you need any further clarification.

CarlosNZ:
> It'd be good to be able to manually run it, too, so it might be a good idea to create a script `yarn cleanup` in `package.json` to manually run the "cleanupFiles" function. (See how it's done for `yarn migrate`)

Don't worry too much about this, we can do it later once the function is all running nicely. For now just test it by setting scheduler test mode to `true` and waiting 30 seconds.

src/components/actions/scheduler.ts
```diff
@@ -43,8 +47,22 @@ export const triggerScheduledActions = async () => {
...
 }
+
+const crawlFileSystem = async (path: string) => {
+  fs.readdirSync(path).forEach(async (file) => {
+    const subPath = basePath.join(path, file)
+    if (fs.statSync(subPath).isDirectory()) crawlFileSystem(subPath)
+    else if ((await DBConnect.checkIfInFileTable(subPath)) === 0) {
+      fs.unlink(subPath, function (err) {
+        if (err) throw err
+        else console.log(`Deleted file at ${subPath}`)
```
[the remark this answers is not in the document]
fergie-nz: I'm struggling to implement this one, as the additional parameters the deleteFile function is after don't exist because in this case we are deleting orphaned files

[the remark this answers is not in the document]
CarlosNZ:
```
const { filePath, thumbnailPath, originalFilename } = file
```
You should be able pass the `filePath`, and use the same value for `originalFilename`. `thumbnailPath` is not required.

Come see me if you're still confused what to do.

src/components/actions/scheduler.ts
```diff
@@ -40,11 +48,29 @@ export const triggerScheduledActions = async () => {
...
+  await crawlFileSystem(basePath, checkFile)
   const deleteCount = await DBConnect.cleanUpFiles()
   if (deleteCount > 0) console.log(`${deleteCount} files removed.`)
+  else console.log('no files removed.')
+}
+
+// Manually launch cleanup with command
+if (isManualCleanup) {
+  cleanUpFiles()
 }
```
CarlosNZ, line 74: When running `yarn cleanup`, the script never exits. That's because it missing the exit instruction, like this one:
Screen Shot 2022-10-03 at 4 32 45 PM

src/components/actions/scheduler.ts
```diff
@@ -40,11 +48,29 @@ export const triggerScheduledActions = async () => {
...
-    'Cleaning up preview files...'
+    'Cleaning up filess and file records...'
   )
+  await crawlFileSystem(basePath, checkFile)
   const deleteCount = await DBConnect.cleanUpFiles()
   if (deleteCount > 0) console.log(`${deleteCount} files removed.`)
+  else console.log('no files removed.')
+}
+
+// Manually launch cleanup with command
```
CarlosNZ, line 71:
```suggestion
// Manually launch cleanup with command `yarn cleanup`
```

src/components/postgresConnect.ts
```diff
@@ -359,6 +359,20 @@ class PostgresDB {
     }
   }
 
+  public checkIfInFileTable = async (subPath: string) => {
+    const text = `
+    SELECT id 
+    FROM file
+    WHERE file_path = '${subPath}'
+    `
```
CarlosNZ, line 367: Please use parameterised queries:
```
SELECT id 
    FROM file
    WHERE file_path = $1
```

See the other database methods for how it works (needs a "values" parameter in `this.query...`).

Also -- this give a brief explanation as to why: https://node-postgres.com/features/queries#parameterized-query

src/components/utilityFunctions.ts
```diff
@@ -85,4 +85,13 @@ export const filterObject = (
...
 
+export const crawlFileSystem = async (newPath: string, checkFile: (newPath: string) => void) => {
+  const files = fs.readdirSync(newPath)
+  for (const file of files) {
+    const subPath = path.join(newPath, file)
+    if (fs.statSync(subPath).isDirectory()) crawlFileSystem(subPath, checkFile)
+    else checkFile(subPath)
+  }
+}
+
```
CarlosNZ, line 96: Cool, looks good. Can you just give the input parameters slightly more accurate names, e.g.
- `newPath` -- why not just `path` (there's nothing particularly "new" about it)
- `checkFile` -- we don't know what this function could be -- any operation on a file path, so shouldn't have such a specific name. Maybe just `fileOperation`, or `handleFile`? (I know you called it `checkFile` for the actual function we're using in this case -- that's fine, but no need for the general parameter to have the same specific name)

src/components/actions/scheduler.ts
```diff
@@ -40,11 +48,29 @@ export const triggerScheduledActions = async () => {
...
+      else console.log(`Deleted file at ${filePath}`)
+    })
+  }
+}
+
 export const cleanUpFiles = async () => {
   console.log(
     DateTime.now().toLocaleString(DateTime.DATETIME_SHORT_WITH_SECONDS),
-    'Cleaning up preview files...'
+    'Cleaning up filess and file records...'
```
CarlosNZ, line 63: Excessssive esssses on fileses, my preciousssss...

src/components/actions/scheduler.ts
```diff
@@ -40,11 +48,29 @@ export const triggerScheduledActions = async () => {
   DBConnect.triggerScheduledActions()
 }
 
+const checkFile = async (filePath: string) => {
+  if (await DBConnect.checkIfInFileTable(filePath)) {
+    fs.unlink(filePath, function (err) {
```
CarlosNZ, line 53: Please use the existing `deleteFile` function here. (You've imported it already, so you were probably heading in that direction....)

CarlosNZ (changes requested):
Cool, getting there.

Just one organisation comment. I think it'd be good to keep the actual "cleanup" script and functions out of the "scheduler" file, and keep just the schedule callers in there. It's getting a bit overloaded with unrelated things now.

(We'll add a schedule for this too).

Also, when I ran `yarn cleanup`, I got this:
```
03/10/2022, 16:30:20 Cleaning up filess and file records...
Deleted file at /Users/carl/GitHub/conforma/conforma-server/files/_generic_thumbnails/noun_File_3764922.png
Deleted file at /Users/carl/GitHub/conforma/conforma-server/files/_generic_thumbnails/noun_Image_1570203.png
Deleted file at /Users/carl/GitHub/conforma/conforma-server/files/_generic_thumbnails/noun_PDF_3283219.png
Deleted file at /Users/carl/GitHub/conforma/conforma-server/files/_generic_thumbnails/noun_word_3515287.png
no files removed.
```
"no files removed", but it's still deleting the generic thumbnails.

Can't remember how we said we were going to handle this, maybe check in with me if you're not sure what's required (I can't remember off the top of my head)

src/components/actions/scheduler.ts
fergie-nz: What's esess, precious? Whats's esses, eh?

PR #857, opened 2022-09-15
#856 revokePermission changes

### LINKED ISSUE #856: Revoke permission action -- needs a "wildcard" for user or org
When we revoke permissions from an org, we need to delete all permissioin_join records with an ordId that matches the selected org, but the `userId` can be anything, including `null`. If we supply `null` to the `userId` parameter, it'll only delete records that have `null` for `userId`. 

For "grantPermissions", we made it that you have to *explicitly* set `null` for a user or an org if we wanted to set `null`, and a missing value (undefined) would throw an error,  by design (for security reasons).

However, for this action, I think it's probably better than an "undefined" parameter should mean "ALL" records. So if we revokePermissions for an orgID and don't provide any userId, then it'll delete ALL permission_join records for that org regardless of user.

CarlosNZ:
Fix #856 

Changes action so that "undefined" (i.e. not provided) will match ALL. `null` will still only match explicit `null` values.

This is so we can revoke ALL permissions related to a company, no matter which user(s) they are linked to.

### DIFF
--- a/documentation/List-of-Action-plugins.md
+++ b/documentation/List-of-Action-plugins.md
@@ -8,10 +8,8 @@
   - [Increment Stage](#increment-stage)
   - [Change Status](#change-status)
   - [Modify Record](#modify-record)
-    - [How the record is built](#how-the-record-is-built)
   - [Modify Multiple Records](#modify-multiple-records)
   - [Generate Text String](#generate-text-string)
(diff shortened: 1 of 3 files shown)


nmadruga (approved):
Looks great and simple. LGTM!

Issue #829, opened 2022-09-01
Data views -- small bug: "value_expression" doesn't override "Basic" fields


CarlosNZ:
It says it does in the docs, so it should.

Basically, if the column name in "data_view_column_definition" corresponds to an *actual* column name, then "value_expression" and "additional_formatting" doesn't work properly. It's only working for "custom" column names.

We can work around it by just giving these columns "custom" names, but it would be easier if it behaved as specified in the docs.

PR #859, opened 2022-09-16
#825 Data views -- discrete (non-overlapping) views

### LINKED ISSUE #825: Data Views -- update to allow discrete "views" rather than combined, permission-wise
The problem: currently, Data Views are defined *per database table*, and the endpoint returns ALL results that the current user is allowed to see. That means if we want to create two separate "views" for a table (e.g. Internal and External users list), it won't work -- even though we can define the view, when the user queries the endpoint `/data-views/table/user`, we get back both views combined.

The obvious solution would be to query a specific "view", not a "table". The parameter for the endpoint would be the view `code` rather than the table name. However, the problem with that (and the reason I didn't do it like that originally), is that if a user has permission to see multiple "views" of a given table, then they'd have multiple entries in their "Database" menu, which could get very ugly. For example, we have a "Public" users view which anyone can see, which only shows people users from their own organisation. And we have a "restricted" users view, which shows all users. It's not obvious how to to NOT show the "public" list to senior staff, other than having a separate permission for it which they don't have (and which would be tedious to administer, and hard to make "public"). Hence the current solution -- they see a single "view" for a table ('user'), which contains a union of all the results (rows and columns) they're allowed to see.

Possible solution (there are others, but this is the one I think is best at the moment):
- allow multiple views with the same "code", and we only return the highest priority one they're allowed to see

The work required to achieve this would be:
- change `/data-views` endpoint to return list of available "views" rather than tables
- change `/data-view/table/<name>` endpoint to use "code" instead of "tableName" and only return the one view
- change `data-view/table/<name>/item/<item>` endpoint similarly
- this would involve changing the "union" logic that currently exists. It'd be a bit of work, but the resulting code would (probably) be simpler.

I don't think there'd be much to do in the front-end (if anything), as the lists and links are all generated by the server -- the front-end just displays what its given.

CarlosNZ:
Fix #825  (and #829)

Requires [front-end PR #1350](https://github.com/openmsupply/conforma-web-app/pull/1350)

This was actually fairly straightforward. It now returns the data view by `code` rather than `tableName`, and it limits to only the view requested, not merged with others.

I had to remove the unique constraint for code per table, as we can now have multiple views with the same code, but it'll return (just) the highest priority one if a user can see more than one.

To test/demo, load snapshot from templates repo: `conforma-templates/dev/snapshots/AA_DataViewImprovements.zip`

This snapshot has the data view configuration pretty much how we want it for Angola, so once this is merged, I'll immediately copy and paste the configurations into the Angola snapshot.

Admin can now view active and cancelled products separately, and internal and external users. External users continue to see only their own products, orgs and fellow users. See the "data_views" table if you wish to inspect the definitions.

See inline comments for more detail.


database/buildSchema/44_data_view.sql
```diff
@@ -12,8 +12,7 @@ CREATE TABLE data_view (
     detail_view_exclude_columns varchar[],
     detail_view_header_column varchar NOT NULL,
     show_linked_applications boolean NOT NULL DEFAULT TRUE,
-    priority integer DEFAULT 1,
-    UNIQUE (table_name, code)
```
CarlosNZ, line 16: Remove constraint -- allows more than one view with the same code. This is now how we can have different views based on permission, but without duplicating entries in the list of available for views for users that happen to have multiple permissions

documentation/API.md
```diff
@@ -372,18 +372,18 @@ The return object contains:
 
 The internal function called by this endpoint is the same one run by the ["generateDoc" action](List-of-Action-plugins.md).
 
-### Data Views
+#### Data Views
```
CarlosNZ, line 375: Just tidying up Header levels for consistency

plugins/action_modify_record/src/modifyRecord.ts
```diff
@@ -2,16 +2,8 @@ import { ActionQueueStatus } from '../../../src/generated/graphql'
...
-import { mapValues, get, snakeCase } from 'lodash'
-import { objectKeysToSnakeCase } from '../../../src/components/utilityFunctions'
-import { singular } from 'pluralize'
-
-// This will be prepended to NEW table created if not already present
-export const DATA_TABLE_PREFIX = 'data_table_'
-
-// These are the only tables in the system that we allow to be mutated with this
-// plugin. All other names will have "data_table_" prepended.
-const ALLOWED_TABLE_NAMES = ['user', 'organisation', 'application', 'file']
```
CarlosNZ, line 14: Since these can be shared elsewhere, I've moved the `allowedTableNames` list into `config.ts`

src/components/data_display/helpers.ts
```diff
@@ -41,44 +40,40 @@ export const getPermissionNamesFromJWT = async (request: any): Promise<JWTData>
...
 
   // Generate graphQL filter object
-  const gqlFilters = getFilters(dataViews, userId, orgId)
+  const gqlFilters = getFilters(dataView, userId, orgId)
 
   // Only for details view
-  const headerColumnName = dataViews[0]?.detailViewHeaderColumn ?? ''
-  const showLinkedApplications = dataViews[0].showLinkedApplications
+  const headerColumnName = dataView.detailViewHeaderColumn ?? ''
+  const showLinkedApplications = dataView.showLinkedApplications
```
CarlosNZ, line 72: Downstream stuff gets simpler now that we only deal with a single data view per request.

src/components/data_display/routes.ts
```diff
@@ -11,26 +11,27 @@ import {
...
   const { permissionNames } = await getPermissionNamesFromJWT(request)
   const dataViews = await DBConnect.getAllowedDataViews(permissionNames)
-  const distinctDataViews = getDistinctObjects(dataViews, 'table_name', 'priority')
+  const distinctDataViews = getDistinctObjects(dataViews, 'code', 'priority')
   const dataViewResponse: DataViewsResponse = distinctDataViews.map(
     ({ table_name, title, code }) => ({
       tableName: camelCase(table_name),
       title,
       code,
+      urlSlug: kebabCase(code),
```
CarlosNZ, line 26: This is what will go into to the front-end url now, instead of the table name, since we're requesting a specific view, not just all the matching ones per table.

src/components/utilityFunctions.ts
```diff
@@ -86,3 +88,18 @@ export const filterObject = (
...
+const ALLOWED_TABLE_NAMES = config.allowedTableNames
+
+export const getValidTableName = (inputName: string | undefined): string => {
+  if (!inputName) throw new Error('Missing table name')
+  if (ALLOWED_TABLE_NAMES.includes(inputName)) return inputName
+  const tableName = snakeCase(singular(inputName))
+  const namePattern = new RegExp(`^${DATA_TABLE_PREFIX}.+`)
+
+  return namePattern.test(tableName) ? tableName : `${DATA_TABLE_PREFIX}${tableName}`
+}
```
CarlosNZ, line 105: Pulled from `modifyRecord` so it can be shared with Data Views

documentation/API.md
```diff
@@ -372,18 +372,18 @@ The return object contains:
...
 
 GET:
 `/data-views`
-`/data-views/table/<tableName>`
-`/data-views/table/<tableName>/item/<id>`
+`/data-views/<dataViewCode>`
+`/data-views/<dataViewCode>/<itemId>`
```
CarlosNZ, line 380: I also took the opportunity to simplify the url structure for these endpoints -- having the word `table` and `item` in there was unnecessarily verbose.

src/components/data_display/helpers.ts
```diff
@@ -306,7 +300,8 @@ export const constructDetailsResponse = async (
   const evaluationFieldArray: string[] = []
   const item = columnDefinitionMasterList.reduce(
     (obj: { [key: string]: any }, { columnName, isBasicField, columnDefinition }) => {
-      if (isBasicField) obj[columnName] = fetchedRecord[columnName]
+      if (isBasicField && !columnDefinition?.valueExpression)
+        obj[columnName] = fetchedRecord[columnName]
```
CarlosNZ, line 304: This is all that's needed to fix #829. This just means we can have a `valueExpression` without needing to change the name of the column to a custom name.

src/components/data_display/helpers.ts
```diff
@@ -227,7 +221,7 @@ export const constructTableResponse = async (
   const tableRows = fetchedRecords.map((record, rowIndex) => {
     const thisRow = columnDefinitionMasterList.map((cell, colIndex) => {
       const { columnName, isBasicField, columnDefinition } = cell
-      if (isBasicField) return record[columnName]
+      if (isBasicField && !columnDefinition?.valueExpression) return record[columnName]
```
CarlosNZ, line 224: Fix for #829

nmadruga (commented):
Aha found the migration code - sorry I got lost in between PRs!

documentation/API.md
nmadruga: You were thinking a bit Githubby for a moment

documentation/Data-View.md
```diff
@@ -35,15 +35,16 @@ Returns an array of data table views the user is allowed to see (based on JWT he
...
     ...etc.
 ]
 ```
 
 This endpoint is called in order to populate the Database menu in the UI.
 
-### `/data-views/table/<tableName>?<queries>`
+### `/data-views/<dataViewCode>?<queries>`
 
 For querying a specific table. Data is returned in the following structure:
```
nmadruga, line 49: Probably needs updating to match the latest response

documentation/Data-View.md
```diff
@@ -198,7 +199,7 @@ Note the different types of data returned and the formatting instructions. Also,
 In order to make outcome data available to the front-end, the minimal configuration requirements are a single entry in the `data_view` table with the following fields provided:
 
 - `table_name` -- name of the table we are displaying
-- `code` a unique (per table) identifier
+- `code` the identifier for the particular "view". It's possible to have multiple views with the same code, but only one will be shown to a given user, based on their permissions, and the `priority` value (see below). You can create multiple views with different codes per `table_name`, for example you might have a `usersExt` view and a `usersInt` view for the 'user' table.
```
nmadruga, line 202: I see : )
It's definitely easier to follow now

documentation/Data-View.md
```diff
@@ -219,7 +220,7 @@ Some things to be aware of:
...
-- `priority`: when multiple layouts match the request (i.e. user has permissions for more than one Layout), the returned columns will be a union of the columns specified in both layouts. However,there will still only be one "Title" and "Header column" returned and each layout may have a different one, so the layout with the highest priority will be used. In general you'd want to give the more "restrcited" permission the higher priority, but most of the time you won't need to consider this as the Title and Header would often be the same. Default value: `1`.
+- `priority`: when multiple views match the `code` (i.e. user has permissions for more than one), the view with the highest priority will be returned the to the user. A typical use case for this would be if you have a particular view that is open to all users (i.e `permissionNames` is `null`), but you want users with certain permissions to see a different view for the same request (e.g Staff might be able to see *all* users, whereas external users can only see those in their own organisation). Default value: `1`.
```
nmadruga, line 223: code and tableName maybe? Or just code?

nmadruga (approved):
👍

documentation/Data-View.md
CarlosNZ: What response, sorry?

CarlosNZ: The response shape hasn't changed, I don't think.

documentation/Data-View.md
CarlosNZ: It's just code. If you have two of the same code, they can be pointing to different tables (not that'd you'd want to in practice). And the highest priority one will be returned, as above.

PR #861, opened 2022-09-19
Update Core_actions in Sample template

nmadruga:
Update missing Core_actions in Sample template - Same one Fergus is using to update snapshot for new project
Uploaded latest version of localisations file. Removed English by accident, needs to revert that.


database/core_templates/localisation/languages.json
```diff
@@ -4,7 +4,7 @@
     "description": "NZ English",
     "code": "en_nz",
     "flag": "🇳🇿",
-    "enabled": true
+    "enabled": false
```
CarlosNZ, line 7: Should be `true` in core ;)

CarlosNZ (approved):
Cool, thanks -- just re-enable the English language file before merging, ta.

PR #862, opened 2022-09-20
Small change on languages fields

nmadruga:
Changes to be used in Reviewing [Back-end PR](https://github.com/openmsupply/conforma-web-app/pull/1355)

### DIFF
--- a/src/components/localisation/routes.ts
+++ b/src/components/localisation/routes.ts
@@ -16,7 +16,7 @@ export type LanguageOption = {
   languageName: string
   description: string
   code: string
-  locale: string
+  locale?: string
   flag: string // To-do: limit to flag emojis
   enabled: boolean
 }
@@ -151,6 +151,8 @@ export const routeRemoveLanguage = async (request: any, reply: any) => {
 export const readLanguageOptions = (): LanguageOption[] =>
   JSON.parse(
     readFileSync(path.join(getAppEntryPointDir(), '../localisation/languages.json'), 'utf8')
+  ).map((option: Partial<LanguageOption>) =>
+    'locale' in option ? option : { ...option, locale: 'en_NZ' }
   )
 
 export const writeLanguageOptions = async (languageOptions: LanguageOption[]) =>
--- a/src/components/localisation/routes.ts
+++ b/src/components/localisation/routes.ts
@@ -7,6 +7,7 @@ import {
   makeFolder,
 } from '../../components/utilityFunctions'
 import config from '../../config'
+import { DateTime } from 'luxon'
 
 const { localisationsFolder } = config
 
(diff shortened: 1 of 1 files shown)


src/components/localisation/routes.ts
```diff
@@ -16,6 +16,7 @@ export type LanguageOption = {
   languageName: string
   description: string
   code: string
+  locale: string
```
CarlosNZ, line 19:
```suggestion
  locale?: string
```
We probably want to make this optional and just have a default. That'll save us having to update existing snapshots.

src/components/localisation/routes.ts
```diff
@@ -168,7 +169,9 @@ const checkLanguageFormat = (language: LanguageOption) => {
     language.description &&
     typeof language.description === 'string' &&
     language.flag &&
-    typeof language.flag === 'string'
+    typeof language.flag === 'string' &&
+    language.locale &&
+    typeof language.locale === 'string'
```
CarlosNZ, line 174:
```suggestion
    (language.locale ? typeof language.locale === 'string' : true)
```

Allows it to be optional but still check type if present.

CarlosNZ (commented):
One more thing to do to make "locale" optional, need to inject a default in "readLanguageOptions" if it's missing (`localisation\routes.ts`):
```ts
export const readLanguageOptions = (): LanguageOption[] =>
  JSON.parse(
    readFileSync(path.join(getAppEntryPointDir(), '../localisation/languages.json'), 'utf8')
  ).map((option: Partial<LanguageOption>) =>
    'locale' in option ? option : { ...option, locale: 'en_NZ' }
  )
```

Then we can be sure that a locale will be returned, even for old snapshots, and don't have to check for it in front-end.

nmadruga:
> One more thing to do to make "locale" optional, need to inject a default in "readLanguageOptions" if it's missing (`localisation\routes.ts`):
> 
> ```ts
> export const readLanguageOptions = (): LanguageOption[] =>
>   JSON.parse(
>     readFileSync(path.join(getAppEntryPointDir(), '../localisation/languages.json'), 'utf8')
>   ).map((option: Partial<LanguageOption>) =>
>     'locale' in option ? option : { ...option, locale: 'en_NZ' }
>   )
> ```
> 
> Then we can be sure that a locale will be returned, even for old snapshots, and don't have to check for it in front-end.

Yea, my work around this was having a default `locale` on front-end... But I can do this way. I'm not sure where would be best

CarlosNZ:
Sorry -- one more thing.

When I put an invalid locale code in, the front-end did an ugly crash on the Notes page. 

I think we need to check that the 'locale' values in `language.json` are valid and report an error in the `getPrefs` route if not. 

I'm not sure where we can get a list of valid Luxon locales though. I'll let you know if I find one.

CarlosNZ:
> Yea, my work around this was having a default `locale` on front-end... But I can do this way. I'm not sure where would be best

Actually, Luxon can use the system locale as its default. So we should look that up here, and insert it as the default if not specified.

CarlosNZ (approved):
Cool, looks good. 

It'd be good to have a check for valid "locale" strings, but not sure how to do that easily, so leave it for now.

(Turns out it should have been "en_nz" not "en_NZ")

PR #863, opened 2022-09-22
updates search key for wiki for bitwarden location

fergie-nz:


fergie-nz (commented):
Nicole and I thought this maybe wasn't necessary now and could remove to avoid confusion.

documentation/Demo-Server-Guide.md
```diff
@@ -91,18 +91,6 @@ _To see the actual Docker commands that are constructued, please inspect the fil
...
-   This will launch one instance. To launch other instances in their own container, run the same command, but change:
-
-  - name
-  - port 8000
-  - WEB_HOST url
-  - JWT_SECRET (this can be any random string, but should be a decent length, say > 24 alphanumeric characters. There's no need to record this key anywhere, as it can change anytime -- that just means existing JWTs become invalid, so users will need to re-login)
-
-  The system will be launched with “core_snapshot” data. Upload and load a new snapshot as required.
-
-  Note that this `docker run` procedure resets the container every time it is restarted, so any data changes are wiped out. In order to avoid this, we need persist the data using Docker volumes. See the [docker-compose](#docker-compose) section below for information regarding this.
```
CarlosNZ, line 105: Why remove this?

fergie-nz: I don't understand fully. 

Nicole, maybe you could comment here? I can't remember and/or didn't fully understand the reasoning of why it may be obsolete.

src/generated/graphql.ts
```diff
@@ -7283,17 +7283,6 @@ export type DeleteDataViewByNodeIdInput = {
   nodeId: Scalars['ID'];
 };
 
-/** All input for the `deleteDataViewByTableNameAndCode` mutation. */
```
CarlosNZ, line 7286: You probably want to get rid of this file from this PR, shouldn't be in here.

documentation/Demo-Server-Guide.md
nmadruga: Hey! Long time after...

The problem with this is that we used it to update an existing instance. Which doesn't work. The code after that does the trick. And I think it can be used in all cases. So suggested removing this "duplication" which is misleading

CarlosNZ: Okay, thanks, I just thought there was (some) info in here that wasn't repeated elsewhere. I'll double-check shortly.

nmadruga (commented):
Ok - since the idea of this PR was to help someone in Fergus place to do the server upgrade with more information I suggested to do a few changes to this file. But I think some things got missing to be replace:

1.  where it says `ubuntu@irims-demo.msupply.org` it should be `https://conforma-demo.msupply.org`
2. Maybe moving this heading up `## docker-compose`
3. Explain better about 8000 being instance running on 50000 and the others...

fergie-nz (commented):
For clarity - this reverts this file to the develop version

documentation/Demo-Server-Guide.md
CarlosNZ: Okay, it seems okay. We don't specifically have the command to run a stand-alone instance (i.e. not a docker-compose one), but we have the command to run locally, and that uses the same thing, so it's probably okay.

CarlosNZ:
> Maybe moving this heading up `## docker-compose`

I think what might be good is to add a line just before all the "Move files" stuff, after line 93. And say something like:

"If the server has been configured previously, you can now skip down to the "docker-compose" section (add link). But if this is a new setup, or you're changing existing configurations, please continue."

fergie-nz:
> you're

That sounds good. I think that is clear from a lay persons perspective

fergie-nz:
> > Maybe moving this heading up `## docker-compose`
> 
> I think what might be good is to add a line just before all the "Move files" stuff, after line 93. And say something like:
> 
> "If the server has been configured previously, you can now skip down to the "docker-compose" section (add link). But if this is a new setup, or you're changing existing configurations, please continue."

That sounds good. I think that is clear from a lay persons perspective

nmadruga:
@fergie-nz I think there was a bit of confusion in the last commit - you made it straight into the submodule on _wiki, but that doesn't get persisted in the repo and doesn't show us the changes. This is only meant to be used when upgrading the child repo called _wiki (you will notice that this folder is ommited on Gitignore). Could you please replicate those changes to the files in the `documentation/Demo-Server-Guide.md` instead please?

fergie-nz:
> @fergie-nz I think there was a bit of confusion in the last commit - you made it straight into the submodule on _wiki, but that doesn't get persisted in the repo and doesn't show us the changes. This is only meant to be used when upgrading the child repo called _wiki (you will notice that this folder is ommited on Gitignore). Could you please replicate those changes to the files in the `documentation/Demo-Server-Guide.md` instead please?

Yes indeed I was quite confused. Let me see about fixing that now

fergie-nz:
> 3. Explain better about 8000 being instance running on 50000 and the others...

I haven't quite internalised this part. Could I discuss with you this afternoon?

nmadruga:
> > 5\. Explain better about 8000 being instance running on 50000 and the others...
> 
> I haven't quite internalised this part. Could I discuss with you this afternoon?

It's ok - I was just suggesting to put somewhere a quick explanation that when we do the `docker compose` command with for example APP_PORT=8000 this is to launch or relaunch the server on 50000.

nmadruga (approved):

PR #865, opened 2022-09-22
#977B(1) - Basic search and sorting for data views

CarlosNZ:
Back-end changes required for Front-end #977 (Part 1)

Front-end PR: https://github.com/openmsupply/conforma-web-app/pull/1356

This covers searching and sorting.

Please use snapshot `AA_DataViewFiltering1` in templates repo. (And make sure migration runs)

Note, I've started a new feature branch for this (off develop), so we can keep all this out of the main work until it's all complete, but it'll probably be several PRs.

I've also started on the Filtering, but it's not ready yet so I'll just mark that in the PR and you can ignore it for now.

See inline comments for specifics


.postgraphilerc.js
```diff
@@ -15,6 +15,7 @@ module.exports = {
     // disableQueryLog: true,
     graphileBuildOptions: {
       connectionFilterRelations: true,
+      connectionFilterAllowEmptyObjectInput: true,
```
CarlosNZ, line 18: This makes life a whole lot easier. There was an ugly bit of logic in the front end that put a dummy filter in if it was empty, but it's much nicer just to allow empty filter objects here.

database/buildSchema/44_data_view.sql
```diff
@@ -8,6 +8,7 @@ CREATE TABLE data_view (
     row_restrictions jsonb DEFAULT '{}',
     table_view_include_columns varchar[],
     table_view_exclude_columns varchar[],
+    table_search_columns varchar[],
```
CarlosNZ, line 11: `table_search_columns` specifies which columns the "Search" component should include. (Must be simple "text" columns)

database/buildSchema/44_data_view.sql
```diff
@@ -25,6 +26,7 @@ CREATE TABLE data_view_column_definition (
     element_parameters jsonb,
     additional_formatting jsonb,
     value_expression jsonb,
+    sort_column varchar,
```
CarlosNZ, line 29: If the view column is not a "basic" column (i.e. doesn't correspond to a physical column on the data table), then it won't be sortable. This allows us to specify an "actual" column to sort by if you want it to be.

For example: the "addressFull" field in the "organisation" view is a combination of multiple database fields. So I've specified the "address" column as the sort column. Now when you click the header for the address, it'll sort by that field.

src/components/data_display/gqlDynamicQueries.ts
```diff
@@ -94,3 +94,26 @@ export const queryLinkedApplications = async (id: number, tableName: string) =>
...
+  try {
+    queryResult = await DBConnect.gqlQuery(graphQLquery, variables, authHeaders)
+  } catch (err) {
+    return {
+      error: { error: true, message: 'Problem with Filter List query', detail: err.message },
+    }
+  }
+
+  return queryResult?.[tableNamePlural]?.nodes
+}
```
CarlosNZ, line 119: Ignore for now -- part of filtering

src/components/data_display/index.ts
```diff
@@ -1,3 +1,8 @@
-import { routeDataViews, routeDataViewTable, routeDataViewDetail } from './routes'
+import {
+  routeDataViews,
+  routeDataViewTable,
+  routeDataViewDetail,
+  routeDataViewFilterList,
```
CarlosNZ, line 5: "filterList" endpoint is for fetching the available options for Filters. It's not complete, so please ignore in this PR.

src/components/data_display/routes.ts
```diff
@@ -125,4 +138,38 @@ const routeDataViewDetail = async (request: any, reply: any) => {
...
+      columnName,
+      gqlFilters,
+      authHeaders
+    )
+  ).map((item: { [key: string]: unknown }) => item[columnName])
+
+  // TO-DO: How to handle very large lists?
+  return reply.send([...new Set(items)])
+}
+
```
CarlosNZ, line 174: Filtering -- ignore

src/server.ts
```diff
@@ -138,8 +143,9 @@ const startServer = async () => {
     server.post('/create-hash', routeCreateHash)
     server.post('/generate-pdf', routeGeneratePDF)
     server.get('/data-views', routeDataViews)
-    server.get('/data-views/:dataViewCode', routeDataViewTable)
+    server.post('/data-views/:dataViewCode', routeDataViewTable)
```
CarlosNZ, line 146: Had to change this one to a "POST" method, since it can now take a JSON body (containing the search filter). "GET" method can only take query parameters, which is not enough for the structured filter object.

I was previously sending query info in parameters (for pagination) and constructing it into a filter object in the back-end. However, I'm not using the same machinery used for the ApplicationList for constructing the search filter (and soon the "filtering" filter), so that has to be passed pre-made from the front-end (and is merged into the parameter-based filter here in back-end).

nmadruga (approved):
Looks good - I'm going to review the Front-end and test now

nmadruga:
Ok to merge 👍

PR #32, opened 2020-09-21
#18 schema documentation permission group

nmadruga:
Fix #18 

## Schema
A few changes on the permission tables to correct to what was defined in the document: https://drive.google.com/file/d/1FhUH9Qb9Lud2nO1nKkEHoKLQZWU57Bzm
* permission_policy: added fields `default_restrictions`
* permission_name: added the field `permission_policy_id`
* template_permission: 
  * removed fields: `permission_join_id` and `permission_policy_id`
  * added field: `permission_name_id`
* Removed one .sql file `template_review_stage` after decided it wasn't needed on our 18-09-2020 team meeting.
* Renamed .sql sequential files with numbers.

## Documentation
Added some documentation about Permissions.

PR #867, opened 2022-09-23
#32T - Part 2 - Check for fully approved review

nmadruga:
Part 2 - Fixes https://github.com/openmsupply/conforma-templates/issues/32

[Base PR](880) - Part 1
[Front-end PR](https://github.com/openmsupply/conforma-web-app/pull/1357)

More details soon... the issue above already try to explain what is being fixed. Would need to run `yarn migrate` on the latest snapshot to test, or use one to be saved in our templates repository.

There is a few new changes to functions and some existing tables and the applicationList - which may improve the performance issue. We realised there were too many calls to function during its the returned results for the list, which have been reduced.


database/buildSchema/16_template_element.sql
```diff
@@ -7,6 +7,7 @@ CREATE TYPE public.template_element_category AS ENUM (
 CREATE TYPE public.is_reviewable_status AS ENUM (
     'ALWAYS',
     'NEVER',
+    'ONLY_IF_APPLICANT_ANSWER',
```
CarlosNZ, line 10: Why do we need this extra value? Currently the default value `NULL` does exactly this. Why change it?

We'd have to change all existing templates if we do, so would need to be a pretty strong case for why it's necessary.

CarlosNZ (commented):
I'll have a proper look at this later -- looks pretty comprehensive. 

Just one initial comment, I'll do the same on the front-end for now.

database/buildSchema/16_template_element.sql
nmadruga: Yes, so the case is that otherwise the query for getting the review_response associated gets a lot more complex to deduce when we need to check the `is_reviewable_status` when it can be NULL but the application_response not NULL or if it's != 'ALWAYS' and application_response = NULL. Please have a look on the function `reviewable_questions` for more details.

database/buildSchema/34B_review_assignable_functions.sql
```diff
@@ -26,8 +26,20 @@ CREATE OR REPLACE FUNCTION public.reviewable_questions (app_id int)
     WHERE
         ar.application_id = $1
         AND te.category = 'QUESTION'
-        AND (te.is_reviewable IS NULL
-            OR te.is_reviewable != 'NEVER')
+        AND ((ar.value IS NULL
+                AND te.is_reviewable = 'OPTIONAL_IF_NO_RESPONSE')
+            OR (ar.value IS NOT NULL
+                AND te.is_reviewable != 'NEVER'))
```
nmadruga, line 32: The previous version didn't quite work to get all reviewable_questions considering what was:
- Submitted by applicant AND is_reviewable != 'NEVER'
- Not submitted by applicant AND is_reviewable = 'ALWAYS'

database/buildSchema/16_template_element.sql
nmadruga: https://github.com/openmsupply/conforma-server/pull/867/commits/e4f7ed6071f837b3de5857cc61c8f480a2c35e33#r980706668

CarlosNZ: Okay, in that case, can you make `ONLY_IF_APPLICANT_ANSWER` the `DEFAULT` and make the field `NON_NULLABLE`? So it'll behave the same as currently, but just with a value instead of null. (Might need to run migration code to change all the existing `null`s first.)

Will that require any front-end updates? Probably, right -- but Typescript should catch them once you re-run the codegen.

CarlosNZ: Screen Shot 2022-09-30 at 4 36 03 PM

I don't think you did migration code for this, cos I'm still seeing `NULL` values after running migration script

database/buildSchema/34B_review_assignable_functions.sql
```diff
@@ -0,0 +1,143 @@
+-- Function to return elements reviewable questions (per application)
```
CarlosNZ, line 1: You have to be careful with numbering "B", as the shell script doesn't do a good job with ordering (I think it puts 34B before 34). When I save core_templates with this script and re-run it, it throws an error because of this ordering problem. (I dealt with this a lot way back when we were using these scripts all the time.)

I *think* it's okay to go 34A then 34B, but I'd probably play it safe and just reshuffle all the subsequent numbers so there's no letters in the numbering.

database/migration/migrateData.ts
```diff
@@ -1060,75 +1060,285 @@ const migrateData = async () => {
...
+    await DB.changeSchema(`
+      ALTER TABLE public.template_element 
+        ALTER COLUMN is_reviewable
+        SET DEFAULT 'ONLY_IF_APPLICANT_ANSWER';`)
+
+    await DB.changeSchema(`
+    UPDATE template_element te SET is_reviewable = 'ONLY_IF_APPLICANT_ANSWER'
+      WHERE te.is_reviewable IS NULL
+        AND te.category = 'QUESTION'`)
+  }
```
CarlosNZ, line 1341: You'll also need to add the NOT NULL constraints, and if the field is NON NULLABLE, then you'll have to remove the `AND te.category = 'QUESTION'` bit from above, I think. 

(I don't think there's much point trying to preserve NULL for non-question elements, as they'll all get the default value when created anyway)

database/buildSchema/41_application_list_view.sql
```diff
@@ -46,16 +41,7 @@ CREATE OR REPLACE FUNCTION application_list (userid int DEFAULT 0)
...
-        assigner_action,
-        -- CASE WHEN is_fully_assigned_level_1 IS NULL THEN
-        --     FALSE
-        -- ELSE
-        --     is_fully_assigned_level_1
-        -- END,
-        -- assigned_questions_level_1,
-        total_questions,
-        total_assigned,
-        total_assign_locked
```
CarlosNZ, line 58: I don't fully understand all this, but it's good to see this simplification!

CarlosNZ (changes requested):
Looks really good, although I'm not that clear about the best way to test it.

Some small SQL/migration niggles that need to be sorted. 

I'll try with front-end now. 👍

database/buildSchema/41_application_list_view.sql
nmadruga: Moved to a different PR now: #879 

I'm going to make that base for this PR, but briefly explanation to you why this got removed - because it's not in use by ApplicationList (it already wasn't in use) I think this was going to be used to be checked for completeness of reviews before but stayed half-way through. The only part actually used in ApplicationList is `reviewerActions` and `assignerActions`

database/buildSchema/34B_review_assignable_functions.sql
nmadruga: Thanks! I have done this on the new base PR - Going to base this PR shortly!

nmadruga:
Closing this due to conflict problems. Will open a new PR correctly based from #32-part 1

PR #871, opened 2022-09-27
#870 Action to remove user from org

### LINKED ISSUE #870: New Action -- remove user from company
Does the opposite of "joinUserOrg"

CarlosNZ:
Fix #870 

Very simple -- just deletes the `user_organisation` record

### DIFF
--- a/plugins/action_remove_user_org/src/removeUserOrg.ts
+++ b/plugins/action_remove_user_org/src/removeUserOrg.ts
@@ -9,7 +9,10 @@ const removeUserOrg = async function ({ parameters, DBConnect }: ActionPluginInp
       userId: userId ?? user_id,
       orgId: orgId ?? org_id ?? organisation_id,
     })
-    if (result.success)
+    if (result.success) {
+      console.log(
+        `User ${userId ?? user_id} removed from organisation ${orgId ?? org_id ?? organisation_id}`
+      )
       return {
         status: ActionQueueStatus.Success,
         error_log: '',
@@ -19,7 +22,7 @@ const removeUserOrg = async function ({ parameters, DBConnect }: ActionPluginInp
           orgId: result.organisation_id,
         },
       }
-    else
+    } else
       return {
         status: ActionQueueStatus.Fail,
         error_log: 'There was a problem removing user from organisation.',
--- a/plugins/action_remove_user_org/src/removeUserOrg.ts
+++ b/plugins/action_remove_user_org/src/removeUserOrg.ts
@@ -2,17 +2,19 @@ import { ActionQueueStatus } from '../../../src/generated/graphql'
 import { ActionPluginInput } from '../../types'
 
 const removeUserOrg = async function ({ parameters, DBConnect }: ActionPluginInput) {
-  const { userId, orgId, user_id, organisation_id, org_id } = parameters
+  const { userId, orgId, user_id, organisation_id, org_id, deletePermissions = true } = parameters
+  const userOrg = {
+    userId: userId ?? user_id,
+    orgId: orgId ?? org_id ?? organisation_id,
+  }
   try {
     console.log(`\Removing user from organisation...`)
-    const result = await DBConnect.removeUserOrg({
-      userId: userId ?? user_id,
-      orgId: orgId ?? org_id ?? organisation_id,
-    })
+    const result = await DBConnect.removeUserOrg(userOrg)
+
+    if (deletePermissions) await DBConnect.deleteUserOrgPermissions(userOrg)
+
     if (result.success) {
-      console.log(
-        `User ${userId ?? user_id} removed from organisation ${orgId ?? org_id ?? organisation_id}`
-      )
+      console.log(`User ${userOrg.userId} removed from organisation ${userOrg.orgId}`)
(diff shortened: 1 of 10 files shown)


plugins/action_remove_user_org/src/removeUserOrg.ts
```diff
@@ -0,0 +1,39 @@
+import { ActionQueueStatus } from '../../../src/generated/graphql'
+import { ActionPluginInput } from '../../types'
+
+const removeUserOrg = async function ({ parameters, DBConnect }: ActionPluginInput) {
+  const { userId, orgId, user_id, organisation_id, org_id } = parameters
+  try {
+    console.log(`\Removing user from organisation...`)
+    const result = await DBConnect.removeUserOrg({
+      userId: userId ?? user_id,
+      orgId: orgId ?? org_id ?? organisation_id,
```
CarlosNZ, line 10: Most of our actions now use `camelCase`, so I wanted to keep this pattern. But we accept `snake_case` versions here too, for consistency with `joinUserOrg`

(Shall I update that one to accept both versions too?)

nmadruga: Yes, that's a good call. As long as it's adding more options, that's good. And also for consistency. Can we also allow all options as `orgId ?? org_id ?? organisationId ?? organisation_id` in both actions please?

nmadruga (changes requested):
I tested and it seems all right. Just a few things I would like to clarify about these changes:
- I agree we should also use this time to update `joinUserOrg` to accept the different combinations for camelCase or snake_case in the input.
- We don't need to also remove permissions in `permissionJoin` for the userId and orgId too?
- On our latest snapshot, after the user is removed from organisation, I can log in but there is no options to apply for the common templates a new user would be able to see:
Screen Shot 2022-09-28 at 12 10 57 PM

nmadruga:
> * On our latest snapshot, after the user is removed from organisation, I can log in but there is no options to apply for the common templates a new user would be able to see:
>   Screen Shot 2022-09-28 at 12 10 57 PM

I think the problem with this one is that when we create an Internal User (using the specific template) we don't create those public available permissions that other external users get. Perhaps same problem on CreateExtUserAndCompany template. I'll have a quick look

plugins/action_remove_user_org/src/removeUserOrg.ts
CarlosNZ: Cool, done: e8d7066c

I also added a uniqueness constraint to the user_org table, and handled the "userRole" parameter (which we had documented but not implemented 😳)

CarlosNZ:
> * We don't need to also remove permissions in `permissionJoin` for the userId and orgId too?

Maybe, I don't think it hurts to leave them in: if a user ever gets re-joined to their company then they'll automatically get their permissions back. It still makes sense to have that permission join valid, it's just that that user-org combination can't currently be logged in.

Do you think it's a problem? And if so, should we make it an optional parameter? (`deletePermissions: true/false`)

> * I think the problem with this one is that when we create an Internal User (using the specific template) we don't create those public available permissions that other external users get. Perhaps same problem on CreateExtUserAndCompany template.

Yeah, that's true, but that's a configuration issue. Should the internal users get permission to apply for company etc.? They can't currently log in without a company when part of system_org, so those permissions won't be active until they've been removed from internal org. Probably an edge case anyway.

Probably the best thing to do (at some point, not right now), is just have additional templates to add permissions to external users the way we can do with internal users currently.

nmadruga:
> Maybe, I don't think it hurts to leave them in: if a user ever gets re-joined to their company then they'll automatically get their permissions back. It still makes sense to have that permission join valid, it's just that that user-org combination can't currently be logged in.
> 
> Do you think it's a problem? And if so, should we make it an optional parameter? (`deletePermissions: true/false`)

Yes, I don't know exactly what issue this could cause, but it is definitely unexpected behaviour to have a user which isn't linked to the company but has all company permission joins... I would say we should add and the default is deletePermissions: true.

CarlosNZ:
> I would say we should add and the default is deletePermissions: true.

Done: 1179a16a

plugins/action_remove_user_org/src/removeUserOrg.ts
```diff
@@ -0,0 +1,41 @@
...
+  const { userId, orgId, user_id, organisation_id, org_id, deletePermissions = true } = parameters
+  const userOrg = {
+    userId: userId ?? user_id,
+    orgId: orgId ?? org_id ?? organisation_id,
+  }
+  try {
+    console.log(`\Removing user from organisation...`)
+    const result = await DBConnect.removeUserOrg(userOrg)
+
+    if (deletePermissions) await DBConnect.deleteUserOrgPermissions(userOrg)
```
CarlosNZ, line 14: I didn't bother collecting the result of this one, because it's not critical, and if the first operation (the actual remove user from org) was successful then I'd consider the action a success. 

Anything else would muddy the waters if the user was actually removed.

Also -- barely any chance of this one failing if the first mutation succeeded.

nmadruga (approved):
Tested quickly. Looks good

Issue #864, opened 2022-09-22
The problem with custom-structured response data in data tables


CarlosNZ:
## The Problem

We store much of the output from an application in the same "shape" as the original application response rather than simple text values. The advantage of this is that we can use exactly the same element to display it in the outcome table and make it look pretty, like so:
![Screen Shot 2022-09-22 at 2 04 38 PM](https://user-images.githubusercontent.com/5456533/191642272-0b103f76-3ef5-4735-9c8f-08a6110b53b1.png)

Other advantages include simple mapping from application form to a single field in the database record.

However there are disadvantages to this:
- it's hard to do more with the data, e.g. Filtering, searching, etc.
- it's difficult to extract specific bits of information from within, particularly when there is an arbitrary number of values, or a complex display structure
- the information is not structured in a way that we would if we were just saving the entity itself. The properties in the response are specific to the *display* element, not the data itself
- it's difficult for other software (e.g. CSV output) to parse in a meaningful way, particularly when every instance of it can be differently structured.

The good news is that most fields are stored as "simple" data types (mainly "text", but a few numbers and dates). It's only really *lists* of things that need to be captured differently.

In particular the most complex response shape is the list builder, as it can have an arbitrary number of other elements contained within it (including other list builders!) The most complex example of it is the "Formulations" response in Product Registration, which as an unlimited number of formulations, each of which has its own lists (of arbitrary length) of active and inactive ingredients.

Here is an example of a single "Formulations" response:

https://gist.github.com/CarlosNZ/42241b213ba4470a74868d0e704bd534

## Use cases possibly affected

In order to know how best to handle the problems caused by this kind of structured data, we need to have a clear understanding of what is required. As far as I can tell so far, these are the main two use cases we need to consider:

- Filtering/searching by deeply nested data in the Data View tables
- Outputting a list of data records as a CSV file (from Grafana)
- Displaying records nicely in Grafana (this might not be necessary if we can get enough functionality into the Data Views in the app)

Please describe any additional considerations, and add further detail to the above and give examples of specific usages of them we might encounter.

## Possible solutions

It's hard to know exactly what the best approach would be, and each of those use cases above probably require a different solution.

Also, none of the ideas listed here would provide a comprehensive solution for all types of complex data -- in particular the formulations list is particularly gnarly.

With that in mind, here a few possible ideas of ways to mitigate the above problems:

[...]

PR #873, opened 2022-09-29
#866 Data view filtering

### LINKED ISSUE #866: Data view Filtering improvements
This is for building on the search and sort work already done in PRs #865 [#1356](https://github.com/openmsupply/conforma-server/pull/865).

After discussion today, and [these filtering requirements](https://docs.google.com/document/d/1vsuJ7HGeQ-Ch8Hr1PhRoqiQayRkibdrpgYfZkhTBD2w/edit), as well as [this discussion](https://github.com/openmsupply/conforma-server/issues/864), here is what I suggest.

I already have the basic mechanism for Filters in the data views (it just co-opts the existing Filter mechanism from the Application List), so this is mostly just schema/design details that will (hopefully) allow us to meet the requirements.

#### **data_view table**:

Add columns:
- `filter_include_columns` / `filter_exclude_columns`: Work exactly the same as the equivalent "table" and "detail" view column fields, but define what columns can be filterable on the table page. Note that this will allow us to create filters for fields even if the column doesn't appear in the actual table (manufacturers, administration routes, for example), as well as make some fields explicitly *not* filterable if required.
- update `/data-view/<table>` endpoint to return this info

#### **data_view_column_definitions**:

Add column:
  - `filter_parameters`: Considering the possible data types, and the way columns are build, I'm thinking these possible parameters (in a single JSON field, all optional -- none are required for a "standard" field):
    - `showFilterList`: If not specified, it will be inferred based on `dataType`: `string` will be `true` -- i.e. the front-end will show the filter as a list of unique items, as per the "How the filters work" section below. However, if `false`, filter won't show any lists and will just work as a simple "search-as-you-type" (or data-type appropriate) filter.
    - `delimiter`: A lot of the time we will have text values such as "necuparanib,temgicoluril,palopegteriparatide". It'd be much nicer if these could be separated in the returned Filter list, for selecting from a full list of all possible ingredients, so the delimiter property would specify how to split the string -- in this case it would be ",". So say for example, we had a list of three text values from three records as follows:  
    "necuparanib,temgicoluril,palopegteriparatide" / "necuparanib,paracetamol" / "palopegteriparatide, paracetamol, acetic acid"  
   Then the returned filter list would be: `["necuparanib", "temgicoluril", "palopegteriparatide", "paracetamol", "acetic acid"]`
     - `searchFields`: by default, we are searching the column name. However, if this is a "non-basic" column, (i.e. a "virtual" column that doesn't correspond to a literal database column), then we can tell it which *actual* column (or columns) to search. Equivalent to the "sort_column" field for sorting. (In fact, it might make sense to move the "sort_column" field into this parameter block too -- dunno) 
     - `booleanMapping`: Similar to the existing front-end "Boolean Filter" -- when the data type is a boolean, this defines what values to show instead of just `true`/`false`. For example the "is_active" column is a boolean value, but we want the user to see "Yes"/"No" in their options list, so we'd provide the following booleanMapping:  
     `{true: "Yes", false: "No"}`

### Snapshot

- we'll need to add additional "text-only" columns to the output for fields we want to filter by (e.g. "Active ingredients", "manufacturers"). Most elements have a useable "text" field already, or the evaluator can help us parse the relevant data from the response. (It would be nice to store some values as an array of strings rather than concatenated strings, but from my experimenting with evaluator expressions, it's not possible to construct a simple array from multiple deeply-nested properties, whereas a comma-separated list of strings is super simple, and with the `delimiter` property (above), they'd function exactly the same) And then we can use one column for display (using the appropriate plugin, as we do currently), and the other for search/filter. I suggest sticking to a standard naming pattern for these, such as:
  - active_ingredients_ui (for the "display" data -- i.e. the current one)
  - active_ingredients_text  
  If we're only pulling a part of the full data for searching, it can just be:
  - manufacturers (the full response)
  - manufacturer_names (a concatenated, comma-seperated text string)

### Front-end

- Modifications to the existing "DataViewSearchableList" component to handle the querying/searching described above.
- ~~Allow *multiple* instances of the *same* filter to be created. (So we don't remove them from the "available filters" drop-down). This way, it will be possible to search for multiple different values from different fields, without them having to be "connected" as part of the same "text" string.~~ Actually, this won't be necessary with the `delimiter` parameter described above -- we can just split compound text fields into their distinct parts.

#### How the filters work (from the front-end):

(Applies to "list" filters only, which most will be)

Because we don't know in advance how long the list of possible values in the filter boxes will be, we need a somewhat dynamic approach. Here's what I propose (after some discussion with @andreievg ):

When you first instantiate a filter, it does a query to the "filterList" endpoint, which returns the first 20 items and a flag saying whether this is the complete list or not (or just the totalCount). If it *is* the full list, the front-end filter box just filters this list *locally* as you type, so it'll work exactly as the current application list filters.

However, if it's *not* the full list, the user won't see any results until they type a minimum (3?) characters? Then it'll work like the "Search" plugin with "query as you type" functionality to get the filter list. It only every returns up to 20 values. If there are more than 20, the front-end will just say "more results..." at the bottom of the list so you know to keep typing.

And it basically works the same after that -- a list of items that you can select (multiple) from.

### Additional comments

- The "text" representations for searching and filtering will need to be generated by the "modifyRecord" action when the record is created. That means if we wanted to create different text representations later on (adding a field as searchable that wasn't before, say), then existing records won't have the required text data. We could fix this by running a one-time custom migration query on the data, which wouldn't be too hard, but would require some "tech support" from us to do. I don't actually think this is a very likely scenario too often, just worth bearing in mind.
- The "text" representations should be adequate for outputting to CSV too, as per @jagoje 's example CSV record. So that's good.

All in all, I think this covers all the reasonably expected scenarios, and will definitely allow for the Angola search/filter requirements in an elegant (from the user's point of view) way.

Please let me know if I missed any use cases, or if there's better ways to achieve the same thing.

CarlosNZ:
Fix #866 & #864

Back-end changes required for implementing data view filtering. 

Front-end PR ??

Can test/demo with snapshot: `/conforma-templates/dev/snapshots/AA_DataViewImprovements.zip` Have a look at the definitions in the data_view tables and notice the effect they have. See the "Database" menu for a wider range of data views (for demo purposes only)

More information in front-end PR, and in [original issue](#866), but basically these back-end changes implement:
- schema changes to accommodate filter definitions
- building Filter definitions as part of the `/data-views/<dataViewCode>` endpoint, and returning them with response
- additional (sub) endpoint (`/data-views/<dataViewCode>/filterList` to provide a searchable list of possible values for individual filters

More explanations inline...


src/components/data_display/helpers.ts
```diff
@@ -203,14 +243,78 @@ const buildColumnDisplayDefinitions = async (
...
+    } = columnDefinitions[column]
+
+    return {
+      column,
+      title: title ?? startCase(column),
+      dataType,
+      showFilterList: showFilterList ?? defaultShowFilterList,
+      searchFields: searchFields ?? [column],
+      delimiter,
+      booleanMapping: valueMap ?? booleanMapping,
```
CarlosNZ, line 302: `booleanMapping` and `valueMap` are the same thing. I prefer the term `valueMap`, but the front-end is already using `booleanMapping` for Boolean filters, so we return it to match the front-end to avoid additional renaming there. But I'd rather define it in configurations as `valueMap`, so I've allowed both names for the same thing.

src/components/data_display/routes.ts
```diff
@@ -152,24 +156,80 @@ const routeDataViewFilterList = async (request: any, reply: any) => {
...
+      ? {}
+      : {
+          or: searchFields.map((col: string) => ({
+            [col]: {
+              includesInsensitive: searchText,
+            },
+          })),
+        }
+
+  const gqlFilters = { ...getFilters(dataView, userId, orgId), ...searchFilter }
```
CarlosNZ, line 171: We probably only need to search "string" fields for filter lists, as other types won't be "list" type (e.g. Boolean -> valueMap, date -> Date picker, number -> range selector).

But we can extend this later if we find a use case where it's necessary to have a non-string filter list.

src/components/data_display/routes.ts
```diff
@@ -152,24 +156,80 @@ const routeDataViewFilterList = async (request: any, reply: any) => {
...
-  // GraphQL query column including row_restrictions
-  const gqlFilters = getFilters(dataView, userId, orgId)
+  let fetchedCount = 0
+  let offset = 0
+  let moreResultsAvailable = true
 
-  const items = (
-    await queryFilterList(
+  while (filterList.size < config.filterListMaxLength) {
+    const { fetchedRecords, totalCount, error } = await queryFilterList(
```
CarlosNZ, line 180: Fetch a few records at a time until we reach the "maxListLength" (currently set to 10). We can't just fetch 10 records though, as it's not a one-to-one mapping, due to:
- we want 10 *distinct* values, which might take more than 10 records to find
- some single records will be split into multiple values, using the "delimiter" parameter

src/components/data_display/routes.ts
```diff
@@ -152,24 +156,80 @@ const routeDataViewFilterList = async (request: any, reply: any) => {
...
-  // TO-DO: How to handle very large lists?
-  return reply.send([...new Set(items)])
+    if (error) return reply.send(error)
+
+    fetchedCount += fetchedRecords.length
+
+    fetchedRecords.forEach((record: { [key: string]: unknown }) => {
+      // For some reason, if a single field value is `null`, Postgraphile
+      // returns null instead of an object with a null value on one field, like
+      // the rest of the results
```
CarlosNZ, line 196: This is annoyingly dumb. 

To see what I mean, check out the results of this query:
```
query getFilterList {
  dataTableProducts {
    nodes {
      administrationRoute
    }
    totalCount
  }
}
```

vs this query:
```
query getFilterList {
  dataTableProducts {
    nodes {
      administrationRoute
      name
    }
    totalCount
  }
}
```

And notice the two different ways the `null` value is presented in the results.

src/components/data_display/routes.ts
```diff
@@ -152,24 +156,80 @@ const routeDataViewFilterList = async (request: any, reply: any) => {
...
+      values.forEach((value) => {
+        if (delimiter && typeof value === 'string') {
+          const splitString = value.split(delimiter).map((e) => e.trim())
+          splitString.forEach((string) => filterList.add(string))
+        } else filterList.add(value)
+      })
+    })
+
+    if (fetchedCount >= totalCount) {
+      moreResultsAvailable = false
```
CarlosNZ, line 211: `moreResultsAvailable` allows us to tell the front-end if we've retrieved the full list or not, and display info in the UI accordingly.

database/migration/migrateData.ts
```diff
@@ -1131,20 +1131,30 @@ STABLE;
     ALTER TABLE user_organisation ADD CONSTRAINT user_organisation_user_id_organisation_id_key UNIQUE (user_id, organisation_id);
     `)
```
nmadruga, line 1133: All below needs to go in statement for versionLessThan 0.4.6

I suggest doing this after rebasing this `feature` branch from `develop` and merging here (there are quite a lot of migration code for 0.4.6)

src/components/data_display/gqlDynamicQueries.ts
```diff
@@ -97,14 +97,18 @@ export const queryLinkedApplications = async (id: number, tableName: string) =>
...
   authHeaders: string
 ) => {
   const tableNamePlural = plural(tableName)
   const filterType = upperFirst(camelCase(tableName)) + 'Filter'
-  const variables = { filter: gqlFilters }
-  const graphQLquery = `query getFilterList($filter: ${filterType}) { ${tableNamePlural}( filter: $filter) { nodes { ${column} }, totalCount}}`
+  const variables = { filter: gqlFilters, first, offset }
+  const graphQLquery = `query getFilterList($first: Int!, $offset: Int!, $filter: ${filterType}) { ${tableNamePlural}(first: $first, offset: $offset, filter: $filter) { nodes { ${columns.join(
+    ','
+  )} }, totalCount}}`
```
nmadruga, line 111: That's very cool!

src/components/data_display/routes.ts
```diff
@@ -152,24 +156,80 @@ const routeDataViewFilterList = async (request: any, reply: any) => {
...
+    results = results.filter(
+      (value) =>
+        (typeof value === 'string' &&
+          new RegExp(searchText.toLowerCase()).test(value.toLowerCase())) ||
+        value === null
+    )
+
+  if (results.length > config.filterListMaxLength) moreResultsAvailable = true
+
+  return reply.send({ list: results.slice(0, config.filterListMaxLength), moreResultsAvailable })
```
nmadruga, line 232: I know that most cases `filterListMaxLength` is defined, but it would be good for safe-guard to have a local const that gets this or something else defined in here.

nmadruga (approved):
I think I half understood what is going on here. I might have a better look later on - but for now seems all good and I liked the solution you gave to sort out filtering on arrays, so all good for me to merge in!

src/components/data_display/routes.ts
CarlosNZ: It's hard-coded in config.ts, so shouldn't be missing. But yeah, due to the typing of the config object, it doesn't actually throw an error when a property is missing. We should really rectify that at some point, but for now I've just added a fallback as you suggested.

c83a5827

PR #875, opened 2022-10-02
Update docker entrypoint to append log on restart

andreievg:
When dockerised server is restarted, logs are disappearing. Looks like `tee` command needs to have `-a` for appending rather then overwriting log file. 
/var/log/conforma volume should also be added, this is for log to persist after upgrades (volume should still persist after docker container restarts, unless container is removed)

### DIFF
--- a/docker/entry.sh
+++ b/docker/entry.sh
@@ -28,7 +28,7 @@ yarn postgraphile \
    --disable-query-log \
    -r graphile_user \
    -q '/postgraphile/graphql' \
-   -i '/postgraphile/graphiql' | tee /var/log/conforma/graphile.log &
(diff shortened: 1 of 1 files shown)


CarlosNZ:
Don't we also need to make sure it's in a connected volume too?

CarlosNZ:
> /var/log/conforma volume should also be added, this is for log to persist after upgrades (volume should still persist after docker container restarts, unless container is removed)

Okay, have done this in upcoming PR.

CarlosNZ (approved):
Looks good, thanks. I've done the volume container config here: https://github.com/openmsupply/conforma-server/pull/878/files#diff-758f61a2f54608f37e0cdf5f6f27a7f575ecc4aad97f914a9ddd2f18be45c66c

PR #878, opened 2022-10-03
876 Nodemailer improvements

### LINKED ISSUE #876: Nodemailer problems
After today's email spamming problems (we got 1500+ emails sent out in matter of minutes for the same event), here are some suggestions for how to mitigate the problem and give us better tools to investigate if it happens again:

- [x] Set docker container console log so that it persists within a linked volume (docker-compose) PR: #875 
- [x] don't exclude notification table from snapshot export (update default and individual snapshot options.json file)
- [ ] (Maybe) set a timeout on `sendMail` function within `sendNotification` action (Need to look more carefully at how the nodemailer transport is being created, and how it being fired off async is affecting the parent action
- [x] Update nodemailer package version (It's in the sendNotification package.json, not the global one)
- [x] Notification table should have a timestamp field
- [x] Remove "email_sent" parameter from action "output" (This is always false since the action doesn't wait for the email to send, so it's misleading)

CarlosNZ:
Fix #876 
Addresses most of what is listed in #876 

> Set docker container console log so that it persists within a linked volume (docker-compose) PR: https://github.com/openmsupply/conforma-server/pull/875

@Andrei addressed part of it in #875, the volume persistence added here

> don't exclude notification table from snapshot export (update default and individual snapshot options.json file)

👍

> (Maybe) set a timeout on sendMail function within sendNotification action (Need to look more carefully at how the nodemailer transport is being created, and how it being fired off async is affecting the parent action

Not for now. It's hard to see how it would be needed, since the action accurately reports an error after a time out. And now we'll explicitly log any server responses, so we can check if something weird is going on.

> Update nodemailer package version (It's in the sendNotification package.json, not the global one)

6.4.6 was the highest version it would let me install, even though [npm has 6.8.0 available](https://www.npmjs.com/package/nodemailer). Probably a node version limitation, but there's no indication of why that I could see.

> Notification table should have a timestamp field

Done, and also added an "email_server_log" field where we will capture the server response of all email sending operation (error or otherwise)

> Remove "email_sent" parameter from action "output" (This is always false since the action doesn't wait for the email to send, so it's misleading)

👍

Also added a "test mode" which can be turned on with a flag at the top of the `sendNotification` action. Makes it use a local "[Mailhog](https://github.com/mailhog/MailHog)" configuration for testing.


plugins/action_send_notification/src/sendNotification.ts
```diff
@@ -73,7 +80,6 @@ const sendNotification: ActionPluginType = async ({ parameters, applicationData,
     // Send email
     if (sendEmail && hasValidEmails) {
       console.log('Sending email...')
-      // Note: Using "any" type as imported @types defintions is incorrect, doesn't recognise some fields on "SentMessageInfo" type
```
CarlosNZ, line 76: Someone seems to have updated and correct the typescript definitions!

database/buildSchema/36_notification.sql
```diff
@@ -8,7 +8,9 @@ CREATE TABLE public.notification (
     subject varchar,
     message varchar,
     attachments varchar[],
+    timestamp timestamptz DEFAULT CURRENT_TIMESTAMP NOT NULL,
```
nmadruga, line 11: Would this field get updated when the email is sent (on Update that runs to set 2 variables `email_sent` & `email_server_log`, from my understanding it would be only set on Insert

nmadruga (commented):
The code looks great for me. I'm going to quickly test and get back to you. Just one question pending please

database/buildSchema/36_notification.sql
CarlosNZ: I think the "timestamp" is the timestamp of the notification record, not the email.

I'm not sure that the timestamp of the email is super-important, it's just the result that matters, isn't it?

(I guess we could append an additional timestamp to the email_server_log string if it seems important?)

nmadruga (approved):
Test with successful and looks good. I just wonder if usage of `-` would be better then `\n` since we are not displaying this info other then on the database for now

CarlosNZ:
> Test with successful and looks good. I just wonder if usage of `-` would be better then `\n` since we are not displaying this info other then on the database for now

I see what you mean, but it's not great to read in a single line either way. I think if I was doing any significant debugging with it, I'd export the log text into a text editor -- in which case the `\n` would be easier to parse into line breaks than anything else (there's already actual `-` in the string for example).

PR #879, opened 2022-10-04
Attempt to improve loading on ApplicationList

nmadruga:
Basically this is the work done in PR #867 which will actually help with this work - by dividing the work load.
I thought I could just removed unused variables from being returned in ApplicationList, but forgot that reviewerActions and AssignerActions still use my changed Functions. So I have added all that to this PR. 

Front-end PR: 

### Testing 
Pretty much just need to load a snapshot and run `yarn migrate`. Use the front-end PR to correctly have the ApplicationList without removed variables/properties. And that should be all.

Please note this is considering next release (to be done today) to be 0.4.6!

### Changes
Moving work to improve functions and remove accounts not used in Application List


nmadruga:
Renaming branch...

PR #880, opened 2022-10-04
#32T Part 1  - Attempt improve loading applications list time

nmadruga:
Part 1 - Fixes https://github.com/openmsupply/conforma-templates/issues/32
Base work for Part 2 - PR #867 which will actually help with this work - by dividing the work load.

I thought I could just removed unused variables from being returned in ApplicationList, but forgot that reviewerActions and AssignerActions still use my changed Functions. So I have added all that to this PR. 

Front-end PR: https://github.com/openmsupply/conforma-web-app/pull/1368

### Testing 
Pretty much just need to load a snapshot and run `yarn migrate`. Use the front-end PR to correctly have the ApplicationList without removed variables/properties. And that should be all.

Please note this is considering next release (to be done today) to be 0.4.6!

### Changes
Moving work to improve functions and remove accounts not used in Application List


database/buildSchema/34B_review_assignable_functions.sql
```diff
@@ -0,0 +1,143 @@
...
+        response_id int,
+        is_reviewable public.is_reviewable_status,
+        response_value jsonb,
+        is_optional boolean
+    )
+    AS $$
+    SELECT DISTINCT ON (code)
+        te.code AS code,
+        ar.id AS response_id,
+        te.is_reviewable AS is_reviewable,
```
CarlosNZ, line 14: Just a thought -- perhaps we should change the name of the "is_reviewable" field to "reviewability", as it's now slightly misleading as it has several states?

Might be a hassle though, only do it if it's not too messy.

database/migration/migrateData.ts
```diff
@@ -1137,6 +1079,341 @@ STABLE;
...
+                  id = review_assignment.application_id) = 'PENDING'
+      GROUP BY
+          review_assignment.application_id,
+          review_assignment.level_number;
+          $$
+          LANGUAGE sql
+          STABLE;`
+    )
+
+    console.log(' - Remove not used 4 fields from application_list')
```
CarlosNZ, line 1347:
```suggestion
    console.log(' - Remove 4 unused fields from application_list')
```

CarlosNZ (approved):
Great, this looks really solid. I can't test it particularly thoroughly, but it's showing the correct things in the front-end and it all looks good in here.

Let's hope it gives us something of a performance improvement 🤞

Are you 100% sure that the migrate code is an exact copy of the .sql scripts? (Nothing changed that you haven't updated in both places?)

CarlosNZ:
Oh one other tiny thing -- it needs a fresh "yarn generate" to work properly. 

Can you run `yarn generate` on a fresh `database_init` on both front and back end after merging to develop? Thanks.

database/buildSchema/34B_review_assignable_functions.sql
nmadruga: I'll have a go now

nmadruga: It was ok to do - just needed to be done before other changes on `migrationData` code

nmadruga:
> Oh one other tiny thing -- it needs a fresh "yarn generate" to work properly.
> 
> Can you run `yarn generate` on a fresh `database_init` on both front and back end after merging to develop? Thanks.

Done!

PR #897, opened 2022-10-06
Small postgraphile config tweak -- Move "watch" parameter to config file

CarlosNZ:
I thought we'd already done this, but then I realised that the parameter should be `watchPg` not `watch`. 

Though why it waited until today to start misbehaving for me is a mystery. 

Anyway, this just completes moving all the postgraphile parameters into the ".postgraphilerc.js" file. "watch" was the only one we had kept in the command line, but now I know why it wasn't working in the js file.

Just double-check you can still load a snapshot that involves a schema change with data inserts to the new table.


nmadruga (approved):
Looks ok when I tried to "CREATE TABLE" using this branch

CarlosNZ:
Hi @nmadruga -- would you mind testing this by saving a snapshot, then loading a different snapshot that has schema changes in it (and some data). (e..g load core_templates, then load Angola, or something)? It's just that I was seeing inconsistent behaviour on my system that I hadn't seen before, and this seemed to sort it. But I'm still not sure why it changed, so just want to make sure that it's working as expected for everyone before I merge.

CarlosNZ:
> Hi @nmadruga -- would you mind testing this by saving a snapshot, then loading a different snapshot that has schema changes in it (and some data). (e..g load core_templates, then load Angola, or something)? It's just that I was seeing inconsistent behaviour on my system that I hadn't seen before, and this seemed to sort it. But I'm still not sure why it changed, so just want to make sure that it's working as expected for everyone before I merge.

@fergie-nz would you mind checking too? Thanks.

Issue #848, opened 2022-09-13
Move email server config to prefs


CarlosNZ:
It's currently part of the repo under the "send_notification" action -> config.json.

I suggest that this be loaded into `applicationData` (back-end) as part of the "environmentData" property, so it's available to all actions (and then can get rid of the plugin's own "config" file).

The only problem with that is there will be no compile-time check for missing config. So the `sendNotification` action will also need to gracefully handle missing email config. I suggest that it should still work (i.e. create the notification), but log a warning in the error_log (and write something to console)

CarlosNZ:
I suggest that this loaded into `applicationData` (back-end) as part of the "environmentData" property, so it's available to all actions (and then can get rid of the plugin's own "config" file).

The only problem with that is there will be no compile-time check for missing config. So the `sendNotification` action will also need to gracefully handle missing email config. I suggest that it should still work (i.e. create the notification), but log a warning in the error_log (and write something to console)

CarlosNZ:
@fergie-nz I think this might be a good one for you to have a go at. Let me know if you need anything explained to you.

fergie-nz:
Is this what you had in mind? I've brought the email config over to be called as environment data. However it is still static.

PR #900, opened 2022-10-09
#848 move email config

fergie-nz:


src/types.ts
```diff
@@ -1,4 +1,5 @@
 import { BasicObject, EvaluatorNode } from '@openmsupply/expression-evaluator/lib/types'
+import { ExecOptionsWithStringEncoding } from 'child_process'
```
CarlosNZ, line 2: Wrong import 🚷

src/components/actions/getApplicationData.ts
```diff
@@ -56,10 +56,20 @@ export const getApplicationData = async (input: {
...
 
+  const Emailconfig = {
+    host: 'server.msupply.foundation',
+    port: 465,
+    secure: true,
+    user: 'irims-dev@sussol.net',
+    defaultFromName: 'Conforma',
+    defaultFromEmail: 'test-new-noreply@email.com',
+  }
+
```
CarlosNZ, line 67: Yeah, good, except this needs to be in "preferences.json" rather than hard-coded in here.

You'll also need to make sure it fails gracefully if the preference is missing. I made a suggestion in the original issue as to what probably should happen.

src/types.ts
```diff
@@ -126,6 +127,14 @@ export interface ActionApplicationData {
   environmentData: {
     appRootFolder: string
     filesFolder: string
+    config: {
```
CarlosNZ, line 130: Probably call this `SMTPConfig` to be clear. ("config" could be anything)

CarlosNZ (changes requested):
Cool, looks good. Basically correct, just check the comments I made. (Mainly -- move the config object to preferences)

CarlosNZ:
Also -- just a heads up that the `sendNotification` code has changed somewhat due to #878. So you'll need to re-base from develop now and be careful to deal with the conflicts.

fergie-nz:
Closed to open new clean branch

PR #903, opened 2022-10-11
#851 case insensitive usernames

### LINKED ISSUE #851: Case-insensitive usernames
Needs to allow two things:

- when a user logs in, they can use any case in their input (e.g. "Carl" should log in for "carl")
- database needs to enforce lower-case only usernames (so we don't accidentally end up with multiple versions of the same username ("Carl" and "carl" for example).

We can update the templates to have lower-case only validation too, but this needs to be a convenience bonus, not the actual constraint. (i.e. even if they enter a mixed-case username, it should save a lower-case one to the database)

### Changes required:
- Database constraint to enforce lower-case only usernames (See here for discussion and ideas: https://stackoverflow.com/questions/9046825/postgresql-auto-lowercase-text-while-or-before-inserting-to-a-column)
- `/login` endpoint should transform username input to lower-case before checking
- `/check-unique` endpoint -- not sure exactly, but probably needs a little tweaking to make consistent with the above.
- Migration code to convert all existing usernames to lower-case only

There *shouldn't* be any front-end changes required (please let me know if I'm wrong about that).

fergie-nz:
Fixes #851 

Seems to be working fine!

I wonder if it would be helpful in future to use something like this which saves and then restores views dependent on a table:

https://gist.github.com/mateuszwenus/11187288

If the migrateData gets a bit unwieldy


fergie-nz:
The behaviour is this:

- Data is stored as it was originally entered.
- Where / matching statements ignore case. 
- Selected/returned data retains the original case.
ie SELECT * WHERE username = "nonr" will return value "nonR"

Maybe we will encounter if we do any check unique or matching logic external to our SQL statements ie for searching through returned data.

CarlosNZ:
> 



> The behaviour is this:
> 
>     * Data is stored as it was originally entered.
> 
>     * Where / matching statements ignore case.
> 
>     * Selected/returned data retains the original case.
>       ie SELECT * WHERE username = "nonr" will return value "nonR"

Awesome! And is the unique constraint working as expected: i.e. Database will reject "Carlos" if "carlos" already exists?

database/buildSchema/02_user.sql
```diff
@@ -10,3 +10,7 @@ CREATE TABLE public.user (
     password_hash varchar
 );
 
+-- adds citext constraint to username for case insensitivity
+CREATE EXTENSION IF NOT EXISTS citext;
+ALTER TABLE public.user ALTER COLUMN username TYPE citext;
+
```
CarlosNZ, line 16: You should be able to put `CREATE EXTENSION IF NOT EXISTS citext;` at the top of the file and then just define the type directly:

```
CREATE TABLE public.user (
    id serial PRIMARY KEY,
    ...
    username citext UNIQUE,
    email varchar,
    ...
);
```

database/migration/migrateData.ts
```diff
@@ -1426,14 +1426,90 @@ const migrateData = async () => {
...
-
     console.log(' - Add timestamp and email_server_log to notification table')
     await DB.changeSchema(`
       ALTER TABLE notification 
         ADD COLUMN IF NOT EXISTS timestamp timestamptz
           DEFAULT CURRENT_TIMESTAMP NOT NULL,
         ADD COLUMN IF NOT EXISTS email_server_log varchar; 
     `)
+
+    console.log('add unique constraint to usernames in database')
```
CarlosNZ, line 1437: Minor nit -- can you format the console text the same as the others in this file?

CarlosNZ (commented):
Mostly working well. Case-insensitivity definitely working nicely. 👍

Couple of thoughts:
- I'm having trouble saving a snapshot (with these DB changes in) and reloading it, it throws an error on "createUser...". Not sure quite what's going on, but it's complaining about the type for the username field. We'll need to figure this out before this is ready for merging. Happy to help you investigate if it's proving difficult.
- Can you please go through the whole user registration application and confirm that it all works okay.
- I have an idea for how we can minimise the migration code and avoid all this duplication, though might be worth doing in separate PR just in case it has problems (branch off this one): Whenever the migration script is just repeating exactly some statement that's already in one of the setup .sql files, we should probably put this statement into its own .sql file (which will require renumbering the whole sequence, hence new PR). But then in the migration script, we can just read in the sql text directly from the relevant file, then run that.

CarlosNZ:
Oh - also, can you please update from develop (merge from develop to this branch, remember?), as there is a mistake in one of the other sql scripts in this branch that has been fixed in develop. (You won't be able to save a snapshot without it)

CarlosNZ:
> * I'm having trouble saving a snapshot (with these DB changes in) and reloading it, it throws an error on "createUser...". Not sure quite what's going on, but it's complaining about the type for the username field. We'll need to figure this out before this is ready for merging. Happy to help you investigate if it's proving difficult.

I'm hoping that once we have the pg_dump-based snapshots in place that this problem will go away. So hold of on doing anything with it right now and we'll wait and see. 🤞

database/migration/migrateData.ts
```diff
@@ -1434,6 +1434,83 @@ const migrateData = async () => {
...
+      FROM
+          permission_name
+          LEFT JOIN permission_join ON permission_join.permission_name_id = permission_name.id
+          JOIN permission_policy ON permission_policy.id = permission_name.permission_policy_id
+          LEFT JOIN "user" ON permission_join.user_id = "user".id
+          LEFT JOIN organisation ON permission_join.organisation_id = organisation.id
+          LEFT JOIN template_permission ON template_permission.permission_name_id = permission_name.id
+          LEFT JOIN "template" ON "template".id = template_permission.template_id
+          LEFT JOIN template_category ON "template".template_category_id = template_category.id);
+    `)
```
CarlosNZ, line 1513: Now that all views are recreated in `43_views_functions_triggers.sql`, you don't have to recreate them here.

(Update from develop)

CarlosNZ:
I just updated this from develop for you, so I'll have another look at it.

CarlosNZ:
> * I'm having trouble saving a snapshot (with these DB changes in) and reloading it, it throws an error on "createUser...". Not sure quite what's going on, but it's complaining about the type for the username field. We'll need to figure this out before this is ready for merging. Happy to help you investigate if it's proving difficult.

Sorted with new snapshot method 👍

> * Can you please go through the whole user registration application and confirm that it all works okay.

Just do this.

> * I have an idea for how we can minimise the migration code and avoid all this duplication, though might be worth doing in separate PR just in case it has problems (branch off this one): Whenever the migration script is just repeating exactly some statement that's already in one of the setup .sql files, we should probably put this statement into its own .sql file (which will require renumbering the whole sequence, hence new PR). But then in the migration script, we can just read in the sql text directly from the relevant file, then run that.

This is done with refactor for new snapshot method 👍

database/buildSchema/02_user.sql
```diff
@@ -1,12 +1,17 @@
...
     last_name varchar,
     full_name varchar GENERATED ALWAYS AS (first_name || ' ' || last_name) STORED,
-    username varchar UNIQUE,
+    username citext UNIQUE,
     email varchar,
     date_of_birth date,
     password_hash varchar
 );
 
+ALTER TABLE public.user ALTER COLUMN username TYPE citext;
```
CarlosNZ, line 16: We don't need this line (it's already correct in the CREATE TABLE now)

database/migration/migrateData.ts
```diff
@@ -560,6 +560,18 @@ const migrateData = async () => {
           DEFAULT CURRENT_TIMESTAMP NOT NULL,
         ADD COLUMN IF NOT EXISTS email_server_log varchar; 
     `)
+
+    console.log(' - Add unique constraint to usernames in database')
```
CarlosNZ, line 564:
```suggestion
    console.log(' - Add case-insensitive unique constraint to usernames')
```

database/migration/migrateData.ts
```diff
@@ -560,6 +560,18 @@ const migrateData = async () => {
...
+
+    //drop and save views relating to user
```
CarlosNZ, line 566:
```suggestion
    // Drop views that rely on user table so it can be changed
```

CarlosNZ (approved):
Beautiful -- works like a charm :) 

Please merge.

Issue #18, opened 2020-09-13
Schema Documentation - Permissions


nmadruga:
Needs to move what is in the Permission proposal 29-07 to the schema documentation

PR #904, opened 2022-10-11
#18 Extend action to remove org join to all users

nmadruga:
Based in: #874 (I think it was the one used for latest build...)
Fixes https://github.com/openmsupply/conforma-templates/issues/18

### Testing
- Load [disableOrg-3 template](https://github.com/openmsupply/conforma-server/files/9760354/disableOrg-3.zip) in latest snap shot. Make sure is set as Available.
- Try removing one organisation. (ex: 13) - Log in as `manager` and use the template (Top Menu => Gerência => Desativar Entidade)
- See how it's removed by running this Graphql query **before** and **after**

### Changes

1. Allow passing `username` and/or `orgName` instead of ides
2. Enforces that some orgId or orgName was passed - it cannot be null
3. When no userId or username is passed (`undefined`) will interpret as remove ALL used connected to org
4. Keeps same logic to remove permissionJoins - for the pair `{userId, orgId}` or for all users connected to org (when no userId is defined).

~~TODO: Update documentation~~

### DIFF
--- a/documentation/List-of-Action-plugins.md
+++ b/documentation/List-of-Action-plugins.md
@@ -97,6 +97,7 @@ Changes the application Stage to the next in the sequence
 |                                          | `applicationStatusHistoryTimestamp` or |
 |                                          | `reviewStatusHistoryTimestamp`         |
 
+
 If we are wanting to change the status of a **review**, the parameter `isReview` should be set to `true`. If `applicationId` or `reviewId` are not provided, plugin will try to retrieve from `applicationData`.
 
 **Note:**: If `isReview` is not provided, it default to `true` if the Action was triggered by a review. However, this means that if we want to set the status of an Application when a review is triggered, then we will need to explicitly set `isReview: false`.
@@ -168,13 +169,11 @@ This will look for a user record with `username = "js"` and update it with the _
 #### How the record is built
 
 The full record to be inserted/updated is built from a combination of the `data` parameter and any other non-specific parameters included in the input. Seperated field parameters can be of any degree of complexity (i.e. [evaluator](Query-Syntax.md) expressions), so may be necessary in some cases. However, the vast majority of data fields inserted will be straight mappings from application responses (and there can be an awful lot of them in some workflows!), so we can use the `data` parameter object as a shorthand to map several fields using a more straightforward syntax all at once -- the `data` object is structured with key-value pairs like so:
-
 ```
 { <field-name> : <property-from-applicationData>, ... }
 ```
 
 For example:
-
 ```
 {
   first_name: "responses.Q1.text",
@@ -185,7 +184,6 @@ For example:
 ```
 
 If we had the above `data` parameter, along with the following additional parameters in the input:
-
 ```
 {
   ...
@@ -193,9 +191,7 @@ If we had the above `data` parameter, along with the following additional parame
   current_time: <fetch-current-timestamp-using-"objectFunctions">
 }
 ```
-
 The resultant record that would be inserted might look something like this:
-
 ```
 {
   first_name: "Fred",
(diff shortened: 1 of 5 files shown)


CarlosNZ:
Looks good. Would you mind updating the reference docs (`List-of-Action-plugins.md`) to reflect the new API, it'll make it easier to review (as well as keeping it accurate)?

Thanks.

nmadruga:
Done 6a303d0

nmadruga:
I think it was using Prettier to format .md files too. Now I have installed this extension, but it does add this automatic sub-headers in the contents list. I'll see how to remove this settings
Screen Shot 2022-10-14 at 9 50 16 AM

nmadruga:
> I think it was using Prettier to format .md files too. Now I have installed this extension, but it does add this automatic sub-headers in the contents list. I'll see how to remove this settings Screen Shot 2022-10-14 at 9 50 16 AM

Sorted ;)
Screen Shot 2022-10-14 at 9 55 16 AM

plugins/action_remove_user_org/src/removeUserOrg.ts
```diff
@@ -1,34 +1,58 @@
...
+    if (removed.success) {
       return {
         status: ActionQueueStatus.Success,
         error_log: '',
-        output: {
-          userOrgId: result.id,
-          userId: result.user_id,
-          orgId: result.organisation_id,
-        },
+        output: removed.pairs,
```
CarlosNZ, line 48: This has to be a named property, not an array. Cos the `output` from multiple actions is cumulatively built into the "cumulativeOutput" object by merging the output of each sequential action object. 

So this will break that.

Probably just do `output: {removedUsers: removed.pairs}`

Also -- can you have a think whether any of our current configs use the output of this action anywhere? Cos if so, those will need to be updated

plugins/types.ts
```diff
@@ -10,7 +10,7 @@ export interface ActionPluginInput {
 export interface ActionPluginOutput {
   status: ActionQueueStatus
   error_log: string
-  output?: { [key: string]: any }
+  output?: { [key: string]: any } | Object[]
```
CarlosNZ, line 13: Please revert, as per above comment.

CarlosNZ (changes requested):
Looks good, thanks.

Just one little thing that will need changing. See comment for explanation.

plugins/action_remove_user_org/src/removeUserOrg.ts
CarlosNZ: Oh, and remember to update the docs accordingly :)

nmadruga: Oh, that makes a lot of sense! Thanks

No, I don't think this is used by any other action as an input

documentation/List-of-Action-plugins.md
```diff
@@ -409,11 +409,11 @@ The opposite of `joinUserOrg`. Removes user from company by deleting the `user_o
...
 - _Action Code:_ **`removeUserOrg`**
 
-| Input parameters<br />(\*required) <br/>   | Output properties                       |
-| ------------------------------------------ | --------------------------------------- |
-| `username` or `userId`                     | `array` of `{userOrgId, userId, orgId}` |
-| `orgName` or `orgId` o `organisation_id`\* |                                         |
-| `deletePermissions` (default `true`)       |                                         |
+| Input parameters<br />(\*required) <br/>   | Output properties                                        |
+| ------------------------------------------ | -------------------------------------------------------- |
+| `username` or `userId`                     | `grantedPermissions: {userOrgId, userId, orgId} [Array]` |
```
CarlosNZ, line 414:
```suggestion
| `username` or `userId`                     | `removedUsers: {userOrgId, userId, orgId} [Array]` |
```
Oops -- don't you mean "removedUsers"? ;)

CarlosNZ (approved):
Great, working nicely, output looks correct.

Can you just fix that little documentation thingy and then good to merge: https://github.com/openmsupply/conforma-server/pull/904#discussion_r995278502

👍

nmadruga:
> Can you just fix that little documentation thingy and then good to merge: [#904 (comment)](https://github.com/openmsupply/conforma-server/pull/904#discussion_r995278502)

Wow - Wasn't I distracted on that day! Thanks! Fixed now, will continue with merge.

nmadruga:
Oh... actually this is pointing to `feature/data-view-filter` branch, since we are using this as the base to create latests builds. I'll wait until that is merged to get this PR in also. I don't want to mess with the merge-history by re-basing this to `develop`

CarlosNZ:
> Oh... actually this is pointing to `feature/data-view-filter` branch, since we are using this as the base to create latests builds. I'll wait until that is merged to get this PR in also. I don't want to mess with the merge-history by re-basing this to `develop`

Okay, they're pretty well separated so I think you'd be fine to rebase to develop. But I can also merge the filters branch today if you're happy with it. Maybe after the meeting?

PR #905, opened 2022-10-12
#38T Update NGINX Config to redirect traffic on HTTP requests

nmadruga:
Fixes https://github.com/openmsupply/conforma-templates/issues/38

Follow an advise to use this error_page 497 (for HTTP requests) to be redirected to the proper ssl url

### DIFF
--- a/docker/demo_server/nginx_config/default
+++ b/docker/demo_server/nginx_config/default
@@ -1,5 +1,4 @@
 # copy to /etc/nginx/sites-enabled/default, using "scp" (might need to copy to /home/ubuntu/default first then log in to server and copy with sudo)
-
 server {
     gzip_types text/plain application/xml text/css application/javascript; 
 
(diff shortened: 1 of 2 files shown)


CarlosNZ (approved):
Cool, thanks Nicole. 

Can I just confirm -- this is the exact file that is being used on the current demo server?

Also -- might be a good idea to add this to the other two config files:
- `docker/demo_server/nginx_config_single/default`
- `docker/nginx_config/conforma.conf`

nmadruga:
> 

Interesting! I though about adding to the one in `demo_server/nginx_config_single` as well. But wondering why we have this third nginx_config... I'll have a look on the docs

CarlosNZ:
> But wondering why we have this third nginx_config

I thought the generic one (in nginx_config) is the generic basis, whereas the two in "demo_server" are the actual implementations based off that (single and multiple variants)

nmadruga:
> > But wondering why we have this third nginx_config
> 
> I thought the generic one (in nginx_config) is the generic basis, whereas the two in "demo_server" are the actual implementations based off that (single and multiple variants)

Found it! Is used when generating the docker - so it's a separated nginx configuration internal to docker.
I'll add the changes to `nginx_config_single`

PR #906, opened 2022-10-12
#848 move SMTP config to prefs

fergie-nz:


CarlosNZ:
This is all looking really good now, thanks. I'll just give it a bit more of a test when I get a chance (hopefully later today), then I'll approve it.

Cheers.

plugins/action_send_notification/src/sendNotification.ts
```diff
@@ -123,6 +124,9 @@ const sendNotification: ActionPluginType = async ({ parameters, applicationData,
           )
           db.notificationEmailError(notificationResult.id, `ERROR: ${err.message}`)
         })
+    } else {
+      console.log('Email not sent - missing email configuration')
+      db.notificationEmailError(notificationResult.id, `WARNING: email configuration not provided`)
```
CarlosNZ, line 129: I gave slightly incorrect advice here -- we just need to discriminate between the 3 different `else` conditions:
- what you've done is correct when `!transporter`
- but if `!sendEmail` -- do nothing
- if `!hadValidEmails` -- do something similar but change message to "WARNING: No valid email addresses"

You can probably just remove the `else` statement here and put them in an `if/elseIf` block in the main block -- I'll leave to you to think about which is the most appropriate order of consideration.

(Let me know if that's confusing)

CarlosNZ (changes requested):
Awesome, looks good and working really well.

Just one little thing to tidy up and we'll be good to go.

plugins/action_send_notification/src/sendNotification.ts
```diff
@@ -124,9 +124,15 @@ const sendNotification: ActionPluginType = async ({ parameters, applicationData,
...
+    }
+    //warnings and errors if send email isn't working
+    if (!transporter) {
       console.log('Email not sent - missing email configuration')
-      db.notificationEmailError(notificationResult.id, `WARNING: email configuration not provided`)
+      db.notificationEmailError(notificationResult.id, `WARNING: Email configuration not provided`)
+    }
+    if (!hasValidEmails) {
+      console.log('Email not sent - no valid email address')
+      db.notificationEmailError(notificationResult.id, `WARNING: No valid email addresses`)
```
CarlosNZ, line 135: Problem here is that if `!transporter` then we're likely to also have `!hasValidEmails`. In which case the second one will just overwrite the first one.

Would suggest `else if` so it only checks the second one if the first condition isn't met:
```ts
if (!transporter) {
      console.log('Email not sent - missing email configuration')
      db.notificationEmailError(notificationResult.id, `WARNING: Email configuration not provided`)
    } else if (!hasValidEmails) {
      console.log('Email not sent - no valid email address')
      db.notificationEmailError(notificationResult.id, `WARNING: No valid email addresses`)
    }
```

CarlosNZ (commented):
Thanks -- one tiny little improvement, sorry.

CarlosNZ (approved):
Awesome, looking great. Thanks.

Can you just make sure you put the SMTPConfig into the Fiji snapshot if you haven't already? (Nicole has already done Angola)

PR #907, opened 2022-10-13
Build plugin fix

CarlosNZ:
Fixes the problem with some plugins not building (and not providing any error notification).

Let's get this in ASAP, and I'll start a new build (after updating the appropriate branches) today.

Can test by running `yarn build` and check that it makes "build" folder with no errors. (Compare it with develop, which throws error)


utils/build_all.sh
```diff
@@ -17,4 +17,4 @@ echo 'copying import/export scripts for snapshots'
 cp -R database/*.sh build/database
 cp -R database/*.sql build/database
 echo 'building plugins'
-node ./utils/pluginScripts/buildPlugins.js
+node --unhandled-rejections=strict ./utils/pluginScripts/buildPlugins.js
```
CarlosNZ, line 20: Andrei provided this -- it should make the whole docker build stop on any error, to avoid seeing this problem again.

utils/pluginScripts/buildPlugin.js
```diff
@@ -35,7 +35,7 @@ const buildPlugin = async (pluginName) => {
   }
 
   const build = () => {
-    execSync('yarn tsc --project ' + tempConfigPath, { cwd: rootDirectory })
+    execSync('yarn tsc --project ' + tempConfigPath, { cwd: rootDirectory, stdio: 'inherit' })
```
CarlosNZ, line 38: From Andrei again -- to make error message bubble up to top console.

utils/build_all.sh
CarlosNZ: I don't think we have time to confirm this before updating our servers, but it'd be good if someone can test it later this week by intentionally making one of the plugins non-compile-able (just change a test), and then running "dockerise" and check that it does fail correctly.

plugins/action_trim_responses/src/trimResponsesWithoutDatabase.test.ts
```diff
@@ -1,7 +1,7 @@
 // Test suite for the testResponses Action.
 
 import DBConnect from '../../../src/components/databaseConnect'
-import { ActionQueueStatus, IsReviewableStatus } from '../../../src/generated/graphql'
+import { ActionQueueStatus, IsReviewable } from '../../../src/generated/graphql'
```
nmadruga, line 4: I think this is still wrong - same for `graphql` generated file... Seems to be getting things from not the current changes in latest build (I'm comparing with the one in tag `v0.4.6-7`

nmadruga (changes requested):
I'll run a quick test to check for the failing docker part, but first I think there is still a fix to do

plugins/action_trim_responses/src/trimResponsesWithoutDatabase.test.ts
CarlosNZ: Oh yeah, should be "Reviewability" -- I did have this changed, but think I reverted accidentally. 🙄

Thanks 68677bf4

nmadruga (approved):

PR #908, opened 2022-10-13
848 move email config to prefs

fergie-nz:


fergie-nz:
Hey Carl,

I've just re pulled this - as old branch got tangled with other dev changes

Issue #909, opened 2022-10-14
Backup/snapshot improvments to use pg_dump/pg_restore


CarlosNZ:
After quite a bit of investigation, I've figured out how to get pg_dump/pg_restore to work without the errors we were getting whenever we've tried it in the past. This will make things so much faster -- we can load the whole Angola database in less than a second!

The bad news is, we need to refactor quite a few of the functions to put the schema prefix `public` in the FROM statements. We've done it for quite a few already, but there are plenty of exceptions. The reasons for this are pretty technical, and I don't fully understand it, but you can get a bit of an idea of it from [here](https://stackoverflow.com/questions/56729192/pg-restore-fails-when-trying-to-create-function-referencing-table-that-does-not#comment121636171_58274381) (The problem described in that issue is the problem we were having with pg_restore.)

This would mean our migration script getting a bit out of control with recreating all the functions that already exist, so I suggest we do something like I mentioned to Fergus in [this comment](https://github.com/openmsupply/conforma-server/pull/903#pullrequestreview-1139793684) (3rd bullet point).

So, my suggested workflow for getting pg_dump-based snapshots and backups working is as follows:

- [x]  **Part 1 (Issue #910)** (PR: #914)
  - move all function (and trigger) definitions out of the individual .sql script files and put them all in a single file at the end of the sequence.
  - refactor them to add `public.` schema prefix to the "FROM" clauses
  - remove all function definitions (except DROPs) from the migration script
  - at the end of the migration script call the functions.sql (the new one we've just created) we just read and execute the "CREATE OR REPLACE" for ALL functions and triggers. Yes, this does mean that we drop and re-create all functions for every migration, but it's actually a pretty trivial operation if we're only doing it for version upgrades. And it will save us a lot of hassle when we have to update functions and not have to define them in two different places like we're currently doing. And will stop the migration script getting ridiculous like it currently is (it will be a lot smaller after this, even though it's doing the same amount of work)

- [ ]  **Part 2 (Issue #911)** (PR: #916)
  - Update the useSnapshot/takeSnapshot scripts to use pg_dump and pg_restore when doing whole database snapshots. We should still keep the current mechanism, but the default behaviour will be the new one. I'll need to figure out exactly how the system decides which mechanism to use -- it might be better to just have two options: "Full snapshot", which will use pg_dump; and "template export", which uses the current mechanism, and nothing in between. This will need a bit of looking at to see which is the most straightforward way to migrate the method.
[...]

PR #914, opened 2022-10-17
#910 Refactor sql scripts and migration script

### LINKED ISSUE #910: Refactor .sql and migration script and fix schema problems for pg_dump
Part 1 of Epic #909

CarlosNZ:
Fix #910 

First part of Epic #909 

This has a large diff, but it's mostly just copy-pastes.

Actual changes are:
- updating all functions to have `public.` schema reference in their FROM clauses in order for pg_dump to work properly
- move all functions and triggers into a common file that runs at the end of the setup script
- update migrateData to remove (almost) all functions, and recreate them all at once at the end (along with indexes). This will ensure that every database gets the latest, fixed versions of all functions, and that we only need to maintain them in one place if making changes to functions in future.

This gets the database into a state where it can be used with pg_dump/pg_restore (next PR)

For testing, I guess maybe try loading some older snapshots, and check that migrations all work correctly.


database/buildSchema/42_application_list_view.sql
```diff
@@ -1,6 +1,6 @@
 -- Aggregated VIEW method of all data required for application list page
 -- Requires an empty table as setof return and smart comment to make orderBy work (https://github.com/graphile/graphile-engine/pull/378)
```
nmadruga, line 2:
```suggestion
-- Table to be used on further VIEW of all data required for application list page
-- Requires an empty table as setof return and smart comment to make orderBy work (https://github.com/graphile/graphile-engine/pull/378)
```

database/buildSchema/45_activity_log.sql
```diff
@@ -26,428 +26,3 @@ CREATE TABLE IF NOT EXISTS public.activity_log (
 -- searched by most often
 CREATE INDEX IF NOT EXISTS activity_log_application_index ON activity_log (application_id);
```
nmadruga, line 27:
```suggestion
CREATE INDEX IF NOT EXISTS activity_log_application_index ON public.activity_log (application_id);
```

database/migration/migrateData.ts
```diff
@@ -3,7 +3,9 @@ import DB from './databaseMethods'
 import { ReviewAssignmentsWithSections } from './types'
 import semverCompare from 'semver/functions/compare'
 import { execSync } from 'child_process'
-import { createIndexes } from './indexes'
```
nmadruga, line 6: I couldn't find where this was moved to, the file `indexes.ts` is also being removed.

database/migration/migrateData.ts
```diff
@@ -1451,9 +553,21 @@ const migrateData = async () => {
   }
   // Other version migrations continue here...
 
-  // Set indexes last:
-  console.log('Updating indexes...')
-  await DB.changeSchema(createIndexes)
+  // Update (almost all) Indexes, Views, Functions, Triggers regardless, since
+  // they can be dropped and recreated, or updated with no consequence:
+  const createIndexesScript = readFileSync(
+    path.join(getAppEntryPointDir(), '../database/buildSchema/49_index.sql'),
```
nmadruga, line 559: Ah! Here, I think this file got missing from a commit. Not in this branch `49_index`

nmadruga (changes requested):
I checked that all functions and views have been correctly moved to the other file. And have noticed that `49_index` file is missing. Besides that, I tried to keep an eye for missing prefixes, not sure if all needed. But thought it would be good to compile in one big list : )

1. To keep consistency, I noticed that some views and functions are missing prefix `public.` not sure if that's not required, but thought it might be helpful to list those here:

- All FUNCTION in file `01_jwt_config`
- All TABLE in file `44_data_view`
- Check missing in `50_view_functions_triggers` file:
  - VIEW user_org_join
  - VIEW permissions_all
  - VIEW application_stage_status_latest
  - ALTER TABLE application_status_history
  - FUNCTION delete_whole_application
  - FUNCTION set_original_response
  - FUNCTION assignment_list
  - FUNCTION assigner_list
  - FUNCTION review_list
  - FUNCTION application_list
  - COMMENT ON FUNCTION application_list
  - FUNCTION application_list_filter_applicant
  - FUNCTION application_list_filter_organisation
  - FUNCTION application_list_filter_reviewer
  - FUNCTION application_list_filter_assigner
  - FUNCTION application_list_filter_stage

2. I assume none of the TRIGGERS need to have the prefix `public` correct?

3. File `16_template_element` and `31_reviews` moving Functions - suggestion:
- Would be good to keep all Functions (except the ones in jwt_config) on the new file, so for `public.get_template_code` and `public.get_template_version` which are in use on the next file to generate the column `template_code` I suggest using something similar to what was done for this `ALTER TABLE application_status_history
    ADD COLUMN IF NOT EXISTS application_id INT GENERATED ALWAYS AS (application_status_history_application_id (application_stage_history_id)) STORED;`

- Same thing to move functions `public.review_application_id`, `public.review_reviewer_id`, `public.review_level`, `public.review_stage`, `public.review_time_stage_created`, `public.review_is_last_level`, `public.review_is_last_stage`, `public.review_is_final_decision`. It will be good to keep all of these types in the final file, instead of keeping them in separated places like that.

4. Minor things - Not sure if required changing: 
- Rename function `set_template_verision` to `set_template_version`
- Add prefix `public.` in FROM statement in VIEW `application_stage_status_latest`
- Add prefix `public.` in all FROM/JOIN statement in VIEW `application_stage_status_all`
- Remove empty files `22_ application_stage_status`, `35_review_assignable_functions`, `39_assignment_list`, `43_application_list_filters`
- Remove commented function `public.template_element_parameters_string`
- Remove **duplicated** function `public.review_assignment_template_id` in file `28_review_assignment`
- Add prefix `public.` in EXECUTE FUNCTION in TRIGGER `set_original_response_trigger`
- Add prefix `public` in FROM statement in VIEWs `reviewable_questions_count`,`assigned_questions_count` and `submitted_assigned_questions_count`
- Not sure if needed (since it's using functions and not tables) - add prefix `public` in SELECT CASE statements in FUNCTION assigner_list 
- Add prefix `public` in FROM statement in FUNCTIONs `application_list_filter_applicant`,
`application_list_filter_organisation`, `application_list_filter_reviewer`, `application_list_filter_assigner`, `application_list_filter_stage`
- Add prefix `public.` in EXECUTE FUNCTION in TRIGGER: `status_activity_trigger`,`outcome_insert_activity_trigger`, `outcome_update_activity_trigger`, `deadline_extension_activity_trigger`, `assignment_activity_trigger`, `review_status_activity_trigger`, `review_decision_activity_trigger`, `permission_insert_activity_trigger` and `permission_delete_activity_trigger`
- Add prefix `public.` in FROM statement in FUNCTION `permission_activity_log`

5. Appearance of FUNCTION/TRIGGER in `migrationData` script:
- `CREATE OR REPLACE FUNCTION application_list (userid int DEFAULT 0)`
- `CREATE TRIGGER application_trigger`
- `CREATE TRIGGER trigger_schedule_trigger`
- `CREATE TRIGGER review_trigger`
- `CREATE TRIGGER verification_trigger`

CarlosNZ:
> I checked that all functions and views have been correctly moved to the other file. And have noticed that `49_index` file is missing. Besides that, I tried to keep an eye for missing prefixes, not sure if all needed. But thought it would be good to compile in one big list : )

Thanks, that's very thorough! I think I've got all the essential ones that require the `public.` prefix -- I kept going until pg_restore worked with no errors.  FWIW, only the `FROM` statements within Functions require it -- triggers and views are fine as they are.

Regarding `49_index` -- that's still there. I've deleted the additional indexes script I made for migration, cos that was just an exact duplicate of `49_index`. So I've just done the same thing I did with functions -- read the original script into migration script and run it directly, to avoid having to maintain in two places.

CarlosNZ:
> * Would be good to keep all Functions (except the ones in jwt_config) on the new file, so for `public.get_template_code` and `public.get_template_version` which are in use on the next file to generate the column `template_code` I suggest using something similar to what was done for this `ALTER TABLE application_status_history ADD COLUMN IF NOT EXISTS application_id INT GENERATED ALWAYS AS (application_status_history_application_id (application_stage_history_id)) STORED;`

Yeah, that might work. I've currently just duplicated the functions in the "functions" script (with a comment) currently. I'll have a play with what you suggested -- I think it should be good, and probably less messy. (though I don't love not having the whole table created at once, it's probably a reasonable trade-off in this case).

CarlosNZ:
> * Rename function `set_template_verision` to `set_template_version`

Yeah, I'll need to be a little careful with renaming a function. Will have to explicitly drop the old one so we don't end up with two versions of the same function. And that'll mean having to drop the trigger at the same time. But worth it, yeah, shouldn't be too messy. Thanks.

>     * Remove empty files `22_ application_stage_status`, `35_review_assignable_functions`, `39_assignment_list`, `43_application_list_filters`

Yeah, I was going to wait until this is ready to merge, then do a tidy-up. I'll delete these files and then renumber all the scripts to be sequential. But hadn't done that yet, cos that would make the diff even worse!

>     * Remove commented function `public.template_element_parameters_string`

Yeah -- can you remember what this was even for? (There's actually a couple of functions I came across that seemed to be redundant. I'll check them with you when I find them again, and hopefully remove them.)

>     * Remove **duplicated** function `public.review_assignment_template_id` in file `28_review_assignment`

Yeah, that's another one of those ones that is needed to be defined before it's used in the table. So I'd have to do the same thing you suggested above and add the column afterwards.

>     * Add prefix `public.` in EXECUTE FUNCTION in TRIGGER `set_original_response_trigger`
> 
>     * Add prefix `public` in FROM statement in VIEWs `reviewable_questions_count`,`assigned_questions_count` and `submitted_assigned_questions_count`
> 
>     * Not sure if needed (since it's using functions and not tables) - add prefix `public` in SELECT CASE statements in FUNCTION assigner_list

Doesn't seem to be needed -- no FROM clause. (I can't actually see where it *could* be added in this one)


>     * Add prefix `public` in FROM statement in FUNCTIONs `application_list_filter_applicant`,
>       `application_list_filter_organisation`, `application_list_filter_reviewer`, `application_list_filter_assigner`, `application_list_filter_stage`

It's also not needed here, as it's FROM a function (`FROM application_list()`) -- and functions are global, not schema-specific.

>     5. Appearance of FUNCTION/TRIGGER in `migrationData` script:
> 
> 
>     * `CREATE OR REPLACE FUNCTION application_list (userid int DEFAULT 0)`
> 
>     * `CREATE TRIGGER application_trigger`
> 
>     * `CREATE TRIGGER trigger_schedule_trigger`
> 
>     * `CREATE TRIGGER review_trigger`
> 
>     * `CREATE TRIGGER verification_trigger`

I think some of these are required, though not 100% clear which. Those last two are needed as they're part of the migration to remove a value from an enum. It's a real hassle, there's no way to remove an enum value other than dropping all the objects that reference it, then renaming, then recreating, then dropping the renamed one, then recreating all the referencing objects (I looked into it quite a bit at the time). So I don't think this whole bit can be removed from migration script (even though it's messy and ugly).

CarlosNZ:
> * `public.get_template_code` and `public.get_template_version` which are in use on the next file to generate the column `template_code` I suggest using something similar to what was done for this `ALTER TABLE application_status_history ADD COLUMN IF NOT EXISTS application_id INT GENERATED ALWAYS AS (application_status_history_application_id (application_stage_history_id)) STORED;`
> 
> * Same thing to move functions `public.review_application_id`, `public.review_reviewer_id`, `public.review_level`, `public.review_stage`, `public.review_time_stage_created`, `public.review_is_last_level`, `public.review_is_last_stage`, `public.review_is_final_decision`. It will be good to keep all of these types in the final file, instead of keeping them in separated places like that.

Okay, I managed to move them all over as you suggested: f5e13b93

Thanks.

(Had to also make the INDEXES script run AFTER the functions script, as one index gets created for those auto-generated columns)

database/migration/migrateData.ts
CarlosNZ: Odd -- should be all good now.

database/migration/migrateData.ts
CarlosNZ: Yes, it just uses the single "49_index.sql" script, rather than repeating it in two places.

CarlosNZ:
Okay, I've done a tidy up as per your suggestions. Thanks for looking at it so closely -- it's a real headache wrangling all this.

I've replied to individual comments above, but basically in summary:
- only Functions require the "publid." schema reference in their "FROM" clause -- everything else that doesn't already have it is fine without.
- Functions that refer to another function in their FROM clause also don't need "public" -- in fact they *can't* have it, as functions are defined globally, not per schema (I think)
- There's a big ugly migration that I've had to leave in there in order to drop a couple of values from enum -- have added a comment in the code
- Moved all those functions out of template_element and review as you suggested, and just added their columns in Functions script as suggested. Working well.

FYI, if you want to be able to be able to run these scripts in full:
- delete "schema_init" (and "schema_diff" if there is one) from any snapshots you want to load -- that'll force it to use the current ".sql" scripts rather than the one saved with the snapshot (Might cause problems on new snapshots though, so probably only reliable with core_templates -- after all that's why we saved an init script per snapshot)
- to force the full migration script to run, set the version to 0.0.0. i.e. run `yarn migrate 0.0.0` and check for errors (You should get a lot of "NOTICE"s, they're fine)

So if you run `yarn database_init` and then `yarn migrate 0.0.0` it will run everthing here.

We should probably create (and save) a snapshot that has a really old version of the database. Currently, even if we run `yarn migrate 0.0.0` it'll still skip past a lot of the elements as they already exist. So if you can find the oldest snapshot you can, let's include it in the repo somehow (Not urgent -- just a thought).

nmadruga:
I was checking on out `conforma-templates` repo and the `dev/snapshots/current` is old enough: 0.2.0-0 if you would like to have a go. I'm not sure exactly how to test this so I'll leave that one with you. I'll approve this now. And check the next PR to see the export/import working.

PR #916, opened 2022-10-17
#911 Update snapshot scripts to use pg_dump

### LINKED ISSUE #911: Update snapshot scripts to use pg_dump/pg_restore
Part 2 of Epic #909

CarlosNZ:
Fix #911 

I haven't tested this extensively, but it's basically working. Please try it out and let me know if you find any problems.

You'll have to save a snapshot before you can load one using the new method. Any existing snapshots should load just the same as they do currently.


src/components/permissions/rowLevelPolicyHelpers.ts
```diff
@@ -71,7 +71,7 @@ const updateRowPolicies = async () => {
       .map(({ policyname, tablename }: any) => `DROP POLICY "${policyname}" ON "${tablename}"`)
       .join(';'),
   })
-  console.log(newPolicies.join(';'))
+  // console.log(newPolicies.join(';'))
```
CarlosNZ, line 74: I think it looks cleaner and easier to follow without this massive bit of text in the console

src/components/snapshots/takeSnapshot.ts
```diff
@@ -45,27 +45,46 @@ const takeSnapshot: SnapshotOperation = async ({
...
+      console.log('Exporting files...done')
+    } else {
+      // Do it the old way, using JSON database object export
+      const snapshotObject = await getRecordsAsObject(options)
+      await fs.writeFile(
+        path.join(newSnapshotFolder, `${SNAPSHOT_FILE_NAME}.json`),
+        JSON.stringify(snapshotObject, null, 2)
+      )
+      await copyFiles(newSnapshotFolder, snapshotObject.file ?? [], options)
+    }
```
CarlosNZ, line 73: There's nothing changed to the existing method, it's just all moved inside this if/else block.

src/components/snapshots/useSnapshot.ts
```diff
@@ -63,42 +60,68 @@ const useSnapshot: SnapshotOperation = async ({
...
       execSync(`rm -rf ${FILES_FOLDER}/*`)
     }
 
-    // Prevent triggers from running while we insert data, but only for full re-init
-    if (options.shouldReInitialise) {
+    if (options.usePgDump) {
+      // The quick way, using pg_restore -- whole database only
+      console.log('Restoring database...')
+      execSync(
+        `pg_restore --clean --if-exists --dbname tmf_app_manager ${snapshotFolder}/database.dump`
```
CarlosNZ, line 71: `--clean` drops everything in the database before restore
`--if-exists` means that it won't throw errors if the objects it's dropping (with `clean`) don't exists (i.e. if loading into an already clean database)

src/components/permissions/routes.ts
```diff
@@ -171,7 +171,7 @@ const routeUserPermissions = async (request: any, reply: any) => {
 }
 
 const routeUpdateRowPolicies = async (request: any, reply: any) => {
-  // TODO, add parameters to only drop specific policies, for now drop and reinstante them all
+  // TODO, add parameters to only drop specific policies, for now drop and reinstate them all
```
nmadruga, line 174: Idk if this TODO should be kept. I think it's fine to reinstate all for now - specially if it's that quick!

nmadruga: Just saying to remove the TODO part - but keep the comment about what this is doing

src/components/permissions/rowLevelPolicyHelpers.ts
nmadruga: Can this be set as a flag? Like we have on other places to define as displayRowLevelPolicies. Hmmm Although we can easily get access to this using the pg command `select * from pg_policies` so maybe not. 
Maybe leave a comment about that? Instead of this console.log can say `Run select * from pg_policies to check current set of row level permissions created` ?

src/components/snapshots/routes/helpers.ts
```diff
@@ -9,7 +9,12 @@ export const getSnaphotList = async () => {
 
   for (const dirent of dirents) {
     if (!dirent.isDirectory()) continue
-    if (!fsSync.existsSync(path.join(SNAPSHOT_FOLDER, dirent.name, `${SNAPSHOT_FILE_NAME}.json`)))
+    if (
+      !(
+        fsSync.existsSync(path.join(SNAPSHOT_FOLDER, dirent.name, `${SNAPSHOT_FILE_NAME}.json`)) ||
+        fsSync.existsSync(path.join(SNAPSHOT_FOLDER, dirent.name, `database.dump`))
+      )
+    )
```
nmadruga, line 17: I suggest a comment at the top, I'm not 100% sure of what this is doing. Is it trying to find a snapshot for the given name, and **if snapshot not found** will check if some `datase.dumb` folder exists in the same snapshots folder?

src/components/snapshots/useSnapshot.ts
```diff
@@ -63,42 +60,68 @@ const useSnapshot: SnapshotOperation = async ({
...
-
-    if (options.resetFiles) {
+    if (options.resetFiles || options.usePgDump) {
       execSync(`rm -rf ${FILES_FOLDER}/*`)
     }
 
-    // Prevent triggers from running while we insert data, but only for full re-init
-    if (options.shouldReInitialise) {
+    if (options.usePgDump) {
+      // The quick way, using pg_restore -- whole database only
```
nmadruga, line 68: I'm a bit confused by `whole database only` comment

nmadruga (changes requested):
FYI I checked that:
- Can still load old snapshot (version 0.4.6-3 & 0.5.6-7)
- Export very quickly using `usePgDump` for snapshots (not much data though)
- Problem to import using `usePgDump`. See error below:
```

Restoring database...
pg_restore: while PROCESSING TOC:
pg_restore: from TOC entry 4590; 0 1065900 TABLE DATA application_status_history postgres
pg_restore: error: COPY failed for table "application_status_history": ERROR:  relation "application_stage_history" does not exist
LINE 5:     application_stage_history
            ^
QUERY:  
  SELECT
    application_id
  FROM
    application_stage_history
  WHERE
    id = $1;


CONTEXT:  SQL function "application_status_history_application_id" during inlining
COPY application_status_history, line 1: "10    10      DRAFT   2021-09-08 19:34:27.123407+12  t"
pg_restore: from TOC entry 4606; 0 1066147 TABLE DATA review postgres
pg_restore: error: COPY failed for table "review": ERROR:  relation "review_assignment" does not exist
LINE 5:         review_assignment
                ^
QUERY:  
    SELECT
        application_id
    FROM
        review_assignment
    WHERE
        id = $1;


CONTEXT:  SQL function "review_application_id" during inlining
COPY review, line 1: "2 201     \N"
pg_restore: from TOC entry 4602; 0 1066060 TABLE DATA review_assignment postgres
pg_restore: error: COPY failed for table "review_assignment": ERROR:  relation "application" does not exist
LINE 5:         application
                ^
QUERY:  
    SELECT
        template_id
    FROM
        application
    WHERE
        id = $1;


CONTEXT:  SQL function "review_assignment_template_id" during inlining
COPY review_assignment, line 1: "183    \N      2       1       225     1     2022-09-01 15:54:36.049+12       AVAILABLE       363     \N     {}      \N    2022-09-13 10:59:24.956754+12    1       \..."
pg_restore: from TOC entry 4581; 0 1065768 TABLE DATA template_element postgres
pg_restore: error: COPY failed for table "template_element": ERROR:  relation "template" does not exist
LINE 5:         TEMPLATE
                ^
QUERY:  
    SELECT
        template.code
    FROM
        TEMPLATE
        JOIN template_section ON template_id = template.id
    WHERE
        template_section.id = $1;


CONTEXT:  SQL function "get_template_code" during inlining
COPY template_element, line 1: "114     14      Q1      0       First Name    QUESTION shortText       true    true    true    true    \N     \N      Markdown text can be added to qu..."
pg_restore: from TOC entry 4291; 2606 1065951 FK CONSTRAINT application_response application_response_template_element_id_fkey postgres
pg_restore: error: could not execute query: ERROR:  insert or update on table "application_response" violates foreign key constraint "application_response_template_element_id_fkey"
DETAIL:  Key (template_element_id)=(114) is not present in table "template_element".
Command was: ALTER TABLE ONLY public.application_response
    ADD CONSTRAINT application_response_template_element_id_fkey FOREIGN KEY (template_element_id) REFERENCES public.template_element(id) ON DELETE CASCADE;


pg_restore: from TOC entry 4326; 2606 1066378 FK CONSTRAINT notification notification_review_id_fkey postgres
pg_restore: error: could not execute query: ERROR:  insert or update on table "notification" violates foreign key constraint "notification_review_id_fkey"
DETAIL:  Key (review_id)=(2) is not present in table "review".
Command was: ALTER TABLE ONLY public.notification
    ADD CONSTRAINT notification_review_id_fkey FOREIGN KEY (review_id) REFERENCES public.review(id) ON DELETE CASCADE;


pg_restore: from TOC entry 4317; 2606 1066268 FK CONSTRAINT review_decision review_decision_review_id_fkey postgres
pg_restore: error: could not execute query: ERROR:  insert or update on table "review_decision" violates foreign key constraint "review_decision_review_id_fkey"
DETAIL:  Key (review_id)=(2) is not present in table "review".
Command was: ALTER TABLE ONLY public.review_decision
    ADD CONSTRAINT review_decision_review_id_fkey FOREIGN KEY (review_id) REFERENCES public.review(id) ON DELETE CASCADE;


pg_restore: from TOC entry 4315; 2606 1066229 FK CONSTRAINT review_response review_response_review_id_fkey postgres
pg_restore: error: could not execute query: ERROR:  insert or update on table "review_response" violates foreign key constraint "review_response_review_id_fkey"
DETAIL:  Key (review_id)=(2) is not present in table "review".
Command was: ALTER TABLE ONLY public.review_response
    ADD CONSTRAINT review_response_review_id_fkey FOREIGN KEY (review_id) REFERENCES public.review(id) ON DELETE CASCADE;


pg_restore: from TOC entry 4316; 2606 1066234 FK CONSTRAINT review_response review_response_template_element_id_fkey postgres
pg_restore: error: could not execute query: ERROR:  insert or update on table "review_response" violates foreign key constraint "review_response_template_element_id_fkey"
DETAIL:  Key (template_element_id)=(4738) is not present in table "template_element".
Command was: ALTER TABLE ONLY public.review_response
    ADD CONSTRAINT review_response_template_element_id_fkey FOREIGN KEY (template_element_id) REFERENCES public.template_element(id) ON DELETE CASCADE;


pg_restore: from TOC entry 4318; 2606 1066296 FK CONSTRAINT review_status_history review_status_history_review_id_fkey postgres
pg_restore: error: could not execute query: ERROR:  insert or update on table "review_status_history" violates foreign key constraint "review_status_history_review_id_fkey"
DETAIL:  Key (review_id)=(2) is not present in table "review".
Command was: ALTER TABLE ONLY public.review_status_history
    ADD CONSTRAINT review_status_history_review_id_fkey FOREIGN KEY (review_id) REFERENCES public.review(id) ON DELETE CASCADE;


pg_restore: warning: errors ignored on restore: 10

```

CarlosNZ:
> Problem to import using `usePgDump`. See error below:

This one is probably because the schema hasn't been migrated. I think if you run `yarn migrate` and then re-export, it *should* import properly.

src/components/snapshots/routes/helpers.ts
CarlosNZ: This is just changing what Andrei already had. It would check that `snapshot.json` was present and only consider it an actual snapshot to list if so. This just extends that to accept `snapshot.json` *or* `database.dump`.

src/components/snapshots/useSnapshot.ts
CarlosNZ: As in, you can't do partial exports (like templates only, or certain tables) like you can using the JSON object -- it's the whole database or nothing. (For now -- it's possible we could do partial export/import with pg_dump, but it'd get very messy and complicated, so no point right now)

CarlosNZ:
> > Problem to import using `usePgDump`. See error below:
> 
> This one is probably because the schema hasn't been migrated. I think if you run `yarn migrate` and then re-export, it _should_ import properly.

I'll need to make sure that it's we don't export pg_dumps that haven't been fully migrated. It should be fine in production, with proper version incrementation, but in development it's a bit of a headache. I've added the thing to artificially bump the internal version in dev mode, so the latest migrations always run, but this won't happen if you've just exported the snapshot you're working on (that got loaded with a different branch).

But if you *load* an existing snapshot using *this* branch, then export it, when you re-import it, it should work correctly. Can you let me know if that's not the case?

src/components/permissions/rowLevelPolicyHelpers.ts
CarlosNZ: This is for the console log when loading snapshots. It already says:
```
updating row level policies ... 
updating row level policies ... done
```

Isn't that enough for the console? 

> Maybe leave a comment about that? Instead of this console.log can say Run select * from pg_policies to check current set of row level permissions created ?

This doesn't seem relevant to the context of this bit of code, nor to this PR.

src/components/permissions/routes.ts
CarlosNZ: I dunno, I'm not really changing anything here, just fixing a typo.

We need to do some more general improvements to the whole update policies area, but I'd rather not put any thought into that for this PR, that's a whole other can of worms.

src/components/permissions/rowLevelPolicyHelpers.ts
nmadruga: > Isn't that enough for the console?

Yes that's enough. Could you delete the commented line also please?

All I said was - if the user would like to see what has been included in the database for rowLevel they can run this command. And therefor what this previous command was outputing is not needed.

src/components/permissions/routes.ts
nmadruga: Fine, leave the TODO which isn't something we plan to do anyway

src/components/snapshots/routes/helpers.ts
nmadruga: Oh I see it's only adding to the list of snapshots to be used

src/components/snapshots/useSnapshot.ts
nmadruga: Could you rephrase the comment though? It's not clear

src/components/permissions/routes.ts
CarlosNZ: It's not a matter of whether we do or don't plan to do anything -- it's just that it has nothing to do with this PR. (You wouldn't have even seen this pre-existing comment if I hadn't fixed a typo in it ;)

src/components/snapshots/useSnapshot.ts
CarlosNZ: Okay, I'm not really sure what's not clear about it. Hopefully this is better: c7fb4011

nmadruga:
Trying with another one which is in version 0.3.0 found in `dev/snapshots/appDeadlineTest_2022_07_16`

Result after trying to:

1. Import using `yarn database_init`
2. Export using the Snapshot page
3. Import without problems ✅

CarlosNZ:
> And as Carl mentioned, any Problem that happens inside this block declared in `50_view_functions_triggers` stops the next additions, so he is looking into fix this now so we can also garantee older version will run. I'm going to try another less old to see if there are any propblems now

Fixed in my latest commit b6120135

I don't know why it's so finicky about replacing views with different columns. You would think that "CREATE OR REPLACE" would be the same as "DROP" then "CREATE", but it seems not the case.

src/components/snapshots/useSnapshot.ts
nmadruga: Hahah it's not that different to be honest. I think if you had added something as you explained in the chat of this review it would have been nice. But anyway, it's not a big deal

nmadruga:
Yey it works now 👍

nmadruga (approved):

PR #918, opened 2022-10-18
#912 Automated backup script

### LINKED ISSUE #912: Create backup script and scheduler (using new pg_dump script)
Part 3 of Epic #909

CarlosNZ:
Fix #912 

Regarding encryption -- I played round with quite a few encryption options and zipping tools and this seemed like the best solution for now. It makes a zip file encrypted with AES encryption, which is standard, so can be unzipped using standard OS zip utilities. Other options were either less secure (built-in command line zip), or had to be encrypted separate to the zipping, so you'd need to use conforma to even look at the backup file, which didn't seem like a good idea (Plus I haven't coded a decryption method yet).

Currently, the encrypted zip ouputs are *not* importable using the snapshot upload UI. So to re-import one of these backups (if encrypted), you'll need to unzip it manually, then re-zip it without the encryption. We should incorporate this into the snapshot upload UI at some point, but since the existing .zip library doesn't support encrypted zips (adm-zip), and the library I've used to create them (`archiver-zip-encrypted`) doesn't handle un-zipping, we'll need to find another package for that, so will do later on. Let me know how important you think this is, but to me it seems low priority as restoring a backup should be a very rare occurrence and should really be done by qualified people (i.e. us).

Don't forget to `yarn install` before testing.

For this PR:
- a "backup" script, which just takes a snapshot (using a configurable naming prefix), and moves it to a new "backups" folder (not saved to repo). If a password is provided, it'll be AES-encrypted, otherwise no encryption (standard snapshot zip)
- scheduler to run the above script -- default daily but can be configured in prefs.
- you can run the backup script manually using `yarn backup <optional-password>`
- when run as part of scheduler, password comes from environment variable, so add `BACKUPS_PASSWORD=your-secret` to your personal `.env` file to test that out.

The (optional) preferences (in preferences.json > server) are as follows (example values):
```
    "backupFilePrefix": "angola_backup",
    "backupSchedule": [0],
    "maxBackupDurationDays": 10
```
But they all have defaults if not provided.

As for mapping the "backups" folder to a host folder, I've made that a [separate issue](https://github.com/openmsupply/conforma-server/issues/917) -- cos we'll need to make a docker build of this one to properly develop and test that solution.


database/snapshotOptions/backup.json
```diff
@@ -0,0 +1,14 @@
+{
+  "filters": {},
+  "includeTables": [],
+  "excludeTables": [],
+  "shouldReInitialise": true,
+  "usePgDump": true,
+  "skipZip": true,
```
CarlosNZ, line 7: Added this `skipZip` parameter. This means it won't make a zip file for the "takeSnapshot" step, as the backup script will be creating its own (encrypted) zip immediately after. With large snapshots and high compression levels, zipping twice could be unnecessarily expensive.

src/components/actions/scheduler.ts
```diff
@@ -2,6 +2,7 @@ import DBConnect from '../databaseConnect'
 import scheduler from 'node-schedule'
```
CarlosNZ, line 2: Just FYI -- I plan to move this file out of the "actions" folder into a more general location, since it contains schedulers for more than just actions now.

However, I thought I'd wait until after Fergus's "file clean-up" script is done, since that adds another scheduler here, and the conflicts would be a headache if I've already moved this file.

package.json
```diff
@@ -63,12 +64,15 @@
...
-    "migrate": "ts-node ./database/migration/migrateData.ts --migrate"
+    "migrate": "ts-node ./database/migration/migrateData.ts --migrate",
+    "backup": "ts-node ./src/components/exportAndImport/backup.ts --backup"
   },
   "dependencies": {
     "@graphile-contrib/pg-simplify-inflector": "^6.0.0",
     "@openmsupply/expression-evaluator": "^1.11.2",
     "adm-zip": "^0.5.5",
+    "archiver": "^5.3.1",
+    "archiver-zip-encrypted": "^1.0.11",
```
CarlosNZ, line 75: So there's two packages here -- "archiver" is a general purpose archiving library, and "archiver-zip-encrypted" is a plugin/extension for "archiver" that supports encryption.

src/components/snapshots/routes/routeUploadSnapshot.ts
```diff
@@ -32,14 +32,20 @@ const routeUploadSnapshot = async (request: FastifyRequest, reply: FastifyReply)
...
       const snapshotMainFileName = `${SNAPSHOT_FILE_NAME}.json`
 
-      if (!zip.getEntries().find(({ entryName }) => entryName === snapshotMainFileName)) {
+      if (
+        !zip
+          .getEntries()
+          .find(
+            ({ entryName }) => entryName === snapshotMainFileName || entryName === 'database.dump'
+          )
+      ) {
```
CarlosNZ, line 45: Just changing what gets recognised as a valid snapshot. Previously it looked for `snapshot.json`, now it'll accept either `snapshot.json` or `database.dump` (should have been part of last PR, really)

src/components/snapshots/takeSnapshot.ts
```diff
@@ -1,12 +1,12 @@
-import fs from 'fs/promises'
-import fsSync from 'fs'
+import fs from 'fs'
```
CarlosNZ, line 1: There was already two different imports of `fs`, I've just combined them into one rather than creating a third.

src/components/snapshots/takeSnapshot.ts
```diff
@@ -130,15 +130,12 @@ const getOptions = async (
...
-      if (error) return reject(error)
-      resolve('done')
-    })
-  )
+  const output = fs.createWriteStream(path.join(SNAPSHOT_FOLDER, `${snapshotName}.zip`))
+  const archive = archiver('zip')
+
+  archive.pipe(output)
+  archive.directory(snapshotFolder, false)
+  archive.finalize()
```
CarlosNZ, line 138: The standard snapshot export is now using the same "archiver" zip library. We can eventually remove `adm-zip`, altogether, but it's still needed to un-zipping when uploading snapshots. (Need to find a replacement that can handle encrypted zips)

PR #920, opened 2022-10-19
#917 Docker compose - configure for backups

### LINKED ISSUE #917: Backups - configure docker-compose to map volume to host folder
Final part of Epic #909.

Making this a separate issue, as we'll need to make a build of the rest of the functionality in order to configure and test this.

Will need to use a "bind-mount" rather than a volume in order to be able to map it to a specified host folder location.

Hopefully the host path can be passed in dynamically as part of Environment variables. https://docs.docker.com/compose/compose-file/compose-file-v3/#volumes

CarlosNZ:
Fix #917 

Last part of Epic #909 

Adds backup folder to docker-compose volumes. You can specify the folder to use in an environment variable "BACKUPS_FOLDER", but it will default to the standard volumes location if not provided.

Have updated the docs to reflect this and the "BACKUPS_PASSWORD" variable.

Have tested with a local Docker build, seems to be working okay, but won't know for sure until we test it out on a server.


CarlosNZ:
This is now live on Demo server 50008 and Angola server, so we'll see how it goes.

CarlosNZ:
Screenshot 2022-10-27 at 4 48 01 PM

FYI, this has been installed on Demo servers (and Angola) for the best part of a week now and seems to be saving nightly backups correctly 👍

nmadruga:
> Have updated the docs to reflect this and the "BACKUPS_PASSWORD" variable.
> 
> Have tested with a local Docker build, seems to be working okay, but won't know for sure until we test it out on a server.

I see. So this one hasn't been included to v0.4.6-9? I think after Reviewing this, if you can start a new build with what we have included today in `develop`... then we can start testing this on instance 50008 and release with other changes the version 0.4.6 by the start of next week! Exciting : )

docker/demo_server/docker-compose.yml
```diff
@@ -18,6 +18,7 @@ services:
       - files:/usr/src/conforma-server/build/files
       - localisation:/usr/src/conforma-server/build/localisation
       - preferences:/usr/src/conforma-server/build/preferences
+      - ${BACKUPS_FOLDER:-backups}:/usr/src/conforma-server/build/backups
```
nmadruga, line 21: No sure why adding an suffix `-backups` to the BACKUP_FOLDER. Better to leave without a suffix IMO

nmadruga (approved):
LGTM

docker/demo_server/docker-compose.yml
CarlosNZ: That's not a suffix -- it's a fallback if `BACKUPS_FOLDER` is not provided ;)

CarlosNZ:
> I see. So this one hasn't been included to v0.4.6-9?

It is, but what's in the repo here doesn't actually affect the build -- the "docker-compose.yml" file needs to be updated on all servers, and the environment variables added. Which I did on Angola and Demo server and seems to be working nicely, as per [above comment](https://github.com/openmsupply/conforma-server/pull/920#issuecomment-1292934514)

PR #915, opened 2022-10-17
Add new function to check for all approved/agreed

nmadruga:
Please check more info in: [Front-end PR](https://github.com/openmsupply/conforma-web-app/pull/1391)

Based on PR (Refactoring for `migrationData`) #914 

### Changes

Adds function to be used in comparison to check if Review (going to be submitted) has all APPROVED/AGREED elements that should be for All Reviews - even if there are other ones involved - for elements which are **reviewable**...

The Condition on IncrementStage now is a bit complex, but I couldn't  figure out how to make a simpler function that could check one against the other. If anyone knows, please add to here, and then we can have a single-graphQL call to check : )

IncrementStage Condition:
```json
{
 "operator": "AND",
 "children": [
  {
   "operator": "OR",
   "children": [
    {
     "operator": "AND",
     "children": [
      {
       "operator": "=",
       "children": [
        {
         "operator": "objectProperties",
         "children": [
          "applicationData.reviewData.latestDecision.decision"
         ]
        },
        "CONFORM"
       ]
      },
      {
       "operator": "=",
       "children": [
        {
         "operator": "graphQL",
         "children": [
          "query approvedAssignedQuestionsCount(\n  $appId: Int!\n  $stageId: Int!\n  $levelNumber: Int!\n) {\n  approvedAssignedQuestionsCount(\n    appId: $appId\n    stageId: $stageId\n    levelNumber: $levelNumber\n  )\n}",
          "graphqlendpoint",
          [
           "appId",
           "stageId",
           "levelNumber"
          ],
          {
           "operator": "objectProperties",
           "children": [
            "applicationData.applicationId",
            null
           ]
          },
          {
           "operator": "objectProperties",
           "children": [
            "applicationData.stageId",
            null
           ]
          },
          {
           "operator": "objectProperties",
           "children": [
            "applicationData.reviewData.levelNumber",
            null
           ]
          },
          "approvedAssignedQuestionsCount"
         ]
        },
        {
         "operator": "graphQL",
         "children": [
          "query reviewableQuestionsCount($appId: Int!) {\n  reviewableQuestionsCount(\n    appId: $appId\n  )\n}",
          "graphqlendpoint",
          [
           "appId"
          ],
          {
           "operator": "objectProperties",
           "children": [
            "applicationData.applicationId",
            null
           ]
          },
          "reviewableQuestionsCount"
         ]
        }
       ]
      }
     ]
    },
    {
     "operator": "=",
     "children": [
      {
       "operator": "objectProperties",
[...]

PR #921, opened 2022-10-19
#32T check for fully approved review

nmadruga:
Please check more info in: [Front-end PR](https://github.com/openmsupply/conforma-web-app/pull/1391)

-- Replacement Branch for #915
Based on PR (Refactoring for `migrationData`) #914 

### Changes

Adds function to be used in comparison to check if Review (going to be submitted) has all APPROVED/AGREED elements that should be for All Reviews - even if there are other ones involved - for elements which are **reviewable**...

The Condition on IncrementStage now is a bit complex, but I couldn't  figure out how to make a simpler function that could check one against the other. If anyone knows, please add to here, and then we can have a single-graphQL call to check : )

IncrementStage Condition:
```json
{
 "operator": "AND",
 "children": [
  {
   "operator": "OR",
   "children": [
    {
     "operator": "AND",
     "children": [
      {
       "operator": "=",
       "children": [
        {
         "operator": "objectProperties",
         "children": [
          "applicationData.reviewData.latestDecision.decision"
         ]
        },
        "CONFORM"
       ]
      },
      {
       "operator": "=",
       "children": [
        {
         "operator": "graphQL",
         "children": [
          "query approvedAssignedQuestionsCount(\n  $appId: Int!\n  $stageId: Int!\n  $levelNumber: Int!\n) {\n  approvedAssignedQuestionsCount(\n    appId: $appId\n    stageId: $stageId\n    levelNumber: $levelNumber\n  )\n}",
          "graphqlendpoint",
          [
           "appId",
           "stageId",
           "levelNumber"
          ],
          {
           "operator": "objectProperties",
           "children": [
            "applicationData.applicationId",
            null
           ]
          },
          {
           "operator": "objectProperties",
           "children": [
            "applicationData.stageId",
            null
           ]
          },
          {
           "operator": "objectProperties",
           "children": [
            "applicationData.reviewData.levelNumber",
            null
           ]
          },
          "approvedAssignedQuestionsCount"
         ]
        },
        {
         "operator": "graphQL",
         "children": [
          "query reviewableQuestionsCount($appId: Int!) {\n  reviewableQuestionsCount(\n    appId: $appId\n  )\n}",
          "graphqlendpoint",
          [
           "appId"
          ],
          {
           "operator": "objectProperties",
           "children": [
            "applicationData.applicationId",
            null
           ]
          },
          "reviewableQuestionsCount"
         ]
        }
       ]
      }
     ]
    },
    {
     "operator": "=",
     "children": [
      {
       "operator": "objectProperties",
       "children": [
        "applicationData.reviewData.latestDecision.decision"
       ]
      },
      "NON_CONFORM"
     ]
    }
   ]
  },
  {
   "operator": "objectProperties",
   "children": [
    "applicationData.reviewData.isLastLevel"
   ]
  }
 ]
}
```


database/buildSchema/50_views_functions_triggers.sql
```diff
@@ -1126,6 +1126,21 @@ $$
...
+        COUNT(*)
+    FROM
+        assigned_questions (app_id, stage_id, level_number) aq
+WHERE
+    aq.decision = 'APPROVE'
+    OR aq.decision = 'AGREE'
+$$
+LANGUAGE sql
+STABLE;
+
```
CarlosNZ, line 1143: Is this function used anywhere? It's not in this PR?

CarlosNZ (commented):
The function seems fine, but I'm not seeing where it's used at all. 🤷‍♂️

database/buildSchema/50_views_functions_triggers.sql
nmadruga: Yes, it's used on the Front-end PR & on the Configuration I added to the Template action `IncrementStage`

CarlosNZ: Oh, so there's a function that is *only* called by a config query? 

I'm not sure if that's a good idea or not. Maybe put a comment mentioning that, otherwise someone like me is going to come along in a few months and go "this function's not being used anywhere, let's delete it"!

nmadruga: No problem - Adding a comment now

CarlosNZ (commented):
Hey, just been having a closer look at this, and I don't think we actually need to make a new function for this. You can just query it using GraphQL, and I don't think it makes the evaluator expression any more compicated than you have it currently.

This would be the query that gets the same result as this function:
```
query MyQuery {
  assignedQuestions(
    appId: 10
    stageId: 10
    levelNumber: 10
    filter: {
      or: [{ decision: { equalTo: APPROVE } }, { decision: { equalTo: AGREE } }]
    }
  ) {
    totalCount
  }
}
```

(Obviously with variables instead of those hard-coded values). And you just return the `totalCount`.

I think if we can avoid making SQL functions that only exist for configs, that would be better.

CarlosNZ:
I also just have one question about the query expression above. I sketched it out so I could understand it better:
IMG_2513

I'm just wondering about the circled bit on the left -- why do we increment the stage if the result is NON_CONFORM?

CarlosNZ:
> I'm just wondering about the circled bit on the left -- why do we increment the stage if the result is NON_CONFORM?

I see this actually different to what you've actually got in the config in your latest snapshot -- so perhaps you've already fixed this?

nmadruga:
> I'm just wondering about the circled bit on the left -- why do we increment the stage if the result is NON_CONFORM?

Ok - I see your point. I'll try that, including on the test snapshot and if all good I'll add to the latest angola snapshot. 

As I mentioned on the Back-end PR we should keep this "code" somewhere to re-use considering this is core to the system we need to make sure the IncrementStage action for ANY template with Consolidation has that, and also anywhere else that has last-level Review stage that can be separated to different reviewers

nmadruga:
> > I'm just wondering about the circled bit on the left -- why do we increment the stage if the result is NON_CONFORM?
> 
> I see this actually different to what you've actually got in the config in your latest snapshot -- so perhaps you've already fixed this?

I'll also have a look on that, seems strange

nmadruga:
> > > I'm just wondering about the circled bit on the left -- why do we increment the stage if the result is NON_CONFORM?
> > 
> > 
> > I see this actually different to what you've actually got in the config in your latest snapshot -- so perhaps you've already fixed this?
> 
> I'll also have a look on that, seems strange

Ok. Have removed this part. I'm still not sure why we had that - I think we used to allow for you to move in to the next Stage without sending a Rejection to the applicant. But for this example - and perhaps for Core_actions we don't want to persist.

I have now added a new issue to update that on the Core_templates snapshot #929 

And agree that this can be done with the suggested graphql Query. Thanks!

PR #924, opened 2022-10-21
1389B Filter tweaks and Documentation

CarlosNZ:
Back-end branch for front-end https://github.com/openmsupply/conforma-web-app/pull/1397

Also includes Reference Docs for Data View Filters!

Items from issue addressed here:

- [x] Allow a "default sort column" to be defined (currently just used ID, which should remain the default)
- [x] Filter definitions column defaults tweak: https://github.com/openmsupply/conforma-server/issues/902 (Also update table view colums to NOT return filterData by default)
- [x] "Can be blank" option for filter lists (filter parameters) (the `includeNull` parameter)
- [x] Documentation (back-end)


src/components/data_display/generateFilterDataFields/generateFilterDataFields.ts
```diff
@@ -138,7 +138,7 @@ export const generateFilterDataFields = async (table: string, fullUpdate: boolea
               // TO-DO: Need to pass Auth headers to evaluator API calls
               graphQLConnection: { fetch, endpoint: graphQLEndpoint },
             })
-            patch[camelCase(column)] = evaluatedResult
+            patch[camelCase(column)] = evaluatedResult === '' ? null : evaluatedResult
```
CarlosNZ, line 141: This is a bit hacky, but it's much better in most cases if empty strings get stored as `null`. There may be a situation where we *don't* want this, but I can't think of one, so this is probably simplest for now.

src/components/data_display/helpers.ts
```diff
@@ -92,7 +97,13 @@ export const buildAllColumnDefinitions = async ({
...
 
   const filterColumns: string[] =
-    type === 'TABLE' ? buildColumnList(dataView, fieldNames, 'FILTER') : []
+    type === 'TABLE'
+      ? buildColumnList(dataView, fieldNames, 'FILTER').map((column: string) =>
+          fieldNames.includes(`${column}${FILTER_COLUMN_SUFFIX}`)
+            ? `${column}${FILTER_COLUMN_SUFFIX}`
+            : column
+        )
+      : []
```
CarlosNZ, line 106: Replaces normal columns with their "filterData" replacement for filter column list.

src/components/data_display/helpers.ts
```diff
@@ -203,7 +215,12 @@ const buildColumnList = (
       ? 'tableViewExcludeColumns'
       : 'filterExcludeColumns'
 
-  const includeColumns = getIncludeColumns(includeField, dataView, allColumns)
+  const includeColumns = getIncludeColumns(
+    includeField,
+    dataView,
+    // Don't include "FilterData" columns in full list
+    allColumns.filter((col) => !new RegExp(FILTER_COLUMN_SUFFIX).test(col))
+  )
```
CarlosNZ, line 223: The opposite of above -- *don't* include "filterData" columns in **table** view column list by default (they can still be explicitly included, though, if this is desired)

src/components/data_display/routes.ts
```diff
@@ -204,7 +210,11 @@ const routeDataViewFilterList = async (request: any, reply: any) => {
       values.forEach((value) => {
         if (delimiter && typeof value === 'string') {
           const splitString = value.split(delimiter).map((e) => e.trim())
-          splitString.forEach((string) => filterList.add(string))
+          splitString.forEach((substring) => {
+            // Once split, we need to exclude substrings that don't match the
+            // search text
+            if (new RegExp(searchText, 'i').test(substring)) filterList.add(substring)
+          })
```
CarlosNZ, line 217: Small bug I picked up in the options list results.

src/components/data_display/routes.ts
```diff
@@ -231,7 +241,11 @@ const routeDataViewFilterList = async (request: any, reply: any) => {
 
   if (results.length > filterListMaxLength) moreResultsAvailable = true
 
-  return reply.send({ list: results.slice(0, filterListMaxLength), moreResultsAvailable })
+  const list = results.slice(0, filterListMaxLength)
+
+  if (includeNull) list.push(null)
```
CarlosNZ, line 246: See docs for explanation of why `includeNull` might be useful sometimes.

documentation/Data-View-Filters.md
```diff
@@ -0,0 +1,383 @@
...
+```
+
+## Filters
+
+In the front end, there are several diffent UI elements for filtering table data. Which one is shown for a given data value is determined by a combination of the data type and explicit filter definitions specified in `data_view_column_definition`.
+
+### Handling complex data structures
+
+One of the main challenges with trying to create filters is that much of the data stored in tables is not easily filterable. Many fields store the entire "response" object that came from an application question, which could be any complex shape. For example, a `genericNames` field that has captured responses from a "Search" element might have data stored like so:
+
```
CarlosNZ, line 71: This is quite technical, and I'd planned to put it *after* the filters section, but it became too hard to talk about much of the filters detail without an explanation of what "filterData" columns are for.

src/components/actions/evaluatorFunctions.ts
```diff
@@ -46,10 +46,21 @@ const getISODate = (date?: Date) =>
...
-export const extractNumber = (input: string) => {
+const extractNumber = (input: string) => {
   const numberMatch = input.match(/(-?(\d+\.\d+))|(-?((?<!\.)\.\d+))|(-?\d+)/gm)
   if (!numberMatch) return 0
   return Number(numberMatch[0])
 }
 
-export default { generateExpiry, getYear, getFormattedDate, getJSDate, getISODate, extractNumber }
+// Remove diacritics (accented characters) from strings
+const removeAccents = (input: string) => input.normalize('NFD').replace(/[\u0300-\u036f]/g, '')
```
CarlosNZ, line 56: Not critical, but we'll need this if we want to implement https://github.com/openmsupply/conforma-web-app/issues/1388

(Will still need config in filter_definitions, but this will be needed too)

nmadruga:
Ok - Considering this is already on a few builds. We have tested it, now just missing a review so we can merge into `develop` and soon to `main` so we can release **v0.4.6**

documentation/Data-View-Filters.md
nmadruga: The descriptions are very good and succinct, so looks good for SystemAdmin users - who this Docs should be used by

documentation/Data-View-Filters.md
```diff
@@ -0,0 +1,385 @@
...
+}
+```
+
+Clearly this is not filterable in any meaningful way with simple database queries. So we've provided a mechanism by which complex data values can be converted into more easily parse-able objects, such as plain text or numbers.
+
+We can define "filter data" columns, which are defined like so:
+
+- an extra column definition in `data_view_column_definitions`. It's just the same as other definitions, but the `column_name` field should have the suffix `FilterData` to be recognised by the system as a filter data column.
+- there are two fields in `data_view_column_definitions` that are of relevance:
+  - `filter_expression`: an [evaluator](Query-Syntax.md) expression defining the value for the filter data column. It's essentially the same as the [`value_expression` column](Data-View.md#data_view_column_definition-table) in that it takes the current record as the evaluator "objects" parameter.
```
nmadruga, line 101: The link to `data_view_column_definition-table` seems broken

documentation/Data-View-Filters.md
```diff
@@ -0,0 +1,385 @@
...
+  ]
+}
+```
+
+Which would return this output based on the original response data: `MedPack, General Manufacturers`
+
+That value is (automatically) stored in a field called "manufacturer_list_filter_data" on the "product" table, which is much more suitable for filtering.
+
+Note that it is entirely possible to create *multiple* filter data columns for a single original data column if there's several data elements in it that you wish to filter by. For example, for an ingredients list, you might want a "names" and a "types" filter data column from a single response object.
+
```
nmadruga, line 173: Maybe an example?

documentation/Data-View-Filters.md
nmadruga: At least I couldn't relate to the place that this link took me. And on Data-View.md the TOC is not taking anywhere in that page for these two subtitles:
Screen Shot 2022-11-03 at 12 34 30 PM

documentation/Data-View-Filters.md
```diff
@@ -0,0 +1,385 @@
...
+- `table`: the name of the database table to update filter_data for. If not provided, it will update *all* filter data in the system
+- `fullUpdate` (`true`/`false`): if `true`, the script will re-compute (i.e. will re-evaluate every record) the filter_data for all existing filter data as well as generating values for missing data. If `false`, it will only update records where all filter data columns are `null` (i.e. these should be *new* records). By default, `fullUpdate` is `false` when a `table` is specified (this is what `modifyRecord` uses), but `true` when no table is specified (i.e. all tables).
+
+
+### Filter columns
+
+The columns that will appear as available filters for a given data view are defined in `filter_include_columns` and `filter_exclude_columns`. These work the same way as the table view and detail view [include/exclude lists](Data-View.md#data_view-table), but with the following differences:
+
+- if nothing is specified (i.e. `null` value), the defaults will be the same as the table view columns that are being returned. This is because, by default, we expect the available filters to be for the same data as the table columns.
```
nmadruga, line 206: Maybe specify the "columns" for completeness

documentation/Data-View-Filters.md
```diff
@@ -0,0 +1,385 @@
...
+- Similarly, the special value `...` matches all *table* columns, not the full set of columns. So it becomes fairly straightfoward to specify a filter set that closely matches the table columns but with, for example, one extra column and one exclusion. It's a good idea to explicitly exclude columns for which the data can't be processed by one of the standard filters -- however, using mapped "filter data" columns (see [above](#handling-complex-data-structures)) it's possible to get meaningful, filter-able content from almost any type of data.
```
nmadruga, line 207: "matches all *table* columns, not the full set of columns." Not entirely sure on what "full set of columns" means

nmadruga: Sorry to be pick, but this text seems a bit long, I would suggest changing this phrase: "So it becomes fairly straightfoward to specify a filter set that closely matches the table columns but with, for example, one extra column and one exclusion."

 to something more explanatory or an example to how to use this special value

documentation/Data-View-Filters.md
```diff
@@ -0,0 +1,385 @@
...
+- `includeNull`: (boolean, default `false`). Sometimes data view columns may contain `null` values. These *will* show up in the filter list (they appear as (Blank)), but only if `null` happens to appear in the top 10 options results (due to 10-item maximum). And typing anything in the Search field will obviously immediately exclude `null`, so it's possible to never see the `null` option in the options list. But if the `includeNull` parameter is set to `true`, then the options list returned from the server will *always* include the `null` "(Blank)" value.
+- `showFilterList`: (boolean, default `true`) as mentioned above, this Options list filter is the default filter for any `string` data. However, if you wish to provide a free-text search instead (often more useful when every item has a different value), set `showFilterList` to `false` and a ["Text search" filter](#text-search) will be displayed instead.
+
+##### API
+
+The options list filter makes its own request to the server to get the list of available options:
+```
+/data-views/<dataViewCode>/filterList/<column>
```
nmadruga, line 245: Hmmm just wondering why not using the same pattern for URL i.e:
`/data-views/<dataViewCode>/filter-list/<column>`

nmadruga (changes requested):
A little unsure about the images showing Pt text, since that's not the most used language for anyone using the App and sometimes it's mixed with English, I think it would be preferable to replace with English version ones. Sorry to point that, but that's the only problem I found with this.

The graphql types also seems needs to be regenerated using `core_templates` snapshot

documentation/Data-View-Filters.md
CarlosNZ: Seems to be working okay for me. Where are you viewing it?

If you look as a Github doc, it seems to work: https://github.com/openmsupply/conforma-server/blob/1389B-filter-tweaks/documentation/Data-View-Filters.md

(And those two TOC links seem okay too: https://github.com/openmsupply/conforma-server/blob/1389B-filter-tweaks/documentation/Data-View.md)

documentation/Data-View-Filters.md
CarlosNZ: > **For example**, for an ingredients list, you might want a "names" and a "types" filter data column from a single response object.

Is that not an example? ;) (sorry, i didn't want to actually make one since we don't currently have a use case for it -- just pointing out that it *can* be done if we need it)

documentation/Data-View-Filters.md
CarlosNZ: Not sure what you mean here -- which "columns" are you referring to?

documentation/Data-View-Filters.md
CarlosNZ: I've already explained how `...` works in the Data View documentation -- this just clarifies how it's subtly different.

documentation/Data-View-Filters.md
CarlosNZ: Not sure what URL you're referring to there -- what you wrote is exactly what I have. 🤷‍♂️

CarlosNZ:
Thanks for having a proper look, @nmadruga. A little confused by a couple of comments -- maybe just clarify with me in person what you mean?

> A little unsure about the images showing Pt text, since that's not the most used language for anyone using the App and sometimes it's mixed with English, I think it would be preferable to replace with English version ones. Sorry to point that, but that's the only problem I found with this.

Come on, Nicole have some pride in your mother tongue! 🇧🇷🇧🇷🇧🇷🇧🇷 😄

For public docs, sure, but this is just a dev reference. I made these filters specifically for the Angola config, so I'm reluctant to spend a long time remaking English versions just for some screenshots. Perhaps I can Photoshop the current ones if it's really a problem, but I wouldn't have thought it's a high priority.

> The graphql types also seems needs to be regenerated using `core_templates` snapshot

Yup, I'll do a full re-gen once we merge these.

documentation/Data-View-Filters.md
nmadruga: I was suggesting to change the logic so the URL shows static text in kebab-case

documentation/Data-View-Filters.md
nmadruga: When you say "will be the same as the table view columns" I though you meant a data_base column, so the reason why 'Specify', but I guess this is about the columns displayed to the user on the DataView page?

documentation/Data-View-Filters.md
nmadruga: I think this is not part of your PR, so don't feel like you have to fix it. But on the page you just linked: https://github.com/openmsupply/conforma-server/blob/1389B-filter-tweaks/documentation/Data-View.md

If you click those subheaders it doesn't take you anywhere, try cicking on the next one to see how it actually takes you to the Subheader selected.

nmadruga:
> Come on, Nicole have some pride in your mother tongue! 🇧🇷🇧🇷🇧🇷🇧🇷 😄
> 
> For public docs, sure, but this is just a dev reference. I made these filters specifically for the Angola config, so I'm reluctant to spend a long time remaking English versions just for some screenshots. Perhaps I can Photoshop the current ones if it's really a problem, but I wouldn't have thought it's a high priority.

Como assim, eu nunca ouvi falar em Português!

I was thinking the example you have for testing filters could do, but you're right. That's the one in Pt. I think it's ok for now, but hopefully we don't get complains about using private data - which I know it isn't since it's all data generated by tests!

documentation/Data-View-Filters.md
CarlosNZ: Ah, gotcha. Weird that the TOC links weren't auto-updated. Anyway, done that now: 1bfeaee9

(I think the one on the Filter page should be fine now too, if it wasn't already)

documentation/Data-View-Filters.md
CarlosNZ: Okay, maybe I can make it a bit clearer. I've added an example of the kind of data value that you might want to make multiple filterData columns from: 3a0dd2bf

Hopefully that clarifies it. (I don't think it needs the actual column definitions spelled out, as that's already explained earlier -- you'd just to the same thing twice.)

documentation/Data-View-Filters.md
CarlosNZ: Yeah, as I've mentioned elsewhere, when I talk about "fields", I mean data base columns. When I talk about "columns" I mean data view columns. Usually these are the same, but the distinction is important when they are different. I've tried to be consistent about this, but please point out where I may have slipped.

documentation/Data-View-Filters.md
CarlosNZ: Yeah, I think it's clear when taken in the context of what came before it. I'm open to explaining it better, but I just need to know specifically which part is not clear. (Tell me in person if you'd like further clarification).

documentation/Data-View-Filters.md
CarlosNZ: Yeah, in the server endpoints, there are quite a few camelCase ones, so probably best to leave as is for now. As long as the user-facing ones are consistent, I'm happy with it. (I'd like to do a bigger tidy-up, but don't want to risk breaking things for the sake of a little cosmetic improvement, but I agree that consistency is better)

CarlosNZ:
> but hopefully we don't get complains about using private data

I think everything you see is just from fake companies/names and lookup tables.

nmadruga (approved):

nmadruga:
Looks great, Carl!

PR #928, opened 2022-10-26
#930 bug fix -- cleanup files action

### LINKED ISSUE #930: Bug: Documents in internal documents tab not persisting after new document is generated
The Carbone documents (Check AIM template) and the emails are sent, but after another one is send - it seems to be when it's send on the same stage - the previous document is removed from the tab

CarlosNZ:
Fix #930

Pretty simple fix -- "cleanupFiles" action was matching too many files -- we haven't updated it since adding a bunch more fields to the "file" table, and it was matching some of the output doc files too.

Surprised it's taken us this long to notice to be honest -- as this problem would have existed ever since the "output docs" field was created, which is a while back now 🥴


plugins/action_files_cleanup/src/databaseMethods.ts
```diff
@@ -9,6 +9,10 @@ const databaseMethods = (DBConnect: any) => ({
     FROM file
     WHERE application_serial = $1
     AND submitted = false
+    AND application_response_id IS NOT NULL
+    AND is_output_doc = false
+    AND is_internal_reference_doc = false
+    AND is_external_reference_doc = false
```
CarlosNZ, line 15: I think it would have been enough to just add the `AND application_response_id IS NOT NULL` line, but no harm in being even more rigorous.

plugins/action_files_cleanup/src/cleanupFiles.ts
```diff
@@ -36,7 +35,7 @@ async function cleanupFiles({ parameters, applicationData, DBConnect }: ActionPl
 
   try {
     // Get all unsubmitted files for current application
-    const files = await db.getApplicationFiles(applicationSerial)
+    const files = await db.getApplicationResponseFiles(applicationSerial)
```
CarlosNZ, line 38: Changed method name to more accurately reflect what it's doing.

plugins/action_files_cleanup/src/cleanupFiles.ts
```diff
@@ -1,4 +1,3 @@
-import fs from 'fs'
```
CarlosNZ, line 1: Leftover from sometime in the distant past.

nmadruga (approved):

PR #933, opened 2022-10-27
#932 fix crashing on history panel

### LINKED ISSUE #932: Bug: Option to ViewReviewHistory showing to applicant and crashing App
_Moved from https://github.com/openmsupply/conforma-templates/issues/51_

The button was showing on the Summary page - when opening there is only a blank page
Please look at the video for example

This is how the crash looks to reviewers: https://drive.google.com/file/d/1sdoXybhIUh5_BOdJyuWDLQS26b6p8Wmo/view?usp=sharing

CarlosNZ:
Fix #932 

This also fixes the problem @nmadruga was having loading my snapshot for https://github.com/openmsupply/conforma-web-app/pull/1386!

Turns out that (somehow) the auto-generated values for the "template_code" and "template_version" on the "template_element" table were missing (had `null` values), and also the "unique" constraint for `template_code, code, template_version` was not being re-generated, and this constraint was required as part of the `templateElementByTemplateCodeAndCodeAndTemplateVersion` query the HistoryPanel uses

This change just drops and recreates the generated columns, and ensures the uniqueness constraint is re-added every time.

In order to test -- download `angola_2022_09_27_1616_server` from the Angola SERVER (50000), load it, and go to the review shown in Jess's screen recording: http://localhost:3000/application/AIM-2022-0084/review/21?activeSections=all (You need to be logged in as 'tecnico2' to see it)

### DIFF
--- a/database/buildSchema/43_views_functions_triggers.sql
+++ b/database/buildSchema/43_views_functions_triggers.sql
@@ -277,12 +277,15 @@ IMMUTABLE;
 -- Add columns (and index) to template_element table that use the above 2
 -- functions
 ALTER TABLE public.template_element
+    DROP COLUMN IF EXISTS template_code,
     ADD COLUMN IF NOT EXISTS template_code varchar GENERATED ALWAYS AS (public.get_template_code (section_id)) STORED;
 
 ALTER TABLE public.template_element
+    DROP COLUMN IF EXISTS template_version,
     ADD COLUMN IF NOT EXISTS template_version integer GENERATED ALWAYS AS (public.get_template_version (section_id)) STORED;
 
-CREATE UNIQUE INDEX IF NOT EXISTS template_element_template_code_code_template_version_idx ON public.template_element (template_code, code, template_version);
+ALTER TABLE public.template_element
+    ADD UNIQUE (template_code, code, template_version);
 
 -- APPLICATION
 --FUNCTION to update `is_active` to false


database/buildSchema/43_views_functions_triggers.sql
```diff
@@ -277,12 +277,15 @@ IMMUTABLE;
...
+    DROP COLUMN IF EXISTS template_code,
     ADD COLUMN IF NOT EXISTS template_code varchar GENERATED ALWAYS AS (public.get_template_code (section_id)) STORED;
 
 ALTER TABLE public.template_element
+    DROP COLUMN IF EXISTS template_version,
     ADD COLUMN IF NOT EXISTS template_version integer GENERATED ALWAYS AS (public.get_template_version (section_id)) STORED;
 
-CREATE UNIQUE INDEX IF NOT EXISTS template_element_template_code_code_template_version_idx ON public.template_element (template_code, code, template_version);
+ALTER TABLE public.template_element
+    ADD UNIQUE (template_code, code, template_version);
```
CarlosNZ, line 288: This is closer to how the table was originally defined. I'm not entirely sure what the difference between a `UNIQUE INDEX` and a `UNIQUE` constraint is (since the constraint also generates an index), but they get called slightly different things `_key` vs `_idx` -- the the latter is what is required to make the GraphQL connection `templateElementByTemplateCodeAndCodeAndTemplateVersion` available.

nmadruga:
> Turns out that (somehow) the auto-generated values for the "template_code" and "template_version" on the "template_element" table were missing (had `null` values)

Oh! I noticed that too... Just didn't realised it would be related. Good that you found this problem quickly

nmadruga (approved):
Cool. I can see this fixes the crash - but probably something to look further into PR for issue https://github.com/openmsupply/conforma-web-app/issues/1384 is why the Review_responses for the Reviewer aren't displayed in the ViewHistory modal...

Check the example below, it's only displaying the Applicant's responses to the question:
 
Screen Shot 2022-10-28 at 11 35 46 AM

PR #934, opened 2022-10-28
Fix for missing generated column data

CarlosNZ:
Okay, so this fixes the problem, but I'm still somewhat mystified as to why the "review" and "template_element" generated columns need to be recreated like this but other ones don't (like the one in "review_assignment", and "application_status_history") -- they seem to work exactly the same way.

### DIFF
--- a/database/buildSchema/27_review_assignment.sql
+++ b/database/buildSchema/27_review_assignment.sql
@@ -1,19 +1,5 @@
 -- review assignment
--- FUNCTION to auto-add template_id to review_assignment
-CREATE OR REPLACE FUNCTION public.review_assignment_template_id (application_id int)
-    RETURNS int
-    AS $$
-    SELECT
-        template_id
-    FROM
-        application
-    WHERE
-        id = $1;
-
-$$
-LANGUAGE SQL
-IMMUTABLE;
-
+--
 -- ENUM
 CREATE TYPE public.review_assignment_status AS ENUM (
     'AVAILABLE',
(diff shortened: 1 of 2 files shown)


database/buildSchema/27_review_assignment.sql
```diff
@@ -1,19 +1,5 @@
...
-        template_id
-    FROM
-        application
-    WHERE
-        id = $1;
-
-$$
-LANGUAGE SQL
-IMMUTABLE;
-
```
CarlosNZ, line 16: Matching the pattern used everywhere else to *only* define the function in the main "functions" script

database/buildSchema/43_views_functions_triggers.sql
```diff
@@ -789,6 +789,14 @@ IMMUTABLE;
...
 -- Add columns to "review" now that these functions are defined:
 ALTER TABLE public.review
+    DROP COLUMN IF EXISTS application_id CASCADE,
+    DROP COLUMN IF EXISTS reviewer_id CASCADE,
+    DROP COLUMN IF EXISTS level_number CASCADE,
+    DROP COLUMN IF EXISTS stage_number CASCADE,
+    DROP COLUMN IF EXISTS time_stage_created CASCADE,
+    DROP COLUMN IF EXISTS is_last_level CASCADE,
+    DROP COLUMN IF EXISTS is_last_stage CASCADE,
+    DROP COLUMN IF EXISTS is_final_decision CASCADE,
```
CarlosNZ, line 799: So weird how this is required here, but not for the "template_id" generated column in `review_assignment` 🤷‍♂️

nmadruga (approved):

Issue #816, opened 2022-08-15
Allow management access to Lookup tables (and Localisation) without full "admin" access


CarlosNZ:
Improving Lookup-tables issue 1: Allow non-admins to edit Lookup tables 

Currently, template builder, snapshots, lookup tables and localisations require "admin" access and are enforced by back-end by being within the "admin" sub-route (which checks for "isAdmin" flag in JWT).

So it's not currently possible to allow access to one of these without granting access to all. Ideally, the lookup tables and localisations control would be considered "management" tasks, so would be accessible to anyone with the "systemManager" permission.

To do this change properly would be a bit of work. Rather than just have a standard "isAdmin" flag marked in the JWT, we'd need to be able to assign specific permissions to specific endpoints, and have additional checking for this in the back-end. 

So we'd probably need to do away with the "admin" sub-route (which checks for "isAdmin" in JWT), and just all endpoints under the main endpoint route ("/api"), and add configurable permissions per endpoint.

So not trivial, but also not a huge amount of work. I'm thinking probably a couple of days, including relevant changes to front-end?

nmadruga:
Good to also remove Admin permission from GrantUserPermission template

CarlosNZ:
Just putting a reminder here -- I've made an additional change to data views to handle a dummy "lookupTables" permission name for matching lookup tables for data views. Just flagging this so we remember to consider it when doing this issue: https://github.com/openmsupply/conforma-server/pull/935#issue-1428733986

nmadruga:
The list of issues to improve Lookup tables includes:

Issue 1: **current**
Issue 2: https://github.com/openmsupply/conforma-web-app/issues/1048
Issue 3: https://github.com/openmsupply/conforma-web-app/issues/1467 - removing the need for exporting first to match IDs

CarlosNZ:
> I'm thinking probably a couple of days, including relevant changes to front-end?

More like an hour and a half 😜 #1040

PR #935, opened 2022-10-30
#1412B Lookup table Dataview display

CarlosNZ:
Back-end tweaks required for front-end issue: https://github.com/openmsupply/conforma-web-app/issues/1412

Adds another column to "data_table" to connect a "data_view" definition to a lookup table.

Also handles the "special" permission names "lookupTables" by inserting data views matching it into the list of "allowed" data views if the auth token contains `isAdmin` (which is what is required for viewing lookup tables, currently)

### DIFF
--- a/src/components/data_display/routes.ts
+++ b/src/components/data_display/routes.ts
@@ -18,6 +18,9 @@ import { ColumnDefinition, LinkedApplication, DataViewsResponse } from './types'
 import { DataView } from '../../generated/graphql'
 import config from '../../config'
 
+// CONSTANTS
+const LOOKUP_TABLE_PERMISSION_NAME = 'lookupTables'
+
 const routeDataViews = async (request: any, reply: any) => {
   const { permissionNames } = await getPermissionNamesFromJWT(request)
   const dataViews = await DBConnect.getAllowedDataViews(permissionNames)
@@ -37,6 +40,7 @@ const routeDataViewTable = async (request: any, reply: any) => {
   const authHeaders = request?.headers?.authorization
   const dataViewCode = camelCase(request.params.dataViewCode)
   const { userId, orgId, permissionNames } = await getPermissionNamesFromJWT(request)
+  if (request.auth.isAdmin) permissionNames.push(LOOKUP_TABLE_PERMISSION_NAME)
   const query = objectKeysToCamelCase(request.query)
   const filter = request.body ?? {}
 
@@ -97,6 +101,7 @@ const routeDataViewDetail = async (request: any, reply: any) => {
   const dataViewCode = camelCase(request.params.dataViewCode)
   const recordId = Number(request.params.id)
   const { userId, orgId, permissionNames } = await getPermissionNamesFromJWT(request)
+  if (request.auth.isAdmin) permissionNames.push(LOOKUP_TABLE_PERMISSION_NAME)
 
   const {
     tableName,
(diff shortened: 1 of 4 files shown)


src/components/data_display/routes.ts
```diff
@@ -97,6 +101,7 @@ const routeDataViewDetail = async (request: any, reply: any) => {
   const dataViewCode = camelCase(request.params.dataViewCode)
   const recordId = Number(request.params.id)
   const { userId, orgId, permissionNames } = await getPermissionNamesFromJWT(request)
+  if (request.auth.isAdmin) permissionNames.push(LOOKUP_TABLE_PERMISSION_NAME)
```
CarlosNZ, line 104: Note that we only add this for the "table", "detail" and "filter list" routes, not the "list" route, as we don't want lookup table definitions appearing in the "Database" menu.

nmadruga:
> Also handles the "special" permission names "lookupTables" by inserting data views matching it into the list of "allowed" data views if the auth token contains `isAdmin` (which is what is required for viewing lookup tables, currently)

Hmmm this part should be flagged for when we plan to implement issue #816

CarlosNZ:
> Hmmm this part should be flagged for when we plan to implement issue #816

Yes, I thought about that. Still not sure what the best solution to that issue is, but yeah good to remember this. Have made a reminder on that issues: https://github.com/openmsupply/conforma-server/issues/816#issuecomment-1301533960

src/components/data_display/routes.ts
nmadruga: I have no idea what do you mean... hahaha

But good that LookupTables don't show there.

nmadruga (approved):
LGTM

- I tried the filters, search, pagination
- Have successfully exported and imported LookupTable with extra record

Wondering - what happens when a new column is added?
And also great you kept the Import/Export on the same page!!!

CarlosNZ:
> Wondering - what happens when a new column is added?

Good question. It should be fine if your data view definition is "simple" (i.e. just uses defaults for most things), but that's a good example of where the `...` in column lists is handy -- as long as you have `...` any new columns will appear just fine. (It's particularly easy for lookup tables as all fields are plain text, so defaults work pretty well in most cases)

However -- can you actually re-import a lookup table with extra columns? I thought you could only add new rows. 🤷‍♂️

nmadruga:
> However -- can you actually re-import a lookup table with extra columns? I thought you could only add new rows. 🤷‍♂️

ok -- trying now -- that's actually a requirement for some of the Lookups we have

CarlosNZ:
> Tested and reported on Telegram - it looks good overall, the new filter is created, but something a little unexpected when using the filter options which didn't filter the new Column

Screenshot 2022-11-03 at 3 55 01 PM

⬆️ Filters working for me now.

I think you maybe needed to refresh your page. When the lookup table is imported, there is a "refetch" to refetch the data, but the filter definitions don't get changed, so you can get some odd behaviour until you refresh the page.

I don't think it's worth changing this right now. The filter definitions only refresh when the data-view code changes, which I think is reasonable as data views are expected to be static definitions. In order to change this I'd have to make a separate flag in LookupTable page and pass it down and watch it in order to re-load the dataview filter definitions, which I think would be quite clunky and messy for something that doesn't happen very often.

As for the slow loading of the filter list, that's because the filter list function keeps looping over and over until it's found 10 distinct items (or reached the end of the list), and this particular list is so sparse that it never finds ten unique elments (and it only fetches in batches of ten at a time -- hence slooow).

Also, having empty strings rather than null is quite crappy, so have made an issue to deal with this at some point: https://github.com/openmsupply/conforma-server/issues/949

nmadruga:
That's it them - All good to merge ❤️

PR #938, opened 2022-11-01
#807 Keep template files linked to active template version

### LINKED ISSUE #807: Bug: Problem keeping template files linked to new versions
It's easy to accidentally lose files linked to templates (e.g. .carbone templates) when exporting/importing templates.

When you export a template, the linked files (i.e. matching template_id) get exported with it. That's fine.

And when you import a template, if the linked file doesn't already exist in the system, it gets added. 

The problem is, when you import a template, if the linked file *does* already exist, it gets the id of the *previous* template version, and now the new version no longer has a linked file. And then if you delete the old version of the template, the file gets deleted, too since it's connected to the wrong version.

Possible solutions:
1. always import a new copy of the file regardless so every template version has a linked file. The problem is, we'll need to generate a new uniqueId for it, in which case the actions that refer to it will no longer work.
2. if the file exists, change the template_id to match the new imported template. The downside is, if we re-activate an old version of the template, it'll no longer be linked to the correct file. This is okay if the new version of the template stays in the system, as the file will still exist and it'll connect correctly with the uniqueID. However, as soon as the other version gets deleted, the file will get deleted too.

It seems like Option 2 is the better one, as creating a new uniqueId for a file that has actions/elements that refer to it is a really bad idea. We'll just need to be vigilant and if we ever want to delete a template while keeping an older version in the system, then we'll need to remember to relink the file before deleting the other version.

Any other suggestions?

CarlosNZ:
Fix #807.

Was able to just add a bit extra to an existing function, so that makes it simpler and no need for a whole new trigger or anything.

Seems to be working pretty well. To test:
- duplicate a template to create a new version
- set the new version to "AVAILABLE" (from DRAFT)
- notice all the files linked to that template have been updated to link the new version (template_id field)
- if you change which version is "AVAILABLE", the files should change their template_id again.


nmadruga:
I can't test this with current snapshots because they are all in versions ahead this one (even that I have merged `develop` in) I think it would be easier to wait until next pre-release using develop this time, to make this task easier

nmadruga (approved):
Looks great! 💟

PR #940, opened 2022-11-02
#70T-Fix-trigger-issue-after-unassignment

nmadruga:
Fixes https://github.com/openmsupply/conforma-templates/issues/70

Try checking using current snapshot loaded on demo instance 50004. Or use the latest on Fiji shared snapshot 

Problem was due to difference on returned (by `getMatchingReviewAssignments` Pg function) of snake_case and camelCase (expected). So now the only properties used are returned and converted to camelCase.

I still noticed that after Unassigning the page doesn't reload - so planning to add this fix to Front-end PR https://github.com/openmsupply/conforma-web-app/pull/1409

### DIFF
--- a/plugins/action_update_review_assignments_status/src/databaseMethods.ts
+++ b/plugins/action_update_review_assignments_status/src/databaseMethods.ts
@@ -1,3 +1,5 @@
+import { ReviewAssignment } from '../../../src/generated/graphql'
+
 const databaseMethods = (DBConnect: any) => ({
   getReviewAssignmentById: async (reviewAssignmentId: number) => {
     const text = `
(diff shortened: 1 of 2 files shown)


CarlosNZ:
> I still noticed that after Unassigning the page doesn't reload - so planning to add this fix to Front-end PR [openmsupply/conforma-web-app#1409](https://github.com/openmsupply/conforma-web-app/pull/1409)

Yes, I think that the trigger is not being waited for properly, which applies to assigning and unassigning. This will be critical once we've fixed [this] too -- the front-end *must* wait for the back-end triggers to complete before refetching --it should never rely on its own logic to deduce what sections are assigned to whom.

fergie-nz (approved):
Nice! Love your work

PR #942, opened 2022-11-02
#66T Remove port requirement from single-instance servers

CarlosNZ:
Fix #66T

Have already implemented this on Fiji server: https://fiji.conforma.nz

Just changing it here for our default single-instance setup for future installations.


nmadruga (approved):
👍 LGTM

PR #944, opened 2022-11-03
#54T Proper back-end enforcement of review_assignments

CarlosNZ:
Back-end part of issue [#54T](https://github.com/openmsupply/conforma-templates/issues/54)

I discovered something important here about trigger functions -- when you're updating the same record (using `UPDATE...`) that caused the trigger to fire, it's very easy to get into an infinite loop without a lot of extra qualifiers to prevent re-firing. This is a problem I had before when creating the `empty_assigned_sections` function (now gone!!) and was having here -- until I discovered that updating the record is the WRONG way to do it. What we should be doing is just mutating the `NEW` variable and _returning_ it, so the values get changed even *before* the first write, and then there's no chance of the "update" causing another trigger event etc etc...

With that in mind, I've also updated the `empty_assigned_sections` function and consolidated it into one "validation" function.

So what this does now, before the record even gets written is:

- remove `assigned_sections` that are already assigned to others (same stage and level)
- set `assigned_sections` to empty if status = "AVAILABLE" (this was already there, but now it's combined into one function, and done nicer.
- set `status` = "AVAILABLE" if `assigned_sections` are empty (the reverse of the above -- and needed for the case when _no_ new `assigned_sections` are valid)
- remove `assigned_sections` if they are not in `allowed_sections` (we weren't checking for this at all before, just relying on the front-end to enforce it)
- update `time_updated` to current time regardless (I presume this is okay, @nmadruga ?)

Then the regular trigger (the one that triggers actions) runs AFTER the record has been written, as normal, but it will have all the _correct_ review_assignment data to work with!

I've only done database changes here. There may be some downstream effects that I haven't considered --  for example, what happens to the actions in the case where there are no valid assignments? So have a think about possible repercussions in that area.

To test: just manually change stuff in the review_assignment table and check that everything automatically gets updated to correctly validated values.

(I'd like to have a look through elsewhere and see if we have any other overly complicated trigger logic that could be simplified with this "RETURN NEW" approach -- something to remember when we have time)


database/buildSchema/43_views_functions_triggers.sql
```diff
@@ -630,21 +630,64 @@ IMMUTABLE;
 ALTER TABLE review_assignment
     ADD COLUMN IF NOT EXISTS template_id INT GENERATED ALWAYS AS (review_assignment_template_id (application_id)) STORED;
 
-CREATE OR REPLACE FUNCTION public.empty_assigned_sections ()
+-- These no longer used as combined into (below) validation function
+DROP FUNCTION IF EXISTS public.empty_assigned_sections ();
+
+DROP TRIGGER IF EXISTS review_assignment_trigger2 ON public.review_assignment;
```
CarlosNZ, line 636: Important to DROP these, so we don't leave them active in existing databases!

database/buildSchema/43_views_functions_triggers.sql
```diff
@@ -630,21 +630,64 @@ IMMUTABLE;
...
+                            unnest(assigned_sections)
+                        FROM
+                            review_assignment
+                        WHERE
+                            stage_id = NEW.stage_id
+                            AND level_number = NEW.level_number
+                            AND reviewer_id <> NEW.reviewer_id)
+                        AND (new_assigned = ANY (NEW.allowed_sections)
+                            OR NEW.allowed_sections IS NULL));
+    END IF;
```
CarlosNZ, line 673: This is quite complicated and I'd love a simpler query if possible? Basically, it combines all other people's existing sections assignments into a single result set (rows), splits the NEW section assignments into rows, then removes any of the new ones that are already in the existing one, and removes any that are not in the "allowed" sections.

database/buildSchema/43_views_functions_triggers.sql
```diff
@@ -630,21 +630,64 @@ IMMUTABLE;
...
+    IF NEW.assigned_sections = '{}' THEN
+        NEW.status = 'AVAILABLE';
+    END IF;
+    NEW.time_updated = NOW();
+    RETURN NEW;
```
CarlosNZ, line 678: This is the magic improvement: `RETURN NEW;` So much better than using `UPDATE` on the record directly as it completely eliminates the chance of an infinite loop.

plugins/action_generate_review_assignment_records/src/generateReviewAssignments.ts
```diff
@@ -459,17 +459,22 @@ const getNewOrExistingAssignmentStatus = (
...
   sectionCodes: string[],
-  existingAssignment?: ExistingReviewAssignment,
+  existingAssignment?: ExistingReviewAssignment
 ): AssignmentState => {
   const isReviewAssigned = existingReviewsAssigned.length > 0
   const isAssigned = existingReviewsAssigned.some(
     ({ userId }) => userId === existingAssignment?.userId
   )
-  // temporarily final decision shouldn't be locked if there are other reviewAssignemt assigned
+  // temporarily final decision shouldn't be locked if there are other reviewAssignment assigned
```
CarlosNZ, line 468: Fixed one tiny typo in here, the rest is all auto-formatting (not sure why it was wrong previously)

nmadruga:
@CarlosNZ I'm finding this problem when running migration code:
Screen Shot 2022-11-07 at 2 34 03 PM

nmadruga (approved):
LGTM

PR #945, opened 2022-11-03
#805 combined

fergie-nz:


src/components/actions/scheduler.ts
```diff
@@ -2,10 +2,18 @@ import DBConnect from '../databaseConnect'
...
+import { filesFolder } from '../files/fileHandler'
+import path from 'path'
+import { getAppEntryPointDir } from '../utilityFunctions'
+import { crawlFileSystem } from '../utilityFunctions'
+import { deleteFile } from '../files/deleteFiles'
 import createBackup from '../exportAndImport/backup'
 
 // Dev config option
 const schedulerTestMode = false // Runs scheduler every 30 seconds
+const basePath: string = path.join(getAppEntryPointDir(), filesFolder) // declares the file storage path
```
CarlosNZ, line 15: This "basePath" is already defined in `./constants.ts` -- you can just import "FILES_FOLDER" instead and it includes the full path.

src/components/actions/scheduler.ts
```diff
@@ -53,15 +61,46 @@ export const triggerScheduledActions = async () => {
   DBConnect.triggerScheduledActions()
 }
 
-export const cleanUpPreviewFiles = async () => {
+const checkFile = async (filePath: string) => {
+  if (await DBConnect.checkIfInFileTable(filePath)) {
+    const subPath = filePath.slice(basePath.length)
+    const file: any = { filePath: subPath }
```
CarlosNZ, line 67: Try not to use `any`. This is a good example of why not. There was a weird error occurring when running the script which I couldn't easily figure out. But as soon as I changed this `any` to a specific type (`{filePath: string}`) it became immediately obvious what the problem was -- this is not the correct input type for the `deleteFile` function (it needed thumbnailPath and a couple of other things). I was able to modify the "deleteFile" function to handle different input, but it was so much easier to figure out what the problem was once Typescript alerted me to it.

src/components/postgresConnect.ts
```diff
@@ -360,6 +360,46 @@ class PostgresDB {
...
 
+  public checkIfInFileTable = async (subPath: string) => {
+    const text = `
+    SELECT id 
+    FROM file
+    WHERE file_path = $1
+    `
+    try {
+      const result = await this.query({ text, values: [subPath] })
+      return result.rows.length === 0
```
CarlosNZ, line 371: Should really be `return result.rows.length !== 0`. Otherwise it's backwards -- returns `false` when the file IS there and vice versa. So either return the opposite, or the name should be `checkIfNotInFileTable`. (I went with the former)

src/components/postgresConnect.ts
```diff
@@ -360,6 +360,46 @@ class PostgresDB {
...
+    SET is_missing = true
+    WHERE file_path = $1
+    `
+    try {
+      const result = await this.query({ text, values: [filePath] })
+      return result
+    } catch (err) {
+      throw err
+    }
+  }
```
CarlosNZ, line 401: I removed both these and found a more straightforward alternative (see #950)

CarlosNZ (commented):
Cool, thanks -- I've made a new PR with some suggested fixes and changes for you to have a look at and incorporate here when you're ready: #950 

(Don't do any more changes on this branch -- just make suggestions on mine if you think of anything else)

CarlosNZ:
Okay, I'm going to have one last test of this with the current Fiji snapshot and check there's no problems, then we can merge this.

CarlosNZ (approved):
Great -- I think this is all working very nicely now.

Let's get it merged!

Thanks for your patience.

PR #950, opened 2022-11-03
#805 Further improvements

CarlosNZ:
Final tweaks on #805 

Hey @fergie-nz I had a good look through and made some additional improvements to [your PR](https://github.com/openmsupply/conforma-server/pull/945). I've added them here separately so you can have a look through and see what was needed.

There was a few things not working right, and a few little extra tweaks needed here and there. But a lot of it is just tidying up the file structure (moving stuff out of "scheduler" that weren't to do with scheduling, etc.) See inline comments below for details.

Feel free to add your own suggestions, or question what I've done, and when you're happy we can merge this into yours and then merge it all to develop -- finally!


database/buildSchema/33_file.sql
```diff
@@ -12,7 +12,6 @@ CREATE TABLE public.file (
     is_output_doc boolean DEFAULT FALSE NOT NULL,
     is_internal_reference_doc boolean DEFAULT FALSE NOT NULL,
     is_external_reference_doc boolean DEFAULT FALSE NOT NULL,
-    is_missing boolean DEFAULT FALSE NOT NULL,
```
CarlosNZ, line 15: We don't need this any more -- I found a better way to process the missing records. I didn't go with the Cursor, as that would need to install a whole extra package [node-pg-cursor](https://www.npmjs.com/package/pg-cursor), but I achieved something similar using GraphQL query with pagination to limit the batch size.

src/components/actions/scheduler.ts
```diff
@@ -2,18 +2,11 @@ import DBConnect from '../databaseConnect'
 import scheduler from 'node-schedule'
 import config from '../../config'
 import { DateTime } from 'luxon'
-import fs from 'fs'
-import { filesFolder } from '../files/fileHandler'
-import path from 'path'
-import { getAppEntryPointDir } from '../utilityFunctions'
-import { crawlFileSystem } from '../utilityFunctions'
-import { deleteFile } from '../files/deleteFiles'
+import cleanUpFiles from '../files/cleanup'
```
CarlosNZ, line 5: Moved clean up script to its own file, just the scheduler launcher left here.

src/components/files/cleanup.ts
```diff
@@ -0,0 +1,79 @@
...
+    filePaths.forEach(({ id, filePath }) => {
+      if (!fs.existsSync(path.join(FILES_FOLDER, filePath))) fileIdsToBeDeleted.push(id)
+    })
+    offset += BATCH_SIZE
+    filePaths = await DBConnect.getFilePaths(BATCH_SIZE, offset)
+  }
+
+  await DBConnect.deleteMissingFileRecords(fileIdsToBeDeleted)
+  return fileIdsToBeDeleted.length
+}
```
CarlosNZ, line 42: Using GraphQL -- this grabs the file records 100 at a time, and keeps a list of the ids of records that should be deleted. Then deletes all those in a single query at the end.

src/components/files/cleanup.ts
```diff
@@ -0,0 +1,79 @@
...
+
+export const cleanUpFiles = async () => {
+  let filesMissingRecords = 0
+
+  // Check if file in database and delete if not
+  const checkFile = async (filePath: string) => {
+    if (path.dirname(filePath) === GENERIC_THUMBNAILS_FOLDER) return
```
CarlosNZ, line 49: Need to remember not to delete the "_generic_thumbnails" folder.

src/components/files/cleanup.ts
```diff
@@ -0,0 +1,79 @@
...
+
+    const relativeFilePath = filePath.replace(FILES_FOLDER + '/', '')
```
CarlosNZ, line 51: The path you were passing into the "checkIfInFileTable" method wasn't correct -- it was the full system path (`/User/<name>/conforma....`) but the file table paths are relative to the "files" folder.

src/components/files/cleanup.ts
```diff
@@ -0,0 +1,79 @@
...
+
+  await DBConnect.deleteMissingFileRecords(fileIdsToBeDeleted)
+  return fileIdsToBeDeleted.length
+}
+
+export const cleanUpFiles = async () => {
+  let filesMissingRecords = 0
+
+  // Check if file in database and delete if not
+  const checkFile = async (filePath: string) => {
```
CarlosNZ, line 48: Moved this "checkFile" function *inside* the main cleanup function just so we can keep track of the count of deleted files. "checkFile" itself can't return anything, as it's being called many times by "crawlFileSystem", so needed a way to keep a count outside of the function but within the scope of "cleanUpFiles".

src/components/files/cleanup.ts
```diff
@@ -0,0 +1,79 @@
...
+      filesMissingRecords++
+    }
+  }
+  console.log(
+    DateTime.now().toLocaleString(DateTime.DATETIME_SHORT_WITH_SECONDS),
+    'Cleaning up files and file records...'
+  )
+  await crawlFileSystem(FILES_FOLDER, checkFile)
+  const recordsMissingFiles = await processMissingFileLinks()
+  const filesCleanedUp = await DBConnect.cleanUpFiles()
```
CarlosNZ, line 64: 3 distinct tasks -- as summarised in the comment at the top of this file.

src/components/files/deleteFiles.ts
```diff
@@ -9,27 +9,27 @@ const { filesFolder, genericThumbnailsFolderName } = config
...
+  thumbnailPath?: string
 }
 
 export const deleteFile = async (file: FileDetail) => {
   const { filePath, thumbnailPath, originalFilename } = file
   try {
     await fsPromises.unlink(path.join(filesPath, filePath))
     // Don't delete generic (shared) thumbnail files
-    if (!thumbnailPath.match(genericThumbnailsFolderName))
+    if (thumbnailPath && !thumbnailPath.match(genericThumbnailsFolderName)) {
```
CarlosNZ, line 22: Modified "deleteFile" method to handle a missing "thumbnailPath" -- this was the source of the error you were getting.

src/components/graphQLConnect.ts
```diff
@@ -186,6 +186,23 @@ class GraphQLdb {
...
+            filePath
+            id
+          }
+        }
+      }
+      `,
+      { first: batchSize, offset }
+    )
+    return data?.files?.nodes || []
+  }
```
CarlosNZ, line 205: GraphQL query is easier for this, as we can use pagination to grab them in batches -- much harder to do with SQL without using a CURSOR.

src/components/postgresConnect.ts
```diff
@@ -360,41 +360,27 @@ class PostgresDB {
...
 
-  public setIsMissing = async (filePath: string) => {
+  public checkIfInFileTable = async (filePath: string) => {
     const text = `
-    UPDATE file
-    SET is_missing = true
-    WHERE file_path = $1
+    SELECT id 
+    FROM file
+    WHERE file_path = $1 OR thumbnail_path = $1
```
CarlosNZ, line 379: This is a bit of a mess to read as it's mixing and matching two functions -- one new one existing. Check the actual file to make better sense of it. But there's not much different here -- "checkIfInFileTable" is basically the same, have replaced "setIsMissing" with "deleteMissingFileRecords"

src/components/utilityFunctions.ts
```diff
@@ -87,15 +87,16 @@ export const filterObject = (
   return Object.fromEntries(filtered)
 }
 
+// Generic function for recursively crawling the file system. Will run method "fileOperation" on all files within "directory"
 export const crawlFileSystem = async (
-  newPath: string,
-  fileOperation: (newPath: string) => void
+  directory: string,
+  fileOperation: (filePath: string) => void
```
CarlosNZ, line 93: Better names -- you were using "newPath" for two different things, which is a bit confusing.

src/components/utilityFunctions.ts
```diff
@@ -87,15 +87,16 @@ export const filterObject = (
...
 ) => {
-  const files = fs.readdirSync(newPath)
+  const files = fs.readdirSync(directory)
   for (const file of files) {
-    const subPath = path.join(newPath, file)
-    if (fs.statSync(subPath).isDirectory()) crawlFileSystem(subPath, fileOperation)
-    else fileOperation(subPath)
+    const subPath = path.join(directory, file)
+    if (fs.statSync(subPath).isDirectory()) await crawlFileSystem(subPath, fileOperation)
+    else await fileOperation(subPath)
```
CarlosNZ, line 99: Couple of missing `await`s here, which were causing problems.

fergie-nz (approved):
Hey Carl,

That is awesome, love your work. Very neat and tidy.

Thank you for your patience in me looking through this -  I wanted to take the time to understand everything that was going on.

Issue #948, opened 2022-11-03
Fiji -- backups not being encrypted


CarlosNZ:
This is working locally, so something is happening on the server to not get the "BACKUPS_PASSWORD" correctly from environment variables.

Need to investigate...

PR #952, opened 2022-11-04
#951 #948 Fix migration / encrypted backups

### LINKED ISSUE #951: Migration code not always working on server
Okay, I know why this is happening now. (and that's what's causing the History panel crash even though it's technically "fixed")

When you import a snapshot that is output by the same version as the server, the migration script doesn't run. This is the expected behaviour. However, due the to problem with some of the generated columns not being restored, we *need* the functions/triggers/etc script to run in order to DROP and recreate those columns. In development this wasn't a problem as the migration script would run every restart regardless, but on the live server it doesn't.

The best way to deal with this is just to make sure the "functions/triggers" sql script runs after every snapshot reload regardless.

I'll do this now.

CarlosNZ:
Fix #951 (which is what is causing https://github.com/openmsupply/conforma-web-app/issues/1417 to persist)

See inline comments for further detail.

Edit: Added a quick fix for #948 as well :)


database/migration/migrateData.ts
```diff
@@ -12,24 +12,15 @@ import { getAppEntryPointDir } from '../../src/components/utilityFunctions'
...
-    ? // In development, pretend the version is one higher than it is,
-      // so we can keep getting changing migrations when testing
-      semverInc(version, 'patch')
-    : version
-
   try {
     databaseVersion = (await DB.getDatabaseVersion()).value
-    // No migration if database version matches current version, but we still
-    // proceed if this is a manual migration
-    if (semverCompare(databaseVersion, appVersion) >= 0 && !isManualMigration) return
```
CarlosNZ, line 32: No longer checking in order to avoid whole script -- the script will run regardless, and we'll be solely relying the version as to what actually gets run.

But the important thing is that `43_views_functions_triggers.sql` gets run after every snapshot install, which will make sure the generated columns get fixed every time.

src/components/snapshots/useSnapshot.ts
```diff
@@ -155,8 +155,10 @@ const useSnapshot: SnapshotOperation = async ({
     delay(1500)
 
     // Migrate database to latest version
-    console.log('Migrating database (if required)...)')
-    await migrateData()
+    if (options.shouldReInitialise) {
+      console.log('Migrating database (if required)...)')
+      await migrateData()
+    }
```
CarlosNZ, line 161: Need to make sure migration *doesn't* run for template imports, otherwise the database will be temporarily non-functional on a live system. (even if only for a few seconds)

docker/demo_server/docker-compose.yml
```diff
@@ -12,6 +12,7 @@ services:
       SMTP_PASSWORD: '${SMTP_SECRET}'
       WEB_HOST: '${WEB_URL}'
       JWT_SECRET: '${JWT_SECRET}'
+      BACKUPS_PASSWORD: '${BACKUPS_PASSWORD}'
```
CarlosNZ, line 15: Oops -- forgot to pass this through, so that's why the backups weren't being encrypted for Fiji.

Have made this fix manually on the server already, so this just persists in the repo for future. (Will need to update other servers too)

nmadruga: Ok - I'll just leave this on a note in our group chat so we remember about this next time we update the other server

nmadruga (approved):
LGTM

PR #954, opened 2022-11-07
#1413B Fix missing review element in History

CarlosNZ:
Back-end for front-end issue https://github.com/openmsupply/conforma-web-app/issues/1413

Turns out the reason why it was missing is that the "stage_number" field was `null` on that response.

Have changed the schema so that the stage_number field is an auto-generated column, so we don't have to worry about entering it.

Basing this in the "final-decision" fixes branch cos it'd be good to check that this works properly when a "Make decision" review is created -- i.e. generates the correct stage number for the NEW review.


database/migration/migrateData.ts
```diff
@@ -581,7 +581,7 @@ const migrateData = async () => {
 
     await DB.changeSchema(`
     ALTER TABLE data_table
-    ADD COLUMN data_view_code varchar;
+    ADD COLUMN IF NOT EXISTS data_view_code varchar;
```
nmadruga, line 584: Nice! 

Just for other errors on - not related - other changes in schema. Could you also fix those 2 please?
After Rename TYPE is_reviewable to reviewability and update field in TABLE template_element

- Problem altering schema: type "public.is_reviewable_status" does not exist 

- Problem altering schema: column "is_reviewable" does not exist

CarlosNZ:
> Although would like you to have another check please, it seems the original ApplicantResponse isn't showing anymore.

Thanks @nmadruga, good catch.

Have fixed on the front-end: https://github.com/openmsupply/conforma-web-app/pull/1424#issuecomment-1309801755

database/migration/migrateData.ts
CarlosNZ: Yeah, I've actually tried to fix this a few times, but neither of those statements ("ALTER TYPE" and "RENAME COLUMN") have an "IF EXISTS / IF NOT EXISTS" option, so it's a pain. Apparently there's ways if you wrap it in a function and catch the error, or do an IF based on a schema lookup, but it's all a bit of a hassle for something tiny, so I've just added an option to silence the errors on "changeSchema" statements: d1f58396

Hope that works okay.

nmadruga: Amazing, thanks!

nmadruga (approved):

Issue #953, opened 2022-11-06
Summary of Final Decision improvements


CarlosNZ:
After some discussion and as a result of the numerous bugs encountered due to overly-complicated final decision logic, we think the following changes are necessary to make the system more stable and robust:

- [x] Part 1 - Ensure that only one person can ever be assigned to a section at each stage/level, so no auto-assigning for final decision (Already done: #944 )
- [x] Part 2 - Update "generateReviewAssignments" to NOT auto-assign multiple reviewers at Final Decision stage https://github.com/openmsupply/conforma-server/pull/955
- [x] Part 3 - Go back to only allowing one level for a Final Decision stage. Otherwise the logic to determine whether or not to show individually-reviewable elements becomes overly complex and is very error-prone (Currently it shows incorrectly for multi-stage Final Decision stages) ~ https://github.com/openmsupply/conforma-web-app/pull/1434
- [x] Fix "re-assign" problem due to simultaneious unassign/assign mutations https://github.com/openmsupply/conforma-web-app/pull/1435
- [x] Finish front-end to properly wait for triggers to process and refetch correct assignments when assigning (@nmadruga this is mostly done, right?)
- [x] Fix how label is showing on Final decision, clearer this is the **previous review** decision made (Fixed in Part2)
![Screen Shot 2022-11-07 at 3 38 00 PM](https://user-images.githubusercontent.com/16461988/200215109-e2a55240-dc17-43c3-90bc-341dcf04581e.png)
- [x] Fix how label on top shows "Locked"  (Fixed in Part2)
![Screen Shot 2022-11-07 at 3 38 06 PM](https://user-images.githubusercontent.com/16461988/200215233-3cc13a6d-8c44-4b9e-9a6a-26ef5612e777.png)
- [x] Fix how Reviewer see Self-assignment option after review has been started by another  (may be fixed in Part3)
![Screen Shot 2022-11-07 at 3 55 04 PM](https://user-images.githubusercontent.com/16461988/200216945-336fdd4e-79d3-480b-a10a-6527d4e26fe3.png)
- [x] Fix option to Self-assign still showing on the list after another user is assigned to make "Final decision" review https://github.com/openmsupply/conforma-server/pull/960
![Screen Shot 2022-11-14 at 9 51 28 AM](https://user-images.githubusercontent.com/16461988/201544018-6f66b2f6-f45c-47b6-9d9a-6baf0c3d8edf.png)
[...]

PR #955, opened 2022-11-07
#953  - Part 2 Generate review assignments action not auto assign final decision

nmadruga:
Part 2 fix for #953

Front-end: https://github.com/openmsupply/conforma-web-app/pull/1428

### Briefing of changes
- Removed action `updateReviewAssignmentStatus` not needed anymore - since we don't change `isLocked` for selfAssignments
- Removed from ENUM assign_actions the column `ASSIGN_LOCKED`
- Add new column auto-generated on Graphql calls to set `available_sections` in ReviewAssignment

### Previous description
Changes to action `generateReviewAssignments` required for FinalDecision to act as any other self-assignable assignment. So only **one reviewAssignment** should stay as ASSIGNED status. The reviewer has to self-assign first and then start the review. This fix should be resolving the existing bug with https://github.com/openmsupply/conforma-server/issues/943

~~I also had to change how `isLocked` persists. Will need further testing to make sure this is consistent with other states.~~ Not needed now that action `updateReviewAssignmentStatus` has been fully removed.


plugins/action_update_review_assignments_status/src/updateReviewAssignmentsStatus.ts
```diff
@@ -35,7 +35,7 @@ async function updateReviewAssignmentsStatus({
 
       return otherSelfAssignments.map(({ id, assignedSections }) => ({
         id,
-        isLocked: isLocked || assignedSections.length > 0,
+        isLocked: isLocked ? assignedSections.length === 0 : false,
```
CarlosNZ, line 38: Do we still need `setOtherSelfAssignmentsLocked` at all? 

With the new database function to ensure consistency of assignments, there shouldn't be any other self-assignments (or any assignments at all) for the same section/level.

Come to think of it -- do we even need the "is_locked" field for review assignments at all? What does it mean for an assignment to be "locked"? They'll be unassigned if someone else takes it, not sure why we need to keep "is_locked" any more?

Maybe have a chat about this in the morn?

CarlosNZ (commented):
Cool, great to see a PR that is mainly removals :) 

See my inline comment about the "is_locked" situation.

plugins/action_generate_review_assignment_records/src/generateReviewAssignments.ts
```diff
@@ -456,41 +454,24 @@ const constructReviewAssignmentObject = (
...
-      isLocked: false,
-      assignedSections: sectionCodes,
-    }
-
-  // Create new OR update ReviewAssignment:
+  // Create NEW or update EXISTING ReviewAssignment:
   // 1. If existing
   //   - keep same status, isSelfAssignable
-  //   - just update isLocked = true (if already assigned to another)
+  //   - keep same isLocked if assigned
```
CarlosNZ, line 466: > `   //   - keep same isLocked if assigned`

Shouldn't existing review_assignments get `isLocked = false` always? (assuming this action runs on submission of applications and reviews). Otherwise, when do the "locked" assignments get "unlocked"?

documentation/List-of-Action-plugins.md
```diff
@@ -461,7 +460,7 @@ See [Grant Permissions](#grant-permissions) above regarding acting on user-only
 
 ### Generate Review Assignments
 
-Generates records in the `review_assignment` table -- i.e. which users (reviewers) are allowed to do a review for the current stage/level (and for which Sections). The records are set with `status` "Available" or "Assigned" and flags for `isSelfAssignable` and `isLocked` to help define when is allowed self-assignment.
+Generates records in the `review_assignment` table -- i.e. which users (reviewers) are allowed to do a review for the current stage/level (and for which Sections). The records are set with `status` "Available" or "Assigned". With the properties to specify the type of assignament: `isSelfAssignable` if it should show for self-assignment when not assigned by another user. The `isLocked` property helps defining that the review can start but not be submitted - after the application has been back to Applicant for ammendments.
```
CarlosNZ, line 463: Maybe bullet-point these properties to make easier to read?

Also:
> `isSelfAssignable` if it should show for self-assignment when not assigned by another user.

Isn't "self-assignable" a template configuration? So maybe should say "determines whether this assignment is self-assignable as determined by the template configuration for this stage/level" ?

CarlosNZ: Typo on `assignament` => "assignment" ;)

CarlosNZ (commented):
See comments. Looking good though :)

database/migration/migrateData.ts
```diff
@@ -583,6 +584,25 @@ const migrateData = async () => {
...
+    await DB.changeSchema(`
+      ALTER TYPE public.assigner_action RENAME to public.assigner_action_old;
+
+      CREATE TYPE public.assigner_action AS ENUM
+      (
+        'ASSIGN',
+        'RE_ASSIGN'
+      );
+
+      DROP TYPE public.assigner_action_old;
```
CarlosNZ, line 610: This won't work as it's still in use.

Unfortunately you have to add commands to change the columns of the old one to the new one. See how I did it for the trigger values -- around line 355 of this file.

database/migration/migrateData.ts
```diff
@@ -6,6 +6,7 @@ import { execSync } from 'child_process'
 import path from 'path'
 import { readFileSync } from 'fs'
 import { getAppEntryPointDir } from '../../src/components/utilityFunctions'
+import dbConnectInstance from '../../src/components/databaseConnect'
```
CarlosNZ, line 9: Accidental (unused) import.

database/migration/migrateData.ts
```diff
@@ -583,6 +584,25 @@ const migrateData = async () => {
     ALTER TABLE data_table
     ADD COLUMN data_view_code varchar;
     `)
+
+    console.log(' - Update existing review_assignment to set is_locked=false when isSelfAssigned')
+    // TODO - Not sure how to do this
```
CarlosNZ, line 589: What happens if we don't do this?

database/buildSchema/43_views_functions_triggers.sql
```diff
@@ -1423,7 +1477,6 @@ LANGUAGE sql
 STABLE;
 
 -- (https://github.com/graphile/graphile-engine/pull/378)
--- Required to make 'orderBy' work in application_list
```
CarlosNZ, line 1426: Why remove this line? Mistake?

database/buildSchema/43_views_functions_triggers.sql
```diff
@@ -700,6 +700,41 @@ CREATE TRIGGER review_assignment_validate_section_trigger
     WHEN (NEW.trigger IS NULL AND OLD.trigger IS NULL)
     EXECUTE FUNCTION public.enforce_asssigned_section_validity ();
 
+-- FUNCTION to set `available_sections` on given review_assignments based on other assignments
```
CarlosNZ, line 703:
```suggestion
-- FUNCTION to return `available_sections` on given review_assignments based on other assignments
```
We never see this as a column in postgres, on in GraphQL, so we should make that clear by emphasising its a function not a column

database/buildSchema/43_views_functions_triggers.sql
```diff
@@ -700,6 +700,41 @@ CREATE TRIGGER review_assignment_validate_section_trigger
...
+CREATE OR REPLACE FUNCTION public.review_assignment_available_sections (assignment public.review_assignment)
+    RETURNS text[]
```
CarlosNZ, line 705: `RETURNS varchar[]`

Mine throws an error with `text` -- doesn't yours?

database/buildSchema/43_views_functions_triggers.sql
```diff
@@ -700,6 +700,41 @@ CREATE TRIGGER review_assignment_validate_section_trigger
...
+                        SELECT
+                            unnest(assigned_sections)
+                        FROM
+                            review_assignment
+                        WHERE
+                            status = 'ASSIGNED'
+                            AND stage_id = $1.stage_id
+                            AND level_number = $1.level_number
+                            AND application_id = $1.application_id
+                            AND id <> $1.id))
```
CarlosNZ, line 733: I think you want to get rid of this last line -- otherwise it still says sections are available for the assigned reviewer, which is not correct.

CarlosNZ: To also handle the allowed_sections, just add a little bit to the final WHERE clause:
```
 WHERE
                    available_sections NOT IN (
                        SELECT
                            unnest(assigned_sections)
                        FROM
                            review_assignment
                        WHERE
                            status = 'ASSIGNED'
                            AND stage_id = $1.stage_id
                            AND level_number = $1.level_number
                            AND application_id = $1.application_id
                          )
                          AND (available_sections = ANY($1.allowed_sections)
                          OR $1.allowed_sections IS NULL)
                          )
```

database/buildSchema/43_views_functions_triggers.sql
CarlosNZ: Reverted

database/buildSchema/43_views_functions_triggers.sql
CarlosNZ: Done

database/buildSchema/43_views_functions_triggers.sql
CarlosNZ: Okay, have taken care of all this -- seems to working very nicely now ;)

database/buildSchema/43_views_functions_triggers.sql
CarlosNZ: Changed. No error anymore.

CarlosNZ:
Okay, @nmadruga I've updated and fixed a bunch of those things so we could get this branch working. Can see my changes in 52ce0d82

I've marked all the comments "Resolved" that I've taken care off -- you can show them again if you want to see what they were. So the only comments still visible should be minor things or questions for you to take a look at. The only one that seems fairly important is [this one](https://github.com/openmsupply/conforma-server/pull/955#discussion_r1015975606 ).

plugins/action_generate_review_assignment_records/src/generateReviewAssignments.ts
nmadruga: Fixed. But basically I added one specific case for when one ReviewAssignment was added while in the state of last submitted Review has locked others (LOQ submitted). So for that reason I have also changed `updateReviewStatus` which will set ReviewAssignment to be locked in this case (and then checked here). Please have a look on these 2 files:
- plugins/action_generate_review_assignment_records/src/generateReviewAssignments.ts
- plugins/action_update_review_status/src/updateReviewsStatuses.ts

documentation/List-of-Action-plugins.md
nmadruga: Thanks. Done

database/migration/migrateData.ts
nmadruga: Removed. 👍

nmadruga:
> Okay, @nmadruga I've updated and fixed a bunch of those things so we could get this branch working. Can see my changes in [52ce0d8](https://github.com/openmsupply/conforma-server/commit/52ce0d82f8b006a65ef6272d16dff279046d1795)
> 
> I've marked all the comments "Resolved" that I've taken care off -- you can show them again if you want to see what they were. So the only comments still visible should be minor things or questions for you to take a look at. The only one that seems fairly important is [this one](https://github.com/openmsupply/conforma-server/pull/955#discussion_r1015975606).

Looks good!

Thanks Carl, I have made suggestions. Please just note the changes flagged on [this comment](https://github.com/openmsupply/conforma-server/pull/955#discussion_r1018561135), I'm happy to merge if you are ok with that. Although I haven't got much further with testing this changes yet

plugins/action_update_review_status/src/updateReviewsStatuses.ts
```diff
@@ -92,10 +92,23 @@ const updateReviewsStatuses: ActionPluginType = async ({
           ReviewStatus.Draft,
         ])
 
+        // Lock ReviewAssignment of Review subimmet as changes required to applicant
```
CarlosNZ, line 95:
```suggestion
        // Lock ReviewAssignment of Review submitted as changes required to applicant
```

plugins/action_update_review_status/src/databaseMethods.ts
```diff
@@ -1,6 +1,23 @@
...
+    `
+
+    try {
+      const result = await DBConnect.query({ text, values: [reviewId] })
+      return result.row[0]
+    } catch (err) {
+      console.log(err.message)
+      throw err
+    }
+  },
```
CarlosNZ, line 19: Not sure why we need an extra database call. We already fetched all the associated reviews at the beginning, you should be able to pull it out of that result.

nmadruga: Reverting this - as per convo in person

Also - I'll be adding a briefing issue on what we have agreed on

CarlosNZ (approved):
LGTM, Thanks.

PR #958, opened 2022-11-11
#1431B assignment validation function fix

CarlosNZ:
Back-end branch for front-end PR https://github.com/openmsupply/conforma-web-app/pull/1435

Function was broken, but hadn't noticed as it worked fine with the *first* assignment for each stage/level. But didn't work for other ones. And also, forget to filter by "application_id" in the WHERE clause.


fergie-nz (approved):
Approve!

PR #959, opened 2022-11-13
#956 Review locking simplifications

### LINKED ISSUE #956: Simplify the "locking" of reviews by moving logic into the database.
It's becoming increasingly problematic to try and figure out all the logic for the "is_locked" field of both "review_assignment" and "review" as part of the "generateReviewAssignments" and "updateReviewStatuses" actions, so we are proposing to remove all considerations of "is_locked" completely from the actions and handle it all in the database before the trigger even gets written.

The "is_locked" value has no affect on back-end operations, it's purely for the front-end to be able to determine whether to show something as assignable, so should be fine to keep it completely out of actions.

First step:
- get rid of the "is_locked" field on the "review" table. This can be derived from the "is_locked" field on review_assignment and it makes it a lot simpler to only have to keep one field updated (not to mention consistent). I don't think there is any reason where these two values would need to be different, so may as well just have one. 

Then add additional triggers that run BEFORE updates on "application" and "review", so they'd update the appropriate records with the locked status before the normal action trigger runs.

Logic would be like this:

**ON_REVIEW_SUBMIT**:
(i.e. updating a "review" record with trigger "ON_REVIEW_SUBMIT")
- If `decision !== LIST_OF_QUESTIONS` or `is_last_level !== true`, do nothing. (i.e. they both must be true to continue)
- Find all review_assignments for the same stage/level ~~that are "ASSIGNED"~~ and set them to "is_locked"

**ON_APPLICATION_SUBMIT**:
(i.e. updating application record with trigger "ON_APPLICATION_SUBMIT")
- all review assignments associated with this application to `is_locked: false` (doesn't matter about level or stage, just do all)

There's one more case to consider, and that's *new* review assignments created as a result of actions triggered by the above events. In the "ON_REVIEW_SUBMIT" case we don't need to do anything, as only next-level review assignments could be created, and they should be all unlocked.

For application submissions, the only new ones created would be if new reviewers have been added or existing reviewers have new permissions. So we probably also need to add a "ON INSERT" trigger to the review assignment table to update the "is_locked" fields based on the above logic.

Is that all?

CarlosNZ:
Fix #956 

(Front-end: https://github.com/openmsupply/conforma-web-app/pull/1436)

As discussed on the issue, this changes:

- "LOCKED" status removed from "review_status" enum (and current "locked" values migrated to PENDING
- Function to return "is_locked" value based on application status
- replace references to "is_locked" value in existing functions with call to new function
- DROP "is_locked" column from review_assignment
- completely remove "updateReviewStatus" action
- update "generateReviewAssignments" to remove anything to do with "isLocked" (have I missed anything?)


nmadruga:
I'm getting this error when running the migration:

``
 - Simplify how we handle "locked" reviews
Problem altering schema: invalid input value for enum review_status_old: "LOCKED"

CarlosNZ:
> I'm getting this error when running the migration:
> 
> ```
>  - Simplify how we handle "locked" reviews
> Problem altering schema: invalid input value for enum review_status_old: "LOCKED" 
> ```

Yeah, you'll get that on every subsequent migration run on a given database. If I can't suppress it with `IF EXISTS` (I don't think I can), I'll need to add that "silenceError" flag again.

CarlosNZ:
> Yeah, you'll get that on every subsequent migration run on a given database. If I can't suppress it with `IF EXISTS` (I don't think I can), I'll need to add that "silenceError" flag again.

Okay, have separated commands out and added some error silencing (as little as possible) for commands that might cause errors: fa1d3768

database/migration/migrateData.ts
```diff
@@ -600,6 +600,51 @@ const migrateData = async () => {
...
+      UPDATE public.review_status_history
+        SET status = 'PENDING'
+        WHERE status = 'LOCKED';
+      `,
+      { silent: true }
+    )
+
+    await DB.changeSchema(`
+        DROP function review_status;
+      `)
```
nmadruga, line 644: Which function is to be dropped?

nmadruga (changes requested):
Stopped half-way through this review because I'm noticing a new trigger error on Assignment. This needs to be investigated.

database/migration/migrateData.ts
CarlosNZ: `review_status()` 🤷‍♂️ 
`-- Function to expose status name field on review table in GraphQL`
(Because the type depends on it, so it must be dropped first before we can drop the type)

CarlosNZ:
> Stopped half-way through this review because I'm noticing a new trigger error on Assignment. This needs to be investigated.

Can you be more specific? Unsure how this could be related to this PR.

nmadruga:
> Can you be more specific? Unsure how this could be related to this PR.

Sorry, yea. So to recreate please use the Fiji-Test4 snapshot (on our shared folder - in folder `SnapshotForPRs` and then login as `Admin` find the applications in this URL: TODO
 and try to Self-assign. When you click Submit there is a trigger error.

CarlosNZ:
> find the applications in this URL: TODO

That's a very helpful URL, @nmadruga LOL

CarlosNZ:
> So to recreate please use the Fiji-Test4 snapshot (on our shared folder - in folder `SnapshotForPRs` and then login as `Admin` find the applications in this URL: TODO and try to Self-assign. When you click Submit there is a trigger error.

So without knowing exactly which application you're talking about, all the ones that are currently self-assignable in that snapshot are in "Final Decision" stage, and have already been auto-assigned using the old behaviour (before last weeks fixes), so are in an illegal state (multiple reviewers all assigned to S1). 

So naturally there are all kinds of errors as a result of this. 

Please try again with an application not affected by this and it'll hopefully work.

CarlosNZ:
Okay, @nmadruga I found a couple of stray GraphQL queries in the back-end that were still using "isLocked" from review_assignment, which I've fixed. Updated from feature branch and tested -- assignment working okay now.

Please try snapshot `Fiji_assignment_with_locking_changes` in the "Snapshots for testing PRs" folder -- quick url to test assignment: http://localhost:3000/application/S-AVH-0114/review?tab=assignment

Seems to be working on the one you tested with earlier too.

nmadruga:
> So without knowing exactly which application you're talking about, all the ones that are currently self-assignable in that snapshot are in "Final Decision" stage, and have already been auto-assigned using the old behaviour (before last weeks fixes), so are in an illegal state (multiple reviewers all assigned to S1).
> 
> So naturally there are all kinds of errors as a result of this.
> 
> Please try again with an application not affected by this and it'll hopefully work.

Yes, this stayed incomplete due to the rush that was yesterday for me. The application I was trying which didn't have all reviewers already assigned was for PR2 template (a simpler copy of PR1). Filling the TODO missing from yesterday:

http://localhost:3000/application/S-BWY-0002/review?tab=assignment

Although trying now it's not showing! So I think the fix you made corrected it! I'm happy I made the comment half-way through anyway and happy it actually helped to display an existing problem : )

nmadruga:
Ok - Noticed something still failing on the `ChangeStatus` after I click to Start (using the "Make decision" action) but we know this is related to changed to the action that will need to be updated in the template. This is looking great! I'll have a look on your example now and re-review before we can approve all and start a new build (hopefully last pre-release) today.

CarlosNZ:
> Ok - Noticed something still failing on the `ChangeStatus` after I click to Start (using the "Make decision" action) but we know this is related to changed to the action that will need to be updated in the template:

Yup, that's why I suggested the snapshot above -- which has this problem corrected.

database/migration/migrateData.ts
```diff
@@ -600,6 +600,59 @@ const migrateData = async () => {
...
       );
     `)
+
+    console.log(' - Remove any illegal (i.e. overlapping sections) assignment state')
+    await DB.changeSchema(`
+      UPDATE public.review_assignment
+      SET status='ASSIGNED',
+      assigned_sections = assigned_sections, \
+      trigger = 'DEV_TEST';
+    `)
```
nmadruga, line 610: Would this be setting one review_assignment assigned per existing application in current review level?
What happens if there is already another review_assignment previously assigned? I imagine that would be persisted. Correct?

nmadruga (approved):
Looks great!

I have also tested the example and it's working well.

nmadruga:
Oh, on a further test, I tried using Re-assign and there is a small issue still. This is the workflow for when the bug is showing:
1. Logged in as Admin - http://localhost:3000/application/S-AVH-0114/review?tab=assignment
2. I selected **Basic Reviewer** to review and submitted
3. Then I re-assigned to **Senior Reviewer** and submitted
4. If I login as **Basic Reviewer** the used seems as could still be Self-assigned
 
Screen Shot 2022-11-15 at 10 54 14 AM

database/migration/migrateData.ts
CarlosNZ: It just makes sure that there's only one assignment per section per level. Where there are overlaps, it'll just keep the first one it finds (which would presumably be by id).

CarlosNZ:
> When I try to select "Yourself" and submit the user gets the Assigment!

Yup, there must be something wrong with the logic or permissions that determine what Basic Reviewer can see. I think we should have a quick chat about this. But not related to this PR, so I'll merge this one now and then we can make a plan for the whole branch.

nmadruga:
Thanks for fixing, Carl! I'll create a separated task for fixing the front-end part now : )

PR #960, opened 2022-11-13
#953 Fix to only show action SELF_ASSIGN while there are available_sections to be assigned

nmadruga:
Based on PR #956 (meant to be included in `feature/final-decision-fixes` branch)

Missing fix to not display Self-assign when there is another reviewer assigned on all available sections...

Check problem described in task list of issue #953

### DIFF
--- a/database/buildSchema/43_views_functions_triggers.sql
+++ b/database/buildSchema/43_views_functions_triggers.sql
@@ -1397,6 +1397,8 @@ STABLE;
 
 -- REVIEWER_ACTION_LIST
 -- Aggregated VIEW method of reviewer action to each application on application list page
+DROP FUNCTION IF EXISTS review_list (integer, integer);
+
 DROP FUNCTION IF EXISTS review_list (integer, integer, public.application_status);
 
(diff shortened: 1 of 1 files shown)


database/buildSchema/43_views_functions_triggers.sql
```diff
@@ -1397,6 +1397,8 @@ STABLE;
 
 -- REVIEWER_ACTION_LIST
 -- Aggregated VIEW method of reviewer action to each application on application list page
+DROP FUNCTION IF EXISTS review_list (integer, integer, public.application_status);
```
CarlosNZ, line 1400: Somehow we've ended up with two different "review_list" functions in the database.

Can you drop the other one too:
```
DROP FUNCTION IF EXISTS review_list (integer, integer);
```

CarlosNZ (approved):
Cool, looks good :) 

Just one little request please (see comment)

database/buildSchema/43_views_functions_triggers.sql
nmadruga: Oh! I see. Thanks!

Issue #852, opened 2022-09-15
Template Builder - Improvements/tasks for future work


nmadruga:
A few issues/improvements notices while working with TB:
- [ ] #167 If we somehow allow the same version to be included it will fail trying to load the application because there are 2 AVAILABLE versions

TO BE CONTINUED

CarlosNZ:
> * If we somehow allow the same version to be included it will fail trying to load the application because there are 2 AVAILABLE versions

We fixed this -- the trigger already exists, but it was being disabled during snapshot import. However, I've actually made a further fix to this which is more appropriate: #850

CarlosNZ:
Close this now?

nmadruga:
No... I'm trying to list all issues we have for Template Builder improvements - not that we are going to work on it now. But it's good to have a list

PR #963, opened 2022-11-15
#961 File handling fixes

### LINKED ISSUE #961: Handle files correctly when exporting templates
This will need a combination of dev tweaks and config I think.

The "generateDoc" action needs to be updated so that it *doesn't* add the "template_id" field for output docs, as we *don't* want them being exported with the template. (Probably behaviour should be "don't save template_id unless explicitly stated in parameters")

And the "fileUpload" template needs to be tweaked so that it has an option to associate the template_id (drop-down list of templates I guess), so that the template_id *will* be associated with the new file and be exported along with it.

CarlosNZ:
Fix #961 

Requires back-end https://github.com/openmsupply/conforma-web-app/pull/1439

Changes required to preserve association between documents and templates so they get exported and re-imported correctly. See inline comments for more info.

[fileUpload-4.zip](https://github.com/openmsupply/conforma-server/files/10009734/fileUpload-4.zip)
Use this template to test.

- Use it to upload a file, setting the subfolder and associated template.
- Notice that the file record in the database has the template_id correctly linked.
- Export the template from the templates page
- If you unzip it, you'll notice the connected file is included (and hopefully nothing else)
- Load a different snapshot (that doesn't have this template)
- Import the exported template
- The linked file should be copied in and correctly entered into the database

### DIFF
--- a/plugins/action_generate_document/src/generateDoc.ts
+++ b/plugins/action_generate_document/src/generateDoc.ts
@@ -13,7 +13,10 @@ const generateDoc: ActionPluginType = async ({
     parameters
   const userId = parameters?.userId ?? applicationData?.userId
   const applicationSerial = parameters?.applicationSerial ?? applicationData?.applicationSerial
-  const templateId = parameters?.templateId ?? applicationData?.templateId
+  // We don't want to include template id normally, as that will link the file
+  // to the template when exporting, which we only want for Carbone docs and the
+  // like. So we'll only use "templateId" if it's explicitly provided.
+  const templateId = parameters?.templateId
 
   // Build full data object
   const allData = {
--- a/src/components/exportAndImport/importFromJson.ts
+++ b/src/components/exportAndImport/importFromJson.ts
@@ -132,6 +132,7 @@ const getForeignKeyReplacements = (
           referenceTableName
         )
         if (referenceValue) foreignKeyReplacements[columnName] = referenceValue
+        else foreignKeyReplacements[columnName] = null
       }
     )
   return foreignKeyReplacements


plugins/action_generate_document/src/generateDoc.ts
```diff
@@ -13,7 +13,10 @@ const generateDoc: ActionPluginType = async ({
     parameters
   const userId = parameters?.userId ?? applicationData?.userId
   const applicationSerial = parameters?.applicationSerial ?? applicationData?.applicationSerial
-  const templateId = parameters?.templateId ?? applicationData?.templateId
+  // We don't want to include template id normally, as that will link the file
+  // to the template when exporting, which we only want for Carbone docs and the
+  // like. So we'll only use "templateId" if it's explicitly provided.
+  const templateId = parameters?.templateId
```
CarlosNZ, line 19: This prevents all the output docs getting the template_id saved, otherwise we copy every single output doc when exporting a template, which we definitely *don't* want.

This won't fix *existing* files though, so if you export a template that already has output docs associated with it, they'll still come along for the ride (just delete them manually from the export zip I guess)

src/components/exportAndImport/importFromJson.ts
```diff
@@ -132,6 +132,7 @@ const getForeignKeyReplacements = (
           referenceTableName
         )
         if (referenceValue) foreignKeyReplacements[columnName] = referenceValue
+        else foreignKeyReplacements[columnName] = null
```
CarlosNZ, line 135: Screenshot 2022-11-15 at 5 46 52 PM

This mitigates this occasional error that I mentioned yesterday. The problem was (in this case), that the file record we were importing had a foreign key reference to application_serial and application_response, but these referenced records were not in the system we're importing into, and they're not part of the template snapshot, so it was throwing this error and not importing the file.

This change just means that instead of failing, it'll just put `null` into the fields it can't reconnect. Which makes sense, as we can't link to an application that doesn't exist (and we don't want to export all linked ones with a template export)!

nmadruga (approved):
Changes seem pretty safe. I'll be testing with Front-end PR now

src/components/exportAndImport/importFromJson.ts
CarlosNZ: > I have added it to this issue: #852

Why? It's not longer an issue :)

PR #965, opened 2022-11-16
Prevent adding columns to application and file tables

CarlosNZ:
We have some protection so that actions can't write to tables other than "data_table_..." and a few key others (in `config.allowedTableNames`

This included "application" table, which we need access to for updating serials, etc. 

However, what we *don't* want is the schema being changed on these tables. Fergus had a "generateText" action that was adding additional columns to the "application" table, which caused errors due to name conflicts.

We need to completely prevent this, as it's a loophole we've overlooked. This PR just makes an additional list of tables that can be written to, but not have columns added. If you try and run an action that would add a column to these, it'll throw an error.


nmadruga (approved):
👍 

@CarlosNZ can you merge this one now please?
I'll start the build after this one is merged in

PR #968, opened 2022-11-16
#967 Template category export/import fix

### LINKED ISSUE #967: Bug: Problem keeping category when importing template
Image

Gets reset to "No category".

This was caused by the change I introduced the other day to fix file import:

```
        if (referenceValue) foreignKeyReplacements[columnName] = referenceValue
        else foreignKeyReplacements[columnName] = null
```

The `else...null` line was added. 

Unfortunately, there's no simple way to fix one problem without causing the other, as there's no clear way to know if the inserted reference for an existing record (i.e. not being imported with the template) is correct in the database or not.

I'm considering adding a check to the `findReferenceValues` function to look for existing records in the database if not found in the "insertedRecords", but this could cause problems if the foreign key exists, but refers to a different item than it did in the database that it was exported from. (i.e. the category could change to the wrong one without anyone noticing).

There are so many little bugs and problems like this which support my belief that the template import/export system needs to be completely be re-written. The problem has always been trying to treat it like a "snapshot" rather than its own special case. (Links should be matched more by name and code rather than foreign key references IMO, with additional logic to handle conflicts.)

Obviously that's quite a big job, so not suggesting doing it right away, but we just need to remember that template import/export will remain buggy while it works the way it does now.

Not sure what the best solution to this immediate problem is. If we have to choose between templates not importing their linked files, and templates being imported with missing categories, I think the latter is preferable as its easier to notice and fix. (In which case, we'd leave it as it is now and just have to manually re-add the category when importing templates). 

Do we agree? (Or does anybody have a better solution?)

CarlosNZ:
Fix #967 

Requires front-end https://github.com/openmsupply/conforma-web-app/pull/1443

This is not a complete fix -- there's still cases you can create where this doesn't work properly, but these problems already exist for existing connected elements (e.g. permissionNames), and fixing them would be a whole lot extra. (For example, if you import a template with categories or permissionNames that don't already exist in the system, it'll only create new ones if the "id" *matches* an existing entry -- you'd think the other way round would be the expected behaviour, but 🤷‍♂️. Again, this needs a complete re-write).

However, this will sort out the case we mostly deal with -- exporting a template to import into another (similar) server.

And it doesn't break the linked file import, so that's good.

### DIFF
--- a/database/snapshotOptions/templateExport.json
+++ b/database/snapshotOptions/templateExport.json
@@ -11,12 +11,19 @@
     "templateStage",
     "templateSection",
     "templateStageReviewLevel",
+    "templateCategory",
     "permissionName",
     "file"
   ],
   "//tablesToUpdateOnInsertFail": "deprecated, in favour of skipTableOnInsertFail",
-  "//skipTableOnInsertFail": "These tables have unique contraint other then ID, they will be skippped if insert fail, and any related records will reference existing record",
-  "skipTableOnInsertFail": ["permissionName", "permissionPolicy", "filter", "file"],
+  "//skipTableOnInsertFail": "These tables have unique constraint other then ID, they will be skipped if insert fail, and any related records will reference existing record",
+  "skipTableOnInsertFail": [
+    "permissionName",
+    "permissionPolicy",
+    "filter",
+    "file",
+    "templateCategory"
+  ],
   "excludeTables": [],
   "insertScriptsLocale": "dev",
   "includeInsertScripts": [],


nmadruga:
> For example, if you import a template with categories or permissionNames that don't already exist in the system, it'll only create new ones if the "id" _matches_ an existing entry

Oh, no! That's not good. But I understand not a fix for now... Maybe we should create an issue already regarding that problem. I imagine this would cause what you had mentioned before in the issue:

> I'm considering adding a check to the findReferenceValues function to look for existing records in the database if not found in the "insertedRecords", but this could cause problems if the foreign key exists, but refers to a different item than it did in the database that it was exported from. (i.e. the category could change to the wrong one without anyone noticing).

nmadruga (approved):
Cool easy fix. Exporting the Category link, nice!

I'll check the Front-end now and test.

CarlosNZ:
> But I understand not a fix for now... Maybe we should create an issue already regarding that problem.

Which brings me back to "This whole thing needs a redesign". I'd be reluctant to spend any significant time making this logic even more convoluted than it already is as that's guaranteed to cause more problems than it fixes (as we're finding out with these little tweaks). I'd rather just bite the bullet and re-design a proper export/import module.

PR #969, opened 2022-11-18
Fix cleanup script when files folder not present yet

CarlosNZ:
Quick fix: 

- change order of startup functions, so "createDefaultDataFolders()" runs before "cleanUpFiles()"
- put clean-up script in try/catch block so it will fail gracefully

Either of those on its own would fix the current problem, so we've some extra redundancy :)


nmadruga (approved):
That's great. Thanks! I'll be starting the testing build in the new version 0.4.7 : )

PR #971, opened 2022-11-21
#1445B - Moving isAdmin, isManager and permission names to returned User

nmadruga:
Fixes [Front-end PR](https://github.com/openmsupply/conforma-web-app/pull/1448)

### Changes
- Move preferences about SystemManager permissionName to be in back-end of `defaultSystemManagerPermissionName` (default) or project-specific to be defined in `preference.json`.
- Return User including values of `isAdmin`, `isManager` and `permissionNames` (it doesn't seem to be used, but still)

### DIFF
--- a/src/components/postgresConnect.ts
+++ b/src/components/postgresConnect.ts
@@ -692,16 +692,18 @@ class PostgresDB {
       WHERE id = $1
       `
     const text2 = `
-      SELECT name as "orgName"
-      FROM organisation
-      WHERE id = $1
+    SELECT distinct("orgName"), "permissionName"
+      FROM permissions_all
+      WHERE ("userId" = $1 OR "userId" IS NULL) 
+      AND "orgId" = $2
       `
     try {
       const result = await this.query({ text, values: [userId] })
       const userData = { ...result.rows[0], orgName: null }
       if (orgId) {
-        const orgResult = await this.query({ text: text2, values: [orgId] })
+        const orgResult = await this.query({ text: text2, values: [userId, orgId] })
         userData.orgName = orgResult.rows[0].orgName
+        userData.permissionNames = orgResult.rows.map(({ permissionName }) => permissionName)
       }
       return userData
     } catch (err) {
--- a/src/components/postgresConnect.ts
+++ b/src/components/postgresConnect.ts
@@ -692,18 +692,16 @@ class PostgresDB {
       WHERE id = $1
       `
     const text2 = `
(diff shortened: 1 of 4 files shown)


src/components/postgresConnect.ts
```diff
@@ -692,16 +692,18 @@ class PostgresDB {
       WHERE id = $1
       `
     const text2 = `
-      SELECT name as "orgName"
-      FROM organisation
-      WHERE id = $1
+    SELECT distinct("orgName"), "permissionName"
+      FROM permissions_all
+      WHERE ("userId" = $1 OR "userId" IS NULL) 
+      AND "orgId" = $2
```
CarlosNZ, line 698: I'm getting an error when calling `getApplicationData` endpoint on an application with no orgId.

The problem is here -- you're only returning the permissions info when there is an orgId provided, but it should return it regardless.

CarlosNZ: This query is also called by `getUserInfo`/`getUserOrgData` (in loginHelpers), so I'm pretty sure it'll break that by changing it here too.

Everything you need is already done in the `getUserInfo` function, so you can probably just use that in "getApplicationData" (instead of `getUserData`) and modify the result into the shape you need it to be in "getApplicationData" (it includes all the permission name info too, I'm pretty sure.)

I'm very reluctant to change any of these queries, as they're very fragile and have subtle implications in lots of different places.

CarlosNZ (changes requested):
See the inline comments.

I'm pretty sure you can get what you need by reverting the postgres queries back to what they were and just calling `getUserInfo` (in loginHelpers) from "getApplicationData" and then just putting it in the shape required for applicationData.

src/components/postgresConnect.ts
CarlosNZ: Okay, have tried my suggestion and it seems to work okay. It manages to get what we need without changing any postGresConnect queries, which I think is important considering how interconnected they are.

Branch is #1445B-move-checks-for-admin-manager-inside-current-user_suggestion

Compare here: https://github.com/openmsupply/conforma-server/compare/%231445B-move-checks-for-admin-manager-inside-current-user...%231445B-move-checks-for-admin-manager-inside-current-user_suggestion

(All the changes in `postgresConnect.ts` are just reverting it back to `develop`)

nmadruga: Nice Carl! That looks great to me

CarlosNZ (approved):
👍

PR #973, opened 2022-11-23
Update core_templates

CarlosNZ:
Couple of things to check:

- Check the actions are correct in "core_actions" template. Have removed the deprecated ones, so check that what's left makes sense.
- Now saved as a pg_dump file. Is this okay? Can change "core_templates" to use the old way if we think it's safer?


CarlosNZ:
Converting to draft -- best to leave it until #976 is sorted.

CarlosNZ:
Closing -- will re-do with the core-actions branch once it's merged.

PR #976, opened 2022-11-24
#975(1) Handle core actions

### LINKED ISSUE #975: Proposal for keeping "core actions" out of template configs
There are a number of actions that *must* run on certain events in order for the system to function correctly, and it's kind of a hassle to make sure they're correctly configured on every template, and too easy to accidentally misconfigure them and break things.

These should really be hard-coded into the system so we don't have to worry about them at all in configs. I suggest adding an extra function in `processTrigger` to insert these "core" actions into the "actionsSequential" sequence. The position in the sequence and the exact parameters might vary somewhat, so there would need to be a hard-coded reference for how to insert the actions into the sequence.

We shouldn't try and do this until we've identified all the core actions, what events they run on, where in the sequence they need to go (some might actually be async?), and what parameters they take (and how to configure them?), so let's start making a list here.

I think this will actually be quite easy to implement, the hard part will be identifying and specifying the details for what the actions are, so let's start a comprehensive list here:

### ON_APPLICATION _CREATE
- `generateTextString` for serial. We need a way for the "pattern" parameter to be configurable, the rest can probably be hard-coded (as per the current "CoreActions" config in core_templates). Perhaps we need a simple "pattern" field in the template builder (with a default)
- `generateTextString` for application name. Just use the "CoreActions" config (names it `<templateName> - <serial>`). If a different name is needed
- `incrementStage` -- simple :) 

These should all go *before* any custom actions that might be added to the template.

### ON_APPLICATION_RESTART
- Change status (as per core templates)

### ON_APPLICATION_SUBMIT
- `changeStatus` to SUBMITTED. Easy
- `Trim duplicate reponses` as per "CoreActions"
- `generateReviewAssignments` as per "CoreActions"
- `updateReviewStatuses`  reverting as per https://github.com/openmsupply/conforma-templates/issues/106
- `cleanupFiles` as per "CoreActions"
- MAYBE "Change Outcome" IF can find condition that considers reviewability?

### ON_REVIEW_ASSIGN
- `changeStatus` "On Re-assignment if review is existing change to DRAFT"
(For case when someone is re-assigned and they have an existing review -- should go back to DRAFT)

### ON_REVIEW_CREATE
- `changeStatus` - review status to "DRAFT"

### ON_REVIEW_RESTART
- as per core templates (Sets to DRAFT)

### ON_REVIEW_SUBMIT
- `changeStatus` -- review status to "SUBMITTED"
- `trimResponses` -- as per "CoreActions"
- `updateReviewStatuses` (as per discussion)
- `generateReviewAssignments` (no condition or parameters necessary)
- `updateReviewVisibility` -- as per core_templates, including Condition
- `incrementStage` -- core if and only if is last level and CONFORM
- `changeStatus` -- change application to "CHANGES REQUIRED" if last level and decision is LOQ
- `changeOutcome` -- Maybe? (But sometimes we need if for "Application Submit")

### ON_REVIEW_UNASSIGN
- change existing review status to "DISCONTINUED"

Any other triggers?

## Summary

It seems like all the "core" actions should be inserted *before* any template-specific actions run, which makes it easier. And as far as I can tell, the only one that would require any configuration is the very first one (pattern for generate Serial string), so we'd just need to come up with a solution for this -- my suggestion is just a simple "serial pattern" field in the main template builder config (and probably a new field on the "template" database table to store this).

I think this would make things a lot better for multiple reasons:
- maintaining these actions would be easy -- any changes just need to be done in one place in the code base with no changes done to template configs
- safer -- it would be very difficult to break functionality by misconfiguration actions
- the template builder Actions page would be a *lot* less crowded and therefore much easier to work with and maintain

CarlosNZ:
Fix #975 

Oh my god, I can't believe we didn't do this ages ago -- it's so much easier to keep a handle on these core actions if they're all in one place like this and it was incredibly simple to implement. Works a treat :) 

I haven't actually got all the actions correct yet -- @nmadruga, you and I should sit down together and go through them carefully, but it's so much easier to know we can do it once and it'll be correct everywhere!

Still to do:

- [x] Fetch serial no "pattern" from template so it can be configurable #978 
- [x] Get exact core action collection sorted (requires #977 )
- [ ] ~~Find a way to show a warning when loading a snapshot with the core actions still in place~~

### DIFF
--- a/src/components/actions/coreActions.ts
+++ b/src/components/actions/coreActions.ts
@@ -70,6 +70,19 @@ const coreActions: CoreActions = {
       },
     },
   ],
+  [Trigger.OnApplicationRestart]: [
+    // Set status = DRAFT
+    {
+      code: 'changeStatus',
+      path: '../plugins/action_change_status/src/index.ts',
+      name: 'Change Status',
+      trigger: 'ON_APPLICATION_SUBMIT',
+      event_code: null,
+      sequence: -1,
+      condition: true,
+      parameter_queries: { newStatus: 'SUBMITTED' },
+    },
+  ],
   [Trigger.OnApplicationSubmit]: [
     // Set status = SUBMITTED
     {
@@ -78,7 +91,7 @@ const coreActions: CoreActions = {
       name: 'Change Status',
       trigger: 'ON_APPLICATION_SUBMIT',
       event_code: null,
-      sequence: -4,
+      sequence: -5,
       condition: true,
       parameter_queries: { newStatus: 'SUBMITTED' },
     },
@@ -89,7 +102,7 @@ const coreActions: CoreActions = {
       name: 'Trim duplicate responses',
       trigger: 'ON_APPLICATION_SUBMIT',
       event_code: null,
-      sequence: -3,
+      sequence: -4,
       condition: true,
       parameter_queries: {},
     },
@@ -100,7 +113,7 @@ const coreActions: CoreActions = {
       name: 'Generate Review Assignment Records',
       trigger: 'ON_APPLICATION_SUBMIT',
       event_code: null,
-      sequence: -2,
+      sequence: -3,
       condition: true,
       parameter_queries: {},
     },
@@ -112,20 +125,39 @@ const coreActions: CoreActions = {
       name: 'Clean up application files',
       trigger: 'ON_APPLICATION_SUBMIT',
       event_code: null,
-      sequence: -1,
+      sequence: -2,
       condition: true,
       parameter_queries: {},
     },
+    // Update review statuses
+    // TO-DO: Un-comment once action is re-implemented
+    // {
+    //   code: 'updateReviewStatuses',
+    //   path: '../plugins/action_update_review_statuses/src/index.ts',
+    //   name: 'Update Review Statuses',
+    //   trigger: 'ON_APPLICATION_SUBMIT',
+    //   event_code: null,
+    //   sequence: -1,
+    //   condition: true,
+    //   parameter_queries: {
+    //     changedResponses: {
+    //       operator: 'objectProperties',
+    //       children: ['outputCumulative.updatedResponses'],
+    //     },
+    //   },
+    // },
   ],
   [Trigger.OnReviewAssign]: [
-    // TO-DO: this is not correct -- needs to be worked out exactly
+    // If any reviews already exist for the current assignment, set them to
+    // "DRAFT" (this would happen if the reviewer had been unasssigned and then
+    // re-assigned)
     {
       code: 'changeStatus',
       path: '../plugins/action_change_status/src/index.ts',
       name: 'Change Status',
       trigger: 'ON_REVIEW_ASSIGN',
       event_code: null,
-      sequence: 1,
+      sequence: -1,
       condition: {
         operator: '!=',
         children: [
@@ -138,15 +170,12 @@ const coreActions: CoreActions = {
       },
       parameter_queries: {
         isReview: true,
-        reviewId: {
-          operator: 'objectProperties',
-          children: ['applicationData.reviewData.reviewId', null],
-        },
         newStatus: 'DRAFT',
       },
     },
   ],
   [Trigger.OnReviewCreate]: [
+    // Set to DRAFT
     {
       code: 'changeStatus',
       path: '../plugins/action_change_status/src/index.ts',
@@ -159,9 +188,63 @@ const coreActions: CoreActions = {
     },
   ],
   [Trigger.OnReviewRestart]: [
-    // Is this even used?
+    // Set to DRAFT
+    {
+      code: 'changeStatus',
+      path: '../plugins/action_change_status/src/index.ts',
+      name: 'Change Status',
+      trigger: 'ON_REVIEW_CREATE',
+      event_code: '',
+      sequence: -1,
+      condition: true,
+      parameter_queries: { newStatus: 'DRAFT' },
+    },
   ],
   [Trigger.OnReviewSubmit]: [
+    // Set review status => SUBMITTED
+    {
+      code: 'changeStatus',
+      path: '../plugins/action_change_status/src/index.ts',
+      name: 'Change Status',
+      trigger: 'ON_REVIEW_SUBMIT',
+      event_code: '',
+      sequence: -7,
+      condition: true,
+      parameter_queries: { newStatus: 'SUBMITTED' },
+    },
+
+    // Remove any review responses that have not changed since previous review
+    {
+      code: 'trimResponses',
+      path: '../plugins/action_trim_responses/src/index.ts',
+      name: 'Trim duplicate review responses',
+      trigger: 'ON_REVIEW_SUBMIT',
+      event_code: '',
+      sequence: -6,
+      condition: true,
+      parameter_queries: {
+        operator: 'objectProperties',
+        children: ['outputCumulative.reviewStatusHistoryTimestamp'],
+      },
+    },
+    // Update review statuses
+    // TO-DO: Un-comment once action is re-implemented
+    // {
+    //   code: 'updateReviewStatuses',
+    //   path: '../plugins/action_update_review_statuses/src/index.ts',
+    //   name: 'Update Review Statuses',
+    //   trigger: 'ON_REVIEW_SUBMIT',
+    //   event_code: null,
+    //   sequence: -5,
+    //   condition: true,
+    //   parameter_queries: {
+    //     changedResponses: {
+    //       operator: 'objectProperties',
+    //       children: ['outputCumulative.updatedResponses'],
+    //     },
+    //   },
+    // },
+
     // If sent back to applicant for further information (LOQ), this sets which
     // review responses are visible to the applicant
     {
@@ -170,7 +253,7 @@ const coreActions: CoreActions = {
       name: "Update Applicant's Review Visibility",
       trigger: 'ON_REVIEW_SUBMIT',
       event_code: '',
-      sequence: -3,
+      sequence: -4,
       condition: {
         operator: 'AND',
         children: [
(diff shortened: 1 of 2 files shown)


src/components/actions/coreActions.ts
```diff
@@ -0,0 +1,239 @@
...
+  [Trigger.OnApplicationCreate]: [
+    // Generate serial for application when first created
+    // Can be over-ridden by specifying a template action
+    {
+      code: 'generateTextString',
+      path: '../plugins/action_generate_text_string/src/index.ts',
+      name: 'Generate Text String',
+      trigger: 'ON_APPLICATION_CREATE',
+      event_code: null,
+      sequence: -3,
```
CarlosNZ, line 27: Using negative numbers to ensure they run before any added template actions (which would normally start from 1, 2...)

src/components/actions/coreActions.ts
```diff
@@ -0,0 +1,239 @@
...
+      condition: true,
+      parameter_queries: {
+        pattern: 'S-[A-Z]{3}-<+dddd>',
```
CarlosNZ, line 30: To-Do: fetch this pattern from template config

src/components/actions/processTrigger.ts
```diff
@@ -30,7 +31,10 @@ export async function processTrigger(payload: TriggerPayload): Promise<ActionRes
   const actionsSequential = resolvedActions.filter(({ sequence }) => !!sequence)
   const actionsAsync = resolvedActions.filter(({ sequence }) => !sequence)
 
-  for (const action of [...actionsAsync, ...actionsSequential]) {
+  // Get core actions for the current trigger
+  const currentCoreActions = coreActions?.[trigger] ?? []
+
+  for (const action of [...actionsAsync, ...currentCoreActions, ...actionsSequential]) {
```
CarlosNZ, line 37: Core actions get loaded the inserted into the action queue along with the rest of the sequential actions. From there on, there's no difference between them -- they get handled exactly the same)

CarlosNZ:
> * Find a way to show a warning when loading a snapshot with the core actions still in place

I actually can't think of a way to do this that's not more trouble than it's worth. Probably best to just be vigilant as we migrate the templates.

nmadruga:
Ready to Review this one @CarlosNZ ?

CarlosNZ:
> Ready to Review this one @CarlosNZ ?

Same answer I gave you on Friday:

> > @CarlosNZ Is this one Ready for review?
> 
> It's ready in a conceptual sense.
> 
> The only thing that's not done is figuring out exactly _what_ the core actions should be.
> 
> But I thought we could go through that together at some point and get it really solid.

CarlosNZ:
Okay, @nmadruga, have updated the core actions as per what we discussed today. Have included "updateReviewStatuses" but it's currently commented out until that action is finished.

Regarding a possible "changeOutcome" on Application Submit, a simple "is it reviewable" condition is not enough. For example, on "User Registration", we actually change the outcome "ON_VERIFICATION", not on application submit. So what I've done is a query to check if there are any other "change outcome" actions associated with this template and, if not, change the outcome  to "APPROVED". Any other outcome changes will need to be specified in template action config.

src/components/actions/coreActions.ts
nmadruga: Is this included on the next PR: #978 ?

CarlosNZ: Yes

src/components/actions/coreActions.ts
```diff
@@ -0,0 +1,493 @@
...
+    //   parameter_queries: {
+    //     changedResponses: {
+    //       operator: 'objectProperties',
+    //       children: ['outputCumulative.updatedResponses'],
+    //     },
+    //   },
+    // },
+
+    // Change outcome to APPROVED if there are no other "changeOutcome" actions
+    // associated with this template
```
nmadruga, line 151:
```suggestion
    // associated with this template. 
    // Used for Non-reviewable templates or when there isn't another specific changeOutcome included
```

nmadruga: I wonder how would this behave when you "create a template" (pretend we don't do copies only) and this changeOutcome is included first, but the user goes and add a new specific one.

src/components/actions/coreActions.ts
```diff
@@ -0,0 +1,493 @@
...
+      parameter_queries: { newStatus: 'DRAFT' },
+    },
+  ],
+  [Trigger.OnReviewRestart]: [
+    // Set to DRAFT
+    {
+      code: 'changeStatus',
+      path: '../plugins/action_change_status/src/index.ts',
+      name: 'Change Status',
+      trigger: 'ON_REVIEW_CREATE',
```
nmadruga, line 229:
```suggestion
      trigger: 'ON_REVIEW_RESTART',
```

src/components/actions/coreActions.ts
```diff
@@ -0,0 +1,493 @@
...
+                    operator: 'objectProperties',
+                    children: ['applicationData.reviewData.latestDecision.decision'],
+                  },
+                  'NON_CONFORM',
+                ],
+              },
+              {
+                operator: 'objectProperties',
+                children: ['applicationData.reviewData.isLastLevel'],
+              },
```
nmadruga, line 421: This condition is a bit different than what is intended to be

nmadruga (changes requested):
Also, could you add the existing descriptions from Core_actions templates in here please? I find it useful to have that displayed on the TB actions' page 🙏

CarlosNZ:
> Also, could you add the existing descriptions from Core_actions templates in here please? I find it useful to have that displayed on the TB actions' page 🙏

I can update the comments (I thought I'd already done most of them, albeit with different wording).

But bear in mind these don't show up in the template builder at all. That's kind of the point, they're completely hidden from configs so the configurer only has to worry about the template-specific actions (which makes the template builder actions page way less cluttered)

src/components/actions/coreActions.ts
CarlosNZ: I think it would be fine. This action wouldn't run, since it'd find another one.

src/components/actions/coreActions.ts
```diff
@@ -0,0 +1,493 @@
...
+    // Set status = DRAFT
+    {
+      code: 'changeStatus',
+      path: '../plugins/action_change_status/src/index.ts',
+      name: 'Change Status',
+      trigger: 'ON_APPLICATION_SUBMIT',
+      event_code: null,
+      sequence: -1,
+      condition: true,
+      parameter_queries: { newStatus: 'SUBMITTED' },
```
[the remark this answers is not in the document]
CarlosNZ: e76a5c86

src/components/actions/coreActions.ts
```diff
@@ -0,0 +1,493 @@
...
+      },
+    },
+  ],
+  [Trigger.OnApplicationRestart]: [
+    // Set status = DRAFT
+    {
+      code: 'changeStatus',
+      path: '../plugins/action_change_status/src/index.ts',
+      name: 'Change Status',
+      trigger: 'ON_APPLICATION_SUBMIT',
```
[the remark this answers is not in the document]
CarlosNZ: e76a5c86

src/components/actions/coreActions.ts
CarlosNZ: e76a5c86

src/components/actions/coreActions.ts
CarlosNZ: Okay, I've updated it to be a bit simpler (although I think the logic is actually the same, just in a more concise expression): e76a5c86

Did you have something else in mind?

nmadruga: No... It isn't there yet. I know that Core-actions had a different version (not fully operational). I'll add my suggestion and start testing this on `core_templates` snapshot

src/components/actions/coreActions.ts
```diff
@@ -0,0 +1,486 @@
...
+      trigger: 'ON_REVIEW_SUBMIT',
+      event_code: null,
+      sequence: -1,
+      condition: {
+        operator: 'AND',
+        children: [
+          {
+            operator: 'objectProperties',
+            children: ['applicationData.reviewData.latestDecision.decision'],
+          },
```
nmadruga, line 389:
```suggestion
          {
            operator: 'objectProperties',
            children: ['applicationData.reviewData.isLastStage'],
          },
          {
            operator: 'objectProperties',
            children: ['applicationData.reviewData.isLastLevel'],
          },
```

CarlosNZ: Ah okay, needs to be last stage as well. Cool. 316aa22f

CarlosNZ:
> I can update the comments (I thought I'd already done most of them, albeit with different wording).

Comments updated. They correspond pretty closely to the original descriptions (with a little updated wording)

nmadruga:
Found while testing - Trigger error when trying to "Start Review"
Screen Shot 2022-11-28 at 3 22 25 PM

CarlosNZ:
> Found while testing - Trigger error when trying to "Start Review" Screen Shot 2022-11-28 at 3 22 25 PM
> 
> You can use `admin` and try to assign and start this review: http://localhost:3000/application/S-SWO-0009/review?tab=assignment

Yes, it's cos of this Action for "ON_REVIEW_ASSIGN", which refers to the old "LOCKED" field

Screenshot 2022-11-28 at 4 03 16 PM

So not related to this PR.

nmadruga:
Pending a small tweak required to get `changeOutcome` for non-reviewable template correctly differentiated (chat through dm)

src/components/actions/coreActions.ts
```diff
@@ -0,0 +1,484 @@
...
```
nmadruga, line 389: Please read carefully the suggestion below...
I advise you had used the suggestion instead of re-doing it, https://github.com/openmsupply/conforma-server/commit/316aa22fe0251ea0d6a5586580b078008629f7e6 is still wrong.

nmadruga (changes requested):

src/components/actions/coreActions.ts
CarlosNZ: I actually did do it, but I committed to #978 instead, sorry.

Done here now: ecb08ebc

CarlosNZ:
> Pending a small tweak required to get `changeOutcome` for non-reviewable template correctly differentiated (chat through dm)

Have done this, though I haven't personally tested it in context yet: f34883c2

nmadruga:
> > Pending a small tweak required to get `changeOutcome` for non-reviewable template correctly differentiated (chat through dm)
> 
> Have done this, though I haven't personally tested it in context yet: [f34883c](https://github.com/openmsupply/conforma-server/commit/f34883c2a496caf9698a1ec0a7cbf23dea4f938a)

Ok - I'm going to re-test all of this now. Can we merge everything to branch `feature/core-actions` now? I think it would be simpler to test and make any further tweaks before merging in to develop today : )

nmadruga (approved):

CarlosNZ:
> Ok - I'm going to re-test all of this now. Can we merge everything to branch `feature/core-actions` now? I think it would be simpler to test and make any further tweaks before merging in to develop today : )

I've merged all the approved PRs to `feature/core-actions`. It's just #980 that is pending now. If you want, I can merge that as well then make a new PR with the whole feature branch? Up to you.

PR #978, opened 2022-11-24
975(2) Core actions serial pattern

CarlosNZ:
Continuing implementation of #975

Edit from front-end using PR https://github.com/openmsupply/conforma-web-app/pull/1453

Adds "serial_pattern" field to "template" table, which is used by the core action to create the initial application serial.

### DIFF
--- a/src/components/actions/coreActions.ts
+++ b/src/components/actions/coreActions.ts
@@ -7,11 +7,28 @@ they've been hard-coded here in order to:
 - prevent accidental misconfiguration or removal
 */
 
+import DBConnect from '../databaseConnect'
 import { Trigger } from '../../generated/graphql'
 import { ActionInTemplate } from '../../types'
 
 type CoreActions = {
   [key in Trigger]?: Omit<ActionInTemplate, 'parameters_evaluated'>[]
+}
+
+export const getCoreActions = async (trigger: Trigger, templateId: number) => {
+  const currentCoreActions = coreActions?.[trigger] ?? []
+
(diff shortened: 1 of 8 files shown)


src/components/actions/coreActions.ts
```diff
@@ -7,13 +7,29 @@ they've been hard-coded here in order to:
...
 
+export const getCoreActions = async (trigger: Trigger, templateId: number) => {
+  const currentCoreActions = coreActions?.[trigger] ?? []
+
+  // Inject configuration over-rides for a limited selection of core action
+  // parameters (currently only serialPattern)
+  switch (trigger) {
+    case Trigger.OnApplicationCreate:
+      const serialPattern = await DBConnect.getTemplateSerialPattern(templateId)
+      if (serialPattern) currentCoreActions[0].parameter_queries.pattern = serialPattern
```
nmadruga, line 26: Maybe also check that this is the coreAction intended (in case we add another one in position 0)

nmadruga (commented):
Just one safe-guard missing

src/components/actions/coreActions.ts
CarlosNZ: But if `currentCoreActions[0]` is not "generateTextString", then we would have no (simple) way to know which one it *should* be.

So I'm not sure what you're suggesting. Throw an error? Seems a bit unnecessary for what is essentially a compile-time check. We shouldn't merge any code where these don't correspond, which should be obvious from a simple review and test (which you're doing now). What am I missing?

nmadruga (approved):

PR #979, opened 2022-11-25
ApplicationData fix

CarlosNZ:
Fixed the problem, but also added stronger Typescript definitions so we don't run in to this again. If we'd had this in place earlier, we wouldn't have been able to make this mistake in the first place. 

In hindsight, we really should have been using an automatic Types generator for PostGres like we have for GraphQL, but alas, we live and learn...

### DIFF
--- a/src/components/actions/getApplicationData.ts
+++ b/src/components/actions/getApplicationData.ts
@@ -1,4 +1,4 @@
-import { ActionApplicationData, ActionPayload } from '../../types'
+import { ActionApplicationData, ActionPayload, BaseApplicationData, ReviewData } from '../../types'
 import DBConnect from '../databaseConnect'
 import { BasicObject } from '@openmsupply/expression-evaluator/lib/types'
 import { getAppEntryPointDir } from '../utilityFunctions'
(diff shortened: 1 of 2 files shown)


src/components/actions/getApplicationData.ts
```diff
@@ -23,28 +23,26 @@ export const getApplicationData = async (input: {
...
+  const userData = {
+    firstName,
+    lastName,
+    orgName: organisation?.orgName || null,
+    orgId: organisation?.orgId || null,
+    username,
+    dateOfBirth,
+    email,
+    permissionNames,
+  }
```
CarlosNZ, line 45: This is the actual bug fix here -- which would never have happened had the other Typescript definitions been in place :)

src/components/actions/getApplicationData.ts
```diff
@@ -23,28 +23,26 @@ export const getApplicationData = async (input: {
...
-      organisation: orgName,
-      username,
-      dateOfBirth,
-      email,
-      permissionNames,
-    },
+    user: { firstName, lastName, organisation, username, dateOfBirth, email, permissionNames },
   } = await getUserInfo({
     userId: applicationData.userId,
-    username: applicationData.username,
```
nmadruga, line 43: We should also send username, since `getUserInfo` will send to `getUserOrgData` and that is getting by either one: orgId or Username

nmadruga (changes requested):
Looks great! Just one suggestion to fix please

src/components/actions/getApplicationData.ts
CarlosNZ: You only need to send one of `userId` or `username`, not both. `getUserOrgData` handles either :)

nmadruga (approved):

PR #980, opened 2022-11-25
#977 Re-write update review statuses action

### LINKED ISSUE #977: Re-write "Update Review Statuses" action
Fixes https://github.com/openmsupply/conforma-templates/issues/106

As per discussion [here](https://github.com/openmsupply/conforma-templates/issues/106)

We will incorporate this into [core actions](https://github.com/openmsupply/conforma-server/pull/976) once it's done, so please base this work off the `feature/core-actions` branch and we'll get everything working nicely before merging it together (and updating the snapshots accordingly)

@nmadruga -- assigning you as I think you're already made a start?

---

From discussion on other issue:

> As far as I can tell, it should do the following only:
> 
> - find any reviews linked to the current application stage/level that have "changedResponses" within the "assigned_sections" of the linked review_assignment and change their status (add new review_status_history) to "DRAFT"
> 
> Is that correct? (Maybe I'm missing something else?)
> 
> - single query to get reviews/review-assignments (don't consider sections)
> - bit of logic to filter out reviews whose assigned_sections don't overlap with the "changedResponses"
> - update the review status for those reviews (just call the existing "changeStatus" action on each one -- don't write additional logic or queries)
>

CarlosNZ:
Fix #977 

~~This is working well for application submissions, but I'm a bit unsure about what exactly should happen for review submissions (if anything?) @nmadruga I'll clarify this with you on Monday?~~

Okay, I think this is working correctly now, @nmadruga. 

You can try it out with snapshot ~~`conforma-templates/dev/snapshots/review_structure_testing.zip`~~ `conforma-templates/dev/snapshots/review_structure_testing_updated.zip`[^1] with the "Review Testing" template as it has 4 very small sections so you can go through application and multi-level review process very quickly. I've tried it out with a variety of reviewing use cases with LOQs and change requests, etc. and it seems to be working as expected. :) 

[^1]: Same as the crossed out one, but with "core_actions" removed from template now that I've merged the core actions from base branch.

I've also updated the documentation for this action as well as the "Core Actions" section, which mostly relates to #976 .


documentation/List-of-Action-plugins.md
```diff
@@ -515,10 +515,10 @@ Whenever an application or review is submitted, this Action "cleans up", by dele
...
 - _Action Code:_ **`trimResponses`**
 
-| Input parameters<br />(\*required) <br/> | Output properties |
-| ---------------------------------------- | ----------------- |
-| `applicationId` _OR_                     | `deletedIds`      |
-| `reviewId`                               | `updatedIds`      |
+| Input parameters<br />(\*required) <br/> | Output properties  |
+| ---------------------------------------- | ------------------ |
+| `applicationId` _OR_                     | `deletedResponses` |
+| `reviewId`                               | `updatedResponses` |
```
CarlosNZ, line 521: This was wrong.

plugins/action_change_status/src/changeStatus.ts
```diff
@@ -131,8 +133,9 @@ const changeReviewStatus = async (
       returnObject.error_log = 'Review Status not changed'
       returnObject.output = {
         status: newStatus,
-        statusId: currentStatus.status_history_id,
+        statusId: currentStatus.id,
         reviewStatusHistoryTimestamp: currentStatus.time_created,
+        reviewId,
```
CarlosNZ, line 138: Small tidy-ups in "changeStatus" action to make sure returned output was consistent whether or not the status actually changed.

plugins/action_update_review_status/src/databaseMethods.ts
```diff
@@ -0,0 +1,108 @@
...
+    try {
+      const result = await DBConnect.query({ text, values: [templateElements] })
+      return result.rows.map(({ code }: { code: string }) => code)
+    } catch (err) {
+      console.log(err.message)
+      throw err
+    }
+  },
+
+  getAssociatedReviewAssignmentNotStarted: async (
```
CarlosNZ, line 49: Everything below here is not (currently) in use. I'll delete them all once I've implemented the correct review submission behaviour.

plugins/action_update_review_status/src/updateReviewsStatuses_OLD.ts
```diff
@@ -0,0 +1,188 @@
+import {
```
CarlosNZ, line 1: Will delete before finishing, just leaving it here now so I can grab stuff from it for review submission.

plugins/action_update_review_status/src/updateReviewsStatuses.ts
```diff
@@ -0,0 +1,184 @@
...
+                reviewResponseDecision === ReviewResponseDecision.Disagree
+            )
+            .map(({ templateElementId }) => templateElementId)
+        )
+
+        const lowerReviewsToUpdate = reviews.filter(
+          ({ levelNumber, assignedSections }) =>
+            levelNumber < thisReviewLevel &&
+            assignedSections.some((section) => disagreedSections.includes(section))
+        )
```
CarlosNZ, line 148: This is updating ALL lower-level reviews. Should it just be level 1?

documentation/List-of-Action-plugins.md
```diff
@@ -698,48 +729,6 @@ The "condition" field (common to all template_actions) can also override the ori
...
-
-#### On Review Submit:
-
-- Change Status (to "Submitted")
-- Trim Responses
-- Update Review Statuses (for other reviews related to this review submission)
-- Increment Stage (if last level reviewer approves)
-- Generate Review Assignments (for next level review)
-- Update review response visibility (for applicant)
-- Change Status (Application, conditional on the review decision)
```
CarlosNZ, line 741: It seemed wasteful to duplicate this information when it's already well-documented/commented in the actual code. (aka I don't want to maintain two lists!)

plugins/action_update_review_status/src/updateReviewsStatuses.ts
```diff
@@ -0,0 +1,184 @@
...
+      const changedSections = await db.getChangedSections(
+        changedResponses.map(({ templateElementId }) => templateElementId)
+      )
+
+      // Filter out reviews that are not level 1 and don't intersect with
+      // changed section codes
+      const reviewsToUpdate = reviews.filter(
+        ({ reviewStatus, levelNumber, assignedSections }) =>
+          levelNumber === 1 &&
+          reviewStatus === ReviewStatus.Submitted &&
```
nmadruga, line 74: Just a though on this specific part - what if we have a Reviewer that is reviewing Section 2 (but still on DRAFT) and then after it was sent to Applicant for RFI there was a new version for one of the responses done by the applicant (even though it wasn't required in the RFI the applicant can change other responses too). Should we also change that Review (which wouldn't be SUBMITTED) to be Pending as well?

plugins/action_update_review_status/src/updateReviewsStatuses.ts
```diff
@@ -0,0 +1,184 @@
...
+      // REVIEW SUBMISSIONS:
+      if (!reviewId) throw new Error('Missing reviewId')
+
+      const thisReviewLevel = reviews.find((review) => review.reviewId === reviewId)?.levelNumber
+
+      if (!thisReviewLevel) throw new Error('Invalid reviewId')
+
+      const nextLevelReview = reviews.filter(
+        ({ levelNumber }) => levelNumber === thisReviewLevel + 1
+      )
```
nmadruga, line 113: Maybe check for not DISCONTINUED Reviews here. In case there was another Review of level +1 which isn't the current ASSIGNED one.

nmadruga: That's to cover the case for Re-assignment on consolidations

plugins/action_update_review_status/src/updateReviewsStatuses.ts
nmadruga: No. As per my previous comment on the documentation.

nmadruga: Also - I believe the same rule applies here. We shouldn't account for reviews that have status set to DISCONTINUED, since those are no longer the current ones... And therefor shouldn't be updated.

nmadruga (changes requested):
A few small tweaks - I'll start testing now also using the snapshot you pointed on description

plugins/action_update_review_status/src/updateReviewsStatuses.ts
CarlosNZ: Cool, I've added an additional filter to remove all "Discontinued" reviews right at the top (with comment and not in docs): 6c8b5b22

plugins/action_update_review_status/src/updateReviewsStatuses.ts
CarlosNZ: I think this would already cover that case, as they'd all be included in the application's "changedResponses", right?

nmadruga (approved):
Approving! Looks pretty solid. 

Needs more testing, but the aim is to do with `feature/core-actions` branch after a build, so it can be properly tested on a demo server also.

PR #986, opened 2022-12-02
Quick fix - remove current review from associated reviews to update

### LINKED ISSUE #985: Bug: When submitting one review it's been changed to PENDING
This is when using `feature/core-actions` branch, so only reporting a bug to be fixed prior to pre-release.

After the review is submitted the message is displayed (In Portuguese due to localisation)
Screen Shot 2022-12-02 at 11 40 13 AM

After that happens there is a "Reavaliar" action for the level 1 reviewer that just submitted
photo_2022-12-02 16 11 59

Basically displaying in the wrong moment - due to a change on status that shouldn’t have happened. I can investigate that in 30 minutes if someone else isn’t looking.

It is supposed to show when working on a consolidation and the lower level review submits, so new responses will be added to your review. So the message says: “You will need to refresh your page"

**Expected Result:**
Should be only changing other reviews ON_REVIEW_SUBMIT if:
- lower level reviews in case a reviewer (of last level) submits a review to get changes from lower level reviewer.
- higher level reviews in case it has started and a new lower level reviewer submits some more work

nmadruga:
Fixes #985
 filter out the current review from list of associated reviews

### DIFF
--- a/plugins/action_update_review_status/src/updateReviewStatuses.ts
+++ b/plugins/action_update_review_status/src/updateReviewStatuses.ts
@@ -60,6 +60,8 @@ const updateReviewStatuses: ActionPluginType = async ({
     const reviews = (await db.getAssociatedReviews(applicationId, stageNumber))
       // Ignore all "Discontinued" reviews
       .filter(({ reviewStatus }) => reviewStatus !== ReviewStatus.Discontinued)
+      // Ignore current review
+      .filter(({ reviewId: id }) => id !== reviewId)
 
     // APPLICATION SUBMISSIONS:
     if (triggeredBy === 'APPLICATION') {


plugins/action_update_review_status/src/updateReviewStatuses.ts
```diff
@@ -60,6 +60,8 @@ const updateReviewStatuses: ActionPluginType = async ({
     const reviews = (await db.getAssociatedReviews(applicationId, stageNumber))
       // Ignore all "Discontinued" reviews
       .filter(({ reviewStatus }) => reviewStatus !== ReviewStatus.Discontinued)
+      // Ignore current review
+      .filter(({ reviewId: id }) => id !== reviewId)
```
CarlosNZ, line 64: Great, but tiny bit more efficient to do it in one filter pass:
```
.filter(({ reviewStatus, reviewId: id }) => reviewStatus !== ReviewStatus.Discontinued &&
     id! == reviewId)
```

CarlosNZ (approved):
Awesome, thanks for the quick fix. One tiny nit pick, but not a big deal to leave as is if you'd rather.

plugins/action_update_review_status/src/updateReviewStatuses.ts
nmadruga: I just did like this so it's separated and clearer... I think will keep like that still

PR #987, opened 2022-12-02
#984 Prevent verification links possibly not being parsed correctly

### LINKED ISSUE #984: Password reset link failed via the create internal user form.
This issue occurred on the latest Angola build and template (with core actions migrations)

Using the create internal user form, the link generated wasn't working.

Going through the password reset form worked. 

Possibly a result of a error in verification link generation

See below second link (from second email) which was working.

6BEAFBD6-A1AC-41E1-9C3A-B6D457C50756

CarlosNZ:
Fix #984

Explanation of the problem this solves in the issue comment: https://github.com/openmsupply/conforma-server/issues/984#issuecomment-1334701491

One thing to consider -- should we do this for any other nano-ids? The only one I can think of is the file unique id, but I can't think of a case where we'd be putting the id as part of a url in a Markdown/text field. Can you? (If so, I'll make a general method)

### DIFF
--- a/plugins/action_create_verification/src/createVerification.ts
+++ b/plugins/action_create_verification/src/createVerification.ts
@@ -9,7 +9,7 @@ const createVerification: ActionPluginType = async ({ parameters, applicationDat
   const {
     applicationId = applicationData?.applicationId,
     expiry = null, // duration in hours
-    uniqueId = nanoid(24),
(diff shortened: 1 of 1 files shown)


fergie-nz (approved):
Looks great and works great!

Issue #983, opened 2022-11-30
Application List Performance


andreievg:
### Explain Postgres Query

* Identify the slow query (should see it in the browser dev tools, network tab)
* Add --allow-explain and --enhanced-graphiql when starting postgraphile, as [per this link](https://www.graphile.org/postgraphile/debugging/#step-4-viewing-the-generated-sql), i think this config goes in [this file](https://github.com/openmsupply/conforma-server/blob/develop/.postgraphilerc.js)
* Run that query in graphiql and see if there are any seq scans, I can help look into the analysis, sometimes it's hard to figure out what it means. You'll probably need to add the token to headers in graphiql

### Why is Docker much slower then Local instance

It's very fast when running locally on our m1 macs. And still very slow when running through Docker image on m1 mac, although the image is created for different architecture (to match our cloud provider virtual machine architecture).

* [ ] Try making image for M1 architecture, is it much slower then running it locally ?
* [ ] Is it postgres ? (try exposing postgres port in the image and running conforma server locally connected to postgres in image)
* [ ] If above is true, try running postgres image from dockerhub (might be a bit trickier to set up), and connecting local server to that image

CarlosNZ:
A good reference query is the following:

- use snapshot `conforma-templates/dev/snapshots/Angola_live_server_2022_11_28_2313_slow_list.zip` from the templates repo
- login as user "fergusInt" (PW: pass1234)
- visit url: `/applications?type=U2-OrgRegistration&user-role=reviewer&page=1&per-page=20`

It should be pretty much instant locally but takes 15+ seconds on the Angola server. (It's much slower with an internal user, presumably because of the reviewer action functions, and also slower with user who has many permissions)

CarlosNZ:
Also for reference: https://github.com/openmsupply/conforma-web-app/issues/1444

CarlosNZ:
Regarding what were were discussing IRL, if we kept intermediate tables or materialized views to store row access information, it would need to be refreshed whenever something changes (insert, update or delete) in the "permission_join" table, and I think that's it.

[...]

PR #989, opened 2022-12-06
983 - Improve application list performance

andreievg:
closes: #983

Please see last comment of issue for details.

I've tested this on the snapshot mentioned by @CarlosNZ in the issue, tried a few users and compared application list results, they all seem to match. I think I've spent majority of the time on the assigner actions, but figured out a pretty simple way to check (see comment and `assigned_sections_by_stage_and_level` view).

Carl suggested that I should remove unused methods, I did a search on both front end and back end (on front end camelCase on backend snake_case, I think it's safe to delete them but not 100% sure).

Tested with opening existing snapshot, seem to upgrade correctly

### DIFF
--- a/database/buildSchema/43_views_functions_triggers.sql
+++ b/database/buildSchema/43_views_functions_triggers.sql
@@ -1168,6 +1168,106 @@ $$
 LANGUAGE sql
 STABLE;
 
+-- Function to return elements of assigned questions for current stage/level
+CREATE OR REPLACE FUNCTION public.assigned_questions (app_id int, stage_id int, level_number int)
+    RETURNS TABLE (
+        review_id int,
+        response_id int,
+        review_assignment_id int,
+        review_response_code varchar,
+        review_response_status public.review_response_status,
+        decision public.review_response_decision,
+        is_optional boolean,
+        is_lastest_review boolean
+    )
+    AS $$
+    SELECT DISTINCT ON (review_response_code)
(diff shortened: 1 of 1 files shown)


nmadruga (changes requested):
Tested and I can see the improvement! It's great.

I have had a look on the changes to Function `application_list` and seems all good to me. Although for the removed functions, I can list where these have been used and please ask to get the ones associated to be replaced 🙏 

- `public.assigned_questions` and (not removed) `public.reviewable_questions` used in Front-end query file `Conforma-web-app/src/utils/graphql/queries/getReviewableQuestionCounts.query.ts`
- `public.assigned_questions` and `public.reviewable_questions_count` (which calls `pubic. reviewable_questions`)
in "IncrementStage" Graphql query (used currently in ProdRegistration on the Angola project - check PR https://github.com/openmsupply/conforma-web-app/pull/1391)

andreievg:
@nmadruga I've reverted function deletes

nmadruga (changes requested):
Andrei, we can remove the ones that were only used by the `application_list`: so `review_list`, `assigner_list` and `assignment_list`. Could you still remove these please?

nmadruga (approved):

nmadruga:
As discussed per DM Andrei will create a separated issue to remove these functions - which aren't in use:
- review_listassigner_list and assignment_list

PR #990, opened 2022-12-07
#109T Fix updateReviewStatuses action not considering its Triggered by Review when passing reviewId

nmadruga:
Fixes https://github.com/openmsupply/conforma-templates/issues/109

Just a change on `updatedReviewStatuses` action that caused a regression. The action was supposed to receive the `reviewId` and interpret this action as triggered by Review (it was considering it was triggered by Application before)

Fix problem with re-submitting a review wasn't changing the consolidation review status

### Testing
- Check sending `Changes Requested` to a lower level reviewer while on a consolidation review
  - The lower level reviewer should be able to re-submit (but set another Rejection) and send again as Non-conform
  - This issue fixes that now the Consolidator should be able to do a Re-review of the consolidation
- Also test sending RFI to Applicant
  - Check that Application re-submission is also working

### DIFF
--- a/plugins/action_update_review_status/src/updateReviewStatuses.ts
+++ b/plugins/action_update_review_status/src/updateReviewStatuses.ts
@@ -46,7 +46,7 @@ const updateReviewStatuses: ActionPluginType = async ({
   // action
   const changedResponses: ChangedResponse[] = parameters.changedResponses || []
 
-  const triggeredBy: TriggeredBy = parameters.triggeredBy || 'APPLICATION'
+  const triggeredBy: TriggeredBy = reviewId ? 'REVIEW' : parameters.triggeredBy || 'APPLICATION'
 
   console.log(
     `Updating statuses of reviews associated with ${
@@ -60,8 +60,6 @@ const updateReviewStatuses: ActionPluginType = async ({
     const reviews = (await db.getAssociatedReviews(applicationId, stageNumber))
       // Ignore all "Discontinued" reviews
       .filter(({ reviewStatus }) => reviewStatus !== ReviewStatus.Discontinued)
-      // Ignore current review
-      .filter(({ reviewId: id }) => id !== reviewId)
 
     // APPLICATION SUBMISSIONS:
     if (triggeredBy === 'APPLICATION') {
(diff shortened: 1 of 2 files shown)


plugins/action_update_review_status/src/updateReviewStatuses.ts
```diff
@@ -60,8 +60,6 @@ const updateReviewStatuses: ActionPluginType = async ({
     const reviews = (await db.getAssociatedReviews(applicationId, stageNumber))
       // Ignore all "Discontinued" reviews
       .filter(({ reviewStatus }) => reviewStatus !== ReviewStatus.Discontinued)
-      // Ignore current review
-      .filter(({ reviewId: id }) => id !== reviewId)
```
CarlosNZ, line 64: Does this not work for application triggers if it's here?

src/components/actions/coreActions.ts
```diff
@@ -336,6 +336,10 @@ const coreActions: CoreActions = {
           operator: 'objectProperties',
           children: ['outputCumulative.updatedResponses'],
         },
+        reviewId: {
+          operator: 'objectProperties',
+          children: ['applicationData.reviewData.reviewId'],
+        },
```
CarlosNZ, line 342: You shouldn't need this -- it gets if from `applicationData` if present.

CarlosNZ: Every action checks applicationData for applicationId and reviewId. You only need to provide it is a parameter if you're wanting to override the current value (which we would rarely want to do).

plugins/action_update_review_status/src/updateReviewStatuses.ts
nmadruga: It is needed because on line 109 it needs to get the `current review level`: const thisReviewLevel = reviews.find((review) => review.reviewId === reviewId)?.levelNumber

src/components/actions/coreActions.ts
nmadruga: It's either this or we need to set `triggerBy` REVIEW, I choose to pass the `reviewId`

CarlosNZ: That's what I'm saying-- you don't need to pass it, it's already there (look at the reviewId=... at the top).

CarlosNZ: I'm confused -- this is for review submissions right?

I really don't see how you passing "applicationData.reviewData.reviewId" would be any different to it getting it automatically from the same applicationData. If it _is_ different, then there's something more fundamentally wrong and we should be fixing that.

PR #995, opened 2022-12-09
Migrate core templates to use core actions

CarlosNZ:
- update all templates to use core actions
- add ["defaultValue" fix ](https://github.com/openmsupply/conforma-server/issues/988#issuecomment-1340285087)to "Password Reset"
- snapshot is now in "pg_dump" format -- hope this is okay?


nmadruga (changes requested):
Just did tests to make sure this is running smoothly, and I suspect we need to have at leat one Reviewer with the existing permissions (to review "Sample template" permission_name `reviewGeneral`) so this action won't break the App: Trigger error notices ON_SUBMIT_APPLICATION

Screen Shot 2022-12-20 at 10 05 18 AM

CarlosNZ:
> Just did tests to make sure this is running smoothly, and I suspect we need to have at leat one Reviewer with the existing permissions (to review "Sample template" permission_name `reviewGeneral`) so this action won't break the App: Trigger error notices ON_SUBMIT_APPLICATION

Okay, makes more sense to protect against this case (action shouldn't fail if it can't find any valid reviewers). Thanks, I'll look into it.

nmadruga:
Thanks! I'll have another look now

CarlosNZ:
> > Just did tests to make sure this is running smoothly, and I suspect we need to have at leat one Reviewer with the existing permissions (to review "Sample template" permission_name `reviewGeneral`) so this action won't break the App: Trigger error notices ON_SUBMIT_APPLICATION
> 
> Okay, makes more sense to protect against this case (action shouldn't fail if it can't find any valid reviewers). Thanks, I'll look into it.

Done: e8d49fc1

(I actually think the "permissions_all" view probably shouldn't be returning records with `null` users (maybe needs a LEFT JOIN rather than a full JOIN), but I'm reluctant to start messing around with that at this stage as too many things depend on it. So this small guard in the action will do for now.)

nmadruga (changes requested):
Another trigger error - this one seems less alarming, but good to fix. 

When I submit a review approving all  - it fails to modify a record

Screen Shot 2022-12-20 at 10 49 25 AM

Can you also remove "Apply" permission from Admin user please? Ir looks confusing on UI when I grant the `reviewGeneral` to admin for testing, and this user can also act as Applicant

CarlosNZ:
> Can you also remove "Apply" permission from Admin user please? Ir looks confusing on UI when I grant the `reviewGeneral` to admin for testing, and this user can also act as Applicant

No, because then there would be no users that can apply for anything. This is the snapshot I use when working on plugins etc, and we need quick access to be able to apply for applications. It's not meant to be used in a proper setting, it's just a dev snapshot (and a starting point for some templates). No one should be using this as a basis for an actual real system.

CarlosNZ:
> Another trigger error - this one seems less alarming, but good to fix.
> 
> When I submit a review approving all - it fails to modify a record
> 
> Screen Shot 2022-12-20 at 10 49 25 AM

The old "logo_url" error that we've fixed a million times on the other snapshots! 🙄

Done: 797397c0

nmadruga (approved):

PR #999, opened 2023-01-12
Generate Filter Data fix

CarlosNZ:
Urgent fix for filter data not being generated for new Chemical Import permits (Fiji).

Reason was (and why I hadn't noticed before) is that the "table name" used in "modifyRecord" was "permit_chemical", whereas this method had been expecting camelCase table names (which is what the data view definition is saved as in the "table" column).

So now I've just made it that it'll work for any type of table name specified in modifyRecord. So all of these inputs will work:
- `permitChemical`
- `dataTablePermitChemical`
- `data_table_permit_chemical`
- `permit_chemical`

### DIFF
--- a/src/components/data_display/generateFilterDataFields/generateFilterDataFields.ts
+++ b/src/components/data_display/generateFilterDataFields/generateFilterDataFields.ts
@@ -96,9 +96,9 @@ export const generateFilterDataFields = async (table: string, fullUpdate: boolea
     }
 
     // Delete unused (no filter definitions) columns
-    // for (const { name } of currentColumns) {
-    //   await db.dropColumn(tableNameFull, name)
-    // }
+    for (const { name } of currentColumns) {
+      await db.dropColumn(tableNameFullSnake, name)
+    }
 
     // Iterate over all data table records and update their filter field values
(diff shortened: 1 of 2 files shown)


src/components/data_display/generateFilterDataFields/generateFilterDataFields.ts
```diff
@@ -87,12 +96,12 @@ export const generateFilterDataFields = async (table: string, fullUpdate: boolea
     }
 
     // Delete unused (no filter definitions) columns
-    for (const { name } of currentColumns) {
-      await db.dropColumn(tableNameFull, name)
-    }
+    // for (const { name } of currentColumns) {
+    //   await db.dropColumn(tableNameFull, name)
+    // }
```
nmadruga, line 101: Not needed?

nmadruga (approved):
Looks good - just one comment

src/components/data_display/generateFilterDataFields/generateFilterDataFields.ts
CarlosNZ: Oops, yes that is needed. I just turned it off temporarily so it wouldn't keep deleting a column. I'll restore it.

CarlosNZ: 6a777798

PR #1002, opened 2023-01-16
Allow entering initialDate to generate an expiry date

nmadruga:
Required for issue https://github.com/openmsupply/conforma-templates/issues/113
to enable generating an expiry_date prior to creation of product - so it's based on the `statusHistoryDate` of currentStage. And that's the same one which will be used afterwards when creating the product.expiry_date on last stage approval


CarlosNZ (approved):
Yeah cool, looks good.

Tiny thing -- maybe structure it the same as the "getFormattedDate" method, just for consistency, but not really a big deal.

Also -- can you please copy the change to the front-end version so they stay in sync (can push that one straight to develop as it'll be exactly the same as this)

PR #1003, opened 2023-01-16
#997 Lookup table type inference

### LINKED ISSUE #997: Lookup tables: Parse JSON strings to objects on import
In order to be able to imported the exported Medsafe data to a Lookup table, the import module will need to be able to parse JSON strings (from the CSV file) and save them as `jsonb` objects in the database. Currently, it treats everything as a string.

CarlosNZ:
Fix #997 

Requires front-end https://github.com/openmsupply/conforma-web-app/pull/1471 in order for the UI to work with this.

Tried to be as minimally invasive to the existing code as possible, so this just intercepts the existing imported data and mutates the fieldMap and data rows in-place to conform to inferred data type. The postgres fields will be one of:
- `varchar` (what they all are by default)
- `integer`
- `double precision`
- `boolean`
- `timestamptz`
- `jsonb` (the one we really want for scraped data import)

Test it with the following.

Import this lookup table, and notice the different data types in the created table:
[test_data.csv](https://github.com/openmsupply/conforma-server/files/10424847/test_data.csv)

Then import this one over the top to see the effect of modifying the exported CSV.
[test_2023-01-16.csv](https://github.com/openmsupply/conforma-server/files/10424848/test_2023-01-16.csv)

You can export it again yourself and change/add some data to check re-import as well.

### DIFF
--- a/src/lookup-table/models/LookupTable.model.ts
+++ b/src/lookup-table/models/LookupTable.model.ts
@@ -159,7 +159,7 @@ const LookupTableModel = () => {
     tableName: string,
     fieldMaps: FieldMapType[]
   ): Promise<boolean> => {
-    const text = `UPDATE data_table SET field_map = $1 WHERE display_name = $2`
+    const text = `UPDATE data_table SET field_map = $1 WHERE table_name = $2`
     try {
       await DBConnect.query({ text, values: [JSON.stringify(fieldMaps), tableName] })
       return true
--- a/src/lookup-table/models/LookupTable.model.ts
+++ b/src/lookup-table/models/LookupTable.model.ts
@@ -8,6 +8,7 @@ import {
   LookupTableStructure,
 } from '../types'
 import config from '../../config'
+import { exportDataRows } from '../utils/dataTypeUtils'
 
 const { dataTablePrefix } = config
 
@@ -17,7 +18,7 @@ const LookupTableModel = () => {
     const fields = fieldMap.map(mappedField).join(',')
     const text = `SELECT ${fields} FROM ${dataTablePrefix}${tableName}`
     const result = await DBConnect.query({ text })
-    return result.rows
+    return exportDataRows(fieldMap, result.rows)
   }
 
   const createStructure = async ({ tableName, displayName, fieldMap }: LookupTableStructure) => {
--- a/src/lookup-table/utils/dataTypeUtils.ts
+++ b/src/lookup-table/utils/dataTypeUtils.ts
@@ -33,6 +33,20 @@ export const setDataTypes = (fieldMaps: FieldMapType[], rows: object[]) => {
   })
 }
 
+// Prepares table export by stringifying jsonb fields
+// Mutates rows *in-place*
+export const exportDataRows = (fieldMaps: FieldMapType[], rows: { [key: string]: any }[]) => {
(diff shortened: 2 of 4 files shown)


src/lookup-table/models/LookupTable.model.ts
```diff
@@ -158,7 +159,7 @@ const LookupTableModel = () => {
     tableName: string,
     fieldMaps: FieldMapType[]
   ): Promise<boolean> => {
-    const text = `UPDATE data_table SET field_map = $1 WHERE display_name = $2`
+    const text = `UPDATE data_table SET field_map = $1 WHERE table_name = $2`
```
CarlosNZ, line 162: This was an existing mistake -- the field maps weren't getting updated when new columns were added due to this.

src/lookup-table/utils/dataTypeUtils.ts
```diff
@@ -0,0 +1,103 @@
...
+        const type = getType(value)
+        if (type !== null) {
+          fieldMaps[index].dataType = type
+          done.push(key)
+        }
+      }
+    })
+
+    if (done.length === fieldMaps.length) break
+  }
```
CarlosNZ, line 20: It just checks the first value of each column to determine its type. However, we can have `null` values, so it iterates through all rows until it's found at least one non-null value for each field, then it breaks out of the loop.

src/lookup-table/utils/dataTypeUtils.ts
```diff
@@ -0,0 +1,103 @@
...
+
+  const fieldMap: { [key: string]: any } = fieldMaps.reduce((acc, curr) => {
+    const name = curr.fieldname
+    const type = curr.dataType
+    return { ...acc, [name]: type }
+  }, {})
```
CarlosNZ, line 26: This is just a compact version of the main `fieldMaps` array, with a simple `{ key: type }` object. Makes the subsequent lookup a bit cleaner.

src/lookup-table/utils/dataTypeUtils.ts
```diff
@@ -0,0 +1,103 @@
...
+
+  // Now mutate rows based on those types
+  rows.forEach((row, index) => {
+    rows[index] = mapValues(row, (value, key) => {
+      if (key === 'id') return value
+      return convertType(value, fieldMap[key])
```
CarlosNZ, line 32: All values coming from CSV files are strings, so this converts the actual data to the correct type (i.e. `"2"` becomes `2`, `"FALSE"` becomes `false` etc.

src/lookup-table/utils/dataTypeUtils.ts
```diff
@@ -0,0 +1,103 @@
...
+// Prepares table export by stringifying jsonb fields
+// Mutates rows *in-place*
+export const exportDataRows = (fieldMaps: FieldMapType[], rows: { [key: string]: any }[]) => {
+  const jsonFields = fieldMaps.filter((e) => e.dataType === 'jsonb').map((e) => e.fieldname)
+
+  rows.forEach((row) => {
+    jsonFields.forEach((jsonField: string) => {
+      row[jsonField] = row[jsonField] !== null ? JSON.stringify(row[jsonField]) : ''
+    })
+  })
```
CarlosNZ, line 46: Without this, the json fields get exported as `[Object object]`, so this stringifies those fields. All other fields get dumped out as strings (like they were when imported).

nmadruga (approved):
Working well
Code looks good

PR #1005, opened 2023-01-18
#1004 Add new filter to remove elements in array

### LINKED ISSUE #1004: Filter out results on arrays before displaying on Plugin element
Required specifically for removing the `admin` preferences to be selected in templates e.g. GrantUserPermission.

### Existing implementation

Currently the list of permissions is displayed by `Checkbox` plugin element, which only uses an array passed on (or received by a API query result) to be displayed, there is no filters applied.
Screen Shot 2023-01-17 at 1 48 06 PM

The current API query used will return ALL permissions based on the orgId and/or userId received:
Screen Shot 2023-01-17 at 1 48 30 PM

It would be tricky to add some filtering to the `Checkbox` plugin or to change the existing `user-permissions` end-point

### Proposed Solution
Create a new evaluator custom function to filter an array, which would be what the `checkboxes` parameter would call, and pass as one of the parameters the result from the API query to `user-permissions`. 

Syntax:

```
filter( valuesArray, options }

Where "options" is:
{
  key?: string
  values: string | string[] (in principle could be numbers or any primitive type)
  rule?: "exclude" | "include (default: "exclude")
}
```

Parameters types explained:
`valuesArray`: array of objects OR strings
`key`: which key to filter (optional -- omitted means the input is an array of strings/primitives not array of objects
`values`: array of values to filter OR single value
`rule`: enum to say if should include or exclude these keys on the result (default "exclude" would you say?)

nmadruga:
Fixes #1004

Required a new evaluator function to remove permissions from the array returned by user-permissions, so not all of the permissions displays in specific templates. Work to be used in configuration of https://github.com/openmsupply/conforma-templates/issues/55

### DIFF
--- a/src/components/actions/evaluatorFunctions.ts
+++ b/src/components/actions/evaluatorFunctions.ts
@@ -8,6 +8,42 @@ Any changes done here should also be replicated in front-end
 
 import { DateTime, Duration } from 'luxon'
 
+type FilterRule = 'exclude' | 'include'
+interface FilterOptions {
+  key?: string
+  rule?: FilterRule
+  values: string | string[]
+}
+
+const filterArray = (valuesArray: Object[] | string[], options: FilterOptions) => {
+  const { key, rule = 'exclude', values } = options
+  if (
+    (valuesArray.length > 0 && key && typeof valuesArray[0] === 'object') ||
+    (!key && typeof valuesArray[0] === 'string')
+  )
+    return key
+      ? (valuesArray as Object[]).filter(
+          (
+            obj // Array of strings
+          ) =>
+            Object.entries(obj).find(
+              ([keyMap, value]) =>
+                keyMap === key &&
+                (rule === 'include' ? values.includes(value) : !values.includes(value))
+            )
+        )
+      : (valuesArray as string[]).filter(
+          (
+            str // Array of objects
+          ) => (rule === 'include' ? values.includes(str) : !values.includes(str))
+        )
+  else {
+    if (valuesArray.length > 0)
+      console.log('Type of elements in array mismatch filter options', options)
+    return valuesArray
+  }
+}
+
 const generateExpiry = (duration: Duration, startDate?: string | Date) => {
   const date = startDate
     ? typeof startDate === 'string'
@@ -65,6 +101,7 @@ const extractNumber = (input: string) => {
 const removeAccents = (input: string) => input.normalize('NFD').replace(/[\u0300-\u036f]/g, '')
 
 export default {
+  filterArray,
   generateExpiry,
   getYear,
   getFormattedDate,
--- a/src/components/actions/evaluatorFunctions.ts
+++ b/src/components/actions/evaluatorFunctions.ts
@@ -25,8 +25,8 @@ const filterArray = (valuesArray: unknown[], options: FilterOptions) => {
   return valuesArray.filter((element) => {
     if (key && isObject(element))
       return Object.entries(element as Object).find(
-        ([keyMap, value]) =>
-          keyMap === key &&
+        ([objKey, value]) =>
+          objKey === key &&
           (rule === 'include' ? compareValues.includes(value) : !compareValues.includes(value))
       )
     else
--- a/src/components/actions/evaluatorFunctions.ts
+++ b/src/components/actions/evaluatorFunctions.ts
@@ -8,40 +8,27 @@ Any changes done here should also be replicated in front-end
 
 import { DateTime, Duration } from 'luxon'
 
-type FilterRule = 'exclude' | 'include'
 interface FilterOptions {
   key?: string
-  rule?: FilterRule
+  rule?: 'exclude' | 'include'
   values: string | string[]
 }
 
-const filterArray = (valuesArray: Object[] | string[], options: FilterOptions) => {
+const filterArray = (valuesArray: unknown[], options: FilterOptions) => {
   const { key, rule = 'exclude', values } = options
-  if (
-    (valuesArray.length > 0 && key && typeof valuesArray[0] === 'object') ||
-    (!key && typeof valuesArray[0] === 'string')
-  )
-    return key
-      ? (valuesArray as Object[]).filter(
-          (
-            obj // Array of strings
-          ) =>
-            Object.entries(obj).find(
-              ([keyMap, value]) =>
-                keyMap === key &&
-                (rule === 'include' ? values.includes(value) : !values.includes(value))
-            )
-        )
-      : (valuesArray as string[]).filter(
-          (
-            str // Array of objects
-          ) => (rule === 'include' ? values.includes(str) : !values.includes(str))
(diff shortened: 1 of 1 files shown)


src/components/actions/evaluatorFunctions.ts
```diff
@@ -8,6 +8,42 @@ Any changes done here should also be replicated in front-end
 
 import { DateTime, Duration } from 'luxon'
 
+type FilterRule = 'exclude' | 'include'
+interface FilterOptions {
+  key?: string
+  rule?: FilterRule
```
CarlosNZ, line 14: Why not just:
```
rule?: 'exclude' | 'include'
```

src/components/actions/evaluatorFunctions.ts
```diff
@@ -8,6 +8,42 @@ Any changes done here should also be replicated in front-end
...
+  values: string | string[]
+}
+
+const filterArray = (valuesArray: Object[] | string[], options: FilterOptions) => {
+  const { key, rule = 'exclude', values } = options
+  if (
+    (valuesArray.length > 0 && key && typeof valuesArray[0] === 'object') ||
```
CarlosNZ, line 21: `valuesArray.length > 0`

No need for this restriction, is there?

src/components/actions/evaluatorFunctions.ts
```diff
@@ -8,6 +8,42 @@ Any changes done here should also be replicated in front-end
...
 import { DateTime, Duration } from 'luxon'
 
+type FilterRule = 'exclude' | 'include'
+interface FilterOptions {
+  key?: string
+  rule?: FilterRule
+  values: string | string[]
+}
+
+const filterArray = (valuesArray: Object[] | string[], options: FilterOptions) => {
```
CarlosNZ, line 18: I don't think it should be restricted to just objects and strings -- any primitive type should work fine.

src/components/actions/evaluatorFunctions.ts
```diff
@@ -8,6 +8,42 @@ Any changes done here should also be replicated in front-end
...
+  const { key, rule = 'exclude', values } = options
+  if (
+    (valuesArray.length > 0 && key && typeof valuesArray[0] === 'object') ||
+    (!key && typeof valuesArray[0] === 'string')
+  )
+    return key
+      ? (valuesArray as Object[]).filter(
+          (
+            obj // Array of strings
```
CarlosNZ, line 27: `obj // Array of strings`
Don't you mean "Array of objects"?

src/components/actions/evaluatorFunctions.ts
```diff
@@ -8,6 +8,42 @@ Any changes done here should also be replicated in front-end
...
+            Object.entries(obj).find(
+              ([keyMap, value]) =>
+                keyMap === key &&
+                (rule === 'include' ? values.includes(value) : !values.includes(value))
+            )
+        )
+      : (valuesArray as string[]).filter(
+          (
+            str // Array of objects
+          ) => (rule === 'include' ? values.includes(str) : !values.includes(str))
```
CarlosNZ, line 38: It's nice `values.includes` works for both arrays and strings, but unfortunately it doesn't work for other primitive types.

Probably best to convert it all to an array at the top:
```
const compareValues = Array.isArray(values) ? values : [ values ]
```

src/components/actions/evaluatorFunctions.ts
CarlosNZ: Probably no need to check the type of the first item.

I would just check *each* item as part of the filter and if it's not an object then:
- exclude it if the rule is "include"
- include it if the rule is "exclude"

Then it follows the rule in a way that makes sense regardless of what type it is :)

src/components/actions/evaluatorFunctions.ts
```diff
@@ -8,6 +8,42 @@ Any changes done here should also be replicated in front-end
...
+        )
+  else {
+    if (valuesArray.length > 0)
+      console.log('Type of elements in array mismatch filter options', options)
```
CarlosNZ, line 42: Following the type check suggestion above, you won't need this at all.

src/components/actions/evaluatorFunctions.ts
```diff
@@ -8,6 +8,42 @@ Any changes done here should also be replicated in front-end
...
+}
+
+const filterArray = (valuesArray: Object[] | string[], options: FilterOptions) => {
+  const { key, rule = 'exclude', values } = options
+  if (
+    (valuesArray.length > 0 && key && typeof valuesArray[0] === 'object') ||
+    (!key && typeof valuesArray[0] === 'string')
+  )
+    return key
+      ? (valuesArray as Object[]).filter(
```
CarlosNZ, line 25: Also, with proper type checking within the filter, you won't need the `as`.

CarlosNZ (changes requested):
Cool, thanks. Seems to mostly be working okay. Few suggestions added.

src/components/actions/evaluatorFunctions.ts
CarlosNZ: This is a good case for using `valuesArray: unknown[]` and then having proper type guards within (as suggested above).

nmadruga:
Carl I tried to address your suggestions in this commit [aa57ea1](https://github.com/openmsupply/conforma-server/pull/1005/commits/aa57ea142fb1b0500ba2c8a0fd01bde38af15e23)

CarlosNZ:
> Carl I tried to address your suggestions in this commit [aa57ea1](https://github.com/openmsupply/conforma-server/pull/1005/commits/aa57ea142fb1b0500ba2c8a0fd01bde38af15e23)

Cool, looking really nice. Looks like it's all correct, but would you mind copying to the front-end branch as well so I can just do some quick tests in the UI?

src/components/actions/evaluatorFunctions.ts
```diff
@@ -8,6 +8,29 @@ Any changes done here should also be replicated in front-end
...
+interface FilterOptions {
+  key?: string
+  rule?: 'exclude' | 'include'
+  values: string | string[]
+}
+
+const filterArray = (valuesArray: unknown[], options: FilterOptions) => {
+  const { key, rule = 'exclude', values } = options
+  const compareValues = Array.isArray(values) ? values : [values]
+  return valuesArray.filter((element) => {
```
CarlosNZ, line 20: `typeof element === 'object'` isn't *quite* enough as both `null` and arrays return true.

It'll need to be `typeof element === 'object' && !Array.isArray(element) && element !== null`

Feel free to abstract this out into a `isObject` utility function if you like.

nmadruga: Hmmm no problem. I'll do this now

CarlosNZ (approved):
Cool, thanks. 

See my tiny comment for front-end.

Approved, assuming both front and back-end will be made identical before merging :) 

👍

PR #1006, opened 2023-01-23
Add column registration on application table

nmadruga:
Fixes #55T

Adding an extra column to `application` table since it cannot be modified using `modifyRecord` due to our own restrictions, we should always add columns in the database when application needs to be altered.

With that in place we can store the generated registration to the application record in a different stage than the one that needs to use the same registration (and add to the product).

### DIFF
--- a/database/buildSchema/18_application.sql
+++ b/database/buildSchema/18_application.sql
@@ -14,11 +14,11 @@ CREATE TABLE public.application (
     org_id integer REFERENCES public.organisation (id) ON DELETE CASCADE,
     session_id varchar,
     serial varchar UNIQUE,
-    registration varchar UNIQUE,
     name varchar,
     outcome public.application_outcome DEFAULT 'PENDING',
+    outcome_registration varchar UNIQUE,
     is_active bool,
     is_config bool DEFAULT FALSE,
-    TRIGGER public.trigger,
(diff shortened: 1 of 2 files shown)


database/buildSchema/18_application.sql
```diff
@@ -14,10 +14,11 @@ CREATE TABLE public.application (
     org_id integer REFERENCES public.organisation (id) ON DELETE CASCADE,
     session_id varchar,
     serial varchar UNIQUE,
+    registration varchar UNIQUE,
     name varchar,
     outcome public.application_outcome DEFAULT 'PENDING',
     is_active bool,
     is_config bool DEFAULT FALSE,
-    TRIGGER public.trigger
+    TRIGGER public.trigger,
```
CarlosNZ, line 22: Shouldn't have final comma (don't think it'll work)

CarlosNZ (approved):
Cool, thanks.

Just one little tweak please. 

Also, was wondering if "registration" was a bit vague, maybe "outcome_registration" or "approval_registration" might be clearer, but I don't love either of those, so up to you.

(We already have a "registration" field on "organisation" which means something slightly different, so just trying to avoid confusion).

nmadruga:
> outcome_registration

Ok - I'll go with this one

PR #1008, opened 2023-01-25
#1007 Template export fix

### LINKED ISSUE #1007: Bug: template export not working
Click to export a template. The .zip file will download, but it's corrupted/unreadable.

Probably to do with the change to the zip library when adding the backup functionality, but will need looking into.

CarlosNZ:
Fix #1007 

Turns out this was mostly a front-end problem -- see PR https://github.com/openmsupply/conforma-web-app/pull/1475

However, there was one `await` missing here. I guess it wasn't essential as it wasn't causing any real problems, but it's slightly more thorough.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "conforma-server",
-  "version": "0.5.3-0",
+  "version": "0.5.3-1",
   "main": "server.js",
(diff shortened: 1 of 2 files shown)


nmadruga (approved):
👍

PR #1011, opened 2023-02-17
1010 rename default value column

### LINKED ISSUE #1010: Rename "default_value" field to "initial_value" to avoid confusion with individual plugin defaults
### Back-end:
- write migration to change field name
- update .sql script as well

### Front-end:
- update handling of "defaultValue" in "ApplicationViewWrapper"
- update documentation to explain it (there's already an explanation of the difference, just needs to be updated -- should be simpler now)
- check if it's used anywhere else? (I don't think so)

CarlosNZ:
Closes #1010 

Not much required for back-end, just a little migration.


nmadruga (approved):
👍

PR #1014, opened 2023-02-28
#1013 Filter list improvements

### LINKED ISSUE #1013: Filter list -- query takes too long in certain conditions
When a data table / lookup table has many thousands of records, yet only a small number of distinct values for a given field (i.e. less than 10), then the query to fetch the filter list is incredibly slow. This is because of the way the back-end builds the filter list -- it fetches records in batches of 10 (the list length) until it finds 10 distinct values. In the case described above, this will result in querying the whole table 10 records at a time, which means hundreds or even thousands of queries just to get a small number of results.

I think the simplest solution here is to just create an additional [Filter parameter](https://github.com/openmsupply/conforma-server/wiki/Data-View-Filters#filter-types-and-available-options) where a filter list can be manually defined. If this parameter is present, the server doesn't make its own queries to build the list, it just returns the pre-defined one. 

Use case is the "Authority" field in the list of Pre-registered provisional products -- the only possible values are "Australia" and "New Zealand", and the lookup table is over 27,000 records long!

So we'd simply define the parameter `{filterList: ["Australia", "New Zealand"]}` to prevent the server having to run this time-consuming process.

Another possibility is to just increase the batch-size to something like 1,000. That means it'll fetch way too many records for most filters (when most of the values are different), but it's probably still performant enough, and better overall.

Probably a good idea to do both of these -- we'd use the first solution for the "Authority" field, and rely on the second one for "Route of Adminstration", which also suffers from this problem, but is even more annoying when you try and shorten the list with filters.

CarlosNZ:
Fixes #1013 

Implements both solutions as outlined in the issue.

Open to other suggestions for ways to make this operation more efficient (No, SQL "DISTINCT" won't work), although this table with over 27,000 records is about as big as it would get.

To test:

Use snapshot `/dev/snapshots/AAA_Fiji_Provisional_registration_2023_03_01_1005.zip` in templates repo, and view the "Pre-registered products" table from the "Database" menu.

Note the performance difference compared with the base branch for the following filter lists:
- "Authority" -- now uses the manually defined "New Zealand" / "Australia" list, so should be no delay
- "Route of Administration" -- the batch size difference makes the filter list load in less than a second as opposed to several seconds with the small batch size..


fergie-nz (approved):
This works really well.

I do feel slightly concerned at raising the configuration overhead, as this will be another thing that needs to be configured for any dynamic data list. A common case might be changing languages.

I wonder if it would be possible to test? Even something simple like: Do these options actually exist in the list they are filtering

CarlosNZ:
> > I do feel slightly concerned at raising the configuration overhead, as this will be another thing that needs to be configured for any dynamic data list. A common case might be changing languages.
> 
> It shouldn't add much -- most of the time you won't need to think about it, it'll be fine with defaults. This is a solution for a pretty edge case, and the batchSize solution makes it even more unnecessary in most cases.
> 
> > I wonder if it would be possible to test? Even something simple like: Do these options actually exist in the list they are filtering
> 
> You talking about if you define a "custom list"? I don't think they even need to exist, you could use it to add filter values that don't exist (yet) but might at some point in the future.

fergie-nz:
> You talking about if you define a "custom list"? I don't think they even need to exist, you could use it to add filter values that don't exist (yet) but might at some point in the future. You can test them as they'll show up immediately in the filter list UI.

I am thinking about a case where we use a snapshot as a base, then change languages (including the authority list) yet don't change the default filter. In this case, I don't think I would think to check the default filters of data views. 

Am I thinking this through correctly?

CarlosNZ:
> I am thinking about a case where we use a snapshot as a base, then change languages (including the authority list) yet don't change the default filter. In this case, I don't think I would think to check the default filters of data views.

Yes, but *everything* to do with a template is not currently translated, including all the existing data view definitions, so it's a lot of work either way. I do want to make templates localisable so they switch with the chosen language, and that would also include data view strings like this, but that's a big job for the future: https://github.com/orgs/openmsupply/projects/8/views/4?filterQuery=locali&pane=issue&itemId=583629 (I made an in-principle working prototype at one point, but it hasn't been a priority to properly do)

Plus, as I said, it's not likely we'd actually use a custom filter list very often -- the batch size fix takes care of most cases.

fergie-nz:
Makes sense, good as gold.

Thank you for explaining all of that.

PR #1017, opened 2023-03-02
#1016 Reload preferences after snapshot load

### LINKED ISSUE #1016: Loading a snapshot doesn't reset preferences
The server preferences only get read on server startup, and when loading a snapshot from the UI, the server doesn't actually restart, so the new snapshot prefs don't get used until the server is manually started and stopped at some point.

Need to come up with a mechanism to handle this.

CarlosNZ:
Fix #1016 

Just mutates the config object directly, which affects all "config" references throughout (since they're all using the same original object).

If you want to test, just add a `console.log(config)` somewhere and notice how it changes after loading a different snapshot (assuming its prefs are different), but doesn't change when on `develop`.

### DIFF
--- a/plugins/action_send_notification/src/sendNotification.ts
+++ b/plugins/action_send_notification/src/sendNotification.ts
@@ -192,7 +192,7 @@ const prepareAttachments = async (
   filesFolder: string
 ): Promise<Attachment[]> => {
   const attachmentInput = Array.isArray(attachments) ? attachments : [attachments]
-  const attachmentObjects = []
+  const attachmentObjects: Attachment[] = []
   for (const file of attachmentInput) {
     if (typeof file === 'object') {
       if (!file?.path || !file?.filename) throw new Error('Invalid attachment')
--- a/src/components/actions/getApplicationData.ts
+++ b/src/components/actions/getApplicationData.ts
@@ -73,7 +73,7 @@ export const getApplicationData = async (input: {
     appRootFolder: getAppEntryPointDir(),
     filesFolder: config.filesFolder,
     webHostUrl: process.env.WEB_HOST,
-    SMTPConfig: config.SMTPConfig,
+    SMTPConfig: config?.SMTPConfig,
   }
 
   const sectionCodes = (await DBConnect.getApplicationSections(applicationId)).map(
--- a/src/components/snapshots/useSnapshot.ts
+++ b/src/components/snapshots/useSnapshot.ts
@@ -1,5 +1,5 @@
 import fs from 'fs/promises'
-import fsSync from 'fs'
+import fsSync, { readFileSync } from 'fs'
 import path from 'path'
 import { execSync } from 'child_process'
 import insertData from '../../../database/insertData'
@@ -10,11 +10,11 @@ import importFromJson from '../exportAndImport/importFromJson'
 import { triggerTables } from './triggerTables'
 import semverCompare from 'semver/functions/compare'
 import config from '../../../src/config'
+import { serverPrefKeys, ServerPreferences } from '../../types'
 // @ts-ignore
 import delay from 'delay-sync'
 import { createDefaultDataFolders } from '../files/createDefaultFolders'
 import migrateData from '../../../database/migration/migrateData'
-
 import {
   DEFAULT_SNAPSHOT_NAME,
   SNAPSHOT_FILE_NAME,
@@ -182,6 +182,8 @@ const useSnapshot: SnapshotOperation = async ({
       })
     }
 
+    refreshPreferences(config)
+
     return { success: true, message: `snapshot loaded ${snapshotName}` }
   } catch (e) {
     return { success: false, message: 'error while loading snapshot', error: e.toString() }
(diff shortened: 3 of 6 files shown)


plugins/action_send_notification/src/sendNotification.ts
```diff
@@ -192,7 +192,7 @@ const prepareAttachments = async (
   filesFolder: string
 ): Promise<Attachment[]> => {
   const attachmentInput = Array.isArray(attachments) ? attachments : [attachments]
-  const attachmentObjects = []
+  const attachmentObjects: Attachment[] = []
```
CarlosNZ, line 195: Unrelated, but fixes unreported Typescript error

src/components/actions/getApplicationData.ts
```diff
@@ -73,7 +73,7 @@ export const getApplicationData = async (input: {
     appRootFolder: getAppEntryPointDir(),
     filesFolder: config.filesFolder,
     webHostUrl: process.env.WEB_HOST,
-    SMTPConfig: config.SMTPConfig,
+    SMTPConfig: config?.SMTPConfig,
```
CarlosNZ, line 76: Could be missing

src/components/snapshots/useSnapshot.ts
```diff
@@ -281,4 +283,15 @@ const copyFiles = async (snapshotFolder: string, fileRecords: ObjectRecord[] = [
   }
 }
 
+// Mutate the active config object to inject new preferences
+type Config = typeof config
+export const refreshPreferences = (config: Config) => {
+  const serverPrefs: ServerPreferences = JSON.parse(readFileSync(PREFERENCES_FILE, 'utf-8')).server
+  serverPrefixKeys.forEach((key) => {
```
CarlosNZ: At first I just iterated over the imported "serverPrefs", but the problem is if a key is absent from the imported server prefs, then it would get left unchanged in "config", when it actually should be erased so that system defaults are used.

So had to make an array of all possible server pref keys and iterate over that to ensure that *all* possible prefs will be updated (or removed).

src/components/snapshots/useSnapshot.ts
```diff
@@ -281,4 +283,15 @@ const copyFiles = async (snapshotFolder: string, fileRecords: ObjectRecord[] = [
   }
 }
 
+// Mutate the active config object to inject new preferences
+type Config = typeof config
+export const refreshPreferences = (config: Config) => {
+  const serverPrefs: ServerPreferences = JSON.parse(readFileSync(PREFERENCES_FILE, 'utf-8')).server
+  serverPrefKeys.forEach((key) => {
```
CarlosNZ, line 290: At first I just iterated over the imported "serverPrefs", but the problem is if a key is absent from the imported server prefs, then it would get left unchanged in "config", when it actually should be erased so that system defaults are used.

So had to make an array of all possible server pref keys and iterate over that to ensure that *all* possible prefs will be updated (or removed).

src/components/snapshots/useSnapshot.ts
```diff
@@ -281,4 +283,15 @@ const copyFiles = async (snapshotFolder: string, fileRecords: ObjectRecord[] = [
...
+    if (serverPrefs[key]) {
+      config[key] = serverPrefs[key] as never
```
CarlosNZ, line 292: I really don't know why the `as never` is needed here. `config[key]` should definitely not be `never` as both `config` and the key values are from properly defined types (and config *includes* all the server pref keys)

src/config.ts
```diff
@@ -1,10 +1,11 @@
 import prefs from '../preferences/preferences.json'
 require('dotenv').config()
 import { version } from '../package.json'
+import { ServerPreferences } from './types'
 const isProductionBuild = process.env.NODE_ENV === 'production'
-const serverPrefs: { [key: string]: any } = prefs.server
+const serverPrefs: ServerPreferences = prefs.server
```
CarlosNZ, line 6: Proper definitions for Server Prefs

src/types.ts
```diff
@@ -243,3 +243,37 @@ export interface Organisation {
...
+  }
+  systemManagerPermissionName?: string
+  previewDocsMinKeepTime?: string
+  previewDocsCleanupSchedule?: number[]
+  backupSchedule?: number[]
+  backupFilePrefix?: string
+  maxBackupDurationDays?: number
+}
+
+export const serverPrefKeys: (keyof ServerPreferences)[] = [
```
CarlosNZ, line 267: I wanted the type of this array to be an *exhaustive* check of all pref keys, so it will warn if any are missing. Best I could do was do show error if incorrect ones are included, but couldn't get it to be exhaustive.

fergie-nz (approved):
Tested and is working as expected.

I feel like it covers a lot of things i don't understand, however, so I can't comment further.

I really appreciate prompts for easy testing validation. Thank you for that.

CarlosNZ:
> Tested and is working as expected.
> 
> I feel like it covers a lot of things i don't understand, however, so I can't comment further.
> 
> I really appreciate prompts for easy testing validation. Thank you for that.

Thanks for taking a look. I might wait till @andreievg has a quick look at this one too, since it's possible there's some unintended side effects I might have forgotten about.

CarlosNZ:
@andreievg I'm going to merge this now as it's been working fine in production for a few weeks.

But if you do spot any issues, please let me know.

PR #1020, opened 2023-03-09
#1019 Fix template export bug

### LINKED ISSUE #1019: Template export fails if there is no template category

CarlosNZ:
Fixes #1019 

Took a while to figure out what was going on here -- bit of a rabbit hole to go down.

The problem was in the way the filter references are generated in `exportToJson.ts` -- because the "template" table has a reference to "templateCategory", a template filter would be generated restricting it to templateCategory ids already found when accessing templateCategory records. But there would be no templateCategory records, cos they had already been restricted by templateID (and in this case the templateCategoryId for the template is `null`) -- so it was a bit circular, and ended up putting a filter like this on template records:
```
{ templateCategoryId: { in: [] }, id: { equalTo: 98 } }
```

The template export query injects a bunch of filters into the `options` object already (from front-end query):
```
 {
      template: { id: equalToTemplateId },
      filter: { templateFilterJoins: allElementsMatchTemplateId },
      permissionName: { templatePermissions: allElementsMatchTemplateId },
      templateCategory: { templates: { some: { id: { equalTo: id } } } },
    }
```
So it's not necessary to additionally filter based on referenced columns.

So my fix here just uses *either* the baseFilter (as above) *or* the generated reference filters, not a merge of both like it previously did. As far as I can tell the baseFilter is always a complete superset of the reference filters (at least it is for template export), which in itself makes the reference filters redundant for those tables.

### DIFF
--- a/src/components/exportAndImport/exportToJson.ts
+++ b/src/components/exportAndImport/exportToJson.ts
@@ -18,8 +18,8 @@ const getRecordsAsObject = async ({
 
   for (const table of tablesToExport) {
     const { tableName } = table
-    const filteredReferences = getFilteredReferences(table, filteredRecords)
     const baseFilter = filters?.[tableName]
+    const filteredReferences = getFilteredReferences(table, filteredRecords, baseFilter)
 
     const { gql, getter, hasFilter } = constructGqlAndGetter(table, filteredReferences, baseFilter)
     const result = await databaseConnect.gqlQuery(gql)
@@ -33,7 +33,8 @@ const getRecordsAsObject = async ({
 
 const getFilteredReferences = (
   { referenceTables, columns }: DatabaseTable,
-  previousFilteredRecords: ObjectRecords
+  previousFilteredRecords: ObjectRecords,
+  baseFilter: object | undefined
 ) => {
   if (referenceTables.length === 0) return {}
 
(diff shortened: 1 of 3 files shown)


src/components/exportAndImport/exportToJson.ts
```diff
@@ -13,19 +13,19 @@ const getRecordsAsObject = async ({
 
   const tablesToExport = filterByIncludeAndExclude(includeTables, excludeTables, databaseTables)
 
-  const fitleredRecords: ObjectRecords = {}
+  const filteredRecords: ObjectRecords = {}
```
CarlosNZ, line 16: Long-standing typo ;)

src/components/exportAndImport/exportToJson.ts
```diff
@@ -66,7 +66,9 @@ const constructGqlAndGetter = (
   filteredReferences: object,
   baseFilter: object
 ) => {
-  const filter = { ...filteredReferences, ...baseFilter }
+  // If filters already defined in options (baseFilter), just use that,
+  // otherwise we can end up with filtering out too much
+  const filter = baseFilter ?? filteredReferences ?? {}
```
CarlosNZ, line 71: This is the actual fix -- very small in the end!

src/components/exportAndImport/exportToJson.ts
```diff
@@ -13,42 +13,49 @@ const getRecordsAsObject = async ({
...
+      isNullable,
       reference: { tableName: referencedTableName, columnName: referencedColumnName },
       columnName,
     }) => {
       if (!isReference) return
+      // If the column is nullable and there is already a base filter defined
+      // (in options) for this table, then don't make additional filters based
+      // on references, otherwise it can filter out to much (particularly for
+      // template export)
+      if (baseFilter && isNullable) return
```
CarlosNZ, line 58: Added an additional check to only ignore filter references if the column is also nullable. 

The previous commit doesn't have this, it just ignores filter references if a baseFilter is present regardless of nullability: baae7b94 (See if you think this is better)

fergie-nz (approved):
Cool looks good. Great pick up too

PR #1022, opened 2023-03-12
1018 Fix for new columns with `null` values

### LINKED ISSUE #1018: New data table can get columns set to wrong data type if the first record has `null` value
When data table is first created as a result of a "modifyRecord" action, the data types of the columns are automatically inferred from the incoming data. However, if a field has a `null` value, the type defaults to "varchar", which might not be correct. And the next record that comes in with valid data can cause an error or incorrectly typed data to be entered.

Simplest solution is to just not create columns if the incoming value is `null` -- it'll get created correctly when the first record with non-null data arrives.

CarlosNZ:
Fix #1018 

Just a problem with the ordering of events -- it was doing the `null` check before the full record had been created, which is why it worked in some cases but not others.

Seems good now.

### DIFF
--- a/plugins/action_modify_record/src/modifyRecord.ts
+++ b/plugins/action_modify_record/src/modifyRecord.ts
@@ -25,16 +25,16 @@ const modifyRecord: ActionPluginType = async ({ parameters, applicationData, DBC
   const valueToMatch = matchValue ?? record[fieldToMatch]
   const applicationId = applicationData?.applicationId || 0
 
-  // Don't update fields with NULL
-  for (const key in record) {
-    if (record[key] === null || record[key] === undefined) delete record[key]
-  }
-
   // Build full record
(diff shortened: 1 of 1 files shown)


fergie-nz (approved):
I don't quite understand why the original check doesn't work.

Is it because there are duplicate keys where one would be recreated after being deleted by the initial check?

CarlosNZ:
> I don't quite understand why the original check doesn't work.
> 
> Is it because there are duplicate keys where one would be recreated after being deleted by the initial check?

Cos how it was previously, it was only looking at fields on `...record`, which are the fields defined directly in the parameters. But there's also the fields defined in the "shortcut" `data` object which get mapped to actual record fields with `...mapValues(data, (property) => get(applicationData, property, null)),`.  And this was happening *after* the null check, so weren't being considered. 

Now it builds the *whole* record first then removes any `null` fields.

PR #1023, opened 2023-03-13
#1021 Scheduled action -- allow specific date input 

### LINKED ISSUE #1021: Scheduled action -- allow specific date input
Currently the "scheduleAction" action only takes a "duration", and the scheduled event uses that duration added to the current date/time.

However, we need to be able to schedule events for a specific time regardless of the current time (e.g. a fixed expiry date for Provisionally registered products), so we need to be able to specify a "date" object instead of a "duration"

CarlosNZ:
Fix #1021 

As well as allowing either `date` or `duration` parameters, I've also tightened up the validation of parameters with slightly more informative error message.

To  test, just use the `runAction` endpoint and update the parameters like so:

```
// Duration as number
{
    "actionCode": "scheduleAction",
    "applicationId": 880, 
    "parameters": {
        "eventCode": "test",
        "duration": 5 // 5 weeks
    }
}

// Duration as object
{
    "actionCode": "scheduleAction",
    "applicationId": 880, 
    "parameters": {
        "eventCode": "test",
        "duration": { "years": 5 } // 5 years
    }
}

// Date as ISO string
{
    "actionCode": "scheduleAction",
    "applicationId": 880, 
    "parameters": {
        "eventCode": "test",
        "duration": "2026-03-01"
    }
}

// Date as object
{
    "actionCode": "scheduleAction",
    "applicationId": 880, 
    "parameters": {
        "eventCode": "test",
        "duration": {"year": 2026, "month": 3, "day": 31 } 
    }
}

// Can't really test JS Date input using HTTP requests
```

Then check the output in the `trigger_schedule` table.

### DIFF
--- a/documentation/List-of-Action-plugins.md
+++ b/documentation/List-of-Action-plugins.md
@@ -663,14 +663,17 @@ A "special" action that allows other actions to be triggered at some time in the
 
 | Input parameters<br />(\*required) <br/> | Output properties |
 | ---------------------------------------- | ----------------- |
-| `duration`\*                             | `scheduledEvent`  |
+| `duration`                               | `scheduledEvent`  |
+| `date`                                   | `scheduledEvent`  |
 | `eventCode`                              |                   |
 | `applicationId`                          |                   |
 | `templateId`                             |                   |
 | `cancel`                                 |                   |
 | `data`                                   |                   |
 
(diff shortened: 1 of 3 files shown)


fergie-nz (approved):
Looks good, nice!

PR #1027, opened 2023-03-16
#1026 Default filters for Data Views

### LINKED ISSUE #1026: Allow default filters for data views
For example, only showing "is_Active" products by default. This would be pretty simple to implement -- just add a new string field to "data_view" table called "default_filter", where a url filter string can be added, e.g. `is-active=true`, which we then append to the URL in the front-end.

CarlosNZ:
Fix #1026.

Pretty straightforward to implement -- just returns a url query filter string for each data view -- front-end handles the rest.

Front-end: https://github.com/openmsupply/conforma-web-app/pull/1484

### DIFF
--- a/database/migration/migrateData.ts
+++ b/database/migration/migrateData.ts
@@ -715,6 +715,14 @@ const migrateData = async () => {
     `)
   }
 
+  if (databaseVersionLessThan('0.5.7')) {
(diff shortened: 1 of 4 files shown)


fergie-nz (approved):
Looks great!

PR #1031, opened 2023-04-06
#1030 Fix core action mutation bug

### LINKED ISSUE #1030: User Registration times out -- database error
Somehow the serial it's trying to write is from the wrong pattern -- it's using whatever one was used last.

A workaround is to explicitly set a pattern in the template (done for Fiji), but need to figure out why the default is using the most recent instead of the default one in "coreActions".

CarlosNZ:
Fix #1030 

Fixed as per comment in code.

### DIFF
--- a/src/components/actions/coreActions.ts
+++ b/src/components/actions/coreActions.ts
@@ -7,6 +7,7 @@ they've been hard-coded here in order to:
 - prevent accidental misconfiguration or removal
 */
 
+import { cloneDeep } from 'lodash'
 import DBConnect from '../databaseConnect'
 import { Trigger } from '../../generated/graphql'
 import { ActionInTemplate } from '../../types'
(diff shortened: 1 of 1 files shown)


fergie-nz (approved):
I haven't quite internalised how a deepClone will fix this issue, but looks great if that is working.

Great catch by Jess and Laurence

CarlosNZ:
> I haven't quite internalised how a deepClone will fix this issue, but looks great if that is working.

The `coreActions` object is meant to be a reference object that doesn't change. However, by injecting different serial patterns, we were inadvertently mutating that original object. And so next time it was used, it would have a different value for `coreActions[0].parameter_queries.pattern` which would get used if the template didn't have it's own serial pattern defined. 

A shallow copy wouldn't be enough to fix this as the mutation is three levels deep, so safest to just make a full ("deep") copy of the coreActions (just for the current trigger, not the whole thing) every time for the "currentCoreActions" (which gets dropped at the end of the functions).

Not sure if that makes sense. Let me know if you'd like any further explanation in person next week :)

PR #1032, opened 2023-04-13
1486F Return defaultFilterString in DataViewTable response

CarlosNZ:
As per title, required for https://github.com/openmsupply/conforma-web-app/pull/1489


fergie-nz (approved):
Looks good as per https://github.com/openmsupply/conforma-web-app/pull/1489

Issue #137, opened 2020-12-06
Create Action to trim unchanged responses on submission


nmadruga:
Should be an automated action - on the list create on #204.

Should receive the id of the `review_status_history` ?

PR #1034, opened 2023-04-27
#137T Add username authorisation check

CarlosNZ:
This prevents a potential privacy problem as outlined in the [templates repo #137](https://github.com/openmsupply/conforma-templates/issues/137).

The username is now included in the JWT token data. And so we can compare the token username to the user data fetched when loading (based on userID). And if they don't match, we don't allow fetching any user data, and the front-end will be logged out.

**To test**:
- On `develop`, log in as userID 100 (ext1) on PNG snapshot. Then (from another browser), load the Fiji snapshot. Then refresh the original browser. You'll now be logged in as a different person (with ID 100).
- Switch to this branch, then repeat the above procedure. You should be logged out when refreshing.


fergie-nz (approved):
Nice looks great 👍

Issue #1036, opened 2023-05-02
Make sure applicationData has org data from when the application was submitted.


CarlosNZ:
We recently had a situation where a user was removed from their organisation after they had submitted an application but before it had been approved. 

And then on approval, the company details were missing from the output record. 

We should update the `getUserOrg` query in `getApplicationData` to get this info from the application itself rather than the user-org View. Same for user data and anything else that might change.

PR #1037, opened 2023-05-02
#1036 #1035 applicationData fix and additional migration

### LINKED ISSUE #1035: Add "application_id" column to "action_queue" table
Won't make any difference to functionality, but it'll make it easier for us to find and diagnose action error for specific applications.

Should only take under an hour.

CarlosNZ:
Fix #1035 and #1036

The migration to add "application_id" column is a bit gnarly, but it's worth it to have this in there -- it also means actions and triggers that no longer have an associated application will automatically get deleted.

### To test

**For #1035**:
- Load various snapshots, and check that after loading/migration, the `trigger_queue` and `action_queue` tables have a populated `application_id` field.
- Try doing a few things that cause triggers (start/submit/assign etc) and check that the actions behave as normal and that they get `application_id` stored in the action and trigger queues

**For #1036**:
- Find a submitted application, then remove the applicant from the organisation they applied on behalf of. Then run the `getApplicationData` endpoint -- you should still get the correct organisation (id and name) data for the application (if you do the same thing on `develop`, the orgId and orgName fields will be `null`)

### DIFF
--- a/database/migration/migrateData.ts
+++ b/database/migration/migrateData.ts
@@ -741,7 +741,7 @@ const migrateData = async () => {
         REFERENCES public.application (id) ON DELETE CASCADE;
       `)
       console.log(' ...and updating trigger_queue data')
-      await DB.populateTriggerQueueApplicationId()
+      await DB.populateQueueApplicationIds('trigger_queue')
     }
     if (!(await DB.checkColumnExists('action_queue', 'application_id'))) {
       await DB.changeSchema(`
@@ -750,9 +750,8 @@ const migrateData = async () => {
         REFERENCES public.application (id) ON DELETE CASCADE;
       `)
       console.log(' ...and updating action_queue data')
-      await DB.populateActionQueueApplicationId()
+      await DB.populateQueueApplicationIds('action_queue')
     }
-    // await DB.populateActionQueueApplicationId()
   }
 
   // Other version migrations continue here...
--- a/database/migration/migrateData.ts
+++ b/database/migration/migrateData.ts
@@ -723,6 +723,38 @@ const migrateData = async () => {
     `)
   }
 
+  // v0.6.0
+  if (databaseVersionLessThan('0.6.0')) {
+    console.log(' - Update FK reference to trigger_queue in action_queue')
+    await DB.changeSchema(`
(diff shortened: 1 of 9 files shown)


database/migration/migrateData.ts
```diff
@@ -723,6 +723,37 @@ const migrateData = async () => {
...
+    console.log(' - Update FK reference to trigger_queue in action_queue')
+    await DB.changeSchema(`
+      ALTER TABLE action_queue DROP CONSTRAINT IF EXISTS   
+        action_queue_trigger_event_fkey; 
+      ALTER TABLE action_queue ADD CONSTRAINT action_queue_trigger_event_fkey
+        FOREIGN KEY (trigger_event) REFERENCES trigger_queue (id) ON DELETE CASCADE;
+    `)
+
+    console.log(' - Add application_id column to trigger_queue & action_queue')
+    if (!(await DB.checkColumnExists('trigger_queue', 'application_id'))) {
```
CarlosNZ, line 737: Added this additional check (in addition to the version check), because this is quite a slow process, so we want to make sure it only ever runs once.

src/components/postgresConnect.ts
```diff
@@ -594,13 +594,19 @@ class PostgresDB {
...
+    status,
+  }: TriggerQueueUpdatePayload): Promise<boolean> => {
+    const text = `UPDATE trigger_queue SET status = $1${
+      application_id ? ', application_id = $3' : ''
+    } WHERE id = $2`
+    const values = [status, id]
+    if (application_id) values.push(application_id)
+
+    try {
+      await this.query({ text, values })
```
fergie-nz, line 609: I don't quite understand the mechanism of this code, but it is working well - I'll take some time later to properly wrap my head around it.

fergie-nz (approved):
Looks great and working well

src/components/postgresConnect.ts
CarlosNZ: This bit is a bit messy, but it's just building a conditional SQL query, because the `application_id` may or may not be present, so the query string has to change accordingly. You'll see other examples of this around such as when the `orgId` might be present or not.

PR #1038, opened 2023-05-04
#1028 (precursor) preferences endpoints

CarlosNZ:
In order to do #1028 properly, we need to have a UI to edit the preferences. This PR just provides a couple of additional Admin endpoints (`get-all-prefs`, `set-prefs`) to facilitate this.

Front-end PR https://github.com/openmsupply/conforma-web-app/pull/1497

Documentation for all preferences included here too.

PR #1039, opened 2023-05-04
#1028 Test mode for emails

### LINKED ISSUE #1028: Test mode -- to avoid sending out notifications to real users
Now that we have real users on a real system, we should be careful about testing with live data. It's very easy to cause actions that can send out notifications to real users, which would be quite uncool.

So we probably need some kind of "test mode" that we can enable whenever a snapshot is installed on a non-live server, which would then re-direct all emails to whoever is testing. Or something like that. 

@Fergie-nz has suggested Mailhog, which might be a good idea, but not sure how that would work on a demo server though.

Something to discuss further.

CarlosNZ:
Fix #1028 

Based on earlier PR #1038, and will need front-end branch from [this PR](https://github.com/openmsupply/conforma-web-app/pull/1497) to be able to interact with the settings.

I've modified (and hopefully simplified) the logic for "testing" mode in "Send Notification" emails now, the main focus being to prevent emails being sent out to real users from a testing or dev instance.

How it works is:

- the system compares the "host" value in `web.siteHost` preference to the current URL (from `WEB_HOST` environment variable) to determine if it's a "live" system. (Note: if no `siteHost` is specified, it assumes it *is* live)
- If it is live, then emails will be sent out as normal
- If not (i.e. a demo/testing server or dev environment), then it looks for a "testingEmail" preference (in server prefs). If found, all email will be sent to this address. This can easily be changed in the UI in order to change it to the email of whomever is currently testing the system.
- If no "testingEmail" value is present, then *no* emails will be sent at all, so you have to configure one if you want to get emails during testing/development.

I've added some additional console logging so we can easily see which email "mode" is being used, along with where the email actually went. There are four "modes" available, as outlined in the comments in `sendNotification`:

```js
/*
Operation modes:

- LVIE: Emails are sent normally according to action configurations
- TEST: All emails are sent to a single address, defined in server
  preferences "testingEmail" property. Used on testing servers or in
  development.
- NONE: No emails are sent at all. Used for automated testing, or when a
  "testingEmail" address is not provided.
- MAILHOG: All emails are relayed through a local MailHog SMTP server (so not
  actually sent). An alternative development mode.
*/
```

(I don't really think we'll use the "MAILHOG" mode much, but it was already there, so I left it in (it can only be enabled by changing a hard-coded value in development))

To test, I suggest using Postman to post to the `runAction` endpoint to run the "sendNotification" action. Then experiment with changing the preferences (namely `web.siteHost`, `web.testingEmail` and the "WEB_HOST" environment variable) to see if email sending behaves as described above.

### Update:

I've added an additional optional server preferences, `emailTestMode`, as sometimes we may want to toggle the "TEST_EMAIL" mode on or off regardless of the other settings. For test servers, it makes sense to let it figure out by itself if it's a live server or not, but for ones like PNG, we might want to have it in test mode while testing it, then turn it "on" once it's ready for others to use.

So, basically, if `emailTestMode` is not specified, the logic will work exactly as described above. However, it `emailTestMode` is explicitly set to either `true` or `false`, then that will take priority (the only caveat being that a `testingEmail` must be defined for it to be in TEST_EMAIL mode).

### Update 2 (17/05):

I've moved the email mode logic into "config.ts", so the info can be used elsewhere, and so we display the email mode in the console when starting or refreshing preferences -- this will help with peace of mind when developing.


plugins/action_send_notification/src/sendNotification.ts
```diff
@@ -18,10 +32,20 @@ const sendNotification: ActionPluginType = async ({ parameters, applicationData,
...
 
+  const testingEmail = config.testingEmail
+
+  const mode = getOperationMode({
+    mailHog: USE_MAIL_HOG,
+    webHostUrl,
+    productionHost,
+    suppressEmail: other?.suppressEmail ?? false,
+    testingEmail,
+  })
```
CarlosNZ, line 48: The logic for each kind of state was getting a bit messy, so I've collapsed them all down to one of four "modes", which makes it easier to follow what situation we're in for each conditional branch.

plugins/action_send_notification/src/sendNotification.ts
```diff
@@ -80,14 +104,20 @@ const sendNotification: ActionPluginType = async ({ parameters, applicationData,
...
+      )
       transporter
         .sendMail({
           from: `${fromName} <${fromEmail}>`,
-          to: toAddressString,
-          cc: ccAddressString,
-          bcc: bccAddressString,
+          to: mode === 'TEST_EMAIL' ? '' : toAddressString,
+          cc: mode === 'TEST_EMAIL' ? '' : ccAddressString,
+          bcc: mode === 'TEST_EMAIL' ? testingEmail : bccAddressString,
```
CarlosNZ, line 121: The testing email is sent as "bcc", I guess just to make it slightly more clear to the recipient that they're not necessarily the intended recipient. But this is somewhat dubious logic, so I'm happy to just make it the `to` field if that seems clearer?

plugins/action_send_notification/src/sendNotification.ts
```diff
@@ -102,11 +132,7 @@ const sendNotification: ActionPluginType = async ({ parameters, applicationData,
             (!rejected || rejected.length === 0) &&
             (!pending || pending.length === 0)
           ) {
-            console.log(
-              `Email successfully sent to: ${envelope.to}\ncc:${envelope.cc || ''}\nbcc: ${
-                envelope.bcc || ''
-              }\nSubject: ${subject || ''}\n`
-            )
+            console.log(`Email successfully sent to: ${envelope.to}\n`)
```
CarlosNZ, line 136: Note the difference here. The earlier "Sending to..." log should show the recipients who the email *would* go to in normal "live" operation, whereas this value shows who it was *actually* sent to based on the operation mode. So compare these values in the console to make sure its behaving as it's intended.

fergie-nz (approved):

PR #1040, opened 2023-05-09
#816 Allow manager to access lookup tables

### LINKED ISSUE #816: Allow management access to Lookup tables (and Localisation) without full "admin" access

CarlosNZ:
Fix #816 

Requires front-end https://github.com/openmsupply/conforma-web-app/pull/1501

Changes "/lookup-tables" endpoint from being an "admin" route to a normal authenticated route. However, it has an additional check for *either* "admin" or "systemManager" permission.

### DIFF
--- a/src/components/permissions/loginHelpers.ts
+++ b/src/components/permissions/loginHelpers.ts
@@ -68,11 +68,14 @@ const getUserInfo = async (userOrgParameters: UserOrgParameters) => {
 
   const returnSessionId = sessionId ?? nanoid(16)
 
-  const isAdmin = !!templatePermissionRows.find((row) => !!row?.isAdmin)
-
   const managementPrefName =
     config?.systemManagerPermissionName || config.defaultSystemManagerPermissionName
-  const isManager = !!templatePermissionRows.includes(managementPrefName)
+
+  const { isAdmin, isManager } = await databaseConnect.getUserAdminStatus(
+    managementPrefName,
+    newUserId,
+    orgId ?? null
+  )
 
   return {
     templatePermissions: buildTemplatePermissions(templatePermissionRows),
--- a/src/components/permissions/rowLevelPolicyHelpers.ts
+++ b/src/components/permissions/rowLevelPolicyHelpers.ts
@@ -7,7 +7,7 @@ import { permissionPolicyColumns } from '../postgresConnect'
 
 export const baseJWT = { aud: 'postgraphile' }
 
-/* Compiles JWT from userInfo and PerissionRows
+/* Compiles JWT from userInfo and PermissionRows
   in { userId: 1, ... }, [
   {
     templatePermissionRestrictions: {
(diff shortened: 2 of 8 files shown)


src/components/permissions/loginHelpers.ts
```diff
@@ -68,11 +68,14 @@ const getUserInfo = async (userOrgParameters: UserOrgParameters) => {
 
   const returnSessionId = sessionId ?? nanoid(16)
 
-  const isAdmin = !!templatePermissionRows.find((row) => !!row?.isAdmin)
-
   const managementPrefName =
     config?.systemManagerPermissionName || config.defaultSystemManagerPermissionName
-  const isManager = !!templatePermissionRows.includes(managementPrefName)
```
CarlosNZ, line 75: This expression didn't work at all!

I've combined `isManager` and `isAdmin` into a single database call that is independent of template permissions. (The previous way, if a user was Admin, but there were no templates that required Admin permission, then `isAdmin` would return `false`. As for `isManager`, well...🙄)

src/components/permissions/rowLevelPolicyHelpers.ts
```diff
@@ -31,9 +31,10 @@ export const baseJWT = { aud: 'postgraphile' }
   }
 */
 const compileJWT = (JWTelements: any) => {
-  const { userId, orgId, username, templatePermissionRows, sessionId, isAdmin } = JWTelements
+  const { userId, orgId, username, templatePermissionRows, sessionId, isAdmin, isManager } =
+    JWTelements
 
-  let JWT: any = { ...baseJWT, userId, orgId, username, sessionId, isAdmin }
+  let JWT: any = { ...baseJWT, userId, orgId, username, sessionId, isAdmin, isManager }
```
CarlosNZ, line 37: We weren't even putting `isManager` in the JWT. We weren't actually using it anywhere, which is why we hadn't noticed the problem. But now we need to check it for look-up table authentication.

src/types.ts
```diff
@@ -271,6 +272,7 @@ export const serverPrefKeys: (keyof ServerPreferences)[] = [
   'hoursSchedule',
   'SMTPConfig',
   'systemManagerPermissionName',
+  'managerCanEditLookupTables',
```
CarlosNZ, line 275: Add a preference for the off chance that we *don't* want the system manager to access lookup tables (it's `true` by default)

fergie-nz (approved):
Nice! I think I get it, and working well

PR #1041, opened 2023-05-23
Data view editor - database changes

CarlosNZ:
~~Please don't review this until other PRs are merged, @fergie-nz. I based this off a combination of all the other PRs to avoid messy conflicts with the migration script, but that means that there's no clean diff to any particular branch.~~

~~But once a few more PRs are merged then this will be cleaner. I'll update this a bit more when that happens, just wanted to get this published so we don't lose track of it.~~

Back-end changes required to implement the data view editor (front-end PR: https://github.com/openmsupply/conforma-web-app/pull/1511)

It's basically just adding an additional field "identifier" to data views, so that the front-end can distinguish between two different views for the same code. It should be a human-readable string explaining what it is (e.g. "Chemical Permits (Internal view)" vs "Chemical Permits (own company)")


CarlosNZ:
Okay @fergie-nz the other PRs are all merged, so this should be good to look at now.

Nearly there! :)

fergie-nz (approved):

PR #1053, opened 2023-06-22
#1052 Fix "un-expire" problem

### LINKED ISSUE #1052: Application not being "un-expired" when extending deadline
Looks like we forgot to add this action to "Core Actions" when we made it a while back.

CarlosNZ:
Fix #1052 

Looks like this problem has been hanging around since we moved "core actions" into the codebase, but we just hadn't noticed it 😳.

To test, download this snapshot: https://www.dropbox.com/s/6f8x3ya1dvan06u/FIJI_expiry_bug.zip?dl=0 (Password is the Fiji admin password) (The Fiji snapshot is now too big to store in the templates repo)

Log in as "Emi" (same as admin pw) and go to: http://localhost:3000/application/PROV-001857/review?tab=overview

If you're on `develop`, then when you press "Extend", the expiry date will move forward, but the application will still stay "Expired".

Switch to this branch and extend it again -- this time the application should go back to "In progress" :)

**UPDATE**: You won't actually see the button to extend, because of the state of the application. To make it appear, just go to "Overview.tsx" in the front-end and change the condition on line 91 to `true`, ie. this:
```
 ((applicantDeadline.isActive && outcome === ApplicationOutcome.Pending) ||
                  (!applicantDeadline.isActive && outcome === ApplicationOutcome.Expired)) && 
```


fergie-nz (approved):
Awesome looks great,

Was it just missing all this time?

CarlosNZ:
> Was it just missing all this time?

Yup! 😆

PR #1055, opened 2023-06-28
#1054 Implement "enabled" status for data views

### LINKED ISSUE #1054: Add an "enabled" status for Data Views
Currently I haven't been able to add new Data Views until the templates for them are active and they have received some applications. This is kind of annoying and has led to some confusion in keeping track of what has been configured where.

It'd be really good if we could create a Data View but set it to "disabled" and then just turn it on when it becomes valid.

### What is required:

- [x] Add new boolean column "enabled" to "data_view" table (with default value `true`) in migration script
- [x] Update the data views endpoint(s) to not return disabled views
- [x] BONUS! Update the "modifyRecord" action to automatically enable data views as part of the "createTable" method -- if any matching data views are disabled when a new table is created, turn them on!
- [x] Update docs

CarlosNZ:
Fix #1054 

Fairly straightforward, code should be self-explanatory (including docs)

~~UPDATE: Note, requires front-end branch https://github.com/openmsupply/conforma-web-app/pull/1533 in order to see the new fields in the data view config ui~~ -- Merged


fergie-nz (approved):
nice nice looks good

PR #1057, opened 2023-07-11
#1056 Better template versioning

### LINKED ISSUE #1056: Better version tracking
It's very easy to get confused with which version of a template is the latest, and then end up with changes on two different versions. We also need to know for sure if a template has been changed so we can meaningfully compare different versions in different systems.

The "version number" is not a whole lot of use, as it's just increased sequentially, so you can't use it to compare versions between instances (e.g. local vs demo vs live).

I think something like a "version_id" field (a random 6-char string) could help, but not 100% sure exactly how it should work.

So just putting this here for further discussion.

CarlosNZ:
Fix #1056 

Note: the base branch here is not the final one, but I needed to build off multiple PRs. This will become correct once those PRs are merged to `develop`.

I'll put most of the testing instructions in the front-end PR (https://github.com/openmsupply/conforma-web-app/pull/1538).

The changes here are primarily schema (new version columns, remove old ones), and tweaks to the template import/export process to accommodate this. 

The basic schema (on the "template" table) is:

- `version_id`: a unique  (per template type), random id for a template version. Once this is set, the version is "locked" or "committed" and can't be changed any more. Versions that are "in-progress" (i.e. can be modified) have the placeholder id `"*"` until they're committed. (Suffixes are added if there is more than one in-progress version)
- `parent_version_id`: the version_id of the version that the current version is "branched" (i.e. duplicated) from. Useful for tracking the sequence of changes.
- `version_timestamp`: same as it currently is, though I've tried to be a bit more careful as to when exactly this gets set/updated
- `version_comment`: a user-specified commit to give some indication as to what has changed in this template version. The equivalent of a "commit message"
- `version_history`: A JSON array of previous versions. Each item contains all of the above values for each previous version. This means that we can see the complete version commit history for a template even if earlier versions have been deleted. Also, the `length` of this array is used to generate the version "number" (We no longer need a specific "version" number field)

Example version history:
```json
[
  {
    "comment": "Minor tweaks",
    "timestamp": "2021-08-10T02:13:36.778Z",
    "versionId": "m6p4da",
    "parentVersionId": null
  },
  {
    "comment": "Updated form elements",
    "timestamp": "2022-09-30T02:12:46.157Z",
    "versionId": "ca54l6",
    "parentVersionId": "m6p4da"
  },
  {
    "comment": "Added email correspondence",
    "timestamp": "2022-09-30T02:56:25.777Z",
    "versionId": "dhwr8u",
    "parentVersionId": "ca54l6"
  },
  {
    "comment": null,
    "timestamp": "2023-06-30T12:02:32.789186+12:00",
    "versionId": "ba39d9",
    "parentVersionId": "dhwr8u"
  }
]
```

### DIFF
--- a/database/buildSchema/43_views_functions_triggers.sql
+++ b/database/buildSchema/43_views_functions_triggers.sql
@@ -149,31 +149,12 @@ CREATE TRIGGER trigger_queue
     EXECUTE FUNCTION public.notify_trigger_queue ();
 
 -- TEMPLATES
--- FUNCTION to generate a new version of template (should run as a trigger)
-CREATE OR REPLACE FUNCTION public.set_template_verision ()
-    RETURNS TRIGGER
-    AS $template_event$
-BEGIN
-    IF (
-        SELECT
-            count(*)
-        FROM
-            public.template
-        WHERE
-            id != NEW.id AND code = NEW.code AND version = NEW.version) > 0 THEN
-        NEW.version = (
-            SELECT
-                max(version) + 1
-            FROM
-                public.template
-            WHERE
-                code = NEW.code);
-        NEW.version_timestamp = CURRENT_TIMESTAMP;
-    END IF;
-    RETURN NEW;
-END
-$template_event$
-LANGUAGE plpgsql;
+
+DROP FUNCTION IF EXISTS public.set_template_verision CASCADE; -- DEPRECATED
+
+-- Enforce unique versionId per template type
+ALTER TABLE public.template
+    ADD UNIQUE (code, version_id);
 
 -- FUNCTION to make sure duplicated templates have 'DRAFT' status
 --   but only if there is another version with 'AVAILABLE' status
@@ -228,14 +209,6 @@ END;
 $template_event$
 LANGUAGE plpgsql;
 
---TRIGGER to generate new version of template on insertion or update
-DROP TRIGGER IF EXISTS set_template_version_trigger ON public.template;
-
-CREATE TRIGGER set_template_version_trigger
-    BEFORE INSERT OR UPDATE ON public.template
-    FOR EACH ROW
-    EXECUTE FUNCTION public.set_template_verision ();
-
 --TRIGGER to make sure duplicates templates have 'DRAFT' status
 DROP TRIGGER IF EXISTS set_template_to_draft_trigger ON public.template;
 
@@ -246,7 +219,6 @@ CREATE TRIGGER set_template_to_draft_trigger
 
 -- TRIGGER to ensure only one template version can be 'AVAILABLE'
 DROP TRIGGER IF EXISTS template_status_update_trigger ON public.template;
-
 CREATE TRIGGER template_status_update_trigger
     AFTER UPDATE OF status ON public.template
     FOR EACH ROW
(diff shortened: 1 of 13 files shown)


database/buildSchema/43_views_functions_triggers.sql
```diff
@@ -149,31 +149,12 @@ CREATE TRIGGER trigger_queue
     EXECUTE FUNCTION public.notify_trigger_queue ();
 
 -- TEMPLATES
--- FUNCTION to generate a new version of template (should run as a trigger)
-CREATE OR REPLACE FUNCTION public.set_template_verision ()
```
CarlosNZ, line 153: No longer required since we've got rid of the "version" field completely.

database/buildSchema/43_views_functions_triggers.sql
```diff
@@ -149,31 +149,12 @@ CREATE TRIGGER trigger_queue
...
-    RETURN NEW;
-END
-$template_event$
-LANGUAGE plpgsql;
+
+DROP FUNCTION IF EXISTS public.set_template_verision CASCADE; -- DEPRECATED
+
+-- Enforce unique versionId per template type
+ALTER TABLE public.template
+    ADD UNIQUE (code, version_id);
```
CarlosNZ, line 157: Just an extra precaution to prevent multiple versions having the same code.

database/buildSchema/43_views_functions_triggers.sql
```diff
@@ -308,11 +281,30 @@ ALTER TABLE public.template_element
...
+        RETURN NULL;
+END;
+$template_element_event$
+LANGUAGE plpgsql;
+
+DROP TRIGGER IF EXISTS template_version_code_update ON public.template;
+CREATE TRIGGER template_version_code_update
+    AFTER UPDATE OF version_id, code ON public.template
+    FOR EACH ROW
+    EXECUTE FUNCTION public.recompute_template_element_code_version ();
```
CarlosNZ, line 306: This was causing a problem -- when I updated the "versionId" when "committing" a version, the generated columns on `template_element` weren't getting updated, which resulted in database errors when a new version used the "*" placeholder versionId. So this just ensures that those values get recomputed whenever the versionId or code are updated in the template table.

database/migration/databaseMethods.ts
```diff
@@ -166,6 +167,63 @@ const databaseMethods = {
       }
     }
   },
+  migrateTemplateVersions: async () => {
```
CarlosNZ, line 170: Quite a substantial migration to move existing templates over to the new versioning schema.

database/snapshotOptions/templateExport.json
```diff
@@ -30,5 +30,6 @@
   "excludeInsertScripts": [],
   "includePrefs": false,
   "includeLocalisation": false,
-  "resetFiles": false
+  "resetFiles": false,
+  "templates": { "checkVersionOnImport": true }
```
CarlosNZ, line 34: An additional option to tell the template snapshot import to check for existing templates with the same versionId, and return and error if there is (so you can't accidentally import two copies of the same version)

src/components/exportAndImport/importFromJson.ts
```diff
@@ -33,6 +35,43 @@ const insertFromObject: InsertFromObject = async (
...
+          timestamp: DateTime.fromISO(template.versionTimestamp),
+          versionId: getTemplateVersionId(),
+          parentVersionId: null,
+        }))
+        template.version = null
+      }
+
+      const existingVersions = await databaseConnect.getTemplateVersionIDs(template.code)
+
+      // Make sure unlocked "*" templates have unique versionId
```
CarlosNZ, line 63: Since the "versionId" field must be unique (per template code), if we have more than one "in-progress" version, this just adds a numeric suffix to preserve uniqueness. e.g.
```
*,
*_1,
*_2,
...
```

PR #1058, opened 2023-07-12
#1539F Multiple trigger protection

CarlosNZ:
An additional fix for https://github.com/openmsupply/conforma-web-app/issues/1539

(Front-end PR: https://github.com/openmsupply/conforma-web-app/pull/1540)

This prevents more than one active version of the same trigger being added to the trigger queue. The front-end PR should solve the problem, but this is extra protection in case there are any other ways to cause duplicate triggers (such as setting triggers via direct API calls/Postman).


fergie-nz (approved):
Nice looks great 👍

PR #1060, opened 2023-07-25
#1059 #1042 File archive management

### LINKED ISSUE #1042: Long-term file archiving/management
**UPDATE: Feel free to read my stream of consciousness ramblings below, but I've gone and distilled this into a more structured, specific proposal in a KDD over here: https://github.com/openmsupply/conforma-server/discussions/1059**

As the size of the stored data increases, the size of the snapshot is getting quite large, which will be problematic for slow internet connections (and even not-so-slow before too long).

Also, the backup system is quite inefficient in that every day's backup contains a complete copy of all the files, most of which are exactly the same, so we'll be consuming hundreds of GB of Dropbox storage that is largely duplication before too long.

Ultimately we're going to need some more sophisticated file management. It'll need some careful design, so just putting this issue here as a starting point for further IRL discussion/planning.

The solution we implement would need to account for (at a minimum) the following use cases:

- We should be able to download a snapshot that is a reasonable size (i.e. most files are not included), but when we restore it, the older files (that we didn't download) should be preserved.
- Each backup snapshot should only include files that have been added since the last backup. However, when restoring from a backup, we should be able to get back *all* the files needed, not just the ones included directly in this backup

CarlosNZ:
Fix #1042 (detailed spec/discussion: #1059)

Big PR, sorry, but I've made lots of small improvements to existing functionality too to make it more consistent and organised.

Front-end PR: https://github.com/openmsupply/conforma-web-app/pull/1541


src/components/exportAndImport/backup.ts
```diff
@@ -18,46 +35,120 @@ const passwordArg: string | undefined = process.argv[3]
...
+  const archivesNotBackedUp = currentSystemArchives.filter(
+    (archive) => !existingArchiveBackupIds.includes(archive.uid)
+  )
+
+  const zipSources: string[] = []
+
+  // Make new archive snapshot with missing archives
+  if (archivesNotBackedUp.length > 0) {
+    const archiveFrom = Math.min(...archivesNotBackedUp.map((a) => a.timestamp))
+    const archiveTo = Math.max(...archivesNotBackedUp.map((a) => a.timestamp))
```
CarlosNZ, line 76: Strictly speaking, this doesn't just get the specific archives found, it collects a whole range between the earliest and latest. But in most normal operation, this will be the same thing.

src/components/files/archive.ts
```diff
@@ -0,0 +1,137 @@
...
+  // Move files
+  for (const file of files) {
+    const newFilePath = path.join(archivePath, 'files', file.file_path)
+    try {
+      await move(path.join(FILES_FOLDER, file.file_path), path.join(FILES_FOLDER, newFilePath))
+      file.archive_path = path.join(archivePath, 'files')
+    } catch {
+      console.log('Problem moving', file.file_path)
+    }
+    const shouldMoveThumbnail = !file.thumbnail_path.startsWith(config.genericThumbnailsFolderName)
```
CarlosNZ, line 74: We don't want to move thumbnails that use the generic system thumbnails (e.g. PDFs)

src/components/files/fileHandler.ts
```diff
@@ -17,9 +17,18 @@ interface HttpQueryParameters {
...
-  if (!result) throw new Error()
-  return result
+  const fileData = await DBConnect.getFileDownloadInfo(uid)
+  if (!fileData) throw new Error('Unable to retrieve file info')
+
+  const isGenericThumbnail =
+    thumbnail && fileData.thumbnail_path.startsWith(config.genericThumbnailsFolderName)
+
+  const filePath = path.join(fileData.archive_path ?? '', fileData.file_path)
+  const thumbnailPath = path.join(
```
CarlosNZ, line 27: The reverse of before -- if using generic thumbnail, don't include the "archive_path" in the full file path.

src/components/files/routeArchiveFiles.ts
```diff
@@ -0,0 +1,15 @@
+import { FastifyReply, FastifyRequest } from 'fastify'
+import { combineRequestParams } from '../utilityFunctions'
+import archiveFiles from './archive'
+
+const routeArchiveFiles = async (request: FastifyRequest, reply: FastifyReply) => {
```
CarlosNZ, line 5: Added an endpoint for this so Archives can be triggered by front-end (on Snapshots page)

src/components/snapshots/takeSnapshot.ts
```diff
@@ -24,89 +32,128 @@ import {
...
+      if (options.usePgDump) {
+        // The quick way, using pg_dump -- whole database only
+        console.log('Dumping database...')
+        execSync(
+          `pg_dump -U postgres tmf_app_manager --format=custom -f ${tempFolder}/database.dump`
+        )
+        // This plain-text .sql script is NOT used for re-import, but could be
+        // useful for debugging when dealing with troublesome snapshots
+        // execSync(
+        //   `pg_dump -U postgres tmf_app_manager --format=plain --inserts --clean --if-exists -f ${tempFolder}/database.sql`
```
CarlosNZ, line 83: Commented out this `.sql` script as we don't use it for anything and it's actually a really large file wasting a lot of space. We can un-comment this if we ever need to debug the database dumps.

src/components/snapshots/useSnapshot.ts
```diff
@@ -59,6 +63,9 @@ const useSnapshot: SnapshotOperation = async ({
       throw `Snapshot was created with version: ${snapshotVersion}\n You can't install a snapshot created with a version newer than the current application version: ${config.version}`
     }
 
+    // Check that we can find all the archives needed:
+    await collectArchives(snapshotFolder)
```
CarlosNZ, line 67: The called function `findArchiveSources` throws an error if it can't find all the archives required to construct the full snapshot, so this won't proceed any further.

fergie-nz (approved):
Looks great.

I haven't internalised everything that is going on here, but looks fantastic and is functioning well.

PR #1061, opened 2023-07-26
1056 improved template versioning

### LINKED ISSUE #1056: Better version tracking

CarlosNZ:
_NOTE: Replaces #1057 as that PR was auto-closed due to base branch being deleted._

Fix #1056 

Note: the base branch here is not the final one, but I needed to build off multiple PRs. This will become correct once those PRs are merged to `develop`.

I'll put most of the testing instructions in the front-end PR (https://github.com/openmsupply/conforma-web-app/pull/1538).

The changes here are primarily schema (new version columns, remove old ones), and tweaks to the template import/export process to accommodate this. 

The basic schema (on the "template" table) is:

- `version_id`: a unique  (per template type), random id for a template version. Once this is set, the version is "locked" or "committed" and can't be changed any more. Versions that are "in-progress" (i.e. can be modified) have the placeholder id `"*"` until they're committed. (Suffixes are added if there is more than one in-progress version)
- `parent_version_id`: the version_id of the version that the current version is "branched" (i.e. duplicated) from. Useful for tracking the sequence of changes.
- `version_timestamp`: same as it currently is, though I've tried to be a bit more careful as to when exactly this gets set/updated
- `version_comment`: a user-specified commit to give some indication as to what has changed in this template version. The equivalent of a "commit message"
- `version_history`: A JSON array of previous versions. Each item contains all of the above values for each previous version. This means that we can see the complete version commit history for a template even if earlier versions have been deleted. Also, the `length` of this array is used to generate the version "number" (We no longer need a specific "version" number field)

Example version history:
```json
[
  {
    "comment": "Minor tweaks",
    "timestamp": "2021-08-10T02:13:36.778Z",
    "versionId": "m6p4da",
    "parentVersionId": null
  },
  {
    "comment": "Updated form elements",
    "timestamp": "2022-09-30T02:12:46.157Z",
    "versionId": "ca54l6",
    "parentVersionId": "m6p4da"
  },
  {
    "comment": "Added email correspondence",
    "timestamp": "2022-09-30T02:56:25.777Z",
    "versionId": "dhwr8u",
    "parentVersionId": "ca54l6"
  },
  {
    "comment": null,
    "timestamp": "2023-06-30T12:02:32.789186+12:00",
    "versionId": "ba39d9",
    "parentVersionId": "dhwr8u"
  }
]
```


fergie-nz (changes requested):
Looks good though seeing error as per review in https://github.com/openmsupply/conforma-web-app/pull/1538#pullrequestreview-1564485290

Maybe can discuss in person as I might not be implementing it right

fergie-nz (approved):

PR #1065, opened 2023-07-28
Improvements to job schedulers and locale/timezone display

CarlosNZ:
A few things I noticed when doing #1060, but rather than make that one any bigger I've done these in a separate PR:

- update docs to get "Server Prefs" up to date
- move schedulers to their own class, and make sure they are only launched *once* by the main server thread.
- added a "reschedule" method to schedulers so they can be updated when updating preferences
- can explicitly set "locale" and "timezone" in server prefs -- this affects the scheduled jobs and the display in the console:
  - "locale" specifies how it should be formatted (in console)
  - "timezone" is the timezone for displayed console times and the reference timezone for the scheduled jobs (Does *not* affect anything in front-end

You should be able to update schedules in prefs -- you can set them to the near future and check they run when expected. For example, to set Backups to run at 2:15 pm, set this in "server" prefs:
```json
"backupSchedule": {"hour": 14, "minute": 15 }
```


documentation/Preferences.md
```diff
@@ -22,14 +22,18 @@ The available properties are as follows (almost  are optional, as the system has
...
 - **`systemManagerPermissionName`**: The "system manager" is a special permission that has certain system management rights (but not as extensive as "Admin"). Any existing permission name can be used for this special permission, in which case it should be specified here. (Default: `systemManager`)
+- **`managerCanEditLookupTables`**: If `true`, then users with the above management permission are also allowed to view/edit lookup tables. (Default: `true`)
 - **`previewDocsMinKeepTime`**: Documents generated as part of the Preview functionality will be periodically cleaned up, as they have no lasting use. It should be a Postgres duration string. (Default: "2 hours").
-- **`previewDocsCleanupSchedule`**: The schedule for cleaning up (deleting) the preview documents, as per the node-schedule syntax above. (Default: daily at 1:05am UTC)
+- **`fileCleanupSchedule`**: The schedule for cleaning up (deleting) and missing files, orphan file database records, and files marked as "to be deleted" (e.g Preview docs), as per the node-schedule syntax above. (Default: daily at 1:05am UTC)
```
CarlosNZ, line 27: Have made the names of these schedules more consistent and accurate:
- actionSchedule
- fileCleanupSchedule
- backupSchedule
- archiveSchedule

This would have an effect on the backwards compatibility of existing systems. But considering we'll mostly just use the defaults and the fact that the timezone was messing up the schedules anyway, this won't matter.

src/components/scheduler.ts
```diff
@@ -0,0 +1,172 @@
+import DBConnect from './databaseConnect'
```
CarlosNZ, line 1: This is not all new, just that I've moved the file out of "actions" (since it is no longer relevant to just actions).

But it's a lot cleaner now, and won't run automatically when imported since the schedulers are now in a class.

src/config.ts
```diff
@@ -98,26 +98,31 @@ export const refreshConfig = (config: Config, prefsFilePath: string) => {
...
   }
-  if (serverPrefs.archiveSchedule) {
-    reschedule('archive', serverPrefs.archiveSchedule)
+
+  //Update scheduled jobs from prefs
+  if (config.scheduledJobs) {
+    config.scheduledJobs.reschedule('action', serverPrefs.actionSchedule)
+    config.scheduledJobs.reschedule('backup', serverPrefs.backupSchedule)
+    config.scheduledJobs.reschedule('cleanup', serverPrefs.fileCleanupSchedule)
+    config.scheduledJobs.reschedule('archive', serverPrefs.archiveSchedule)
```
CarlosNZ, line 119: Ideally we would compare these to the existing and only update if changed, but it's not trivial to compare them due to the input structure being quite different to how they're stored internally, so not really worth the hassle. There's not really a problem rescheduling them to the same time (if unchanged)

src/server.ts
```diff
@@ -48,16 +49,24 @@ import { routeTriggers } from './components/other/routeTriggers'
 import { extractJWTfromHeader, getTokenData } from './components/permissions/loginHelpers'
 import migrateData from '../database/migration/migrateData'
 import routeArchiveFiles from './components/files/routeArchiveFiles'
+import { Schedulers } from './components/scheduler'
 require('dotenv').config()
 
-// Fastify server
+// Set the default locale and timezone for date-time display (in console)
+Settings.defaultLocale = config.locale ?? Intl.DateTimeFormat().resolvedOptions().locale
```
CarlosNZ, line 56: This is right at the top of everything in the entry point file, so the locale and timezone will be correctly set for _all_ subsequent DateTime instances.

src/server.ts
```diff
@@ -178,6 +189,9 @@ const startServer = async () => {
       process.exit(1)
     }
     console.log(generateAsciiHeader(config.version))
+    console.log(DateTime.now().toLocaleString(DateTime.DATETIME_HUGE_WITH_SECONDS))
+    console.log('Locale:', Settings.defaultLocale)
+    console.log('Timezone:', Settings.defaultZoneName)
```
CarlosNZ, line 194: Useful to print this on server start so we can be completely sure what timezone the server thinks it's in and what the current time is in that timezone.

database/migration/databaseMethods.ts
```diff
@@ -224,6 +227,32 @@ const databaseMethods = {
...
+        await DBConnect.query({
+          text: `UPDATE file SET file_size = $1
+          WHERE id = $2`,
+          values: [size, fileId],
+        })
+      } catch {
+        console.log('Problem updating file size for:', fullPath)
+      }
+    }
+  },
```
CarlosNZ, line 255: This is not really part of this PR, and not strictly required, but I find it useful for inspecting the file table.

CarlosNZ: Closes #1067

fergie-nz (approved):
Nice :)

PR #1073, opened 2023-10-30
#1070 External API endpoint

### LINKED ISSUE #1070: Endpoint for configurable access to secure APIs
When accessing authenticated APIs from the front-end, we don't want to have to expose our private API key to the front-end. So we'll need some way for the back-end to do the actual API access and have the front-end just call a server endpoint to "relay" this data.

Use cases:
- Angola IMA API
- Fiji mSupply item lookups

### Initial requirements:
- configurable -- this will probably be the same mechanism we use to make calls to mSupply or HSH, for example
- API keys must be stored on server, probably in a JSON file
- some way to configure permissions and restrict access, e.g. an applicant might be restricted to only do a name search using their own name or NationalID as the parameter
- ???

Becomes a bit trickier if we need to connect to servers that require re-authentication periodically, but I think (hope?) these initial two will be able to use a static API key or JWT for initial implementation.

CarlosNZ:
Fix #1070 

Provides a new endpoint `/external-api` that allows restricted, authenticated requests to external servers. All authentication credentials and restriction configurations are defined in the back end.

Please see the Documentation page "External-API-Access" for an explanation of how it works, with detailed examples.

To try out a working example, please ask me and I'll send you an Angola snapshot with info about how to use it.

**Please note:** the diff here is very big, but most of the changed files are due to having to upgrade Typescript to use Axios, so you can ignore most of these. The files directly relevant to this issue are:
- `documentation/External-API-Access.md`
- `documentation/API.md`
- `src/components/external-apis/routes.ts`
- `src/components/external-apis/helpers.ts`
- `src/components/external-apis/types.ts`
- `src/server.ts`
- `src/types.ts`
- `src/components/actions/evaluatorFunctions.ts`


package.json
```diff
@@ -42,8 +42,8 @@
     "readline-sync": "^1.4.10",
     "rimraf": "^3.0.2",
     "ts-jest": "^26.2.0",
-    "ts-node": "^10.1.0",
-    "typescript": "^3.9.7"
+    "ts-node": "^10.9.1",
+    "typescript": "^4.0.3"
```
CarlosNZ, line 46: Had to upgrade Typescript to work with latest version of Axios. TS 4.0.3 is what we're currently using in front end.

documentation/External-API-Access.md
```diff
@@ -0,0 +1,201 @@
...
+
+- `queryParams`: these are a set of query parameters (`?key=value` in url) that will be inserted into every query to this route. The `value` of each can be an [Evaluator expression](Query-Syntax.md), so they can be generated dynamically for each request. Other query parameters can be supplied by the client request, although the ones defined here will take precedence in the event of the same key being used.
+- `allowedClientQueries`: if defined, the front-end client can only use these query keys. Any others will be ignored.
+- `permissions`: if defined, the client request must have one of these permissions (in its JWT token) in order to proceed. 
+- `returnProperty`: the name of the property from the data returned by the API to return to the client. If the API returns more data than we would like the client to have access to, we could restrict it here. and we only want, say, a last name, we could use `"name.lastName"` for example.
+- `returnProperty`: the name of the property from the data returned by the API to return to the client. If the API returns more data than we would like the client to have access to, we could restrict it here. (E.g. `name.lastName`)
```
andreievg, line 62: returnProperty listed twice here

documentation/External-API-Access.md
```diff
@@ -0,0 +1,201 @@
...
+        returnProperty: string
+        validationExpression?: <EvaluatorExpression>
+    }
+}
+```
+
+Looking at these properties in more detail:
+
+- `queryParams`: these are a set of query parameters (`?key=value` in url) that will be inserted into every query to this route. The `value` of each can be an [Evaluator expression](Query-Syntax.md), so they can be generated dynamically for each request. Other query parameters can be supplied by the client request, although the ones defined here will take precedence in the event of the same key being used.
+- `allowedClientQueries`: if defined, the front-end client can only use these query keys. Any others will be ignored.
```
andreievg, line 59: allowedClientQueryParams ? Or too long ?

src/components/actions/previewActionsRoute.ts
```diff
@@ -57,6 +57,7 @@ export const createDisplayData = (actionsOutput: ActionResult[]): ActionResultDi
             result.output?.document?.filename ??
             'Generated Document',
           fileId: result.output?.document?.uniqueId,
+          filename: result.output?.document?.filename,
```
andreievg, line 60: Is this from another PR ?

src/components/external-apis/routes.ts
```diff
@@ -0,0 +1,130 @@
...
+    queryParams,
+    allowedClientQueries,
+    returnProperty,
+    additionalAxiosProperties,
+    validationExpression,
+  } = routeConfig
+
+  if (permissions) {
+    const { permissionNames } = await getPermissionNamesFromJWT(request)
+    const validPermissions = permissions.filter((permission) =>
```
andreievg, line 51: should all permissions be validated ?

const notMatchedPermission = permissions.find(permission) => !permissioneName.includes(permission))

src/components/external-apis/routes.ts
```diff
@@ -0,0 +1,130 @@
...
+    )
+
+    if (validPermissions.length === 0) {
+      reply.status(403)
+      return reply.send('Not authorized to view this resource')
+    }
+  }
+
+  // Construct data object for subsequent expression evaluator
+  const { userId, orgId } = request.auth as { userId?: number; orgId?: number }
```
andreievg, line 62: do we need 'as' here, it looks like the type should already have this as per method header

src/components/external-apis/routes.ts
```diff
@@ -0,0 +1,130 @@
...
+  const { user } = await getUserInfo({ userId, orgId })
+  const evaluatorData: {
+    user: typeof user
+    applicationData?: ActionApplicationData
+    functions: typeof functions
+  } = { user, functions }
+
+  // ApplicationData only available if an applicationId is provided as a query
+  // parameter
+  const { applicationId } = request.query as { applicationId?: number }
```
andreievg, line 72: can this also be in request type definition in method header ?

src/components/external-apis/routes.ts
```diff
@@ -0,0 +1,130 @@
...
+  if (applicationId) evaluatorData.applicationData = await getApplicationData({ applicationId })
```
andreievg, line 73: Do we need to check user permission to application ?

src/components/external-apis/routes.ts
```diff
@@ -0,0 +1,130 @@
...
+    request.query as QueryParameters,
+    queryParams,
+    allowedClientQueries,
+    evaluatorData
+  )
+
+  if (method === 'post') {
+    const { bodyJson, allowedClientBodyFields } = routeConfig as PostRoute
+    if (request.body || bodyJson) {
+      axiosRequest.data = await constructQueryObject(
```
andreievg, line 91: I think we might need to differentiate between queryParams and post body json object

andreievg (commented):
Looks great, well documented, easy to read and design seems solid. I have a couple of questions, maybe we can discuss in person.

Thanks for mentioning about the typescript upgrade in the PR description (although I still prefer separate PR, but this is the best alternative)

documentation/External-API-Access.md
CarlosNZ: 👍

src/components/actions/previewActionsRoute.ts
CarlosNZ: Yeah, forgot to mention this, forgot it was in here. It's just for the front-end "Document Preview" modal, we need this additional field, but didn't justify a PR of its own.

src/components/external-apis/routes.ts
CarlosNZ: I'm not quite clear what you're suggesting here. Do you mean that the user should have *all* of the specified permissions? I was thinking *any* of the specified permissions would be fine (so *OR* not *AND*), for 3 reasons:
- this is consistent with how it works for Data Views (if you have any of the listed permissions you can view)
- it seems to make more sense that we'd be defining in terms of groups of people who can view something. For example, if we had the `applyCompany` permission required, and then we wanted to include `admin` so Admins could do it too, then we wouldn't want suddenly exclude company applicants.

Or am I missing what you meant?

src/components/external-apis/routes.ts
CarlosNZ: It's just that `request.auth` can be undefined, so it then complains about the destructuring. But in reality, it couldn't actually be undefined by this point, as it's checked upstream and only authenticated requests even get this far.

I've changed it to this, which works, albeit adding a tiny bit of runtime overhead vs purely compilational:
```
const { userId, orgId } = request.auth ?? {}
```

src/components/external-apis/routes.ts
CarlosNZ: The "query" property is `unknown` in Fastify types, so we have to define it somewhere, or add a runtime check (which seems unnecessary in this case).

In fact, I'm pretty sure I copied this style of typing the request query with `as` in its first usage from you: 

Screenshot 2023-11-12 at 1 23 25 PM

😉

src/components/external-apis/routes.ts
CarlosNZ: As discussed IRL, they're both handled separately, but in essentially the same way. Have added `allowedClientBodyFields` to docs to clarify (was missing before)

src/components/external-apis/routes.ts
CarlosNZ: Yeah, cool. Done. Easiest way to make sure the front-end user can access a certain application, though, was to just make an attempt to fetch the application with user's JWT. If no result, then no go.

CarlosNZ:
Okay, @andreievg, have addressed those comments now. Take another look at your convenience. 7ddd98f9

Thanks for the review.

src/components/external-apis/routes.ts
CarlosNZ: I've made it so that if the application doesn't pass permission test, it just silently ignores applicationData and carries on with the request. You think it'd be better to return error?

CarlosNZ:
I added one more small change -- the SMTP config now uses the same "env" variable replacement method that this uses, for consistency. 05bb966e

src/components/external-apis/routes.ts
andreievg: Ahh ok, sorry i missed: "if defined, the client request must have one of these permissions (in its JWT token) in order to proceed"

src/components/external-apis/routes.ts
andreievg: I've been anti ` as ` lately (as i've seen it misused in open mSupply), in general I think ` as ` is fine as long as we don't 'leak' the type further, i.e. when we cast to as and return that type somewhere else and use it, it's hard to validate that the type is indeed correct, have to jump to a lot of places to confirm

src/components/external-apis/routes.ts
andreievg: Hmm, well technically user should not be on the application if they don't have the permission right ? So either way is fine i guess

andreievg (approved):
Looks good to me, thanks for the changes.

A couple of things you may want to do:

https://github.com/msupply-foundation/conforma-server/compare/1070-external-api-server...1070-external-api-server(Rename-allowedClientQueries-to-allowedClientQueryParams-in-readme)

And this one is nit:, I wanted to know how it's done in fastify:

https://github.com/msupply-foundation/conforma-server/compare/1070-external-api-server...1070-external-api-server(Remove-castin-to-as-for-fastify-request)

src/components/external-apis/routes.ts
CarlosNZ: Yeah, and if the query returns valid content, then the applicationData wasn't needed. And if the applicationData is needed, then they'll get an error anyway if it can't be evaluated. :)

CarlosNZ:
> [1070-external-api-server...1070-external-api-server(Rename-allowedClientQueries-to-allowedClientQueryParams-in-readme)](https://github.com/msupply-foundation/conforma-server/compare/1070-external-api-server...1070-external-api-server(Rename-allowedClientQueries-to-allowedClientQueryParams-in-readme))

Good catch, thanks.

> And this one is nit:, I wanted to know how it's done in fastify:

> [1070-external-api-server...1070-external-api-server(Remove-castin-to-as-for-fastify-request)](https://github.com/msupply-foundation/conforma-server/compare/1070-external-api-server...1070-external-api-server(Remove-castin-to-as-for-fastify-request))

Nice, but the only trouble is that now it doesn't like the `auth` property on the request:
Screenshot 2023-11-23 at 5 18 27 PM

Adding an "auth" property to your `AccessExternalApiQuery` doesn't help either, since "auth" is something we've added into the request (`server.ts`, line 99) so it doesn't think that it's a valid property of the Generic.

Did you manage to get it compile, @andreievg ?

CarlosNZ:
> Did you manage to get it compile, @andreievg ?

I can get it to compile if I use this ugly `as` here, but that's barely better than what we started with.

Screenshot 2023-11-23 at 5 24 04 PM

Any idea how to get it to accept additional injected props on FastifyRequest upstream?

andreievg:
As per telegram:

Andrei, [23/11/2023 at 5:36:35 PM]:
I got it to build. Maybe that's different then running it in dev ?

It's not important

I should have left it

Just wanted to know

PR #1076, opened 2023-11-09
#1200F Token expiry check on server

CarlosNZ:
Back-end required for front-end issue: https://github.com/msupply-foundation/conforma-web-app/issues/1607

(Front-end issue coming shortly)

Just does a basic check on requests to REST server to only allow requests with a current JWT. Expired tokens will return a 401 unauthorised error.

Note, this does **NOT** yet work for Postgraphile requests. To do that we need to add some middleware to serve Postgraphile through the main Fastify server (something we want to do when we have time -- @guatemartin is this something you know much about?) 

To test, just set the `logoutAfterInactivity` preference in server preferences to something small, then try and make REST requests after expiry.


src/components/permissions/loginHelpers.ts
```diff
@@ -105,6 +106,9 @@ const getUserInfo = async (userOrgParameters: UserOrgParameters) => {
       isManager,
     },
     orgList,
+    tokenExpiry:
+      parseInt(String(Date.now() / 1000)) +
+      (config.logoutAfterInactivity ?? DEFAULT_LOGOUT_TIME) * 60,
```
CarlosNZ, line 111: We're returning the expiry time to the front-end, although I'm not actually using it (the front-end knows how long the token lifetime is via preferences and refreshed accordingly)

src/components/preferences/routes.ts
```diff
@@ -33,7 +33,15 @@ export const routeGetPrefs = async (request: any, reply: any) => {
   const languageOptions = readLanguageOptions()
   const latestSnapshot = await databaseConnect.getSystemInfo('snapshot')
   const allowedTableNames = config.allowedTableNames
-  reply.send({ preferences: prefs.web, languageOptions, latestSnapshot, allowedTableNames })
+  const logoutAfterInactivity = prefs.server.logoutAfterInactivity ?? DEFAULT_LOGOUT_TIME
+
+  reply.send({
+    preferences: { ...prefs.web, logoutAfterInactivity },
```
CarlosNZ, line 39: This preference value is used by both server and client, so we need to pop it in here from the server values.

guatemartin:
> Back-end required for front-end issue: [msupply-foundation/conforma-web-app#1607](https://github.com/msupply-foundation/conforma-web-app/issues/1607)
> 
> (Front-end issue coming shortly)
> 
> Just does a basic check on requests to REST server to only allow requests with a current JWT. Expired tokens will return a 401 unauthorised error.
> 
> Note, this does **NOT** yet work for Postgraphile requests. To do that we need to add some middleware to serve Postgraphile through the main Fastify server (something we want to do when we have time -- @guatemartin is this something you know much about?)
> 
> To test, just set the `logoutAfterInactivity` preference in server preferences to something small, then try and make REST requests after expiry.

@CarlosNZ - Not in the context of a Fasify server  or Postgraphile for that matter.  I can look at it though if needed.

PR #1079, opened 2023-11-27
1077 improve application list performance

andreievg:
Partially fixes: https://github.com/msupply-foundation/conforma-server/issues/1077

Replaced `review_assignment_available_sections` (which is called for every row), with a view and function that return array and an array comparison, approximate 1.5x improvement without row level permissions.

Creates a view permission_flattened, which is used in policies instead of a in statement that extracts array from jwt (i think postgres can't quite optimise in statements with arrays of values) approximate 2.5x improvement.

@CarlosNZ, snapshot loading is super fast (considering the size of datafile), thanks for doing the work on that, super cool.

We may want to remove some unused stuff from trigger/helpers (like review_actions functions etc.., it's a carry over from previous application_list). Also probably should remove jwtPermission_bigint_ and jwtPermission_array_bigint_ from permission policy generation


database/buildSchema/43_views_functions_triggers.sql
```diff
@@ -1516,6 +1516,74 @@ GROUP BY
...
+    CASE WHEN allowed_sections IS NULL THEN
+        array_agg(distinct(template_section.code))
+    ELSE
+        allowed_sections
+    END AS allowed_sections
+    FROM review_assignment
+        JOIN application ON review_assignment.application_id = application.id
+        JOIN template ON application.template_id = template.id
+        JOIN template_section ON template_section.template_id = template.id
+    WHERE review_assignment.status = 'AVAILABLE'
```
andreievg, line 1541: pre filtered by the same filters that are also in the CASE WHEN clause, similar with the method below

database/migration/migrateData.ts
```diff
@@ -854,6 +856,83 @@ const migrateData = async () => {
     }
   }
 
+  // v0.8.0
+  if (databaseVersionLessThan('0.8.0')) {
+    console.log('Migrating to v0.8.0...')
+
+    console.log(' - Creating permission_flattened view')
+    await DB.changeSchema(`
+        CREATE INDEX IF NOT EXISTS "i_review_status_history_status" ON review_status_history (status);
```
andreievg, line 865: I also tried this with a table, where user_id, organisation_id and policy_id are indexed (individually and as one index), and it was around 1.2x, which is small improvement but requires a bit of maintenance of the table (I had it refreshing for a user when they log in or change or, i was worried that indexes might get slower the more we drop and re-insert rows, and it's more moving parts, so just went with a view)

database/migration/migrateData.ts
```diff
@@ -854,6 +856,83 @@ const migrateData = async () => {
...
+    console.log(
+      ' - Updating policies to replace jwtPermission_array_bigint_template_ids with query rathern then in statement matching arrays in JWT'
+    )
+    const policies = (
+      await DBConnect.query({
+        text: `SELECT id, rules FROM permission_policy`,
+      })
+    ).rows as { id: number; rules: object }[]
+
+    // This is added as an optimisation, rather then doing "in" clause with array values from JWT
```
andreievg, line 884: I was wondering if I need to add this $in replacement to some 'core' place for policies, but I think they should get updated with migration ? Also any future policies should use this

package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "conforma-server",
-  "version": "0.6.1-30",
+  "version": "0.8.0",
```
andreievg, line 3: There was a migration for 0.7 but version is set at 0.6, was that an issue ?

database/migration/migrateData.ts
```diff
@@ -892,10 +971,12 @@ const migrateData = async () => {
...
-  migrateData().then(() => {
-    console.log('Done!\n')
-    process.exit(0)
-  })
+  migrateData()
+    .then(() => {
+      console.log('Done!\n')
+      process.exit(0)
+    })
+    .catch((e) => console.log('Error while migrating', e))
```
andreievg, line 979: I had a problem with migration, thought it was an error, but didn't realise i need to restart server before re-loading snapshot, while diagnosing adding this, can remove, but I think it doesn't hurt having it here ?

database/migration/migrateData.ts
```diff
@@ -854,6 +856,83 @@ const migrateData = async () => {
...
+      })
+    ).rows as { id: number; rules: object }[]
+
+    // This is added as an optimisation, rather then doing "in" clause with array values from JWT
+    // we "select" from permission_flattened table, pre populated with userId, organisationId, permissionPolicyId and templateIds
+    // from: template_id in any (string_to_array(COALESCE(current_setting('jwt.claims.pp6_templateIds, true), '0'), ',')::integer[])
+    // to: template_id in select template_id from permission_flattened
+    //         where user_id = COALESCE(current_setting('jwt.claims.userId, true),''),'0')::integer
+    //         and permission_policy_id = 6
+    //         and (orgainisation_id = COALESCE(current_setting('jwt.claims.orgId, true),''),'0')::integer or organisation_id is null)
```
andreievg, line 890: This comment should explain how the change works

database/migration/migrateData.ts
CarlosNZ: Can you put all views etc in `43_views_functions_triggers.sql` -- we drop all of them and re-create them every time, which prevents a lot of dependency problems, and is super fast. Save the migration script for actual schema changes or data migrations.

database/migration/migrateData.ts
```diff
@@ -854,6 +856,83 @@ const migrateData = async () => {
...
+      const replacementAsJson = JSON.parse(
+        jsonAsString.replace(/"jwtPermission_array_bigint_template_ids"/g, replacementAsString)
+      )
+
+      await DBConnect.query({
+        text: `UPDATE permission_policy set rules = $1 where id = $2`,
+        values: [replacementAsJson, id],
+      })
+    }
+    await updateRowPolicies()
```
CarlosNZ, line 933: This gives an error, as it's actually making a request to an endpoint that doesn't exist yet (migrate script runs before server starts). 

It's actually confusing -- there's two different `updateRowPolicies` defined -- one is in `rowLevelPolicyHelpers.ts` and is the actual function we want. But the other is in `updateRowPolicies.ts` and just queries the endpoint which runs that function. We need to refactor this so we don't keep making the same mistake (it happens on snapshot load during the docker build too) -- so you should be calling the actual `updateRowPolicies` function, and we might even be able to drop the other one (we shouldn't be calling a function that queries an endpoint in the back-end, and the front-end should just be querying the endpoint). 

I can fix this if you like, if you could just change this to import the correct version.

(In fact, we should go a step further and always run this on server restarts/snapshot reloads, so I'll change it to do that properly after this is sorted)

database/migration/migrateData.ts
CarlosNZ: It's designed to work two ways:
- you can manually run migration from the command line with `yarn migrate <version>`, (that's `isManualMigration`)
- it runs every time when the server starts (and checks for current version in DB)

I think you probably only needed to restart the server due to the `updateRowPolicies` error I mentioned above.

package.json
CarlosNZ: The version here shouldn't be changed except when we do releases/builds. The latest migrations should always be higher than the current version (but less or equal to the next intended release). The migrations are all safe in that they can be re-run over and over without causing problems (they either fail silently if they try to change something that's already changed, or we've put an explicit check before executing).

CarlosNZ (changes requested):
Hey, looks good. I haven't had a chance to take a proper look at the actual logic yet, but I've just made a couple of comments regarding processes and things that aren't working right.

We should probably talk through the main logic in the office tomorrow if you're around?

andreievg:
> We should probably talk through the main logic in the office tomorrow if you're around?

Yep that's sounds good, thanks for review so far

database/buildSchema/43_views_functions_triggers.sql
```diff
@@ -1516,6 +1516,74 @@ GROUP BY
...
+            AND ra_inner.level_number = ra_outer.level_number
+            AND ra_inner.status = 'ASSIGNED')
+    WHERE ra_outer.status = 'AVAILABLE' 
+        AND ra_outer.is_self_assignable = TRUE
+        AND application.outcome = 'PENDING'
+        AND ra_outer.reviewer_id = userId
+    GROUP BY 1
+$$
+LANGUAGE sql
+STABLE;
```
CarlosNZ, line 1584: We already do something very similar to this with the `enforce_asssigned_section_validity` function (near the top of this script). I can't help thinking it would be good to just add another couple of columns to the review_assignment table where we can store the result of this calculation too. The trigger is already in place to update it whenever it changes, and it's been pretty robust, so I don't think we'd have a problem keeping it up to date.

Let's discuss...

database/migration/migrateData.ts
andreievg: Removed: 3afc7948d6291e4a0a4816b573c5ba86ce013aea

package.json
andreievg: I changed back to what's on develop: c85e0b83761e78f7e4c80dda10d7ba28ac98716c

But I don't understand, there is migration for 0.7 before i added 0.8, but the version was below that, so: ` latest migrations should always be higher than the current version (but less or equal to the next intended release)` wasn't true. 

So to test current migration work, temporary change package version, but don't commit ?

database/buildSchema/43_views_functions_triggers.sql
andreievg: This would be helpful, but it would only work for subset of reviewer actions, although I think it would still speed it up a bit (since these are the slow ones)

andreievg:
I rebased to: https://github.com/msupply-foundation/conforma-server/pull/1080, and I think did all of the change request, then check out https://github.com/msupply-foundation/conforma-server/pull/1080, and loaded snapshot. Stopped server, checked out this branch, change version in package.json to 0.8.0. Started server, and saw the performance increase, i think it's working as expected with migrations

andreievg:
There is a carry over TODO, to remove unused views/functions, there is also a carry over to only expose things we are updating/viewing in API

database/buildSchema/43_views_functions_triggers.sql
CarlosNZ: Yeah, I'll keep thinking about this, try and document all the relevant events/actions and see if it makes sense.

database/migration/migrateData.ts
```diff
@@ -850,6 +851,75 @@ const migrateData = async () => {
...
   }
 
+  // v0.8.0
+  if (databaseVersionLessThan('0.8.0')) {
+    console.log('Migrating to v0.8.0...')
+
+    console.log(' - Creating permission_flattened view')
+    await DB.changeSchema(
+      `CREATE INDEX IF NOT EXISTS "i_review_status_history_status" ON review_status_history (status);`
+    )
```
CarlosNZ, line 861: Can this one go in `44_index.sql` please -- probably next to this line:
```
CREATE INDEX IF NOT EXISTS "i_review_status_history_review_id_fkey" ON review_status_history (review_id);
```

Just keeps all the same things organised in one place.

package.json
CarlosNZ: > But I don't understand, there is migration for 0.7 before i added 0.8, but the version was below that, so: ` latest migrations should always be higher than the current version (but less or equal to the next intended release)` wasn't true.

Yeah, the current version is `0.6.1-pre-releaseX`. So any new migrations go in `0.7.0` block. After this stuff is merged, I'll make the new build 0.7.0. Then the next lot of migrations will go in new `<0.8.0` block.

CarlosNZ (approved):

PR #1080, opened 2023-11-28
232T & docker script refactoring

CarlosNZ:
Includes changes required for https://github.com/msupply-foundation/conforma-templates/issues/232, plus some improvements to the docker scripts and the updateRowPolicies.

## Summary of changes

- New Action: "generateTextFile" which allows us to output JSON/text files to anywhere on the system
- New Action: "getValues" -- the simplest action yet, but also probably one of the most useful. If we're querying the same value in several different actions in sequence, then it makes sense to just fetch this value once and pass it along to subsequent actions in the output. This makes the evaluator expressions for subsequent actions way simpler. In this case, I wanted to query the next available port number from the database and it seemed silly (not to mention tedious and error-prone) to do it 3 or 4 different times in sequence.
- Update Dockerfile to be able to specify a snapshot -- if defined the image will be built with this as the base snapshot rather than "core_templates"
- Get rid of the unnecessary "updateRowPolicies" function -- all it did was query an endpoint which in turned called the *actual* "updateRowPolicies" function. Have added `updateRowPolicies()` to server start (it was already in snapshot load), so we no longer need the separate command line version of if either as part of the docker scripts
- Made environment variables available to "applicationData" -- but only those explicitly defined in server preferences `envVars` field


database/insert_data.sh
```diff
@@ -17,8 +17,6 @@ fi
 PID=$!
 wait $PID
 
-yarn ts-node ./database/updateRowPoliciesCLI.ts &
```
CarlosNZ, line 20: No longer needed -- we run `updatePolicies` at the end of snapshot load, plus whenever the server restarts.

database/migration/migrateData.ts
```diff
@@ -7,11 +7,7 @@ import path from 'path'
 import { readFileSync } from 'fs'
 import bcrypt from 'bcrypt'
 import { errorMessage, getAppEntryPointDir } from '../../src/components/utilityFunctions'
-import {
-  loadCurrentPrefs,
-  routeGetAllPrefs,
```
CarlosNZ, line 12: Unused import

docker/Dockerfile
```diff
@@ -27,6 +27,7 @@ RUN apt-get install lsof -y
 RUN apt-get install nginx -y
 
 # --- LIBRE OFFICE
+RUN apt-get update
```
CarlosNZ, line 30: Libre office install was failing, but this seems to fix it.

docker/Dockerfile
```diff
@@ -90,10 +91,15 @@ COPY postgres_config/postgresql.conf .
 # RUN echo "listen_addresses = '*'" >> postgresql.conf
 COPY postgres_config/pg_hba.conf .
 
-# --- INITIALISE DATABASE (core templates)
+# --- INITIALISE DATABASE (default: core templates)
+ARG INITIAL_SNAPSHOT=""
+WORKDIR /usr/src/conforma-server/database/_snapshots
+COPY snapshots/${INITIAL_SNAPSHOT}/ ./${INITIAL_SNAPSHOT}
```
CarlosNZ, line 97: We copy the specified snapshot into the container's "snapshots" folder, so it can be loaded with `useSnapshot`

(Note: for now, snapshot (unzipped) needs to be places in `/docker/snapshots`)

docker/Dockerfile
```diff
@@ -90,10 +91,15 @@ COPY postgres_config/postgresql.conf .
...
+
 WORKDIR /usr/src/initialise_database
 COPY database.sh .
-RUN ./database.sh
+RUN ./database.sh ${INITIAL_SNAPSHOT}
+RUN rm -rf /usr/src/conforma-server/database/_snapshots/${INITIAL_SNAPSHOT}
```
CarlosNZ, line 102: And then remove it so we don't have a whole duplicated copy of the snapshot included in image.

docker/database.sh
```diff
@@ -24,7 +24,11 @@ echo '--- RUNNING POST INSTALL'
...
-rm -r localisation
-rm -r preferences
+# Loading a snapshot from here puts these folders in the repo root rather than
+# the "build" folder so, need to move them in:
+rm -r build/files
+rm -r build/preferences
+rm -r build/localisation
+mv files build
+mv preferences build
+mv localisation build
```
CarlosNZ, line 34: These were getting re-set to defaults, which didn't matter with `core_templates` as they were empty/default, but with a custom snapshot we need to keep the snapshot versions.

docker/run.sh
```diff
@@ -2,8 +2,8 @@
 
 # To exit container, type 'exit'
 
-# Get SMTP password from .env file
-export $(grep -v '^#' ../.env | xargs)
+# Get SMTP password etc. from .env file
+source ../.env
```
CarlosNZ, line 6: Same thing, but nicer?

documentation/List-of-Action-plugins.md
```diff
@@ -606,6 +606,33 @@ The output object `document` provides the `uniqueId`, `filename`, and `filepath`
...
+| `filename`\*                             |                              |
+| `outputPath`                             |                              |
+| `registerInDatabase`                     |                              |
+| `fileData`                               |                              |
+
+The file contents can be provided by *either* `text` (`string`), or `data` (`object`), with `text` taking precedence if both are present. The `data` object is serialised to JSON for text output.
+
+A `filename` is required.
+
+The `outputPath` can be either a relative path (e.g. `text/myFolder`) or an absolute one (e.g. `/Users/me/my/path`). If relative, then if the file is not to be registered in the database (i.e. it won't be referenced by Conforma in future), the path will be relative to the Home folder. If it *is* to be registered in the database, it will be placed relative to the system's "files" folder. If no `outputPath` is provided, it will default to `_textFiles`.
```
CarlosNZ, line 628: Any files to be kept in database *must* be in `files` folder since the database "path" field is relative to this.

But we don't want such a restriction for other files, hence they'll get output relative to "home" folder.

src/components/actions/getApplicationData.ts
```diff
@@ -71,6 +71,9 @@ export const getApplicationData = async (input: {
       }
     : {}
 
+  const env = Object.fromEntries(
+    Object.entries(process.env).filter(([key, _]) => config.envVars?.includes(key))
+  )
```
CarlosNZ, line 76: Expose environment variables, but only if defined in system preferences.

src/components/files/fileHandler.ts
```diff
@@ -107,7 +115,32 @@ export async function saveFiles(data: any, queryParams: HttpQueryParameters) {
   return filesInfo
 }
 
-export async function registerFileInDB({
+// For registering a file that already exists on disk
```
CarlosNZ, line 118: The existing `registerFileInDB` function assumed the files were coming from uploading via endpoint. However, this is not the case any more, so I've changed that function to be just `saveToDB`, which is essentially all it was doing. Now "register" normalises the file data based on limited input and then passes to `saveToDB`.

This made it much easier for calling from "generateTextFile" action

src/components/snapshots/useSnapshot.ts
```diff
@@ -5,7 +5,7 @@ import path from 'path'
 import { execSync } from 'child_process'
 import insertData from '../../../database/insertData'
 import DBConnect from '../../../src/components/databaseConnect'
-import updateRowPolicies from '../../../database/updateRowPolicies'
+import { updateRowPolicies } from '../permissions/rowLevelPolicyHelpers'
```
CarlosNZ, line 8: We only ever use this one now -- the other is 💨 !

src/server.ts
```diff
@@ -73,6 +74,7 @@ const startServer = async () => {
   await loadActionPlugins() // Connects to Database and listens for Triggers
   createDefaultDataFolders()
   await cleanUpFiles() // Runs on schedule as well as startup
+  await updateRowPolicies()
```
CarlosNZ, line 77: Run it on server start too, just to be sure.

src/types.ts
```diff
@@ -325,6 +327,7 @@ export interface WebAppPrefs {
   brandLogoFileId?: string
   brandLogoOnDarkFileId?: string
   defaultListFilters?: string[]
+  userRegistrationCode?: string
```
CarlosNZ, line 330: Small change to front-end (already merged) -- allows us to specify the code of the special "user registration" form rather than having it hard-coded (still has same default of `UserRegistration`, but I've changed it to `demoRegister` for the demo system)

documentation/List-of-Action-plugins.md
```diff
@@ -606,6 +606,33 @@ The output object `document` provides the `uniqueId`, `filename`, and `filepath`
...
+
+Set `registerInDatabase` to `true` in order to record the file record in the database. Note that you cannot use an absolute path when registering in database, as all database files are referenced relative to the "files" folder.
+
+Do **not** put any files in the "files" folder if they are not registered in the database, as they'll be deleted in the next scheduled "file cleanup".
```
andreievg, line 632: Do you think it's worth adding getValue one in this file ?

src/components/actions/getApplicationData.ts
andreievg: Should it be exposeEnvVarsToApplicationData ? And mentioned in config.ts ?

src/components/files/fileHandler.ts
andreievg: But we don't have a use case to save generated text file in db ? 

I haven't reviewd this change, is it necessary ?

andreievg (commented):
Thanks, quite easy to review, although I didn't review the saveToDb changes, approving everything else, do you have existing or planned use case for saveToDb without upload (to register generated file in db ?)

andreievg (approved):
Looks good, had a quick look at registerFileInDb, I think it if it's only used in that action that should be fine, but I am still not convinced that we should be implementing this sort of functionality pro-actively, but ofcourse trust your judgement on that.

Approving but pending the getValue doc and some reference to the config.envVars (and potentially renaming of that to specify what it means)

documentation/List-of-Action-plugins.md
CarlosNZ: Yup, done: 334bb5f5

CarlosNZ:
> Approving but pending the getValue doc and some reference to the config.envVars (and potentially renaming of that to specify what it means)

Cool, have addressed those. Thanks for taking a look. Will merge now.

PR #1084, opened 2023-12-19
Trigger throttling

CarlosNZ:
Required for https://github.com/msupply-foundation/conforma-templates/issues/242

It became apparent that the system can't really handle it when an event causes a lot of applications (>1000) to be triggered all at once. This is what happens when we run an SQL script to auto-update the current pending applications as described in https://github.com/msupply-foundation/conforma-templates/issues/242.

Which means we also need some kind of trigger throttling for when all the provisional products expire. It's not until 2026, but if we leave it as is then the expiries won't work reliably on that date.

I've created a new mechanism ensure triggers are processed sequentially when a whole bunch of triggers are fired all at once (by the database notifications). I tried to find an existing package to do this, but I could only find "debounce" type throttling -- i.e. the function is limited from running more than X often (and the extra ones get ignored). However, we want something a bit different -- every invocation needs to run, but they just need to be "spaced out" rather than all being called simultaneously. So I've made my own throttle mechanism. Basically, every "trigger notification" that comes from the database, instead of directly calling `processTrigger()` like we did previously, the `processTrigger` method gets passed to the Throttle object and queued up if there are too many.

See `throttle.ts` for more detailed comments explaining the mechanism.


src/components/actions/resumeProcessing.ts
```diff
@@ -0,0 +1,99 @@
+/**
+ * If the server is shut down while there are still triggers being processed in
+ * the "Throttle" queue, this will ensure that they get re-started. This should
+ * ONLY be called at start-up, before any other trigger/action related methods.
+ */
```
CarlosNZ, line 5: This shouldn't be needed in normal functioning, but if the server *does* happen to get shut down (or crash) during a large trigger batch process, then incomplete triggers would be lost and never complete.

PR #1089, opened 2024-02-06
Use non buffered zip handler

EthanMcQ-TMF:
An [existing issue](https://github.com/cthackers/adm-zip/issues/334) in `adm-zip`. The library will load the zip file data into a buffer, which means it can't handle a file bigger than `buffer.constants.MAX_LENGTH`, which is 2GB on 64-bit systems.

Our full snapshots can get larger than that limit which causes errors, which we can fix with a zip library that streams data.


CarlosNZ:
Hey, thanks for this. Looks good, except for one annoying bug which I can't figure out.

So if you export a snapshot and re-upload it using this it works just fine. However, if you export a template (from the "Templates" configuration) and re-import it, you get an error when unzipping:
Screenshot 2024-02-08 at 8 32 57 PM

The template export/import uses the same mechanism as snapshot (an exported template is basically a snapshot with different export settings), so I really can't tell what the difference is here.

See if you can figure out what's going on there, I've had a good play round with it and can't see where it might be going wrong.

(FYI -- you can upload a template .zip from the "snapshots" page, it works just the same but it won't install and delete the template, so we'll be comparing the exact same process -- probably easier for debugging)

CarlosNZ:
This seems to be some kind of bug in the new library, as far as I can tell. Try this:

- Export template from templates page
- Go to "Snapshots" page, try to upload, see error
- Edit the .zip file on your local system (I use BetterZip but other things would do)
- Add some irrelevant files into the zip, don't change what's already there.
- Resave .zip, then upload to Snapshots again -- it should work.

It also seems to be something to do with the size of the zip file. I can even link a file to a template export within Conforma (I can show you how to do this), and then now when I export the template exactly the same way, the .zip file is bigger and now it re-uploads again just fine.

Very odd...

src/components/snapshots/routes/routeUploadSnapshot.ts
```diff
@@ -8,9 +8,9 @@ import {
 import fs from 'fs'
 import { pipeline } from 'stream'
 import { promisify } from 'util'
-import zipper from 'adm-zip'
 import { timestampStringExpression } from './helpers'
 import { DateTime } from 'luxon'
+import StreamZip = require('node-stream-zip')
```
CarlosNZ, line 13:
```suggestion
import StreamZip from 'node-stream-zip'
```
Better to not use `require` any more.

EthanMcQ-TMF:
@CarlosNZ Looks like the error disappears if we specifically create the destination dir prior to extracting.

Not entirely sure why only some zips are affected. Presumedly the smaller archives are extracted quicker and causing a race condition in which stream is in charge of creating the dir.

CarlosNZ:
> Presumedly the smaller archives are extracted quicker and causing a race condition in which stream is in charge of creating the dir.

Yeah that makes sense. Thanks for the fix! :)

CarlosNZ (approved):
Great, thanks for the fixes. Works nicely :)

PR #1105, opened 2024-05-27
#1103 Memoise list actions

### LINKED ISSUE #1103: Memoise application list actions to speed up performance
There are four columns returned by the application_list function that are causing significant performance burdens:

- assigners (list of users who can assign current stage)
- reviewers (list of reviewers who *are* assigned to current stage)
- reviewer_action (what the current user/reviewer can do next to an application)
- assigner_action (what the current user/assigner can do next to an application)

The last two in particular make large demands, as they return different values depending on the user. Plus they are calculated (using fairly convoluted functions) for every application (of a particular type) on every request, even ones that don't appear in the results (as they must be filterable on those fields)

It makes sense to memoise these values and store them in the database, so we only need to calculate them once, and update them whenever an event occurs that would change them (for a specific user).

I suggest storing this info as follows:

- In existing `application` table, add two new columns for the lists:
  - `reviewer_list` (array of strings)
  - `assigner_list` (array of strings)
- New table: `application_reviewer_action` with the following columns:
  - `user_id`
  - `application_id`
  - `reviewer_action`
  - `assigner_action`
(Probably add a UNIQUE constraint on the four fields (as a combo) as well)

Then it's just the matter of capturing the events to trigger updates. I think we just need to trigger on
- create/update/delete of `review_assignments`
- create of `review_status_history` records

The trigger would then call a function that does the following:
- get the `reviewer_id`, `assigner_id` and the `application_id` related to the record in question (for `review_assignment` these are all directly on the table, for `review_status_history` and extra sub-query would be required)
- calculate the "reviewer_action" and "assigner_action" for each (using existing PG functions)
- delete all `application_reviewer_action` for this users/application combo (assigner and reviewer)
- re-insert records with new info
- fetch reviewer_list and assigner_list for this application_id (using existing PG functions)
- update the `reviewer_list` and `assigner_list` fields of the application

We don't need to worry about permissions -- we'll only be doing this for a particular user/application combo at one time, and will only be triggered by interacting with an application, which they must have already had permission to do.  

From there, we can ditch the current `application_list` function and just create a view that joins `application` to `application_reviewer_action`, and the list query would just be a simple query of this view (we might not even need a view, it could be fine to just connect them using graphQL query shape)

I think all the complicated logic to _determine_ these actions is already done and working fine -- so this refactor would only reduce the number of times we need to call these functions, but not change the underlying logic.

I also think that after this, we can fetch the total count as part of the same single query to the list, but we'll see.

CarlosNZ:
Fix #1103 

I'll add some explanatory comments in the PR (in addition to the in-code comments)

To test, play round with:
- assigning (to self and others)
- starting review
- submitting review
- re-submit application after change requests (as applicant)

After each of these events, note the changes to the `application_reviewer_action` table, and the `reviewers`/`assigners` lists on the `application` table.


database/buildSchema/43_views_functions_triggers.sql
```diff
@@ -1542,164 +1542,377 @@ $$
 LANGUAGE sql
 STABLE;
 
+-- Does the "heavy lifting" to deduce the assigners/reviewers list for a single
+-- application as well as the available actions for a particular
+-- reviewer/assigner for that application. Used to populate the
+-- `application_reviewer_action` table and the reviewer/assigner lists on each
+-- application whenever an event occurs that would change these values
+CREATE OR REPLACE FUNCTION single_application_detail (applicationId int, userId int DEFAULT 0)
```
CarlosNZ, line 1550: This function is a pure copy/paste from the existing list function. Differences are
- only for a single application (passed in `applicationId` parameter, and additional `WHERE` clause).
- removed all columns other than these 4 complicated ones -- the others are all just joined in the list query as before.

CarlosNZ: Still has a `DEFAULT` value since when generating the assigners and reviewers lists, we don't need to specify a user.

database/buildSchema/43_views_functions_triggers.sql
```diff
@@ -1542,164 +1542,381 @@ $$
...
+            WHERE id = NEW.application_id;
+    END IF;
+
+    -- Get list of reviewer_ids to update for 
+    IF TG_OP = 'UPDATE' THEN
+        reviewer_ids = ARRAY(
+            SELECT reviewer_id FROM review_assignment
+            WHERE application_id = NEW.application_id
+        );
+    ELSE
```
CarlosNZ, line 1685: When `review_assignments` are created, it'll run once for each assignment as rows are inserted. But when *updating*, we need to run for all potentially affected reviewers, hence we loop over the list of connected users.

database/buildSchema/43_views_functions_triggers.sql
```diff
@@ -1542,164 +1542,381 @@ $$
...
+        reviewer_ids = ARRAY[NEW.reviewer_id];
+    END IF;
+
+    FOREACH rev_id IN ARRAY reviewer_ids LOOP
+        -- Remove existing
+        UPDATE public.application_reviewer_action
+            SET reviewer_action = NULL
+            WHERE application_id = NEW.application_id
+            AND user_id = rev_id;
```
CarlosNZ, line 1694: I tried a few different approaches for handling the existing records, ended up with:
- Set existing value to `NULL` (the reason it's not deleted is that it processes the reviewer action separately to the assigner action, so one of them might not be getting changed)
- Deduce reviewer_action, update record
- Deduce assigner_action, update record
- Once complete, if any records end up with `NULL` values in both fields, delete them (to keep the list from growing forever and filling up with obsolete records)

database/buildSchema/43_views_functions_triggers.sql
```diff
@@ -1542,164 +1542,381 @@ $$
...
+        INSERT INTO public.application_reviewer_action
+            (user_id, application_id, reviewer_action)  
+        VALUES(
+            rev_id,
+            NEW.application_id,
+            (SELECT reviewer_action FROM reviewer_action)
+         )
+         ON CONFLICT (user_id, application_id)
+         DO UPDATE
+            SET reviewer_action = (SELECT reviewer_action FROM reviewer_action);
```
CarlosNZ, line 1710: Basically an "UPSERT"

database/buildSchema/43_views_functions_triggers.sql
```diff
@@ -1542,164 +1542,381 @@ $$
...
+    AFTER INSERT OR UPDATE ON public.review_assignment
+    FOR EACH ROW
+    EXECUTE FUNCTION public.update_application_reviewer_stats ();
+
+-- Function to update insert/update reviewer/assigner actions on
+-- application_reviewer_action when review_status is updated, by making calls to
+-- the above `single_application_detail` function for the appropriate
+-- user/application combo(s)
+DROP FUNCTION IF EXISTS public.update_reviewer_stats_from_status CASCADE;
+CREATE OR REPLACE FUNCTION public.update_reviewer_stats_from_status ()
```
CarlosNZ, line 1778: There's a bit more work for this one, as the relevant ids (application, reviewer, assigner) are not stored in the review_status_history table, so we fetch them first. After that, it's basically the same as the earlier previous function, except:
- no need to update the reviewer/assigner lists, they won't change here
- only need to update the single reviewer, so don't loop over other connected reviewers like this previous one

There's a bit of code duplication here, but I thought abstracting into new functions would make it more complicated than it needs to be -- unnecessary DRY ;)

package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "conforma-server",
-  "version": "0.9.0-1",
+  "version": "0.10.0-0",
```
CarlosNZ, line 3: Will publish as `v1.0.0` once this work is all done and merged.

CarlosNZ:
Accidentally closed cos I deleted the base branch. Have re-made on https://github.com/msupply-foundation/conforma-server/pull/1106

PR #1106, opened 2024-05-28
#1103 Memoise list actions

### LINKED ISSUE #1103: Memoise application list actions to speed up performance

CarlosNZ:
Fix #1103 

I'll add some explanatory comments in the PR (in addition to the in-code comments)

To test, play round with:
- assigning (to self and others)
- starting review
- submitting review
- re-submit application after change requests (as applicant)

After each of these events, note the changes to the `application_reviewer_action` table, and the `reviewers`/`assigners` lists on the `application` table.


database/buildSchema/43_views_functions_triggers.sql
```diff
@@ -1542,12 +1542,332 @@ $$
 LANGUAGE sql
 STABLE;
 
+-- Does the "heavy lifting" to deduce the assigners/reviewers list for a single
+-- application as well as the available actions for a particular
+-- reviewer/assigner for that application. Used to populate the
+-- `application_reviewer_action` table and the reviewer/assigner lists on each
+-- application whenever an event occurs that would change these values
+CREATE OR REPLACE FUNCTION single_application_detail (applicationId int, userId int DEFAULT 0)
```
CarlosNZ, line 1550: This function is a pure copy/paste from the existing list function. Differences are
- only for a single application (passed in `applicationId` parameter, and additional `WHERE` clause).
- removed all columns other than these 4 complicated ones -- the others are all just joined in the list query as before.

CarlosNZ: Still has a `DEFAULT` value since when generating the assigners and reviewers lists, we don't need to specify a user.

database/buildSchema/43_views_functions_triggers.sql
```diff
@@ -1542,12 +1542,332 @@ $$
...
+            WHERE id = NEW.application_id;
+    END IF;
+
+    -- Get list of reviewer_ids to update for 
+    IF TG_OP = 'UPDATE' THEN
+        reviewer_ids = ARRAY(
+            SELECT reviewer_id FROM review_assignment
+            WHERE application_id = NEW.application_id
+        );
+    ELSE
```
CarlosNZ, line 1685: When `review_assignments` are created, it'll run once for each assignment as rows are inserted. But when *updating*, we need to run for all potentially affected reviewers, hence we loop over the list of connected users.

database/buildSchema/43_views_functions_triggers.sql
```diff
@@ -1542,12 +1542,332 @@ $$
...
+        reviewer_ids = ARRAY[NEW.reviewer_id];
+    END IF;
+
+    FOREACH rev_id IN ARRAY reviewer_ids LOOP
+        -- Remove existing
```
CarlosNZ, line 1690: I tried a few different approaches for handling the existing records, ended up with:
- Set existing value to `NULL` (the reason it's not deleted is that it processes the reviewer action separately to the assigner action, so one of them might not be getting changed)
- Deduce reviewer_action, update record
- Deduce assigner_action, update record
- Once complete, if any records end up with `NULL` values in both fields, delete them (to keep the list from growing forever and filling up with obsolete records)

database/buildSchema/43_views_functions_triggers.sql
```diff
@@ -1542,12 +1542,332 @@ $$
...
+        INSERT INTO public.application_reviewer_action
+            (user_id, application_id, reviewer_action)  
+        VALUES(
+            rev_id,
+            NEW.application_id,
+            (SELECT reviewer_action FROM reviewer_action)
+         )
+         ON CONFLICT (user_id, application_id)
+         DO UPDATE
+            SET reviewer_action = (SELECT reviewer_action FROM reviewer_action);
```
CarlosNZ, line 1710: Basically an "UPSERT"

database/buildSchema/43_views_functions_triggers.sql
```diff
@@ -1542,12 +1542,332 @@ $$
...
+    AFTER INSERT OR UPDATE ON public.review_assignment
+    FOR EACH ROW
+    EXECUTE FUNCTION public.update_application_reviewer_stats ();
+
+-- Function to update insert/update reviewer/assigner actions on
+-- application_reviewer_action when review_status is updated, by making calls to
+-- the above `single_application_detail` function for the appropriate
+-- user/application combo(s)
+DROP FUNCTION IF EXISTS public.update_reviewer_stats_from_status CASCADE;
+CREATE OR REPLACE FUNCTION public.update_reviewer_stats_from_status ()
```
CarlosNZ, line 1778: There's a bit more work for this one, as the relevant ids (application, reviewer, assigner) are not stored in the review_status_history table, so we fetch them first. After that, it's basically the same as the earlier previous function, except:
- no need to update the reviewer/assigner lists, they won't change here
- only need to update the single reviewer, so don't loop over other connected reviewers like this previous one

There's a bit of code duplication here, but I thought abstracting into new functions would make it more complicated than it needs to be -- unnecessary DRY ;)

package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "conforma-server",
-  "version": "0.9.0-1",
+  "version": "0.10.0-0",
```
CarlosNZ, line 3: Will publish as `v1.0.0` once this work is all done and merged.

andreievg (changes requested):
Thanks Carl, looks great. I think assigner_ids should come from assigner join.

database/migration/databaseMethods.ts
```diff
@@ -445,6 +445,77 @@ const databaseMethods = {
       `,
     })
   },
+  createApplicationReviewerActionRecords: async () => {
```
andreievg, line 448: Do you know aprox how long this takes to run ?

database/buildSchema/43_views_functions_triggers.sql
```diff
@@ -1576,137 +1896,57 @@ CREATE TABLE IF NOT EXISTS application_list_shape (
...
-    12,
-    13,
-    stage_status.stage_id
-
-$$
-LANGUAGE sql
-STABLE;
+                AND ts.is_active = TRUE
+                AND ts.event_code = 'applicantDeadline'
+        LEFT JOIN (
```
andreievg, line 1936: in the other left joins we filter out in 'on' statement, not sure about the performance implications but it seems more directr to

```sql
LEFT JOIN application_reviewer_action as ara ON ara.user_id = user_id AND ara.id = app.application_id 
```

database/migration/migrateData.ts
```diff
@@ -1088,6 +1088,36 @@ const migrateData = async () => {
...
+      ALTER TABLE public.application   
+        ADD COLUMN IF NOT EXISTS reviewer_list VARCHAR[],
+        ADD COLUMN IF NOT EXISTS assigner_list VARCHAR[];
+    `)
+
+    console.log(' - Adding application_reviewer_action table')
+    await DB.changeSchema(`
+      CREATE TABLE IF NOT EXISTS public.application_reviewer_action (
+        id serial PRIMARY KEY,
+        user_id integer REFERENCES public.user(id)
```
andreievg, line 1106: I think we need to index user_id and application_id (indexing actions might also help)

database/buildSchema/43_views_functions_triggers.sql
andreievg: > There's a bit of code duplication here, but I thought abstracting into new functions would make it more complicated than it needs to be -- unnecessary DRY ;)

Sounds good, thanks

database/buildSchema/43_views_functions_triggers.sql
```diff
@@ -1549,12 +1549,332 @@ $$
...
+    DELETE FROM public.application_reviewer_action
+    WHERE application_id = NEW.application_id
+    AND reviewer_action IS NULL
+    AND assigner_action IS NULL;
+
+RETURN NULL;
+
+END;
+$$
+LANGUAGE plpgsql;
```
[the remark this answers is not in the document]
andreievg: Btw really like the insert to and on conflict, I think it avoids re-building indexes if the fields were otherwise dropped

database/buildSchema/43_views_functions_triggers.sql
CarlosNZ: Not quite sure what you mean. It's only doing one thing I think -- gets the values for the reviewer/assigner lists. The list actions come from the get list function, which return the pre-computed values that this function calculated earlier.

database/buildSchema/43_views_functions_triggers.sql
```diff
@@ -1549,12 +1549,332 @@ $$
...
+         ON CONFLICT (user_id, application_id)
+         DO UPDATE
+            SET reviewer_action = (SELECT reviewer_action FROM reviewer_action);
+    END LOOP;
+    
+    
+    -- ASSIGNERS
+    IF TG_OP = 'UPDATE' THEN
+        assigner_ids = ARRAY(
+            SELECT assigner_id FROM review_assignment
```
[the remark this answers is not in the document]
CarlosNZ: That would be *possible* assigners, whereas this finds the *actual* assigner who's already done the assigning. Looking at the previous version, it finds:
```
LEFT JOIN public."user" AS assigner_user ON assigned_assignment.assigner_id = assigner_user.id
    LEFT JOIN public."user" AS reviewer_user ON assigned_assignment.reviewer_id = reviewer_user.id
```
which suggests we're getting the users who did they assigning/reviewing, not who *could* have done it.

database/buildSchema/43_views_functions_triggers.sql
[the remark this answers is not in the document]
CarlosNZ: > Another approach is to do calculation for all assigners and users for this application, I did a quick bench it's about 2 seconds on demo4, to calculate actions for all reviewers/assigners for one application, this would be added to all upsert/insert operations on review assignments, hmm.

The problem I had is that the review_assignment records get inserted one by one and so trigger an event for each. So we don't want to update them *all* for *every* insert, hence inserts only update actions for that one assignment record. Whereas an update is done individually, so we can update *all* records associated with that application.

I think with your suggestion, we'd run the whole thing for every assignment being added, which would be very inefficient, not to mention pointless as it would generate an incomplete list for every insert until the last one is added.

database/buildSchema/43_views_functions_triggers.sql
CarlosNZ: I think I did this intentionally, as the "application_reviewer_action" table can get huge, and we'll only ever be joining to a very small selection of its lines at any time. However, I originally had it just joining on `ara.user_id = user_id` and putting a WHERE at the end to filter it by application, which seemed really inefficient to me. 

But I didn't think of doing a two-condition Join like you suggested, so I guess that would probably be basically the same. Have updated: d5cd051f

database/migration/migrateData.ts
CarlosNZ: Yeah good call: 9bd2a3c1

We never query the table by the actions, so probably wouldn't help to index those columns at this point.

CarlosNZ:
Thanks for the review @andreievg -- have replied to or fixed your suggestions. Couple of them I wasn't totally clear what you're referring to, so come chat IRL if you want to clarify further.

database/buildSchema/43_views_functions_triggers.sql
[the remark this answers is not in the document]
andreievg: The assigner actions looks at possible assign actions for an application/user (even before assigner has been assigned to the review_assignment):

https://github.com/msupply-foundation/conforma-server/blob/57d5111f378733e9b4bee80d6b29036588387e44/database/buildSchema/43_views_functions_triggers.sql#L1615-L1616

https://github.com/msupply-foundation/conforma-server/blob/57d5111f378733e9b4bee80d6b29036588387e44/database/buildSchema/43_views_functions_triggers.sql#L1644

https://github.com/msupply-foundation/conforma-server/blob/57d5111f378733e9b4bee80d6b29036588387e44/database/buildSchema/43_views_functions_triggers.sql#L1462-L1467
https://github.com/msupply-foundation/conforma-server/blob/57d5111f378733e9b4bee80d6b29036588387e44/database/buildSchema/43_views_functions_triggers.sql#L1473-L1475

database/buildSchema/43_views_functions_triggers.sql
andreievg: 🙏

database/buildSchema/43_views_functions_triggers.sql
andreievg: It seems to be called separately with userId = 0 for assigners and reviewers, i think below can be separated

```sql
            CASE WHEN count(assigner_user.id) = 0 THEN
                NULL
            ELSE
                ARRAY_AGG(DISTINCT (CONCAT(assigner_user.first_name, ' ', assigner_user.last_name)::varchar))
            END AS assigners,
            CASE WHEN count(reviewer_user.id) = 0 THEN
                NULL
            ELSE
                ARRAY_AGG(DISTINCT (CONCAT(reviewer_user.first_name, ' ', reviewer_user.last_name)::varchar))
            END AS reviewers,
```

And the everything below

```sql
 -- REVIEWER ACTIONS
            CASE WHEN COUNT(*) FILTER (WHERE action_review_status_history.status = 'CHANGES_REQUESTED') != 0 THEN

...
```

database/buildSchema/43_views_functions_triggers.sql
```diff
@@ -1683,78 +1681,48 @@ BEGIN
...
-         ON CONFLICT (user_id, application_id)
-         DO UPDATE
-            SET reviewer_action = (SELECT reviewer_action FROM reviewer_action);
-    END LOOP;
-    
-    
-    -- ASSIGNERS
+    -- REVIEWER/ASSIGNER actions
+    -- When updating review_assignments, we re-calculate for every
+    -- reviewer/assigner associated with this application.
```
CarlosNZ, line 1686: @andreievg -- have updated so that it gets a single list of all reviewers/assigners (from the assigner_join table) and updates them in one pass.

database/buildSchema/43_views_functions_triggers.sql
```diff
@@ -1683,78 +1681,48 @@ BEGIN
...
+            ON ra.id = raaj.review_assignment_id
+            WHERE application_id = NEW.application_id);
     ELSE
-        CASE WHEN NEW.assigner_id IS NULL THEN
-                assigner_ids = ARRAY[];
-            ELSE
-                assigner_ids = ARRAY[NEW.assigner_id];
-            END CASE;
+    -- But when inserting, just update for the new reviewer
+        user_ids = ARRAY[NEW.reviewer_id];
```
CarlosNZ, line 1698: This bit doesn't quite work, I think: We also need to get the assigners linked with the new review_id, but because they're created with an Action, we have no guarantee that they'll be created by the time this runs.

Any suggestions?

CarlosNZ (commented):
See my additional comments, @andreievg

database/buildSchema/43_views_functions_triggers.sql
CarlosNZ: Okay, @andreievg , I think I've got it going properly now. Added an additional (very similar) function that runs on review_assignment_assigner_join INSERTS (only). 

So the triggers are:
- on review_assignment INSERT: run original function (in principle gets assigners from assigner_join too, but in practice those records probably aren't created yet, so will only be for reviewer action). Runs once for *each* insert.
- on review assignment UPDATE: run same function, this time will also update for assigner_joins -- single update re-calculates for *all* related reviewers/assigners
- on review_assignment_assigner_join INSERT: run slightly different function, but does the same as the above but just for the assigner. No need to run on UPDATE as the review_assignment update takes care of it (plus these don't really get updated as a matter of course)

It's somewhat inefficient -- ideally we'd wait for all relevant records to be inserted and then run a function once, but this is not really possible without moving this functionality out into the core Actions, but this would be quite a bit more complicated at this point. We'll see how it goes, speed wise, on the demo site.

database/buildSchema/43_views_functions_triggers.sql
CarlosNZ: Yeah, I was able to just copy and paste all this logic from the existing function, so I'm reluctant to start picking it apart, even a little bit at this point -- potentially very messy and time-consuming.

database/buildSchema/43_views_functions_triggers.sql
```diff
@@ -1549,12 +1549,382 @@ $$
...
+    END IF;
+
+    -- REVIEWER/ASSIGNER actions
+    -- When updating review_assignments, we re-calculate for every
+    -- reviewer/assigner associated with this application.
+    IF TG_OP = 'UPDATE' THEN
+        user_ids = ARRAY(
+            -- Creates an array of distinct user ids for all assigners/reviewers
+            -- who can interact with this application
+            SELECT DISTINCT UNNEST(ARRAY[reviewer_id, raaj.assigner_id])
```
andreievg, line 1691: that's cool, better then my union example !

database/buildSchema/43_views_functions_triggers.sql
andreievg: Thanks Carl, this looks pretty solid, I think it's ok to approve and get it deployed on demo server for performance testing.

I do think that there will be quite a lot of excessive updates, and potential slow down in all operations that have the added trigger, i did a quick check, to see how many assigners are linked to each review_assignment:

```sql
select number_of_assgners, count(*) from 
(select review_assignment.id, count(raj.assigner_id) number_of_assgners from review_assignment join review_assignment_assigner_join raj on raj.review_assignment_id = review_assignment.id group by 1) 
group by 1
```

 number_of_assgners | count
--------------------+-------
                 11 |   976
                  9 |   350
                 15 |   746
                  5 |    12
                  4 |     9
                 10 |    68
                  6 |  6273
                 14 |  6929
                 13 |  8571
                 16 |     9
                  7 |  2319
                 12 | 17188
                  8 |    66

This suggest that majoriy have at least 14 assigners.

I think a queue would be great, doesn't need to be core action queue, can be separate. But yeah lets wait to see the results, thanks for the massive effort

andreievg (approved):

CarlosNZ:
Cool, @andreievg , thanks for a thorough look through

> This suggest that majoriy have at least 14 assigners.

I actually think that's probably too high in practice. They've been a bit loose with handing out permissions in Fiji, we can probably get that down.

PR #1107, opened 2024-06-06
1104 improve row level policy performance

### LINKED ISSUE #1104: Review Performance
Related: https://github.com/msupply-foundation/conforma-server/issues/1103

I ran demo server with large data set without row level policies, and it was very snappy, for everything, including review (apart from some delay when review was finalised, i think due to some action/triggers, 5 or so seconds, which should be acceptable).

There are a couple of things we can do in short term to improve Review Performance:

The first step is to improve row lever policy speed by re-instating changes in this PR: https://github.com/msupply-foundation/conforma-server/pull/1079, that were later remove here: https://github.com/msupply-foundation/conforma-server/commit/8a0d9e52c89cc39c8cf8c95c1aa73d19b61e4737.  Due to org permissions not being matched correctly. (1 days of work)

Next step is to relocate `getReviewResponses` query to a rest endpoint. This endpoint will use user JWT to do a query to specified reviewAssignment (to make sure user has required permissions), and then use admin access to do the actual query, ignoring row level policies. (2 days of work)

Ideally we would do most of the re-mapping that front end is doing, in back end (consumer focused API), but that would require quite a bit more work.

andreievg:
Partially fixes #1104

PR rewires row level policy check to use sub query when checking template ids for a permission, rather then checking against an array that is parsed from a JWT 

**How things were**

JWT would have the following (based on permission joins to user/org, through permissions_all view)

`"pp2_template_ids": "73,153,59,61,77,121,123,130,144,147,154,157,161,162,163,185,186,188,189",`

And the permission policy configuration would have this:

`"template_id": "jwtPermission_array_bigint_template_ids"` -> which is turned into SQL for row level policies: `template_id = any (string_to_array(COALESCE(current_setting('jwt.claims.pp2_template_ids', true), '0'), ',')::integer[])`

**How things now**

A new table `user_org_policy_template`, which is populated at startup (for all user/org combo) and during login (for user/org), this table contains templateIds for all permissions for user and org in this format `{userId}.{orgId || 0}.{permissionPlicyId}, templateId`. if we take the above example, for user 54 and orgId 1 we would have entries in `user_org_policy_template` table as such:

```
54.1.2, 73
54.1.2, 153
54.1.2, 59
54.1.2, 61 
..
```

And the permission policy configuration now has this instead:
```
   "template_id": {
                    "$": {
                        "__": "= ANY (SELECT template_id FROM user_org_policy_template WHERE user_org_policy = jwtUserDetails_text_userId || '.' || jwtUserDetails_text_orgId || '.7')"
                    }
                }
```
`$` and `__` is for inline sql

Which translates to the following in SQL for row level policies:

```
template_id  = ANY ( 
SELECT template_id FROM user_org_policy_template
    WHERE user_org_policy = (
        COALESCE(current_setting('jwt.claims.userId'::text, true), '') 
        || '.'  
        || COALESCE(current_setting('jwt.claims.orgId'::text, true), '') || '.5')
     )
```

**Why user_org_policy_template table ?**

In previous implementation of this performance improvements, a view was used (similar to permissions_all), but the ANY/IN statement didn't actually match existing selection of templateIds (see [user permission](https://github.com/msupply-foundation/conforma-server/blob/09592461d0ef1ec8fe4d9370ef7f8133e0911de8/src/components/postgresConnect.ts#L974) and [org permission](https://github.com/msupply-foundation/conforma-server/blob/09592461d0ef1ec8fe4d9370ef7f8133e0911de8/src/components/postgresConnect.ts#L1000) queries), one of the checks included `isUserCategory` which is extracted from a template in permission_all view, i thought this would be slow and decided to create a table that stores pre-calculated template ids for user/org/permission policy

**Extra considerations**

What happens currently, when a new permission is added to user while they are logged in ? We would need to re-log in for this in this version. 

Does user get new token every time they refresh a page ?


andreievg:
Btw all of this leads to 3x improvements for speed on demo server we were using for getting review assignments. And around 20% for application list.

database/migration/migrateData.ts
```diff
@@ -1088,6 +1079,35 @@ const migrateData = async () => {
...
+  // v0.9.0
+  if (databaseVersionLessThan('0.9.0')) {
+    console.log('Migrating to v0.9.0...')
+
+    await DB.changeSchema(`
+      CREATE TABLE IF NOT EXISTS user_org_policy_template (
+        user_org_policy TEXT,
+        template_id INTEGER 
+      );
+    `)
```
CarlosNZ, line 1091: Let's make this v1.0.0

I've already release an 0.9.0 build. And I've targeted v1.0.0 in [my PR](#1106)

(Plus we're due a v1 release once this is done 😆)

CarlosNZ: Also, can you add a console.log that states the change, as per the other migrations.

Something like:
```
console.log(
      ' - Create user_org_policy_template table for easier filtering for row-level permissions'
    )
```

database/migration/migrateData.ts
```diff
@@ -1088,6 +1079,35 @@ const migrateData = async () => {
...
+
+    // Update row level policy configs replacing : "jwtPermission_array_bigint_template_ids" with
+    // = ANY (SELECT template_id FROM user_org_policy_template WHERE user_org_policy = jwtUserDetails_text_userId || '.' || jwtUserDetails_text_orgId || '.${permissionPolicyId}'))
+    // see sqlFromJson.test.ts for example of final query
+    // Perviously jwtPermission_array_bigint_template_ids was translated to
```
CarlosNZ, line 1096:
```suggestion
    // Previously jwtPermission_array_bigint_template_ids was translated to
```

src/components/permissions/loginHelpers.ts
```diff
@@ -56,6 +56,11 @@ const getUserInfo = async (userOrgParameters: UserOrgParameters) => {
       return { orgId, orgName, userRole, registration, address, logoUrl, isSystemOrg }
     })
 
+  // The logic for template permissions for the user is as follows
+  // templates allowed for user and org based on permission policy joins (flattened via permissions_all view)
+  // where logged in user and organisation matched permissions
+  // where logged in user matches permission and template is user category
+  // where logged in organisation matches permissions and permission does not specify user (org wide permission)
```
CarlosNZ, line 63:
```suggestion
  // The logic for template permissions for the user is as follows:
  // - templates allowed for user and org based on permission policy joins
  //   (flattened via permissions_all view)
  // - where logged in user and organisation matched permissions
  // - where logged in user matches permission and template is user category
  // - where logged in organisation matches permissions and permission does not
  //   specify user (org wide permission)
```

src/components/permissions/routes.ts
```diff
@@ -40,6 +40,7 @@ const routeLogin = async (request: any, reply: any) => {
     if (!(await bcrypt.compare(password, passwordHash as string)))
       return reply.send({ success: false })
 
+    await databaseConnect.updateUserOrgPolicyTemplate(userId, null)
```
CarlosNZ, line 43: I have slight concern that we've captured all the events when this needs to be updated.

What if we create a new template version? When would a user's access to that be updated?

src/components/postgresConnect.ts
```diff
@@ -1012,6 +1012,54 @@ class PostgresDB {
     }
   }
 
+  // Permissions policies rely on a entries for user and org in user_org_policy_template table
```
CarlosNZ, line 1015:
```suggestion
  // Permissions policies rely on entries for user and org in user_org_policy_template table
```

src/components/postgresConnect.ts
```diff
@@ -1012,6 +1012,54 @@ class PostgresDB {
...
+  // in the format {userId}.{orgId || 0}.{permissionPolicyId}, templateId
```
CarlosNZ, line 1016: Just wondering why concatenate userId.orgId.policyId as opposed to just having a separate column for each?

CarlosNZ (changes requested):
Looks good, thanks for this.

I'll probably get you to talk me through a couple of things next week when you're in. 

Few comments, mostly cosmetic. My only main concern is knowing that we've covered all the events where the user_org_policies need to be updated.

andreievg:
closing in favour of https://github.com/msupply-foundation/conforma-server/pull/1110

@CarlosNZ sorry about having to review this PR and me closing it, the other one should be easier to review i think

PR #1110, opened 2024-06-12
#1104 Row level policy improvements and docs

### LINKED ISSUE #1104: Review Performance

andreievg:
Fixes #1104 

Ended up making methods with SECURITY DEFINER and using that in row level policies (see the test methods for examples of row policy output). With this method most queries were almost as fast as without row level policies (review responses, from 9 seconds to 700ms, get templates from 3 seconds to 1, and twice as fast for most application list queries)

It turns out in statement with array of text values wasn't that slow, I benched it (see example of pgbench in the .md file).

It looks like the real problem was with nested permission checks, here is an examples:

```SQL
create table one (
id text primary key,
info text
);

create table two (
id text primary key,
info text,
one_id text references one(id)
);

insert into one VALUES ('1', '1');
insert into one VALUES ('2', '2');
insert into one VALUES ('3', '3');

insert into two VALUES ('1', '1', '1');
insert into two VALUES ('2', '2', '2');
insert into two VALUES ('3', '3', '3');

ALTER TABLE one ENABLE ROW LEVEL SECURITY;
ALTER TABLE two ENABLE ROW LEVEL SECURITY;

CREATE POLICY "one_check" ON "one" FOR SELECT USING (id = '1');
CREATE POLICY "two_check" ON "two" FOR SELECT USING (one_id IN (select id from one where ID in ('2', '3')));
```

in another pgAdmin query window do

```SQL
set role graphile_user;
select * from two;
```
Nothing will show, since in the two_check policy `select id from one..` will apply the `one_check` policy on `one` (and we have quite a few policies that will compound when we checking say review_assignment table) 

Back to first pgAdmin query window
```SQL
CREATE OR REPLACE FUNCTION public.two_check ()
    RETURNS setof two
    SECURITY DEFINER
    AS $$
    	select * from two where two.one_id in ('2', '3')
$$
LANGUAGE sql
STABLE;
DROP POLICY "two_check" ON two;
CREATE POLICY "two_check" ON "two" FOR SELECT USING (id IN (select id from two_policy ()));
```

in another pgAdmin query window do

```SQL
set role graphile_user;
select * from two;
```

Now you should see 2 and 3.

Btw our policy configuration (`select json_agg(rules) from permission_policy;`) are meant to be independent of each other, i.e. for reviewer

```JSON
"application": {
            "view": {
                "id": {
                    "$in": {
                        "$select": {
                            "$from": "review_assignment",
                            "$where": {
                                "reviewer_id": "jwtUserDetails_bigint_userId"
                            },
                            "application_id": true
                        }
                    }
                },
                "template_id": "jwtPermission_array_bigint_template_ids"
            }
        }
```


database/buildSchema/43_views_functions_triggers.sql
```diff
@@ -1,3 +1,5 @@
+-- SCHEMA to hide internal functionality from postgraphile
+CREATE SCHEMA IF NOT EXISTS private;
```
andreievg, line 2: Should this be in migrations ?

src/components/permissions/rowLevelPolicyHelpers.ts
```diff
@@ -167,18 +176,33 @@ const compileRowLevelPolicy = (
...
+    `(COALESCE(current_setting('jwt.claims.${permissionAbbreviation}', true),'') != '' AND ${condition})`
 
-  // If using clause is not used return nothing
-  // otherwise return USING from usingCondition (view rule, see compileRowLevelPolicies) or condition (current rule condition)
+  // If USING clause is not used return nothing
+  // otherwise return USING from usingCondition (which comes from the defined 'view' rule))
   const getUsingClause = () => {
     if (!ruleSettings.using) return ''
-    const conditionToUse = ruleSettings.withCheck ? usingCondition : condition
-    return `USING ${addBracketsAndPermissionCheck(conditionToUse)}`
```
andreievg, line 181: Logic here is slightly changed, but usingCondition is always available (if view permissions are configured in rules, see https://github.com/msupply-foundation/conforma-server/blob/028193d5d11e2ae9b30484c108e9d11ac642f64d/src/components/permissions/rowLevelPolicyHelpers.ts#L133-L137

CarlosNZ:
Hey @andreievg,

I'm getting `new row violates row-level security policy for table "review"'` error when trying to start a review (as user Ben). Have loaded a fresh "Fiji" snapshot while on your branch. 

Did you have it working properly?

database/buildSchema/43_views_functions_triggers.sql
```diff
@@ -1,3 +1,5 @@
+-- SCHEMA to hide internal functionality from postgraphile
```
CarlosNZ, line 1:
```suggestion
-- Private SCHEMA to put the policy_check functions in order to hide them from Postgraphile
```

CarlosNZ (changes requested):
Looks really good, but I haven't been able to successfully test it locally yet, as per [this comment](https://github.com/msupply-foundation/conforma-server/pull/1110#issuecomment-2164132709)

How extensively have you tested this in the app? As in, logged in as different users and checked that they can see/edit what they're supposed to?

database/buildSchema/43_views_functions_triggers.sql
CarlosNZ: Actually, maybe not. Perhaps even DROP the whole schema first, so we know the policy functions are getting created from scratch each time.

documentation/Performance-Tweaks-Guide.md
```diff
@@ -0,0 +1,138 @@
...
+select * from application where template_id = any
+(string_to_array(COALESCE(current_setting('jwt.check', true), '0'), ',')::integer[])
+COMMIT;
+```
+
+Then `ctrl + x, y`
+
+```bash
+pgbench -f bench.sql --log --transactions=30 tmf_app_manager
+```
```
CarlosNZ, line 114: If it's not too much trouble, can you also paste the `pg_bench` output in here?

src/components/permissions/rowLevelPolicyHelpers.ts
```diff
@@ -167,18 +176,33 @@ const compileRowLevelPolicy = (
   usingCondition: string,
   condition: string
 ) => {
+  // Create function for permission so that row level policies are checked independently
+  // SECURITY DEFINER is used to make sure permission check does not take into account existing permissions
+  // i.e. without this function, if we have something like this "restrict review by review_assignment where user is reviewer"
+  // and then we have another restriction of "review_assignment where user is assigner" in the former policy check the latter check will also run
+  const functionName = `private.policy_check_${permissionAbbreviation}_${tableName}`
```
CarlosNZ, line 183: Bit of a naming inconsistency here. `permissionAbbreviation` is called both `permissionPolicyAbbreviation` and `policyAbbreviation` above. I think having the word `policy` in there in important. `policyAbbreviation` works for me, but go with the longer `permissionPolicyAbbreviation` if you want to be 100% clear at the expense of brevity.

src/components/permissions/rowLevelPolicyHelpers.ts
```diff
@@ -158,7 +158,16 @@ export const compileRowLevelPolicies = (
 /* Compiles single row level permission
   in "application", "pp2pn2tp2", "view", "true", 
       "user_id = jwt_get_bigint('userId') AND template_id = jwt_get_bigint('pp2pn2tp2_templateId')" 
-  out `CREATE POLICY "view_pp2pn2tp2" ON "application" FOR SELECT USING (jwt_get_boolean('pp2pn2tp2') = true and user_id = jwt_get_bigint('userId') AND template_id = jwt_get_bigint('pp2pn2tp2_templateId'))`
+  out `
```
CarlosNZ, line 161: A comment about the "Drop existing policies" above. The query to get the existing ones is not getting the new ones as it's only matching view/create etc:
```
public getAllGeneratedRowPolicies = async () => {
    const text = `
      SELECT policyname, tablename 
      FROM pg_policies 
      WHERE policyname LIKE 'view\_%'
      OR policyname LIKE 'update\_%'
      OR policyname LIKE 'create\_%'
      OR policyname LIKE 'delete\_%'
    `
```

Did you want to include the new function-based ones too?

CarlosNZ:
> Looks really good, but I haven't been able to successfully test it locally yet, as per [this comment](https://github.com/msupply-foundation/conforma-server/pull/1110#issuecomment-2164132709)

Hmm, seems to be working okay once I make a branch combining both of our work.

I'll make a build of that today and chuck it up and we'll see how it goes :)

src/components/permissions/rowLevelPolicyHelpers.ts
```diff
@@ -119,11 +119,55 @@ const generateRowLevelPolicies = (permissionRows: permissionPolicyColumns[]) =>
...
+  matchFn: (key: string, value: object) => boolean,
+  modifyFn: (value: object) => string
+): object => {
+  if (typeof obj != 'object') {
+    return obj
+  }
+  if (Array.isArray(obj)) {
+    console.warn('Arrays are not supported')
+    return obj
+  }
```
CarlosNZ, line 133: We have an `isObject` function in `utilityFunctions.ts` that can deal with this:
```ts
// Returns true if input is a "proper" object (i.e. not an array or null)
export const isObject = (element: unknown): element is Object =>
  typeof element === 'object' && !Array.isArray(element) && element !== null
```

In fact, can you put this whole function in `utilityFunctions` and add a comment explaining what it does -- seems generic enough to keep in a common resource.

src/components/permissions/rowLevelPolicyHelpers.ts
```diff
@@ -119,11 +119,55 @@ const generateRowLevelPolicies = (permissionRows: permissionPolicyColumns[]) =>
...
+  )
+}
+
+const updateRulesUseViewsInsteadOfTables = (rules: object) => {
+  const tablesToTurnIntoViews: { [key: string]: boolean } = {}
+
+  return {
+    rules: modifyValueInObject(
+      rules,
+      (key, value) => key == '$from' && typeof value == 'string',
```
CarlosNZ, line 150: As discussed, can you add  `&& !(value.startsWith('private.'))` qualifier too?

documentation/Performance-Tweaks-Guide.md
```diff
@@ -0,0 +1,138 @@
...
+
+```bash
+docker container ls
+```
+
+This would list container and ids and the port mappings, you can find which port is mapped to which demo instance, check out NGINX config
+
+```bash
+cat /etc/nginx/sites-enabled/default
+```
```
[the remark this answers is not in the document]
andreievg: 353c8c39da4ea74e8d1e3ee73b149b65e354c09c, 🙏 also added description that container name can be found via `docker container ls`

documentation/Performance-Tweaks-Guide.md
```diff
@@ -0,0 +1,138 @@
+## Performance Tweaks Guid
+
+During the latest round of performance improvements we had to do quite a lot of performance related test, which were best done on the actual demo server.
```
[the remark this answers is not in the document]
andreievg: ba905443a3baa6daf9f82c065e641c17a6ab387a

documentation/Performance-Tweaks-Guide.md
```diff
@@ -0,0 +1,138 @@
...
+
+After updating the file locally, and doing `yarn build`
+
+Copy newly built files from local computer
+
+```bash
+scp -i {key} {repo_root}/conforma-server/build/src/components/permissions/rowLevelPolicyHelpers.js ubuntu@d{demo server}:/home/ubuntu
+```
+
+Then from demo server
```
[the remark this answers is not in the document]
andreievg: Not quite, just means from the server running docker container, not from actual docker container 3c0ca63dd8fe8dadb023020f362a97c90baf5321

documentation/Performance-Tweaks-Guide.md
andreievg: Good idea. I ran above on demo4 (actually had some errors in the sql, fixed), and added example: b9c35dadb964cb0fde2c5eb4fdd91481cf244163

src/components/permissions/rowLevelPolicyHelpers.ts
andreievg: 🙏 5c653792f190c5a0b3ef076be653d9606b97c003, so much more confident about refactor with isObject when there are tests for modifyValueInObject

src/components/permissions/rowLevelPolicyHelpers.ts
andreievg: 2b09c25e54d6929bca01b80f2cfb6d41a0a3bca1

database/buildSchema/43_views_functions_triggers.sql
andreievg: this would mean row level policies will need to be rebuilt on startup or after every migration, this happens on startup now and snapshot load, would that be all of the use cases ?

database/buildSchema/43_views_functions_triggers.sql
andreievg: https://github.com/msupply-foundation/conforma-server/pull/1110/commits/57765679c5e3bdd2b3fcadfdd4e75e16d5f02257

database/buildSchema/43_views_functions_triggers.sql
andreievg: https://github.com/msupply-foundation/conforma-server/pull/1110/commits/57765679c5e3bdd2b3fcadfdd4e75e16d5f02257

database/buildSchema/43_views_functions_triggers.sql
```diff
@@ -1,3 +1,9 @@
+-- SCHEMA to hide internal functionality from postgraphile graphql schema
+-- for example views that expose tables used in row level policies to avoid compounding
+-- permission queries will be in private schema and not accessible via graphql
```
CarlosNZ, line 3:
```suggestion
-- SCHEMA to hide internal functionality from postgraphile graphql schema
-- for example, views that expose tables used in row level policies (to avoid
-- compounding permission queries) will be in private schema and not accessible
-- via graphql
```

documentation/Performance-Tweaks-Guide.md
```diff
@@ -0,0 +1,156 @@
...
+
+## Getting inside the docker container
+
+After ssh to the demo server.
+
+```bash
+docker container ls
+```
+
+If container name is not indicative of the url it is exposed in, then you can find more information via, `cat /etc/nginx/sites-enabled/default` for NGINX or `~/demo_server/env_files` for Caddy
```
CarlosNZ, line 14: The env files were there for nginx too, so its slightly confusing to mention them separately. I would just say that the port that each container is mapped to is defined in the `env_files` and not mention the server apps here.

src/components/utilityFunctions.ts
```diff
@@ -156,3 +156,21 @@ export const errorMessage = (err: unknown, property?: string) => {
 
   return 'Unknown error'
 }
+
```
CarlosNZ, line 159:
```suggestion

// Recursively searches an object for criteria defined in `matchFn`, and
// modifies matching values according to `modifyFn`
export const modifyValueInObject = (
```

CarlosNZ (approved):
Cool, thanks for all this. Approving now as we should probably get it tested thoroughly in demo. Made a couple of very small formatting/comment suggestions (you can probably just merge the suggestions in most cases). Merge when ready :)

documentation/Performance-Tweaks-Guide.md
andreievg: Cool, i just removed the mention of nginx or caddy entirely since the mapping should be in the name: [Remove mention of caddy or nginx](https://github.com/msupply-foundation/conforma-server/pull/1110/commits/e893655e4a16da091d35b38533d8a33863db9d8c)

andreievg:
Thanks for Review and suggestions @CarlosNZ !

PR #1122, opened 2024-08-29
#1120 Improvements for Reviewer stats updates

### LINKED ISSUE #1120: Major bug: Assigning/Unassigning taking unreasonably long time
What we know:
- it's connected to the "update_application_reviewer_stats" function, which is triggered on review assignment updates (if I disable this trigger, it happens in the normal, timely manner)

What we don't know
- why only on the Live server (it's fast on Demo3 with the exact same snapshot, and that's a slower system)?
- why so slow? Have checked locally, and the number of database operations caused by the trigger shouldn't take this long, even on a slower system.

### Possible solution

In order to prevent this function blocking the update, we could re-implement it on the Server (rather than as a Postgres function), and trigger it from a notification channel. And then run is `async`, so it doesn't block the main mutation.

Still doesn't really address why it's causing such a block, but we'll be able to add more logging to it this way and have a better idea of what's causing it to run so many times.

CarlosNZ:
Fix #1120 

### Main changes:
- Remove the "Insert" trigger for `review_assignment` and `review_assignment_assigner_join`, as this was causing the update function to run way too many times[^1] whenever review assignments were generated. Instead it just gets run once and is called by the `generateReviewAssignments` Action.
- Move the "updateReviewerStats" function to a server function, triggered by database notification (and other things) rather than a PG function. Makes it much easier to see when it's running
- Made sure "updateReviewerStats" gets called async and send to the Throttle queue, which means it has to wait until the current action sequence is complete, and won't block the UI (as the Action triggers don't wait for it)

### Additional related changes
- Make `updateReviewerStats` callable by command line, so we can run manually on existing applications if desired
- Make Throttle a global object so can be used by more than just the database trigger functions (and update types accordingly)
- Data migration to re-calculate reviewer actions for existing applications (that have reviewer actions currently), as we had a few where they were showing up as "Assignable" even though they were approved.

[^1]: It was running for a single reviewer for each insert, whereas now it runs once for all reviewers. I didn't realise quite how bad it was, due to two things:
    - even though it only runs for a single reviewer, the reviewer and assigner lists for the application table were being re-calculated every time, which was very wasteful
    - the `review_assignment_assigner_join` captures a many-to-many relation, so by calling the function on every insert, the number of potential assigners for each application can be quite large (they get repeated for *each* review assignment, so there was 112 in the case that was making this problem noticeable on the Fiji server). And because a lot of them were self-assigning, it's the same person for a lot of them. So we would be getting well over a hundred calls for what boiled down to about 10 actual users. i.e. if it had only been running once per person, it might have been okay, but once per assigner-assignment join was too much!

### DIFF
--- a/plugins/action_generate_review_assignment_records/src/generateReviewAssignments.test.ts
+++ b/plugins/action_generate_review_assignment_records/src/generateReviewAssignments.test.ts
@@ -4,7 +4,7 @@
 // database has entered correct information
 
 import { map, omit } from 'lodash'
-import DBConnect from '../../../src/components/databaseConnect'
+import DBConnect from '../../../src/components/database/databaseConnect'
 import { ActionQueueStatus, ReviewAssignmentStatus } from '../../../src/generated/graphql'
 import { action as generateReviewAssignments } from './index'
 import { ResultObject } from './types'
--- a/src/components/actions/throttle.ts
+++ b/src/components/actions/throttle.ts
@@ -19,10 +19,14 @@
  */
 const THROTTLE_QUEUE_THRESHOLD = 200 // ms
 
-type EventObject<T, U> = { name: string; data: T; action: (data: T) => Promise<U> | U }
+interface EventObject<T> {
+  name: string
+  data: T
+  action: (data: T) => unknown
+}
 
 export class EventThrottle {
-  queue: EventObject<any, any>[]
+  queue: EventObject<any>[]
   timerId: NodeJS.Timeout | undefined
   queueActive: boolean
 
@@ -50,7 +54,7 @@ export class EventThrottle {
     this.queueActive = false
   }
 
-  public add<T, U>(event: EventObject<T, U>): void {
+  public add<T>(event: EventObject<T>): void {
     // Timer is used solely to determine whether to put new events at the front
     // or end of the queue.
     const currentTimer = this.timerId
--- a/src/components/actions/throttle.ts
+++ b/src/components/actions/throttle.ts
@@ -21,8 +21,8 @@ const THROTTLE_QUEUE_THRESHOLD = 200 // ms
 
 type EventObject<T, U> = { name: string; data: T; action: (data: T) => Promise<U> | U }
 
-export class EventThrottle<T, U> {
-  queue: EventObject<T, U>[]
+export class EventThrottle {
+  queue: EventObject<any, any>[]
   timerId: NodeJS.Timeout | undefined
   queueActive: boolean
 
@@ -50,7 +50,7 @@ export class EventThrottle<T, U> {
     this.queueActive = false
   }
 
-  public add(event: EventObject<T, U>): void {
+  public add<T, U>(event: EventObject<T, U>): void {
     // Timer is used solely to determine whether to put new events at the front
     // or end of the queue.
     const currentTimer = this.timerId
--- a/src/components/actions/throttle.ts
+++ b/src/components/actions/throttle.ts
@@ -47,7 +47,11 @@ export class EventThrottle {
       if (next) {
         const { data, action, name } = next
         console.log('Processing queued event', name)
-        await action(data)
+        try {
+          await action(data)
+        } catch (e) {
+          console.log('Error while running throttled action: ', e)
+        }
         console.log(`${this.queue.length} events remaining in queue`)
       }
     }
--- a/src/components/database/updateReviewerStats.ts
+++ b/src/components/database/updateReviewerStats.ts
@@ -14,7 +14,8 @@
 
 import DBConnect from '../../components/database/databaseConnect'
 import { errorMessage } from '../utilityFunctions'
-import config from '../../config'
+
+const isManualUpdate: Boolean = process.argv[2] === '--update-reviewer-stats'
 
 interface NotificationPayload {
   tableName: string
@@ -22,36 +23,10 @@ interface NotificationPayload {
   operation: 'UPDATE' | 'INSERT'
 }
 
-export const updateReviewerStats = async ({ tableName, data, operation }: NotificationPayload) => {
-  let applicationId: number = 0
-  let reviewerIds: number[] = []
-
+export const updateReviewerStats = async (applicationId: number, userIds?: number[]) => {
   const db = databaseMethods
 
-  console.log({ tableName, data, operation })
-  console.log(`Updating reviewer action stats due to ${operation} trigger on table: ${tableName} `)
-
-  switch (tableName) {
-    case 'review_assignment':
-      applicationId = data.application_id as number
-      if (operation === 'INSERT') {
-        reviewerIds.push(data.reviewer_id as number)
-      }
-      if (operation === 'UPDATE') {
-        reviewerIds = await db.getAllStaffForApplication(applicationId)
-      }
-      break
-    case 'review_assignment_assigner_join':
-      applicationId = await db.getApplicationIdFromReviewAssignment(
-        data.review_assignment_id as number
-      )
-      reviewerIds.push(data.assigner_id as number)
-      break
-    case 'review_status_history':
-      break
-    case 'application':
-      break
-  }
+  let reviewerIds = userIds ?? (await db.getAllStaffForApplication(applicationId))
 
   // Update reviewer/assigner lists on application table
   console.log('Updating reviewer/assigner lists for application:', applicationId)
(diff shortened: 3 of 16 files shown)


plugins/action_generate_review_assignment_records/src/generateReviewAssignments.test.ts
```diff
@@ -4,7 +4,7 @@
 // database has entered correct information
 
 import { map, omit } from 'lodash'
-import DBConnect from '../../../src/components/databaseConnect'
+import DBConnect from '../../../src/components/database/databaseConnect'
```
CarlosNZ, line 7: Should have been part of previous PR (#1122), but didn't notice until I tried to build and got errors

src/components/actions/throttle.ts
```diff
@@ -21,8 +21,8 @@ const THROTTLE_QUEUE_THRESHOLD = 200 // ms
 
 type EventObject<T, U> = { name: string; data: T; action: (data: T) => Promise<U> | U }
 
-export class EventThrottle<T, U> {
-  queue: EventObject<T, U>[]
+export class EventThrottle {
+  queue: EventObject<any, any>[]
```
CarlosNZ, line 25: Okay, this has to be a truly legit use of `any`. I wanted to keep the Generic on the "EventObject", to enforce that the input "data" has to be consistent with the "action" function, but I *don't* want to have to explicitly specify a type when creating the "Event Throttle" object, as it should handle any type of Event object.

But, annoyingly, it requires a type in this definition, so it has to be `any`, right?

src/components/database/updateReviewerStats.ts
```diff
@@ -0,0 +1,231 @@
...
+      break
+    case 'application':
+      applicationId = data.id as number
+      // Update ALL staff by leaving reviewerIds empty
+      break
+  }
+  updateReviewerStats(applicationId, reviewerIds.length === 0 ? undefined : reviewerIds)
+}
+
+const databaseMethods = {
```
CarlosNZ, line 85: SQL methods here all basically copied as-is from the previous PG function.

src/types.ts
```diff
@@ -383,6 +384,7 @@ interface ConfigBase {
   isLiveServer: boolean
   emailMode: EmailOperationMode
   maintenanceMode: boolean
+  Throttle: EventThrottle
```
CarlosNZ, line 387: The Throttle used to be only used in one place, but now we need to make the single instance globally available, hence part of config object.

src/components/actions/throttle.ts
```diff
@@ -19,10 +19,14 @@
...
+interface EventObject<T> {
+  name: string
+  data: T
+  action: (data: T) => unknown
+}
 
-export class EventThrottle<T, U> {
-  queue: EventObject<T, U>[]
+export class EventThrottle {
+  queue: EventObject<any>[]
```
CarlosNZ, line 29: Okay, this has to be a truly legit use of `any`. I wanted to keep the Generic on the "EventObject", to enforce that the input "data" has to be consistent with the "action" function, but I *don't* want to have to explicitly specify a type when creating the "Event Throttle" object, as it should handle any type of Event object.

But, annoyingly, it requires a type in this definition, so it has to be `any`, right?

See [this](https://www.typescriptlang.org/play/?jsx=0#code/PTAEHUFNQCwQwG7TgO1AewEYCtIGMAXUAgTwAdoB3ASwJmMvVDICd0KWDrIBnAGgBQIUKgAmoSlVRECTSCgBm6Fnmh04RTOjrM2HLrxEto6BcRjQecALZrykQcMwBXIpNCj0KAORE8cYwl4GXs+UB4md2ceZFAAcXlIFmo8ADpQAAVjAlJwgjEA8R4CZwUFVIFqFAIkhThVUAARDTgAeRx8AgAeABUAPlAAbwFQXXYk0gBGAC5QHpGx-RIAJln5gF8BITAAIVdzeQloeCQRFG0LFgwOwmJQoJT6fzQYdAAbcVQjFjgSDDM6JBtk0Wu1cLcsOCCDx0lAPF5fBJpMQmMYAI7OaiBdQyCwg-Jgzo8UCYZBvN7-A7Aqy2O4UMIRWDoSgSdDOD5HeHmajEmh0NlEABEqBIgoA-JVqrV6tAAMJefJVJKE27DUYoGyQWbFZIoADmC1ELQAkjVrDxZs0CTduiK+gBtAC6AnWQA) for a simplified example of the case.

src/components/database/updateReviewerStats.ts
```diff
@@ -0,0 +1,231 @@
...
+      const { application_id, reviewer_id, assigner_id } = await db.getApplicationDataFromReview(
+        data.review_id as number
+      )
+      applicationId = application_id
+      reviewerIds.push(...new Set([reviewer_id, assigner_id]))
+      break
+    case 'application':
+      applicationId = data.id as number
+      // Update ALL staff by leaving reviewerIds empty
+      break
```
andreievg, line 80: default case would make applicationId = 0 ?

src/components/actions/throttle.ts
andreievg: Could do something like this: https://github.com/msupply-foundation/conforma-server/compare/1120-fix-trigger-block-for-review-stats-update...1120-fix-trigger-block-for-review-stats-update(No-any-example)

andreievg:
I think it's worth handling an error in action queue: https://github.com/msupply-foundation/conforma-server/compare/1120-fix-trigger-block-for-review-stats-update...1120-fix-trigger-block-for-review-stats-update(Error-handling)

src/components/database/updateReviewerStats.ts
```diff
@@ -0,0 +1,231 @@
...
+                SELECT id FROM public.review_assignment
+                WHERE application_id = $1
+            );
+        `
+      const result = await DBConnect.query({
+        text,
+        values: [appId],
+        rowMode: 'array',
+      })
+      return result.rows.flat()
```
andreievg, line 124: it's possible that reviewers will also be assigners ? I think we would want to dedup

CarlosNZ: The `DISTINCT` in each subquery and the `UNION` results in a list of unique IDs, no dupes (assigners and reviewers combined). You would think the `UNION` would combine them, but it doesn't in this case 🤷‍♂️

andreievg: Ohh true, looks like union does distict on rows https://www.postgresql.org/docs/current/queries-union.html

andreievg (approved):
Looks great thanks, I had a quick look and it seems pretty solid to me. I do have a slight concern about, trying to do granular updates, like when [review_status_history](https://github.com/msupply-foundation/conforma-server/blob/a3afd1ed21ef124d2b231b4f1fa135d740ecfb9d/src/components/database/updateReviewerStats.ts#L70-L74) has changed, we only look at reviewer and assigner, I think it would be simpler to just do updates for all of the application staff, but that could lead to excessive updates, especially if application is changed soon after. In which case I think i would create a new queue in database, with just application_ids and auto increment, and do a simple polling (every x seconds) to process it (grouping application_ids updates) <- I don't think it will come to that as the current logic looks pretty solid, thanks !

Is there a change that updateReviewStatsFromDBEvent will run multiple times for the same DB event ? (like when review_status_history update or review_assignment or application will be updated multiple times in sequence ?)

src/components/actions/throttle.ts
CarlosNZ: Yeah, except I don't really like having to wrap the `action` into an additional anonymous function call just to keep Typescript happy. I like being able to pass the `action` and `data` in separately and calling `action` on `data` directly, and using the Generic to enforce consistency between action method and its input.

Minor thing, slightly prefer my way. (Plus, any thoughts about the even more general case in the [playground example](https://www.typescriptlang.org/play/?jsx=0#code/PTAEHUFNQCwQwG7TgO1AewEYCtIGMAXUAgTwAdoB3ASwJmMvVDICd0KWDrIBnAGgBQIUKgAmoSlVRECTSCgBm6Fnmh04RTOjrM2HLrxEto6BcRjQecALZrykQcMwBXIpNCj0KAORE8cYwl4GXs+UB4md2ceZFAAcXlIFmo8ADpQAAVjAlJwgjEA8R4CZwUFVIFqFAIkhThVUAARDTgAeRx8AgAeABUAPlAAbwFQXXYk0gBGAC5QHpGx-RIAJln5gF8BITAAIVdzeQloeCQRFG0LFgwOwmJQoJT6fzQYdAAbcVQjFjgSDDM6JBtk0Wu1cLcsOCCDx0lAPF5fBJpMQmMYAI7OaiBdQyCwg-Jgzo8UCYZBvN7-A7Aqy2O4UMIRWDoSgSdDOD5HeHmajEmh0NlEABEqBIgoA-JVqrV6tAAMJefJVJKE27DUYoGyQWbFZIoADmC1ELQAkjVrDxZs0CTduiK+gBtAC6AnWQA)?)

src/components/database/updateReviewerStats.ts
CarlosNZ: Yeah, which would do nothing (tested the sql, doesn't throw any errors, but updates no records)

CarlosNZ:
> I think it's worth handling an error in action queue: [1120-fix-trigger-block-for-review-stats-update...1120-fix-trigger-block-for-review-stats-update(Error-handling)](https://github.com/msupply-foundation/conforma-server/compare/1120-fix-trigger-block-for-review-stats-update...1120-fix-trigger-block-for-review-stats-update(Error-handling))

Yeah, good call: 833703de

CarlosNZ:
> Is there a change that updateReviewStatsFromDBEvent will run multiple times for the same DB event ? (like when review_status_history update or review_assignment or application will be updated multiple times in sequence ?)

Not sure why that would happen. Previously there was a bit of a problem with the trigger field being set/updated/reset in quick succession causing multiple executions, hence I've tightened up those triggers (on `application` and `review_assignment`) to only respond to particular columns (i.e. NOT the trigger column).

And `review_status_history` only ever gets inserts, never updates (and only when the status changes).

CarlosNZ:
Thanks for the review @andreievg 👍🏼

